PR #3831, opened 2015-12-23
Uninstall nano

rjw1:
Since d65f973b7114d0eefe1aa3429c0d5042f627c996 we have not been bothered by
nano being installed on our systems. No one who is on call should need to
use nano to help them fix things anymore. Therefore we should remove nano.


mattbostock:
See https://github.com/alphagov/govuk-puppet/pull/3972

PR #3874, opened 2016-01-08
Validate TLS certificates

mattbostock:
Use the `validate_x509_rsa_key_pair()` function from Puppet stdlib to
verify the following for our TLS certificates:

- the private key matches the certificate
- the private key is correctly formatted, i.e. it is a private key and
  is not malformed
- the certificate is correctly formatted, i.e. it is a PEM-encoded
  certificate and is not malformed

We've seen each of these cases happen in the past, which this change
will guard against. If validation fails, the Puppet run will abort with
an error.


deanwilson:
Still not present in stdlib version 4.11.0.

mattbostock:
Closing this for now until a new version of stdlib is released.

PR #3876, opened 2016-01-11
Improvements to automatic app restarting

alexmuller:
I haven't tested this in vagrant yet, so it's probably best to not merge it unless you're supremely confident it's ok.

### DIFF
--- a/hieradata/staging.yaml
+++ b/hieradata/staging.yaml
@@ -507,6 +507,7 @@ hosts::production::router::hosts:
     legacy_aliases:
       - "draft-router-api.%{hiera('app_domain')}"
 
+icinga::config::enable_event_handlers: true
 icinga::nginx::enable_basic_auth: false
 
 #Do not enable this without speaking to Brad first


modules/monitoring/files/usr/lib/nagios/plugins/reload_service
```diff
@@ -0,0 +1,15 @@
...
+# Script that NRPE runs when it's notified by the monitoring machine that an application
+# has crossed the high memory threshold.
+#
+set -eu
+
+APPNAME=$1
+
+logger --tag govuk_icinga_event_handler "Reloading service ${APPNAME} because monitoring has detected it using too much memory"
+
+sudo /sbin/initctl reload ${APPNAME}
```
mattbostock: I think the `sudo` here is unneeded, as it's the `nagios` user we're allowing in the sudoers config. I'll raise a PR to remove.

rjw1: @mattbostock are you sure about that?

mattbostock: @rjw1 Can you please be less cryptic?

rjw1: how does it know to escalate privileges if you don't call it with sudo?

mattbostock: Good point :-D Thanks @rjw1

PR #3877, opened 2016-01-11
Reboot MongoDB machines overnight

alexmuller:
This script emulates the checks we perform when rebooting a MongoDB machine using Fabric:

- Make sure the cluster is healthy
- Make sure the candidate to reboot is not the primary

I've explicitly not introduced the logic around stepping down the primary to reboot it so that the first version of this is less complicated. This is still safe, it just means that we won't reboot the primary overnight until an election takes place at some other time.


mattbostock:
:eggplant:

PR #3885, opened 2016-01-12
Remove leavers from Vagrant GPG recipients file

mattbostock:
Kush and Tim no longer work on GOV.UK. I haven't re-encrypted the one
encrypted credential in `vagrant_credentials.yaml` because it's not a
secret; it's just used as a canary to ensure that Hiera eYAML GPG is
working.


alexmuller:
Merging in spite of the test failures because Travis is broken.

PR #3886, opened 2016-01-12
Move creation of CDN logs processed directory

alexmuller:
Moving this into the `govuk::apps::govuk_cdn_logs_monitor` class removes duplication. I missed this `file` resource when I changed the directory recently.


deanwilson:
Looks good. Merge when ready.

PR #3888, opened 2016-01-12
[#109899450] Service dependency for vpn hosts

surminus:
**Affects monitoring so careful rollout required**

In 6f49a3512e905a156e8f8ca35dbcc9dee5fac721 we created a check which failed if the monitoring server was unable to ping the remote gateway of a VPN tunnel. 

This looks to add dependency for any nodes that rely on a VPN to communicate with the monitoring server (EFG, Licensify or any DR nodes). 

Each node creates an exported resource which specifies itself and all it's services to be dependent on monitoring-1 and the gateway service for the same vDC in which it lives. In it's current form it will create the configuration inside of `/etc/icinga/conf.d/icinga_host_api-mongo-4.api.dev.gov.uk/` (as an example) which I think helps keep the configuration split correctly between the hosts, but thoughts welcomed on this approach.

The `vdc` Facter fact is key in helping to determine which machines need to export their resource and in assigning the right service to be dependent upon.

---

Testing in Vagrant creates this configuration file in the location specified:
```
vagrant@monitoring-1:/etc/icinga/conf.d$ cat icinga_host_licensify-frontend-1.licensify.dev.gov.uk/dependency_vpn_licensify-frontend-1.cfg
define servicedependency {
        host_name                       monitoring-1.management.dev.gov.uk
        service_description             Unable to ping vpn_gateway_licensify
        dependent_host_name             licensify-frontend-1.licensify.dev.gov.uk
        dependent_service_description   *
        execution_failure_criteria      n
        notification_failure_criteria   w,c
        }
```

In terms of `execution_failure_criteria` and `notification_failure_criteria`, this means that the nodes will _always_ be actively checked, but will stop "notifications" if the `service_description` goes into a Warning or Critical state. **I welcome discussion if this is the correct thing, and requires some testing in Integration to make sure it does the correct thing**

### DIFF
--- a/modules/govuk/lib/facter/vdc.rb
+++ b/modules/govuk/lib/facter/vdc.rb
@@ -1,0 +1,13 @@
+# Fact which specifies the name of the vDC of a node
+
+Facter.add("vdc") do
+  setcode do
+    env_octet = Facter.value(:ipaddress).split('.')[2].to_i
+    vdc = Facter.value(:domain).split('.').first
+    if env_octet > 7
+      vdc + '_dr'
+    else
+      vdc
+    end
+  end
+end
--- a/modules/icinga/manifests/client/checks.pp
+++ b/modules/icinga/manifests/client/checks.pp
@@ -1,7 +1,17 @@
-# FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
+# == Class: icinga::client::checks
+#
+# This class provides the default checks for a client machine.
+#
+# === Parameters:
+# [*disk_time_warn*]
+#   The disk time in milliseconds that will causing a warning state.
+#
+# [*disk_time_critical*]
+#   The disk time in milliseconds that will cause a critical state.
+#
 class icinga::client::checks (
-  $disk_time_warn = 100, # milliseconds
-  $disk_time_critical = 200 # milliseconds
+  $disk_time_warn = 100,
+  $disk_time_critical = 200,
 ) {
 
   include icinga::client::check_cputype
--- a/modules/icinga/manifests/service_dependency.pp
+++ b/modules/icinga/manifests/service_dependency.pp
@@ -1,0 +1,73 @@
+# == Define: icinga::service_dependency
+#
+# This creates a service dependency for the host service that other host
+# services are dependent on.
+#
+# Most of these params relate to Nagios service configuration directives:
+#
+# https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/objectdefinitions.html#servicedependency
+#
+# === Parameters:
+#
+# [*dependent_host_name*]
+#   This host is dependent on `host_name`. The title of an `Icinga::Host`
+#   resource, usually `$::fqdn`.
+#
+# [*host_name*]
+#   The host upon which the service that causes the dependency runs on.
+#
+# [*service_description*]
+#   Description of the service that is dependent services rely upon.
+#
+# [*check_command*]
+#   Command used to check whether the service is healthy.
+#
+# [*dependent_service_decription*]
+#   Description of the service where functionality is changed depending on the
+#   status of `service_description`.
+#
+# [*execution_failure_criteria*]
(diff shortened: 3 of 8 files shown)


modules/icinga/manifests/service_dependency.pp
```diff
@@ -0,0 +1,89 @@
...
+
+    $check_content = template('icinga/service_dependency.erb')
+
+    Icinga::Host[$host_name] -> Icinga::Service_dependency[$title]
+
+  } else {
+    $check_content = ''
+  }
+
+  file { $check_filename:
```
deanwilson: Should we write an essentially empty config file out or should this resource be inside the if condition?

surminus: @deanwilson agreed, I've updated it to much simpler logic.

surminus:
I ran the a test in Vagrant, bringing up a licensify node to show the example configuration that is created:
```
vagrant@monitoring-1:/etc/icinga/conf.d$ cat /etc/icinga/conf.d/icinga_host_licensify-frontend-1.licensify.dev.gov.uk/dependency_vpn_licensify-frontend-1.cfg
define servicedependency {
        host_name                       licensify-frontend-1.licensify.dev.gov.uk
        service_description             *
        dependent_host_name             monitoring-1.management.dev.gov.uk
        dependent_service_description   Unable to ping vpn_gateway_licensify
        execution_failure_criteria      n
        notification_failure_criteria   w,c
        }
```

It would be good to know that the host_name and dependent_host_name are the right way round as per http://docs.icinga.org/latest/en/dependencies.html#example

modules/icinga/manifests/client/checks.pp
```diff
@@ -1,7 +1,17 @@
...
+#   The disk time in milliseconds that will causing a warning state.
+#
+# [*disk_time_critical*]
+#   The disk time in milliseconds that will cause a critical state.
+#
 class icinga::client::checks (
-  $disk_time_warn = 100, # milliseconds
-  $disk_time_critical = 200 # milliseconds
+  $disk_time_warn = 100,
+  $disk_time_critical = 200,
```
alexmuller: Nice tidy up here :+1:

modules/govuk/lib/facter/vdc.rb
```diff
@@ -0,0 +1,13 @@
...
+  setcode do
+    env_octet = Facter.value(:ipaddress).split('.')[2].to_i
+    vdc = Facter.value(:domain).split('.').first
+    if env_octet > 7
+      vdc + '_dr'
+    else
+      vdc
+    end
+  end
+end
```
alexmuller: It'd be good to add some tests for this Facter logic - we've got a couple of examples like in `fqdn_metrics_spec.rb`. Happy to pair on that if you want.

modules/icinga/manifests/service_dependency.pp
```diff
@@ -0,0 +1,73 @@
...
+
+  if $service_description == undef {
+    fail("Must provide a \$service_description to icinga::Service_dependency[${title}]")
+  }
+  if $dependent_service_description == undef {
+    fail("Must provide a \$dependent_service_description to icinga::Service_dependency[${title}]")
+  }
+
+  if $dependent_host_name == undef {
+    fail("Must provide a \$host_name to icinga::Service_dependency[${title}]")
```
alexmuller: If I can be really, really nitpicky, I think the 3 `fail()` methods in this file should have a capital I for Icinga :)

alexmuller:
This all looks very cool :+1: 

I think your choices for `execution_failure_criteria` and `notification_failure_criteria` are definitely correct as a first attempt. My understanding is that `n` for `execution_failure_criteria` means that Icinga will always continue to run checks against the dependent service but `w,c` for the `notification_failure_criteria` means we won't send notifications if the main (Icinga [describes this as 'master'](http://docs.icinga.org/latest/en/objectdefinitions.html)) check is warning or critical.

deanwilson:
The code looks fine to me. Integration here we come!

PR #3890, opened 2016-01-12
Remove token performanceplatform_big_screen_view

alexmuller:
`govuk::apps::performanceplatform_big_screen_view` is not a proper app (it only includes the `app::package` and `app::nginx_vhost` classes) so it can't have environment variables.

This may not be the correct way to solve the fact that Puppet isn't running on the performance-frontend machines. Does this app need to communicate with the publishing-api?

/cc @danielroseman @alphagov/team-publishing-platform

### DIFF
--- a/modules/govuk/manifests/apps/performanceplatform_big_screen_view.pp
+++ b/modules/govuk/manifests/apps/performanceplatform_big_screen_view.pp
@@ -9,13 +9,8 @@
 #   Whether the app is enabled.
 #   Default: true
 #
-# [*publishing_api_bearer_token*]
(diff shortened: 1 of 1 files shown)


danielroseman:
Thanks Alex. The reason I added this var is that there is a special route registered in the Publishing API from this app, which is usually done at deploy time. However looking at the app itself I can't see any evidence of the registration code, so I don't know how that route got there.

It does look like it's fine to remove this token though.

alexmuller:
> I don't know how that route got there

I wouldn't be surprised if it was manually put there at some point in the past :disappointed:

PR #3892, opened 2016-01-13
[108858698] Update daily sync time in Integration

surminus:
The machines in Integration are switched off over night, which means that the daily sync of Whitehall attachments (ensuring any that may have been missed for $reasons) have not been caught and the asset-slave nodes are out of sync with the asset-master node.

The job itself isn't directly impacting so 09:00 should be fine to sync them up.


mattbostock:
:+1:

PR #3896, opened 2016-01-14
Change attachment sync in integration to 11am

alexmuller:
The data sync for attachments from production to integration has happened between these times for the last 4 days:

- 10:49 to 10:58
- 10:45 to 10:56
- 10:37 to 10:46
- 10:41 to 10:49

We want this final rsync of attachments to run after the data sync so that the number of alerts in integration is minimised. This commit will make it run between 1100 and 1200 so it should only alert for a few minutes in integration.


alexmuller:
Pushed a 2nd commit to this branch which very slightly tweaks the way the disk space alert is generated because it was too trivial to be its own pull request.

surminus:
:+1:

PR #3899, opened 2016-01-14
Export the `reload_service` Icinga plugin

mattbostock:
Story: https://trello.com/c/8U5ou5rr/77-gracefully-kill-unicorn-workers-if-they-use-too-much-ram

The `reload_service` plugin needs to be installed on every machine, so
make it an exported resource that's collected by the `icinga::client`
class.

### DIFF
--- a/modules/icinga/manifests/client.pp
+++ b/modules/icinga/manifests/client.pp
@@ -44,4 +44,5 @@ class icinga::client {
 
   Icinga::Nrpe_config <| |>
   Icinga::Plugin <| |>
+  Icinga::Plugin <<| |>>
 }
--- a/modules/monitoring/manifests/event_handler/app_high_memory.pp
+++ b/modules/monitoring/manifests/event_handler/app_high_memory.pp
@@ -10,13 +10,13 @@ class monitoring::event_handler::app_high_memory () {
     require => File['/usr/local/bin/event_handlers/govuk_app_high_memory.sh'],
   }
 
-  @icinga::plugin { 'reload_service':
-    source => 'puppet:///modules/monitoring/usr/lib/nagios/plugins/reload_service',
-  }
-
   file { '/usr/local/bin/event_handlers/govuk_app_high_memory.sh':
     source => 'puppet:///modules/govuk/usr/local/bin/event_handlers/govuk_app_high_memory.sh',
     mode   => '0755',
+  }
+
+  @@icinga::plugin { 'reload_service':
+    source => 'puppet:///modules/monitoring/usr/lib/nagios/plugins/reload_service',
   }
 
 }


alexmuller:
I understand that `<| |>` is a resource collector and `<<| |>>` is an exported resource collector ([docs](https://docs.puppetlabs.com/puppet/latest/reference/lang_collectors.html)) but I can't really connect the dots here.

Changing to a `@@` means the plugin is exported from the monitoring machine and now will be collected on all machines, which makes sense. But then what does the line `Icinga::Plugin <| |>` actually do?

mattbostock:
`<| |>` will realise the virtual resources; `<<| |>>` will collect the exported resource I've created. The `Icinga::Plugin <| |>` is still needed as many of our modules have virtual `Icinga::Plugin` resources which need to be realised by that line.

deanwilson:
Is there any reason this needs to be exported and can't just be a normal Icinga::package resource?

mattbostock:
@deanwilson `monitoring::event_handler::app_high_memory` is only installed on the monitoring box, whereas this plugin needs to be installed on all boxen. I wanted to keep its definition close to the thing that uses it in the code for better encapsulation, rather than separate it into another class.

mattbostock:
What can I do to help get this merged? Happy to talk through the code with someone.

boffbowsh:
This makes sense to me. Would also like to get this merged. @alexmuller @deanwilson are you happy with it?

boffbowsh:
Pleeeeeeease?

deanwilson:
Can we add a tag to the export and collection please? That'll stop all the plugins being realised when that code is hit. https://github.com/voxpupuli/puppet-lint-spaceship_operator_without_tag-check explains the thinking behind it.

boffbowsh:
@deanwilson what should the tag be? Is `virtual == true` sufficient in this case or should it be something specific?

deanwilson:
Something context specific. I'm being a little vague as naming things is difficult and to be honest the name isn't as important as the existence of the restricting tag.

surminus:
We should get this merged today. I'm going to speak to @deanwilson and gang later.

mattbostock:
@deanwilson From what I understand of the [bug report](https://tickets.puppetlabs.com/browse/PUP-3825), the bug only applies to virtual resources (as opposed to exported resources, which this PR adds) and only poses a problem if specifying dependencies. Is that correct? So from my understanding, the code added here would not make the problem worse.

If so, can I suggest that we tackle it in a separate PR so that we can unblock this one?

deanwilson:
After another discussion we've looked at having the plugin as a distinct resource from the handler so it can be reused by other handlers, such as high CPU. This also removed the complexity of exporting and collecting.

mattbostock:
Superseded by https://github.com/alphagov/govuk-puppet/pull/3899.

MatMoore:
The above link should be #4074.

PR #3901, opened 2016-01-14
[#108858698] Enable new Whitehall attachment processing

alexmuller:
We've tested this in integration and we believe it works as we expect.

This might be easier to review by adding `?w=1` to the URL to remove whitespace changes.


surminus:
LGTM, very exciting :fireworks:

alexmuller:
Pushed a commit for the CPU check.

rjw1:
https://govuk.zendesk.com/agent/tickets/1229635 may have a bearing on this. So hold off for a while.

alexmuller:
We've had a look at that and we think it's unrelated.

mattbostock:
:+1: Nice work! :eggplant:

PR #3904, opened 2016-01-14
Rotate ssh key for paulbowsher

boffbowsh:


mattbostock:
Hi @boffbowsh, I notice that this is a 2048-bit key. That's sufficient, but you might consider upgrading this to a 4096-bit key, which _may_ be more secure, e.g.:

```
ssh-keygen -t rsa -b 4096
```

boffbowsh:
Upgraded, thanks @mattbostock

PR #3913, opened 2016-01-14
Rotate ssh key for Steve Laing

steventux:
Due to CVE-0216-0778


mattbostock:
Hi @steventux, I notice that this is a 2048-bit key. That's sufficient, but you might consider upgrading this to a 4096-bit key, which _may_ be more secure, e.g.:

```
ssh-keygen -t rsa -b 4096
```

steventux:
@mattbostock thanks for the headsup - I've amended the commit with a 4096-bit key

PR #3925, opened 2016-01-14
Rotate ssh key for matmoore

MatMoore:
As requested due to CVE-0216-0778:
http://undeadly.org/cgi?action=article&sid=20160114142733


mattbostock:
Thanks @MatMoore, would you mind squashing these two commits?

MatMoore:
@mattbostock that's done now.

PR #3951, opened 2016-01-15
Remove users

mattbostock:
For various reasons, e.g. annual leave, these users are unable to rotate
their SSH key immediately (as mitigation against CVE-2016-0777).

Until they are able to rotate their keys, we may as well remove their
users (rather than have a public key of `REPLACE ME`).

* * *

Also, remove users that have left GOV.UK.

### DIFF
--- a/modules/users/manifests/chrisroos.pp
+++ b/modules/users/manifests/chrisroos.pp
@@ -1,8 +1,0 @@
-# Creates the chrisroos user
-class users::chrisroos {
-  govuk::user { 'chrisroos':
-    fullname => 'Chris Roos',
-    email    => 'chris.roos@digital.cabinet-office.gov.uk',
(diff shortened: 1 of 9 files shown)


daibach:
:+1: LGTM. Merge when ready.

surminus:
puppet is now failing to run in Integration: 
```
Jan 18 16:11:14 monitoring-1.management puppet-agent[17050]: Could not retrieve catalog from remote server: Error 400 on SERVER: Could not find class users::chrisroos for monitoring-1.management.integration.publishing.service.gov.uk on node monitoring-1.management.integration.publishing.service.gov.uk
```

mattbostock:
Thanks @surminus, addressed in https://github.com/alphagov/govuk-puppet/pull/3958.

PR #3952, opened 2016-01-18
Set bearer token for pp-big-screen-view.

danielroseman:
Since this does not use the govuk::app class, we need to set create the env dir explicitly.

### DIFF
--- a/modules/govuk/manifests/apps/performanceplatform_big_screen_view.pp
+++ b/modules/govuk/manifests/apps/performanceplatform_big_screen_view.pp
@@ -9,8 +9,13 @@
 #   Whether the app is enabled.
 #   Default: true
 #
+# [*publishing_api_bearer_token*]
+#   The bearer token to use when communicating with Publishing API.
+#   Default: undef
+#
 class govuk::apps::performanceplatform_big_screen_view (
   $enabled = false,
+  $publishing_api_bearer_token = undef,
 ) {
   include govuk::deploy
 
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/performanceplatform_big_screen_view.pp
```diff
@@ -29,5 +34,18 @@
       ssl_only        => true,
       app_port        => 3058,
     }
+    # Since this is not a full app, we need to ensure env dir exists.
+    file { "/etc/govuk/${app_name}/env.d":
+      ensure  => $ensure_directory,
+      purge   => true,
+      recurse => true,
+      force   => true,
+      notify  => Govuk::App::Service[$title],
```
alexmuller: I think this `notify` line needs to go - because the app is just static HTML there is no service to notify.

alexmuller:
There will need to be a related pull request on the deployment repo to set the value, but this looks fine otherwise :+1:

danielroseman:
@alexmuller updated. As discussed the value is already present in the deployment repo.

modules/govuk/manifests/apps/performanceplatform_big_screen_view.pp
```diff
@@ -29,5 +34,17 @@
       ssl_only        => true,
       app_port        => 3058,
     }
+    # Since this is not a full app, we need to ensure env dir exists.
+    file { "/etc/govuk/${app_name}/env.d":
```
alexmuller: I think this line needs to be:

```
file { ["/etc/govuk/${app_name}", "/etc/govuk/${app_name}/env.d"]:
```

because the parent directory of the `env.d` directory also isn't created.

modules/govuk/manifests/apps/performanceplatform_big_screen_view.pp
```diff
@@ -29,5 +34,17 @@
...
+      ensure  => 'directory',
+      purge   => true,
+      recurse => true,
+      force   => true,
+    }
+    govuk::app::envvar { "${app_name}-PUBLISHING_API_BEARER_TOKEN":
+      app     => $app_name,
+      varname => 'PUBLISHING_API_BEARER_TOKEN',
+      value   => $publishing_api_bearer_token,
```
alexmuller: Final thing (sorry this is such a hassle :disappointed:), this `govuk::app::envvar` needs a parameter of `notify_service => false,` because otherwise the `govuk::app::envvar` defined type will attempt to notify the service `performanceplatform-big-screen-view` which doesn't exist:

```
Error: Could not find dependent Govuk::App::Service[performanceplatform-big-screen-view] for File[/etc/govuk/performanceplatform-big-screen-view/env.d/PUBLISHING_API_BEARER_TOKEN] at /var/govuk/govuk-puppet/modules/govuk/manifests/app/envvar.pp:24
```

danielroseman:
Thanks, updated again.

PR #3954, opened 2016-01-18
Add user gemma leigh

gemmaleigh:
Re-add user gemma leigh, removed in
https://github.com/alphagov/govuk-puppet/commit/598c6d4


deanwilson:
Looks fine to me.

PR #3956, opened 2016-01-18
Move to saz limits

deanwilson:
In an attempt to move to mature, 3rd party, modules this PR covers a possible migration from harden::limits to saz/limits. It should produce the same settings as using `Harden` but with each value in its own files, and none bunched in to limits.conf

The story is located at https://trello.com/c/nVs0krua


rjw1:
```
==> backend-2.backend: Error: Cannot alias File[/etc/security/limits.d/] to ["/etc/security/limits.d"] at /var/govuk/govuk-puppet/vendor/modules/limits/manifests/init.pp:12; resource ["File", "/etc/security/limits.d"] already declared at /var/govuk/govuk-puppet/vendor/modules/harden/manifests/init.pp:54
```
when run under vagrant

Puppetfile
```diff
@@ -49,7 +50,7 @@ mod 'gdsoperations/gor',
   :ref => 'v0.2.0'
 mod 'alphagov/harden',
   :git => 'git://github.com/alphagov/puppet-harden.git',
-  :ref => 'fa2c51e994b87e71e7b424b6c0048b70c2df36b0'
+  :ref => '69bcbb3'
```
mattbostock: We should probably cut a new release of puppet-harden and use that.

mattbostock:
This PR should also remove the old copy of the vendored puppet-harden module.

mattbostock:
@deanwilson It'd be good to say in your commit messages and/or PR description if this change makes any functional difference; are any new limits set which weren't previously?

deanwilson:
I've updated the PR to mention the functional side. Nothing should really change but the harden module was inconsistent in how it set things so there will be a minor change as settings break out in to individual files to be consistent with all the other usages of `limits`.

mattbostock:
Thanks @deanwilson :+1: It makes it easier for people to test if they know what to look for.

deanwilson:
Must have missed one, which vendored file needs to be removed?

mattbostock:
You can find it using the old SHA from the Puppetfile.lock, it's:
`vendor/puppet/source/fa2c51e994b87e71e7b424b6c0048b70c2df36b0.tar.gz`

deanwilson:
Removed the old file. Thanks.

mattbostock:
LGTM, are you able to test @rjw1's comment in Vagrant?

rjw1:
:+1: lgtm

PR #3961, opened 2016-01-19
creating a new manifest for matteograssotti

mgrassotti:


bradwright:
Welcome @mgrassotti, glad to have you onboard!

Two things:

- Firstly, your change has lint issues, specifically: `class not documented` - if you copy the docs from eg [my user](https://github.com/alphagov/govuk-puppet/blob/master/modules/users/manifests/bradleyw.pp#L1) it'll pass
- Secondly your user isn't being added to any servers - if you add yourself to integration [here](https://github.com/alphagov/govuk-puppet/blob/master/hieradata/integration.yaml#L46) and [here](https://github.com/alphagov/govuk-puppet/blob/master/hieradata/integration.yaml#L236) your user will actually be usable

Let me know if you have any questions!

mgrassotti:
Thanks @bradwright ! I am doing the changes...

mattbostock:
Welcome @mgrassotti!

There was an OpenSSH [vulnerability](http://undeadly.org/cgi?action=article&sid=20160114142733) announced last week and, as mitigation, we're asking that all developers add the line `UseRoaming no` to their SSH client configuration, e.g. in `~/.ssh/config`:

```
Host *
  UseRoaming no
```

Can you please check that you have this setting?

mgrassotti:
@mattbostock thanks! I added those lines to my ssh config files. Sorry for the late reply, I was not at work.

PR #3962, opened 2016-01-19
[#111875397] Trigger GDS_Production_Backup after Puppet deployment

alexmuller:
Puppet is now hosted on public GitHub, so we should back it up to GitHub Enterprise like we do with all of our other apps.


surminus:
:+1: LGTM, merge when ready

rjw1:
:+1:

PR #3964, opened 2016-01-19
Update default execution_failure_criteria

surminus:
This changes the default behaviour so that if the master service is either warning or critical, then active checks are disabled for the dependent services.

This is changed from just notification behaviour previously, so even if a master service was warning or critical, the Icinga dashboard was still spammed with alerts that we did not want to see, but notifications (i.e sending out emails) were disabled.


deanwilson:
Looks fine. Merge away.

mattbostock:
:+1:

PR #3966, opened 2016-01-19
Add end-of-line to manifest

mattbostock:
Adding a trailing newline to a file makes it more compatible with shell
tools such as `cat` and maintains consistency with our other manifests.


surminus:
:+1: merge when ready

PR #3967, opened 2016-01-19
Use the current version of puppet-syntax

deanwilson:
We should be testing our own code. The other updates are from the
bundle update puppet-syntax command and seem to cause no issues when
running tests.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -25,7 +25,7 @@ GEM
     crack (0.4.2)
       safe_yaml (~> 1.0.0)
     diff-lcs (1.2.5)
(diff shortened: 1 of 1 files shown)


alexmuller:
Did you need to make a change to the `Gemfile` too or was it just a case of running `bundle update puppet-syntax`?

alexmuller:
For example, I don't understand why the version of `facter` has changed here.

deanwilson:
This is just from `bundle update puppet-syntax`. I'm happy to close this PR and someone with more bundler experience can have a look.

PR #3968, opened 2016-01-19
Use capitalised first letter in 'Process'

mattbostock:
...so that the service description matches the service description in
the Icinga check. These descriptions are case sensitive, so we'll see
false freshness alerts for this check if they don't match case.

From:
http://docs.icinga.org/latest/en/nsca.html

> Host name (and service description, if applicable) are case sensitive
> and have to match the definition in Icinga.


surminus:
:+1: merge when green

rjw1:
The bikeshed should be purple!

merge when green :+1:

alexmuller:
Merge when purple :eggplant:

PR #3969, opened 2016-01-19
Allow Rummager to listen to RabbitMQ messages

bishboria:
Currently, changes to topics and other tags in content-tagger don't cause the search index (accessed via rummager) to be updated correspondingly. We want such changes to be reflected as soon as possible in the search index.

This change adds rabbitMQ settings to rummager

Part of: https://trello.com/c/lRnxI2x5/472-make-rummager-index-the-tags-in-the-links-hash

### DIFF
--- a/modules/govuk/manifests/apps/rummager.pp
+++ b/modules/govuk/manifests/apps/rummager.pp
@@ -17,14 +17,34 @@
 #   Whether to enable the procfile worker service.
 #   Default: true
 #
+# [*enable_publishing_api_document_indexer*]
+#   Whether to enable the procfile indexing service.
+#   Default: false
+#
 # [*publishing_api_bearer_token*]
 #   The bearer token to use when communicating with Publishing API.
 #   Default: undef
 #
+# [*rabbitmq_hosts*]
+#   RabbitMQ hosts to connect to.
+#   Default: localhost
+#
+# [*rabbitmq_user*]
+#   RabbitMQ username.
+#   Default: rummager
+#
+# [*rabbitmq_password*]
+#   RabbitMQ password.
+#   Default: rummager
+#
 class govuk::apps::rummager(
   $port = '3009',
   $enable_procfile_worker = true,
+  $enable_publishing_api_document_indexer = false,
   $publishing_api_bearer_token = undef,
+  $rabbitmq_hosts = ['localhost'],
+  $rabbitmq_user = 'rummager',
+  $rabbitmq_password = 'rummager',
 ) {
   include aspell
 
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/apps/rummager.pp
```diff
@@ -67,5 +98,15 @@
     "${title}-PUBLISHING_API_BEARER_TOKEN":
       varname => 'PUBLISHING_API_BEARER_TOKEN',
       value   => $publishing_api_bearer_token;
+    'AMQP_EXCHANGE':
```
rboulton: I think the queue and exchange name belong in the code repo - they feel like equivalent level "configuration" to a table name in a database.  The queue and exchange won't vary depending on deployment decisions (unlike the hosts, user and password settings).

That is the pattern followed by other consumers, such as the panopticon one.

modules/govuk/manifests/apps/rummager.pp
```diff
@@ -67,5 +98,15 @@
...
+      value => $rabbitmq_exchange;
+    'AMQP_MESSAGE_QUEUE':
+      value => $rabbitmq_queue;
   }
+
+  # Add this?
```
rboulton: I think this needs to be `govuk::procfile::worker`  with the `process_type` parameter set to the name of the worker (ie, the first bit of the line in the Procfile).

modules/govuk/manifests/apps/rummager.pp
```diff
@@ -52,10 +67,22 @@
...
+    password => $rabbitmq_password,
+  }
+
   govuk::procfile::worker { 'rummager':
     enable_service => $enable_procfile_worker,
   }
 
+  govuk::procfile::worker { 'rummager-publishing-api-document-indexer':
+    setenv_as      => 'rummager',
+    enable_service => $enable_procfile_worker,
```
rboulton: I think we want to feature-flag this separately from the existing procfile worker, so we can initially turn it on only in `integration` while we test it out.  So `enable_service => $enable_publishing_api_document_indexer` and around line 38 set a default parameter for that to false.  Then we also need to modify `hieradata/integration.yaml` to set that parameter to true in the `integration` environment (by adding a line something like `govuk::apps::rummager::enable_publishing_api_document_indexer: true`)

rboulton:
There are a few more things that need to be done.

 - hieradata needs to be updated to set the correct value for rabbitmq_hosts in each environment.  You can copy the way this is done for another consumer, eg, look at the output of `git grep panopticon::rabbitmq_hosts`

 - The `rabbitmq::monitor_consumers` check should be configured.  This will raise an alert in icinga if the consumer fails to process messages for any reason.  For an example, look in `modules/govuk/manifests/apps/email_alert_service.pp`.  It works by checking the time that a message was last processed on the queue (we send regular "no-op" heartbeat messages out from publishing-api, so this value should never be significantly higher than 1 minute).

There will also need to be some changes made in the deployment repository: passwords will need to be set for each environment.  I'll try and help with that later, or 2nd line may be able to if things are quiet.

modules/govuk/manifests/apps/rummager.pp
```diff
@@ -52,10 +72,26 @@
...
     enable_service => $enable_procfile_worker,
   }
 
+  govuk::procfile::worker { 'rummager-publishing-api-document-indexer':
+    setenv_as      => 'rummager',
+    enable_service => $enable_publishing_api_document_indexer,
+    process_type   => 'publishing-api-document-indexer',
+  }
+
+  govuk_rabbitmq::monitor_consumers {'rummage-publishing-api-document-indexer':
```
rboulton: typo here: needs to be `rummager-publish...` not `rummage-...`

PR #3971, opened 2016-01-19
Make use of validate_ip_address() function

mattbostock:
As of version 4.10.0 of Puppet's stdlib, there's a
`validate_ip_address()` functional that allows us to easily validate
IPv4 or IPv6 addresses. Use that instead of `is_ip_address()`.


alexmuller:
This is much nicer code, big :+1:

For the `Puppetfile` change, my understanding was always that the `Puppetfile` should specify the application dependencies (eg we require stdlib above version 4.10.0 because we're using something introduced then) while the `Puppetfile.lock` just specifies the versions that happened to be installed based on those requirements.

An example of when this might break is if we bumped another one of our dependencies which required an older stdlib. `librarian-puppet` would happily downgrade the stdlib because we don't specify we need a newer version.

mattbostock:
Thanks @alexmuller, your explanation was really clear. I hadn't thought of that use case - I've updated the `Puppetfile` to pin the appropriate version.

alexmuller:
Looks good to me, merge whenever :+1:

PR #3972, opened 2016-01-19
Revert "Uninstall nano"

mattbostock:
This reverts commit 5b39f2997abd88e948fc2c9891e6d2788b004454.

We're already specifying that vim should be used where possible, so
there's no need to uninstall Nano:
https://github.com/alphagov/govuk-puppet/blob/d5fdb18fd95a5202f13cf0634a8f6e66bcd0f54c/modules/base/manifests/packages.pp#L25-L28

Also, the code being reverted:
- uninstalls Nano by including a class called 'nano', which is confusing
- adds an external module, which we vendor, to remove a package which
  could be achieved with a package resource
- removes a package unnecessarily - there are lots of other packages we
  don't use from the Ubuntu base image but we don't remove those, so
  this adds extra complexity

I haven't done anything here to add Nano back as we don't need it, but I
don't think that keeping this code is beneficial, hence the revert.


surminus:
:+1: merge!

PR #3974, opened 2016-01-20
Add UFW rule for asset slave using NFS

alexmuller:
The asset slave machines need to connect to the asset master using NFS.

asset-slave-1 is currently allowed to connect because of the broad rule in `s_asset_base` which allows all backend machines to connect to the asset-master.

I'd rather not extend that broad rule to include asset-slave-2 (running on a different IP range in the disaster recovery organisation) so instead I've set up new UFW rules which are different:

- They only apply to the asset_master rather than the asset_base class
- They're explicit about being for the asset_slave only

In an ideal world I'd like to limit these rules by port number too, but I don't know enough about what ports NFS uses (it seems to use lots of high numbered ports).

### DIFF
--- a/hieradata/class/asset_master.yaml
+++ b/hieradata/class/asset_master.yaml
@@ -1,5 +1,11 @@
 ---
 
+govuk::node::s_asset_master::asset_slave_ip_ranges:
+  asset_slave_1_nfs:
+    from: "%{hiera('environment_ip_prefix')}.3.21"
+  asset_slave_2_nfs:
+    from: "%{hiera('environment_ip_prefix')}.11.21"
+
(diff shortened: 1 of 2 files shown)


surminus:
Looks good to me, but what do the asset-slaves require for NFS? I thought that requirement was deprecated?

alexmuller:
It's now [just asset-manager which we need sort out][1].

[1]: https://github.com/alphagov/govuk-puppet/blob/6278dde2271c5f46463c72cf407f8c47c781e8d1/modules/govuk/manifests/node/s_asset_slave.pp#L32

surminus:
Ah! Otherwise LGTM, :+1:, :aubergine:, merge when green, merge when purple, etc

PR #3975, opened 2016-01-20
Add HTML verification page for DWP YouTube channel

mattbostock:
Story:
https://trello.com/c/HfhQgKbR/18-youtube-verification-page-for-dwp
Zendesk ticket: https://govuk.zendesk.com/agent/tickets/1199172

The Department for Work and Pensions (DWP) have asked us to add a HTML
verification page so that YouTube will allow them to link to GOV.UK from
a YouTube video.

I don't yet have the content for the HTML verification page, so the file
I'm adding here is empty.

It seems a shame to have do to this in Puppet; ideally we'd have an
easier way of doing this. Doing it this way is preferable to DNS
verification as we'd have to request a DNS record change from JANET.

### DIFF
--- a/modules/router/templates/router_include.conf.erb
+++ b/modules/router/templates/router_include.conf.erb
@@ -61,6 +61,12 @@ location / {
 
   <%- end -%>
 
+  # HTML verification for DWP YouTube channel
+  location = /dla-ending/google6db9c061ce178960.html {
+    add_header Content-Type text/html;
+    return 200 '';
+  }
+
   location /robots.txt {
     root /usr/share/nginx/www;
   }


modules/router/manifests/nginx.pp
```diff
@@ -125,6 +125,13 @@
     require => File['/usr/share/nginx/www'],
   }
 
+  # HTML verification for DWP YouTube channel
+  file { '/usr/share/nginx/www/google6db9c061ce178960.html':
+    ensure  => present,
+    source  => 'puppet:///modules/router/google6db9c061ce178960.html',
```
alexmuller: I think the contents can just be an empty string. We could do `content => ''` so that we don't need a separate file.

mattbostock: I was going to do that but having read the documentation from Google, I suspect that the file may have contents we need to add. If not, I'll come back and tidy this up after.

rjw1:
the contents is likely to just be 
```
google-site-verification: google6db9c061ce178960.html
```

alexmuller:
Based on the latest update in Zendesk I think this just needs to return an empty `200 OK`. Could we just do this in the nginx config without having to put a file on disk?

mattbostock:
Amended to return a synthetic response from Nginx without serving a file.

alexmuller:
Looks correct to me, feel free to merge if the tests pass :+1:

rjw1:
:+1: merge away

bradwright:
> It seems a shame to have do to this in Puppet; ideally we'd have an easier way of doing this. 

We've discussed providing a way to serve static content via the content-store/router, and this seems like a good use case to plan that work for the future.

PR #3979, opened 2016-01-21
Adds user David Silva

Davidslv:
I'm following step 7 https://github.gds/gds/development#7-create-a-user-account-on-our-remote-servers


danielroseman:
Could you squash these commits into one? It's better to keep the commit history clean.

Davidslv:
thank you @danielroseman, I've squashed the commits.

PR #3981, opened 2016-01-21
Add second SSH key for paulhayes

fofr:
Add a second key for a second machine.


fofr:
For reference, the way @alext used multiple keys: https://github.com/alphagov/govuk-puppet/commit/e940cf687e405470b12567a4da0c286d22df5e01

PR #3983, opened 2016-01-22
Add second key for bob

rjw1:
So that bob can connect from home.


surminus:
I like it when bob can connect from home. :+1: please merge.

PR #3984, opened 2016-01-22
Add `display_name` to mirror hostnames

alexmuller:
These hosts were added at about the same time the `display_name` property was added so it wasn't configured on these 4 hosts.

If the `display_name` isn't provided it defaults to the FQDN of the monitoring machine which makes our monitoring quite confusing.


mattbostock:
Merge when ready.

PR #3985, opened 2016-01-22
Create vDC gateways as host parents

surminus:
In it's current state we have VPN vDC gateways set up as a separate external ping check on the monitoring server, and hosts with service dependencies checking against that service; the issue is that if the VPN goes down, then we will still get alerts and notifications related to those hosts. We only want a notification that there is a problem with the VPN. 

This looks to change that model, and create each vDC gateway as it's own host, with a host check behind that, and then make the nodes "children" of these hosts. The behaviour of this is that if the vDC gateway goes down, then the children nodes behind it go into "unreachable" state rather than "down". You can disable default alert/notifications of nodes in "unreachable" state, which should hopefully do what we want.

Things worth noting for review:

* Removing "unreachable" state as default - does anything else rely on being "unreachable"?
* The gateway hosts check use `host_name` parameter to know what to ping - does it do a DNS lookup, or use the underlying host config (which contains the IP address) to ping the correct thing?

---
Example output:

```
vagrant@monitoring-1:/etc/icinga/conf.d$ cat icinga_host_licensify-frontend-1.licensify.dev.gov.uk.cfg
define host {
        use          generic-host
        alias        licensify-frontend-1.licensify.dev.gov.uk
        address      10.0.2.15
        display_name licensify-frontend-1.licensify.dev.gov.uk
        host_name    licensify-frontend-1.licensify.dev.gov.uk
        hostgroups   all
        parents      vpn_gateway_licensify

}
vagrant@monitoring-1:/etc/icinga/conf.d$ cat icinga_host_vpn_gateway_licensify.cfg
define host {
        use          generic-host
        alias        vpn_gateway_licensify
        address      10.5.0.1
        display_name monitoring-1.management.dev.gov.uk
        host_name    vpn_gateway_licensify
        hostgroups   all
}
vagrant@monitoring-1:/etc/icinga/conf.d$ cat icinga_host_vpn_gateway_licensify/check_ping_vpn_gateway_licensify.cfg
define service {
        use                       govuk_high_priority
        host_name                 vpn_gateway_licensify
        service_description       unable to ping
        check_command             check_ping!100.0,20%!500.0,60%
        notification_period       24x7
}
```


alexmuller:
I really really like this. The commit history is easy to review and I think it's almost entirely correct. I'm going to go through and put a couple of inline comments in but there's nothing major.

> The gateway hosts check use `host_name` parameter to know what to ping - does it do a DNS lookup, or use the underlying host config (which contains the IP address) to ping the correct thing?

I don't know the answer to this question.

modules/icinga/templates/host.erb
```diff
@@ -4,5 +4,7 @@ define host {
         address      <%= @address %>
         display_name <%= @display_name %>
         host_name    <%= @hostalias %>
-        hostgroups   all
+        hostgroups   all<% if @parents %>
+        parents      <%= @parents %>
+        <% end %>
```
alexmuller: This could be more clearly written as:

```
hostgroups   all
<%- if @parents -%>
parents      <%= @parents %>
<%- end -%>
```

The hyphens on the edge of the ERB blocks ensure that the lines containing the ERB don't show up in the file.
```

**Comment by rjw1** [comment]
it will use the ip address you set.

File: modules/monitoring/manifests/network_checks.pp [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+# Sets up vDC gateways as separate hosts so they can become parents of child
+# hosts dependent on the state of their gateway.
+#
+define monitoring::network_checks (
+  $address,
+) {
+    icinga::host { $title:
+      host_name => $title,
+      hostalias => $title,
+      address   => $address,
```
alexmuller: We need to pass in the `display_name` parameter here otherwise it will default to the `fqdn_short` of the monitoring machine (I'm becoming less sure about the `display_name` change as time goes on, still tempted to revert it).

modules/monitoring/manifests/checks/vpn_tunnels.pp
```diff
@@ -1,29 +0,0 @@
...
-# === Parameters
-#
-# [*enabled*]
-#   Should we monitor the VPN tunnels?
-#
-# [*endpoints*]
-#   Hash of the endpoints to monitor
-#
-class monitoring::checks::vpn_tunnels (
-  $enabled = true,
```
alexmuller: There's a bit of hieradata in the staging and production YAML files which could do with tidying up if this is being removed.

modules/icinga/manifests/host.pp
```diff
@@ -1,10 +1,39 @@
...
+#  [*hostalias*]
+#    The title of the host within Icinga, usually `$::fqdn`.
+#
+#
+#  [*address*]
+#    The IP address which is used to report into Icinga, usually
+#    `$::ipaddress`.
+#
+#  [*use*]
+#    FIXME: Cannot find documentation on this parameter for host definitions.
```
alexmuller: The `use` parameter in Icinga is to do with [which object it should inherit from](http://docs.icinga.org/latest/en/objectinheritance.html); `generic-host` is a host provided by the Icinga package which lives in `/etc/icinga/conf.d/generic-host_nagios2.cfg` and specifies a bunch of defaults.

modules/monitoring/manifests/network_checks.pp
```diff
@@ -0,0 +1,22 @@
+# == Class: monitoring::network_checks
+#
+# Sets up vDC gateways as separate hosts so they can become parents of child
+# hosts dependent on the state of their gateway.
+#
```
mattbostock: The description here mentions vDC gateways but the name of the defined type is much more general. Perhaps this could be renamed `monitoring::vpn_gateway` (singular) to make its intent clearer?

modules/icinga/manifests/client.pp
```diff
@@ -36,10 +36,19 @@
...
+      address      => $::ipaddress,
+      display_name => $::fqdn_short,
+      parents      => "vpn_gateway_${::vdc}",
+    }
+  } else {
+    @@icinga::host { $::fqdn:
+      hostalias    => $::fqdn,
+      address      => $::ipaddress,
+      display_name => $::fqdn_short,
+    }
```
mattbostock: Rather than repeating the `icinga::host` block twice, it would be preferable to always set a `parents` parameter but leave it empty or undefined for EFG and Licensify DR hosts.

modules/monitoring/manifests/network_checks.pp
surminus: @mattbostock, I think the description should be updated, I set it as "network_checks" in the (unlikely) event there were other network devices we may want to monitor in the future.

mattbostock: Yep, that'd work :+1:

alexmuller:
This looks awesome. I think we should try it in integration :+1:

surminus:
The commits are a bit incorrect and my tests are failing, when that's fixed I'll merge and test.

PR #3986, opened 2016-01-22
Add a param to enable check_support_default_queue_size for the support app

deanwilson:
This functionality is never used in staging but is used in integration
and production so we want to gate its deployment by environment. Defaults to
true so only stage should see a change.


alexmuller:
@deanwilson: Merge whenever :+1:

PR #3989, opened 2016-01-25
Update parameters for Bouncer Jenkins job

alexmuller:


alexmuller:
Actually, I'm still not sure this job is safe to include on staging. If somebody left the `HOSTNAMES` variable blank staging would grab the production hostname list and try to configure Fastly with it.

alexmuller:
I've removed the more controversial commit 87a37701e833b4b9f42c5732e760755eb62a6ac0 from this pull request so I think this is good to be reviewed now.

mattbostock:
:+1: Merge when green.

PR #3990, opened 2016-01-26
Add a jenkins job to publish existing content

ssatish:
Add a jenkins job to publish hardcoded content which is
currently on the site to content store.
Add job to envs - integration, staging and production.


bradwright:
@ssatish can you please fix the commit message length? We try to keep the first line to 72 characters so it doesn't wrap.

ssatish:
@bradwright fixed comment

PR #3991, opened 2016-01-26
Remove content register

elliotcm:
DB will need to be deleted manually.


alexmuller:
This is really cool, thanks @elliotcm. Feel free to click merge whenever and we'll try it on integration :+1:

Just to reiterate the things you mentioned:

- There will need to be more changes once this is deployed to remove the leftover stuff
- The database will need to be dropped manually

PR #3992, opened 2016-01-26
Jenkins job to publish dept. projects content

ssatish:


bradwright:
This appears to be based off #3990. Are you planning on closing that?

Also the PR description needs work as this change is now bigger than just the giturl change.

ssatish:
@bradwright. I plan to close #3990. Thats why I pinged you to figure out what your preference would be. This PR incorporates both changes. 
Let me know what you want me to do.

bradwright:
@ssatish let's do this one as it's actually right, but squash the commits (they should be one change, to add the job).

ssatish:
@bradwright They are two different commits semantically.

ssatish:
First one hanldes publishing existing content. And second one is fixing a url bug for two different tasks.
I can still squash commits if you prefer

bradwright:
In that case you should change the other commit to describe the change, eg `Fix Git remote for existing content job`

ssatish:
@bradwright done.

bradwright:
Thanks! :+1:

PR #3993, opened 2016-01-26
Remove rabbitmq user and permissions in the right order.

elliotcm:
Puppet will loop forever trying to remove permissions for a user which is already gone.


alexmuller:
Looks right to me :+1:

PR #3994, opened 2016-01-26
Use publishing-api token on whitehall frontend.

danielroseman:
The deploy scripts run the publish_special_routes task on both admin and frontend machines, so the token needs to be present in both cases.


bradwright:
Ideally this would be a different token (so we can separate privileges in the future), but this will do for now.

PR #3995, opened 2016-01-26
Revert "Remove content register"

elliotcm:
Reverts alphagov/govuk-puppet#3991


elliotcm:
Needs a more granular revert.

PR #3996, opened 2016-01-26
Revert remove content register

elliotcm:
Turns out we still need this.


alexmuller:
On the plus side, we've fixed bugs in the `ensure => absent` stuff so next time it will be more reliable :+1:

PR #3997, opened 2016-01-26
Ensure service is stopped before removing script

alexmuller:
The ordering arrow (`->`) that we added previously will apply changes in this order:

1. `service` will schedule the stopping of the service. The scheduling will cause the service to be stopped at the end of this Puppet run.
2. `file` resource will remove the init file.
3. The stopped of the service that was scheduled in (1) will run and fail because the file resource has been removed.

Instead of doing that, we can run an `exec` to stop the service manually (but only if it's running) and then remove the file after that command has executed.


elliotcm:
:+1:

PR #3999, opened 2016-01-26
Upgrade rsyslog module from 2.2.1 to 2.3.0

alexmuller:
It's good to keep our dependencies up-to-date.

https://github.com/saz/puppet-rsyslog/compare/v2.2.1...v2.3.0

I can't see anything too scary in the diff but note that I haven't tested this pull request using [vagrant-govuk](https://github.com/alphagov/vagrant-govuk) yet. Just opening it so people can see.


mattbostock:
The diff is much easier to review [without whitespace changes](https://github.com/saz/puppet-rsyslog/compare/v2.2.1...v2.3.0?w=1). Merge when ready :+1:

alexmuller:
In integration our syslog from various machines is being collected in Kibana so I think this is working as expected.

PR #4001, opened 2016-01-27
Refactor govuk_rbenv

alexmuller:
Currently we have two classes which are included on different machines. `govuk_rbenv` installs a basic set of rubies and `govuk_rbenv::all` installs a more complete set.

Instead of doing this with multiple classes I think it's more clear to this by configuring a single class.

`govuk_rbenv` with parameter `rubies: all` has the same behaviour as including `govuk_rbenv::all`.

I've moved this to YAML because it's machine configuration. I appreciate that this means we're no longer configuring the rubies in `govuk::node::s_ruby_app_server`, but I think that class is broken and should be deprecated and eventually removed (for example, why on earth are we including `mysql::client` on the draft cache machines?

Functionally this pull request should have no change.


mattbostock:
It feels like this PR is doing two things:

- deprecating `s_ruby_app_server` (and using Hiera data to specify which machines are Ruby app servers)
- refactoring `govuk_rbenv`

I agree that `s_ruby_app_server` is probably doing too much but I think the current way of configuring rbenv is cleaner in that we're relying on including a class to determine which Rubies are installed rather than a string parameter that is effectively an enum. 

A smaller change might be to determine which app servers need the additional packages provided by `s_ruby_app_server` and split those out so that `s_ruby_app_server` could be removed.

Can you split 31fdc34 into a separate PR?

alexmuller:
> Can you split 31fdc34 into a separate PR?

Sure, I've opened #4013 and #4014 so that we can discuss the more interesting change here.

I think this approach is more clear because the versions of Ruby that we want to install on a machine don't need to be specified in a Puppet class.

As bob pointed out yesterday, I think we should be aiming to put a list of Ruby versions in hieradata so that not every machine needs to include the same versions of Ruby (and importantly doesn't include Rubies it doesn't need). I don't think the basic/all distinction should stick around forever, but I think it's a better solution while we still have improvements to make to our hiera setup (for example by doing merging of arrays).

mattbostock:
I think the changes that you split out were useful improvements. For this PR, I find the original code is easier to understand.

alexmuller:
I think the new code is nicer, but fair enough! :)

PR #4003, opened 2016-01-27
Add notification_period to icinga::host

alexmuller:
We're using `icinga::host` to set up hosts for our VPN gateways, but we don't want the alerting for these to be as noisy as for normal hosts.

This commit allows us to set a notification_period on an Icinga host and makes sure that the VPN gateways only alert in office hours.


mattbostock:
:+1: Merge when ready.

PR #4004, opened 2016-01-27
Purge dictionary packages

alexmuller:
These packages were added years ago with the following commit messages:

- Passphrase entropy checkign needs a dictionary.
- Add in missing dictionary pkg

However, dictionaries-common and miscfiles aren't installed on the dev VM so applications can't need them anymore.

wbritish-small is on the dev VM, but I would be surprised if we still needed it for anything.

### DIFF
--- a/modules/govuk/manifests/node/s_development.pp
+++ b/modules/govuk/manifests/node/s_development.pp
@@ -149,7 +149,7 @@ class govuk::node::s_development (
 
   package {
     'sqlite3':        ensure => 'installed'; # gds-sso uses sqlite3 to run its test suite
-    'wbritish-small': ensure => installed;
+    'wbritish-small': ensure => purged; # FIXME: Remove once purged from most dev VMs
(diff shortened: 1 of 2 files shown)


deanwilson:
Looks fine. Merge away.

bradwright:
Looks like EFG is broken as a result of this:

    An Errno::ENOENT occurred in password_expired#update: 
    
      No such file or directory @ rb_sysopen - /usr/share/dict/words 
      lib/devise/models/strengthened.rb:15:in `strong_enough_password?'

bradwright:
I've suggested to the EFG team that they follow https://github.com/alphagov/signonotron2/pull/122 as a guide to migrating - I'll see what they come back with.

alexmuller:
Sorry :see_no_evil:

deanwilson:
On the plus side we can add them to the manifest for that application so we know where the requirement comes from rather than have it more 'free floating'

bradwright:
They've changing the app, no problems :+1:

PR #4006, opened 2016-01-27
Move non-creds from deployment hieradata

alexmuller:
These are stored in the deployment repo, but they're only hiera lookups so they can actually go in here.


deanwilson:
Looks fine. Merge when ready.

mattbostock:
:+1: Merge when ready.

PR #4008, opened 2016-01-27
Warn the user if it’s been > 3 days since a puppet run

tuzz:
I wrote a little script that warns the user if it's been more than 3 days since they last ran puppet. I'm not too familiar with puppet. Please could someone check that I've done things correctly, e.g. only including this module for use on the dev vm.

### DIFF
--- a/modules/govuk/manifests/node/s_development.pp
+++ b/modules/govuk/manifests/node/s_development.pp
@@ -25,6 +25,7 @@ class govuk::node::s_development (
   include govuk_rabbitmq
   include redis
   include tmpreaper
+  include last_puppet_run
 
   include govuk::python
   include govuk::testing_tools
--- a/modules/govuk/manifests/node/s_development.pp
+++ b/modules/govuk/manifests/node/s_development.pp
@@ -16,16 +16,16 @@ class govuk::node::s_development (
   include golang
   include govuk_apt::disable_pipelining
   include govuk_mysql::libdev
+  include govuk_rabbitmq
   include hosts::development
   include imagemagick
+  include last_puppet_run
   include memcached
   include mongodb::server
   include mysql::client
   include nodejs
-  include govuk_rabbitmq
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/node/s_development.pp
```diff
@@ -25,6 +25,7 @@
   include govuk_rabbitmq
   include redis
   include tmpreaper
+  include last_puppet_run
```
alexmuller: Please could we keep this list alphabetised?

mattbostock: Can you sort the include alphabetically?

mattbostock:
:+1: Implementation looks fine but I don't use the dev VM frequently enough to comment on the utility/intention.

modules/last_puppet_run/files/etc/profile.d/last_puppet_run.sh
```diff
@@ -0,0 +1,13 @@
...
+last_run=`sudo stat -c %Y /var/lib/puppet/state/last_run_report.yaml`
+time_now=`date +%s`
+difference=$((time_now - last_run))
+three_days=259200
+
+if [ "$difference" -gt "$three_days" ]; then
+  yellow="\e[0;33m"
+  reset="\e[0m"
+
+  echo -e "${yellow}It has been more than 3 days since you last ran govuk_puppet.${reset}"
```
fofr: Could it say the number of days rather than "more than 3"?
(Friday to Monday is more than 3 days)

tuzz: It could do, but my bash-fu isn't particularly strong.

Perhaps this is ok as a first pass?

fofr: Sure :+1:

PR #4009, opened 2016-01-27
Update Staging daily cron hour

surminus:
Interested to hear if changing the time of cron daily will cause any other problems elsewhere; it may also be worth setting this to late in the day/first thing in the morning in Integration as some of these tasks are missed due to the servers being shut down.

### DIFF
--- a/hieradata/staging.yaml
+++ b/hieradata/staging.yaml
@@ -3,6 +3,8 @@ app_domain: 'staging.publishing.service.gov.uk'
 
 base::supported_kernel::enabled: true
 
+cron::daily_hour: 6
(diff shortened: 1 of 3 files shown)


mattbostock:
Why is 4am better? It looks like the sync job will still be running at that time, e.g.:
https://deploy.publishing.service.gov.uk/job/Copy_Data_to_Staging/195/

surminus:
@mattbostock You're completely right, I got that wrong, maintenance tasks need to be later! Will amend.

alexmuller:
Looks good to me :+1:

PR #4010, opened 2016-01-27
Update puppetlabs_spec_helper to 1.0.1

deanwilson:


mattbostock:
@deanwilson Can you please say why you're upgrading it in your commit message?

deanwilson:
No single reason. Upgrading because we're very behind and bug fixes are good.

mattbostock:
Thanks @deanwilson - even if there's no specific reason, it's useful to say why in the commit message so that others aren't left wondering if something breaks or behaviour changes.

deanwilson:
Does mean I'm clear to merge?

mattbostock:
Sorry - I should have been clearer. I was suggesting that you say in the commit message that you're upgrading just to keep up-to-date.

mattbostock:
Superseded by #4012. Thanks @deanwilson

PR #4012, opened 2016-01-28
Update puppetlabs_spec_helper to version 1.0.1

deanwilson:
Old releases can have bugs that newer software may have fixes for so
it is good to update software you use. We have no specific reason
to pick this version to upgrade to, this is more a general maintenance
update.

The diff in versions is here
https://github.com/puppetlabs/puppetlabs_spec_helper/compare/0.4.1...1.0.1


rjw1:
:+1: merge when ready

PR #4013, opened 2016-01-28
Stop blacklisting `s_test` spec

alexmuller:
The manifest that would try to be tested was removed in aab2c3b318b1df7f2a1f287edca0b468384c36e1.


mattbostock:
:+1:

PR #4018, opened 2016-01-28
Add new app share-sale-publisher

mathildathompson:
@alexmuller: we have made changes so we are no longer adding a new machine for this app. This app needs to be externally accessible - does anything need to be done for this? Cheers

### DIFF
--- a/hieradata/class/backend.yaml
+++ b/hieradata/class/backend.yaml
@@ -36,6 +36,7 @@ govuk::node::s_base::apps:
   - release
   - search_admin
   - service_manual_publisher
+  - share_sale_publisher
   - short_url_manager
   - sidekiq_monitoring
   - signon
--- a/modules/govuk/manifests/apps/share_sale_publisher.pp
+++ b/modules/govuk/manifests/apps/share_sale_publisher.pp
@@ -1,0 +1,43 @@
+# == Class: govuk::apps::share_sale_publisher
+#
+# App to publish shares sale.
+#
+# === Parameters
+#
+# [*port*]
+#   The port that publishing API is served on.
+#   Default: 3119
+#
+# [*publishing_api_bearer_token*]
+#   The bearer token to use when communicating with Publishing API.
+#   Default: undef
+#
+
+class govuk::apps::share_sale_publisher(
+  $port = 3119,
+  $enabled = false,
+  $publishing_api_bearer_token = undef,
+) {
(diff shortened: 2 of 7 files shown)


modules/govuk/manifests/apps/share_sale_publisher.pp
```diff
@@ -0,0 +1,29 @@
+# FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
```
alexmuller: Could we add some class documentation here instead of a `FIXME`? There are some examples in the other Puppet manifests in this directory.

hieradata/class/share_sale_publisher.yml
```diff
@@ -0,0 +1,4 @@
+---
+
+govuk::node::s_base::apps:
+  - share_sale_publisher
```
alexmuller: Are we really adding new machines just for this app (that question isn't necessarily directed at you @mathildathompson)? If so, should the machines be named `share-sale-publisher-n.backend`? This isn't something we've done before. There will also need to be a node definition added (`modules/govuk/manifests/node/s_foo.pp`) and host entries for these machines (hieradata).

bradwright: We don't need new machines for these apps - they can go on the backend machines with other publishing tools.

bradwright:
@mathildathompson you need to add `share_sale_publisher` to `govuk::node::s_base_apps` on the `backend` machines too: https://github.com/alphagov/govuk-puppet/blob/master/hieradata/class/backend.yaml#L12

mathildathompson:
@bradwright done.

bradwright:
@alphagov/team-infrastructure-admins anything else needed?

hieradata/class/backend.yaml
```diff
@@ -36,6 +36,7 @@ govuk::node::s_base::apps:
   - release
   - search_admin
   - service_manual_publisher
+  - share-sale-publisher
```
bradwright: This needs to match the Puppet class name, eg `share_sale_publisher` (you're using hyphens rather than underscores)

alexmuller:
I can't think of anything more that's needed here except for @bradwright's comment :+1:

mathildathompson:
@alexmuller should be ready now :-)

PR #4019, opened 2016-01-28
Remove Jenkins job to publish existing content. 

ssatish:
Since the app has gone live and content store already has the content. Going ahead, we use the
github publishing mechanism to change. Also, we don't want to risk accidently running this job.


alexmuller:
This class needs to be removed from the 3 places it's included when it's removed:

```
$ ack departmental_projects_publish_existing_content
hieradata/integration.yaml
90:  - govuk_jenkins::job::departmental_projects_publish_existing_content

hieradata/production.yaml
94:  - govuk_jenkins::job::departmental_projects_publish_existing_content

hieradata/staging.yaml
33:  - govuk_jenkins::job::departmental_projects_publish_existing_content
```

PR #4023, opened 2016-01-29
Signon: Enable new-style logging

boffbowsh:
This updates the log collection to collect the application logs from
stdout and stderr instead of reading from logfiles that the app creates
itself.

To be deployed after https://github.com/alphagov/signonotron2/pull/464

### DIFF
--- a/modules/govuk/manifests/apps/signon.pp
+++ b/modules/govuk/manifests/apps/signon.pp
@@ -27,14 +27,14 @@ class govuk::apps::signon(
   $app_name = 'signon'
 
   govuk::app { $app_name:
-    app_type           => 'rack',
-    port               => $port,
-    vhost_ssl_only     => true,
-    health_check_path  => '/users/sign_in',
-    legacy_logging     => false,
-    vhost_aliases      => ['signonotron'],
-    asset_pipeline     => true,
-    deny_framing       => true,
+    app_type          => 'rack',
+    port              => $port,
+    vhost_ssl_only    => true,
+    health_check_path => '/users/sign_in',
+    legacy_logging    => false,
+    vhost_aliases     => ['signonotron'],
+    asset_pipeline    => true,
+    deny_framing      => true,
   }
 
   Govuk::App::Envvar {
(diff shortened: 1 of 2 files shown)


surminus:
:+1: merge when green

modules/govuk/manifests/apps/signon.pp
```diff
@@ -37,7 +37,7 @@
     port               => $port,
     vhost_ssl_only     => true,
     health_check_path  => '/users/sign_in',
-    log_format_is_json => true,
+    legacy_logging     => false,
     vhost_aliases      => ['signonotron'],
     logstream          => $ensure_logstream,
```
alexmuller: For reference, this param (set in hieradata) was [the one @bradwright was referring to](https://github.com/alphagov/signonotron2/pull/464#issuecomment-177001002). Do we want to set it to true?

alexmuller:
There's lint on this pull request which I think is just the fat arrows not lining up.

modules/govuk/manifests/apps/signon.pp
bradwright: I think we do.

boffbowsh:
Fixed

modules/govuk/manifests/apps/signon.pp
boffbowsh: Should we explicitly set that to true here, or just in hierdata/

alexmuller: The `logstream` param on `govuk::app` defaults to true so I think we can delete the param from this class and from hieradata entirely.

boffbowsh: Done (will remove from hieradata when this is deployed)

alexmuller:
The change looks good to me but the commit to fix arrow alignment also does a whole lot more (sorry to nitpick commit history)

boffbowsh:
@alexmuller fixed I am bad at rebasing

alexmuller:
Looks like Puppet to me, merge at will :+1:

PR #3958, opened 2016-01-18
Remove users

mattbostock:
These users' manifests were removed in https://github.com/alphagov/govuk-puppet/pull/3951 but I omitted to remove them from Hiera data.

PR #4024, opened 2016-02-01
Add new SSH key for tijmenbrommet

tijmenb:
My account was removed in #3951 because I was on leave and could not rotate my SSH key. This commit adds me again with a freshly generated key.


bradwright:
You also need to undo: https://github.com/alphagov/govuk-puppet/pull/3958/files

tijmenb:
@bradwright thanks, missed that. Updated.

PR #4026, opened 2016-02-01
Update daily hour for maintenance

surminus:
While machines are meant to boot up at 04:30, this seeks to ensure that daily tasks are always run if the machines take a while to come up.


MatMoore:
:+1:

bradwright:
Are these still splayed (eg random times across the 6am hour)? We've had issues in the past with IO due to many machines doing the same thing at once.

surminus:
@bradwright Yup: https://github.com/alphagov/govuk-puppet/blob/master/modules/cron/manifests/init.pp#L40-L43

PR #4028, opened 2016-02-01
Move seed-crawler cron to cron.daily

MatMoore:
Cron.daily runs first thing in the morning, whereas before this ran at 6pm.

We have an alert to ensure this has been run in the last 24h which is
currently never be the case on Mondays, because integration is down over the
weekend.

### DIFF
--- a/modules/govuk_crawler/manifests/init.pp
+++ b/modules/govuk_crawler/manifests/init.pp
@@ -182,14 +182,11 @@ class govuk_crawler(
     }
   }
 
-  cron { 'seed-crawler':
-    ensure      => $seed_ensure,
-    user        => $crawler_user,
-    hour        => 18,
-    minute      => 0,
-    environment => ['MAILTO=""'],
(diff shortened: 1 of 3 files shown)


boffbowsh:
LGTM but will let an infra folk check and merge.

deanwilson:
It means a change to the format of the cronjob but in general I'd like to move all our cronjobs from the cron type to a file under cron.d rather than cron.daily. Cron entries in cron.d have the exact time they run and the user they run as inside the file so it's possible to know pretty much everything about it from a single file. Would you mind resubmitting this as a cron.d job please?

https://help.ubuntu.com/community/CronHowto#Advanced_Crontab

Thank you for taking the first step towards this.

MatMoore:
@deanwilson no problem, I'll do that and resubmit.

MatMoore:
@boffbowsh @deanwilson I've changed cron.daily to cron.d now, and verified it gets picked up ok in vagrant.

deanwilson:
That's excellent, thank you @MatMoore. Are you OK with me to merge this now?

MatMoore:
Yep that's fine with me.

PR #4031, opened 2016-02-01
Increase Postgres backup partitions

surminus:
In both Production and Staging the partitions for Postgres backups are right on the edge of filling up in space, and presumably will continue to rise as time goes by.

This change seeks to double the size of the disks so we don't have to worry about this for a while.

Related to https://github.gds/gds/govuk-provisioning/pull/515


alexmuller:
This looks correct to me but I don't know how smoothly Puppet will make this change. Might require some devops :+1:

PR #4032, opened 2016-02-01
Tidy .gitignore

mattbostock:
The `.gitignore` file contains a number of files or directories that
are more appropriate in a user's global gitignore rather than in this
project-specific `.gitignore`.

For example, we are ignoring files generated by:

- bundler, depending on your configuration
- emacs
- geppeto (a Puppet IDE)
- rubygems
- IntelliJ
- `.DS_Store` files created by OS X

...and so on. Rather than clutter `.gitignore` with files specific to
each editor, leave it to developers to configure their global
`.gitignore`. This makes it easier for these files to be accidentally
committed, but then they could just as easily accidentally be committed
other repositories worked on by those users and maintaining `.gitignore`
files across all our repositories in this way is undesirable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,6 @@
 .rspec
 # Written by Cap and read by puppet_config_version
 REVISION
-.vagrant
+/development/.vagrant
 /development/Vagrantfile.localconfig
 /gpg/random_seed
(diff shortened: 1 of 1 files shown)


jamiecobbett:
Could we move these things to a default global .gitignore that gets generated on dev VMs?

mattbostock:
@jamiecobbett I think that's a great idea, I'll look into that.

mattbostock:
Closing while I look at adding these to the dev VM.

PR #4034, opened 2016-02-01
Extract govuk::logstream and dependent classes

mattbostock:
Extract the `govuk::logstream` defined type and the related
`govuk::tagalog` class into a new module called `govuk_logging`. This is
to break up the monolithic `govuk` module.

This will likely cause all of the logstream services we have configured in Upstart to restart, which should not cause any issues.


mattbostock:
_bump_

rjw1:
looks good. merge when your ready.

PR #4035, opened 2016-02-01
Add the puppetlint trailing comma check

deanwilson:
In PR review missing trailing commas are raised and so rather
than having a person look for them we should let the computer do it.

This PR also bulk fixes the current missing commas.


benlovell:
Did you mean puppet lint, not pupptline?

alexmuller:
Looks good to me. Does anyone object to this being added to the build?

Gemfile
```diff
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 gem "rake"
 gem "puppet-syntax", '2.1.0'
 gem "puppet-lint"
+gem 'puppet-lint-trailing_comma-check', :require => false
```
mattbostock: Nice use of `require`, hadn't seen that before :+1:

mattbostock:
:+1: This is great. It would be good to say in the commit message why we like to use trailing commas.

alexmuller:
I've updated this so if tests pass it should be merged :+1:

mattbostock:
:tada:

PR #4036, opened 2016-02-01
Update apt module

alexmuller:
In order to keep our dependencies up to date and to make it easier to install a newer version of the PostgreSQL module, we should update the apt module we use.

https://github.com/puppetlabs/puppetlabs-apt/compare/1.7.0...1.8.0

There are no breaking changes mentioned in the changelog for this release:

https://github.com/puppetlabs/puppetlabs-apt/blob/master/CHANGELOG.md#2015-03-17---supported-release-180

One change in behaviour is that the module now throws warnings when using a short key, so I've modified all of the places where we use keys with apt sources to make them full-length.


deanwilson:
Looks fine to me. Merge away.

PR #4037, opened 2016-02-01
Change the default alert threshold for application resident memory usage

mattbostock:
The majority of GOV.UK applications are using far less memory than the
default thresholds of 2GB for a warning and 3GB for a critical alert.
Currently, many GOV.UK applications could use three to four times their
usual memory and we would not be alerted.

This graph shows the maximum memory usage over 3 months from Nov 1st
2015 for all GOV.UK applications:

render

The dashed red line indicates the 51 percentile (representing a
majority) of memory usage by all applications, which averages 292 MB.
This graph shows the 51 percentile plotted over the same period:

render 1

This graph shows that only 18 out of the 78 applications we have
currently use more than 700MB of resident memory:

render 2

You can see all GOV.UK applications whose memory usage has never
exceeded 1GB over the 3-month time period in this graph:

render 3

As such, reduce the default threshold to 700MB for a warning and 800MB
for a critical alert. Where necessary, I have set appropriate thresholds
for the 18 high-memory applications to reflect their current memory
usage, using the current, maximum and minimal values in this graph to
determine the thresholds:

render 4

I'd expect that over time we can tweak down the 700MB/800MB default, but
for now this should be a reasonable improvement.

* * *

Best reviewed [without white-space changes](https://github.com/alphagov/govuk-puppet/compare/lower_memory_threshold?w=1), commit-by-commit.

I plan to tweak the memory alert threshold for Whitehall in a separate pull request because the current values are much higher than is necessary.


mattbostock:
Note I think I've conflated the concepts of 51 percentile and a 'majority' in the one of the above graphs but I don't think that affects my conclusion - most apps are using below 700MB.

I'm interested if people think the [`to_bytes()`](https://github.com/puppetlabs/puppetlabs-stdlib#to_bytes) function is preferable to the slightly hacky validation I've done in eb01cf6. Using `to_bytes()` would make comparison with Graphite more difficult as it doesn't use SI units.

mattbostock:
Rebased.

surminus:
Although the graphs say otherwise, I am wary of the additional noise this change could potentially create.

I want to fully understand the reasoning behind this change: it is so we would be alerted to an app, for example with a memory leak, quicker than we are currently? Is this something has caused an issue in the past in spotting problems with apps? 

Otherwise if an app does not have a drop in performance when it's using more memory, is the lower threshold required?

mattbostock:
The reason for this change is that currently our thresholds are so high that they would never alert. There's a good question as to whether we need alerting on individual application memory usage. Possibly, it's something that application teams themselves should watch (and we alert just on total machine memory usage).

bradwright:
> Possibly, it's something that application teams themselves should watch

I think teams should set a threshold they expect their app to consume, and then be alerted based on that. Without notifications people aren't going to proactively watch their app usage.

With that in mind, this PR does the right thing.

Long term the correct thing to do would be to have a moving window of memory alerts based on historical usage (eg if the app's usage spikes 20% above the average)

rjw1:
I think this is useful and may uncover issues we are currently hiding.
:+1: merge when ready. I would suggest at a useful time for 2ndline

alexmuller:
I don't think this should have been deployed to production so quickly. We are seeing alerts in all environments.

mattbostock:
@alexmuller: Fix is in https://github.com/alphagov/govuk-puppet/pull/4110

PR #4038, opened 2016-02-02
Log seed-crawler output to a file

MatMoore:
Currently the stdout of seed-crawler is just sent as part of the icinga
message. Unfortunately everything after the first newline is discarded,
and the first line is always the same, so when the alert fails we don't see
anything useful.


surminus:
:+1: merge when green

alexmuller:
Just to confirm I understand, this will send the output to syslog?

MatMoore:
@alexmuller Yep. Which in turn sends it to a file under /var/log/govuk.

PR #4039, opened 2016-02-02
Move backup hour back

surminus:
Due to the change in 0a82ba86220778486e9386dbc26ff094eeb6f2b0 the backup conflicted with the local backup task, so the backup wasn't pulled off the server.

I knew in my gut that the previous "simple change" would have knock-on effects somewhere. Perhaps it would be better to move the env-sync-and-backup task back, and add a lockfile to servers that are receiving the sync so they do not reboot (which is why I think the sync takes place at the time it does)?


surminus:
@deanwilson like thus?

deanwilson:
Looks good data type-wise.

deanwilson:
Merge away.

PR #4040, opened 2016-02-02
Restrict boto timeout/retry settings

MatMoore:
This should change CRITICAL alerts to UNKNOWNs when we have issues talking to AWS.

We use boto for SES in the check_aws_quote icinga plugin.

By default boto is very patient - failed requests can be retried 5 times, and
the http socket timeout is a minute.

The check itself times out at a minute so we want to shorten this so that it
can be handled by the plugin itself. We already have a catch all exception
handler that reports the result as UNKNOWN.

https://trello.com/c/WYsEieXt/12-improve-amazon-web-services-aws-simple-email-service-ses-quota-alert

### DIFF
--- a/modules/monitoring/files/etc/boto.cfg
+++ b/modules/monitoring/files/etc/boto.cfg
@@ -1,0 +1,3 @@
+[boto]
+http_socket_timeout = 5
+num_retries = 2
--- a/modules/monitoring/manifests/checks/ses.pp
+++ b/modules/monitoring/manifests/checks/ses.pp
@@ -36,6 +36,12 @@ class monitoring::checks::ses (
         provider => pip,
     }
 
+    file { '/etc/boto.cfg' :
+      ensure => present,
+      mode   => '0744',
+      source => 'puppet:///modules/monitoring/etc/boto.cfg'
+    }
+
     icinga::plugin { 'check_aws_quota':
         source  => 'puppet:///modules/monitoring/usr/lib/nagios/plugins/check_aws_quota',
         require => Package['boto']


mattbostock:
Can we set these configuration options in the [plugin itself](https://github.com/alphagov/govuk-puppet/blob/master/modules/monitoring/files/usr/lib/nagios/plugins/check_aws_quota) rather than in a global config file? We might have other use cases for boto in future which could require different settings.

mattbostock:
What do you think @MatMoore? Just bumping this as there's a few PRs open at the moment and I'm trying tie them off.

MatMoore:
Oops, missed your comment last week @mattbostock. If there's an easy way to do that I'm happy to make the change. When I last looked into it I couldn't find a way to pass the config to boto from the python code, only via the config files (which can be user specific as well as global).

I think these particular options aren't going to cause too much of a problem but if we keep it this way and plan to use boto elsewhere it might be good to extract that to do a different module in puppet to avoid surprises.

mattbostock:
I was curious so I tried this out https://github.com/alphagov/govuk-puppet/commit/62bab73d7dd5ec9b3a4145098feef7153b01b1f2. The script runs, though I haven't verified if the timeout actually applies.

mattbostock:
The API used by wal-e also exists in the boto test suite, so it looks legitimate:
https://github.com/boto/boto/blob/eeb7f237e0801bce9a87148c9364ba3069c6d6da/tests/integration/s3/test_https_cert_validation.py#L73-L75

MatMoore:
Happy to rework this. I'll close for now and raise a follow up PR.

mattbostock:
Superseded by https://github.com/alphagov/govuk-puppet/pull/4087

PR #4043, opened 2016-02-02
Errbit environment config for share-sale-publisher

bradwright:
Adds Errbit configuration so the app can report errors and deployments.

- [x] Merge deployment#878

### DIFF
--- a/modules/govuk/manifests/apps/share_sale_publisher.pp
+++ b/modules/govuk/manifests/apps/share_sale_publisher.pp
@@ -12,10 +12,20 @@
 #   The bearer token to use when communicating with Publishing API.
 #   Default: undef
 #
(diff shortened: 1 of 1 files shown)


mattbostock:
The build is failing because `errbit_environment_name` is mandatory and isn't set in Hiera's `development.yaml` for the dev VM.

bradwright:
@mattbostock turns out that's [globally defined](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/deploy/config.pp#L87) - will remove.

bradwright:
@mattbostock updated. Will see what the tests say.

bradwright:
@mattbostock ready for re-review.

PR #4047, opened 2016-02-02
Update logrotate conf

surminus:
This seeks to change a couple of issues with logs growing to a size where we are alerted to a disk space issue on the server. More detail in commits.


deanwilson:
Looks fine, and taught me about `su`, so merge away.

mattbostock:
:+1: Nice work :eggplant:

surminus:
I updated this because I realised that I was adding `su` for logs that are in `/var/log` which is incorrect.

PR #4048, opened 2016-02-03
Reinstate my previous key

benlovell:
I revoked my previous key (after SSH-gate, so it's safe!) as I was
returning my laptop for repair. Turns out everything was fine so I'm
reinstating this key for my normal work machine.

I've left my other key here incase I have to use that machine again.
Happy to remove it if that's a problem for anyone.


benlovell:
Thanks :+1:

PR #4052, opened 2016-02-03
Set sudo to false for travis.

rjw1:
`bundler` fails to install `hiera-eyaml` on the precise image the non
container travis infrastructure uses by setting false we get the trusty
image which works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: ruby
-sudo: true
+sudo: false
 rvm:
   - 1.9.3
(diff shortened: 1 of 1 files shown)


rjw1:
[build fail](https://travis-ci.org/rjw1/govuk-puppet/builds/106733384)
[build pass](https://travis-ci.org/rjw1/govuk-puppet/builds/106734115)

deanwilson:
Looks fine. Merge away.

PR #4054, opened 2016-02-03
Set up email alert monitoring in Icinga

tijmenb:
We want to verify that emails are sent for all [drug alerts and medical device alerts](https://www.gov.uk/drug-device-alerts).

This commit sets up a Jenkins job, which runs every hour. It also sets up a Icinga "passive alert", which will notify us if the job has not run successfully in the last hour and a half.

The jenkins job runs a script that actually checks if the emails have been sent: https://github.com/alphagov/email-alert-monitoring/pull/1

The code for the [tagging migration verifier](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/manifests/job/tagging_migration_check.pp) is used as an example for this PR.

We enable the alert only on production because only production sends out actual emails.

Trello: https://trello.com/c/GGGYpRXd

### DIFF
--- a/hieradata/production.yaml
+++ b/hieradata/production.yaml
@@ -96,6 +96,7 @@ govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::job::deploy_licensify
   - govuk_jenkins::job::deploy_puppet
   - govuk_jenkins::job::deploy_router_data
+  - govuk_jenkins::job::email_alert_check
   - govuk_jenkins::job::fetch_prototype_taxonomy_data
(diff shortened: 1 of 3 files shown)


MatMoore:
Looks good to me, but will let someone else look at it before merging.

tijmenb:
Pushed a new version that adds the correct environment variables that are needed by the script. @rboulton will add these to the relevant private repos.

Also changed the cron job to run on the 21st minute instead of on the hour, which may be a busy time for the servers.

mobaig:
LG

PR #4056, opened 2016-02-04
Add pre-command to asset-manager job

surminus:
This was added in cae7c5576ee030332103f18a6499a943bbcdbaa1, and should be added for the other job that runs on the server.


mattbostock:
:+1: Merge when ready.

PR #4057, opened 2016-02-04
Keep CDN logs for only 30 days

surminus:
**This change is raised as a direct result of a monitoring alert due to low disk space on logs-cdn-1**

We only need to keep CDN logs for 30 days at most (as confirmed by @daibach). This change updates the daily rotation period from 365 to 30 days to meet this demand.

This will change rotation from 365 daily to 30 daily for the following logs:

`cdn-assets.log`
`cdn-govuk.log`

As per the change @jennyd made in cf5a8e4e the `cdn-bouncer.log` are exempt from the daily rotation, and so are not affected by this change.

We also have quite a few logs within the `/mnt/logs_cdn/old` directory which, as the name suggests, are quite old:

paulmartin@production-logs-cdn-1:~$ ls /mnt/logs_cdn/old/ |wc -l
2068
paulmartin@production-logs-cdn-1:~$ du -sh /mnt/logs_cdn/old/
64G	/mnt/logs_cdn/old/

I would be interested to know what we should be doing with these logs as that'll help clear disk space if they are no longer needed.

### DIFF
--- a/modules/govuk_cdnlogs/spec/classes/govuk_cdnlogs_spec.rb
+++ b/modules/govuk_cdnlogs/spec/classes/govuk_cdnlogs_spec.rb
@@ -100,7 +100,7 @@ ruleset\(name="cdn-giraffe"\) \{
       end
       it 'should rotate daily' do
         is_expected.to contain_file('/etc/logrotate.d/cdnlogs')
-          .with_content(/rotate 365$/)
+          .with_content(/rotate 30$/)
       end
       it { is_expected.to contain_file('/etc/logrotate.cdn_logs_hourly.conf')
             .without_content(/rotate/) }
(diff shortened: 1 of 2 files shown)


mattbostock:
@surminus: We can delete the 'old' logs. I've confirmed this with recently with @daibach.

mattbostock:
Looks like the build failed because ci-slave-1 ran out of disk space.

:+1: Merge when green.

mattbostock:
Bouncer is configured to keep logs for 365 days (365 * 24): https://github.com/alphagov/govuk-puppet/blob/5a4d144ef1265f7cf96e5fbcd5a6427b6c56a39e/modules/govuk_cdnlogs/templates/etc/logrotate.cdn_logs_hourly.conf.erb#L7

Could we reduce that to 30 days also? /cc @jennyd

mattbostock:
Related: https://github.com/alphagov/govuk-puppet/pull/4084

PR #4064, opened 2016-02-05
Fix bundle command for email alert check

tijmenb:
Running `bundle install` doesn't seem to work, because it tries to install the gem to a location it doesn't have permission to. Setting the path fixes this.

This is used by all [other jenkins jobs](https://github.com/alphagov/govuk-puppet/blob/321d0449b2e582252c080649c93650d2d1146a2f/modules/govuk_jenkins/templates/jobs/deploy_router_data.yaml.erb#L20) that run `bundle install`.

Trello: https://trello.com/c/GGGYpRXd


surminus:
:+1: merge when green

PR #4067, opened 2016-02-08
Add Jenkins job to update CDN edge dictionaries

mattbostock:

We've added a script to update CDN [edge dictionaries][] at Fastly so
that we can block IP addresses at the edge without redeploying our VCL:
https://github.gds/gds/cdn-configs/pull/108

Add a Jenkins job which will update the edge dictionary using the data
in the cdn-configs repository.

[edge dictionaries]: https://www.fastly.com/blog/announcing-edge-dictionaries-make-faster-decisions-edge


surminus:
LGTM, merge when ready

PR #4068, opened 2016-02-08
Add missing Rabbitmq configuration for Rummager

bishboria:
The following changes were missing/in the wrong place in order for RabbitMQ and Rummager to work together.

### DIFF
--- a/hieradata/class/rabbitmq.yaml
+++ b/hieradata/class/rabbitmq.yaml
@@ -7,6 +7,7 @@ govuk::node::s_rabbitmq::apps_using_rabbitmq:
   - govuk_crawler_worker
   - panopticon
   - publishing_api
+  - rummager
   - stagecraft
 
 govuk::safe_to_reboot::can_reboot: 'careful'
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -77,6 +77,7 @@ govuk::node::s_development::apps:
   - 'router'
   - 'router_api'
   - 'rummager'
+  - 'rummager::rabbitmq'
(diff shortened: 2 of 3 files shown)


modules/govuk/manifests/apps/rummager.pp
```diff
@@ -98,10 +98,6 @@
     process_type   => 'publishing-api-document-indexer',
   }
 
-  govuk_rabbitmq::monitor_consumers {'rummager-publishing-api-document-indexer':
```
rboulton: I think we still want to monitor the consumer.  Was this meant to have been moved into the `rummager/rabbitmq.pp` file?

hieradata/class/rabbitmq.yaml
```diff
@@ -8,6 +8,7 @@ govuk::node::s_rabbitmq::apps_using_rabbitmq:
   - panopticon
   - publishing_api
   - stagecraft
+  - rummager
```
alexmuller: Nitpick, this needs to be bumped up one line to be alphabetical :)

rboulton:
Aren't there some test settings that need to be added too?  Or are those not actually needed?

alexmuller:
There was a bit of trouble with adding the test exchange because `govuk::apps::publishing_api::rabbitmq` also defines an exchange called `published_documents_test@/` and Puppet doesn't like duplicate resources named the same thing.

rboulton:
:+1: please merge if you're happy.

rboulton:
Actually, I know you're happy, so I'm merging myself.

PR #4070, opened 2016-02-09
Add CONTRIBUTING.md

alexmuller:
@rboulton pointed out that the Infrastructure team has not been good at communicating that we've moved to a self-merge approach for Puppet pull requests.

@deanwilson suggested that we do it because Puppet is a repo which has lots of people touching it, and it's nice to be able to control when your changes are merged (and therefore deployed).

This file seems like the best way to document that.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,16 @@
+# Contributing guidelines for GOV.UK's Puppet repository
+
+## Pull requests
+
+You can propose changes using a pull request.
+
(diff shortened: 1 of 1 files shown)


bradwright:
In favour of both the approach and the addition of the file. :+1:

mattbostock:
:+1:

rjw1:
This is a good start. :+1:
I wonder if we need to be explicit about our stance on contributions from outside GOV.UK and mention the coding in the open thing. 

Also possibly mention jenkins and travis which I should turn on again I suppose.

PR #4072, opened 2016-02-09
Fix monitoring user in development

alexmuller:
The monitoring user in development is set incorrectly. There's no point in parameterising this username because it can't be changed at the moment; it's used in a defined type.


mattbostock:
:+1: Merge when ready.

PR #4073, opened 2016-02-09
Backup processed CDN logs

alexmuller:
We're going to start deleting historical raw CDN logs, which means the processed files will be the only data we retain. We should back them up.

I've forced the `processed_data_dir` parameter to be in hieradata so that I can do a hiera lookup in `backup.yaml` and avoid some duplication.

https://trello.com/c/xbsYTau0/33-back-up-processed-cdn-logs

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -182,6 +182,8 @@ govuk::apps::email_alert_service::rabbitmq_hosts:
 
 govuk::apps::finder_frontend::enabled: true
 
+govuk::apps::govuk_cdn_logs_monitor::processed_data_dir: '/mnt/logs_cdn/data'
+
 govuk::apps::govuk_crawler_worker::airbrake_endpoint: "https://errbit.%{hiera('app_domain')}/notifier_api/v2/notices"
 govuk::apps::govuk_crawler_worker::airbrake_env: "%{hiera('govuk::deploy::config::errbit_environment_name')}"
 # FIXME: The API can be crawled when https://github.com/alphagov/govuk_crawler_worker/issues/97 is fixed.
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -79,6 +79,12 @@ backup::offsite::jobs:
     hour: 8,
     minute: 13,
     # No encryption because of size and sensitivity
+  'govuk-cdn-logs':
+    sources: "/data/backups/logs-cdn-1.management.%{hiera('app_domain')}/*"
+    destination: "rsync://%{hiera('backup::offsite::dest_host')}//srv/backup-cdn-logs/"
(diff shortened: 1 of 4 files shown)


hieradata/common.yaml
```diff
@@ -79,6 +79,13 @@ backup::offsite::jobs:
     hour: 8,
     minute: 13,
     # No encryption because of size and sensitivity
+  'govuk-cdn-logs':
```
mattbostock: Should this be `processed-cdn-logs` for consistency with the name above?

hieradata/common.yaml
```diff
@@ -79,6 +79,13 @@ backup::offsite::jobs:
...
+    sources: "/data/backups/logs-cdn-1.management.%{hiera('app_domain')}/*"
+    destination: "rsync://%{hiera('backup::offsite::dest_host')}//srv/backup-cdn-logs/"
+    weekday: 6
```
mattbostock: We don't backup our Graphite data daily because of the size of the data. Is there a good reason not to do daily backups of our processed CDN logs?

hieradata/common.yaml
```diff
@@ -79,6 +79,13 @@ backup::offsite::jobs:
     hour: 8,
     minute: 13,
     # No encryption because of size and sensitivity
+  'govuk-cdn-logs':
+    sources: "/data/backups/logs-cdn-1.management.%{hiera('app_domain')}/*"
+    destination: "rsync://%{hiera('backup::offsite::dest_host')}//srv/backup-cdn-logs/"
```
mattbostock: Should this path also mention 'processed'?

hieradata/common.yaml
mattbostock: On second thoughts probably not, since this will copy any directories from the logs-cdn-1 machine including the original path.

hieradata/common.yaml
mattbostock: Please disregard this comment.

hieradata/common.yaml
alexmuller: No, it's a good point. I'm not sure that the match for `"/data/backups/logs-cdn-1.management.%{hiera('app_domain')}/*"` is the best way of doing it. It could also be "/data/backups/*%{hiera('govuk::apps::govuk_cdn_logs_monitor::processed_data_dir')}". Not sure which is nicer.

mattbostock: I think it's fine as-is, it will make future changes easier.

rjw1:
@alexmuller where you going to make this daily?

alexmuller:
Have done, thanks for the nudge :)

rjw1:
:+1: merge when ready

PR #4075, opened 2016-02-09
Remove fail2ban.

deanwilson:
After a discussion we no longer want to run Fail2Ban.

https://trello.com/c/7lnrxg1a


surminus:
:eggplant:

deanwilson:
Force pushed new commit with more details/context from the requesting ticket.

mattbostock:
The cached module needs to be removed.

mattbostock:
:+1: Merge when ready.

PR #4076, opened 2016-02-09
Remove erroneous "minus" character from ERB

surminus:
This caused the logrotate configuration to incorrectly format like:

```
su deploy deploy daily
missingok
```

This fixes it. :person_frowning:


mattbostock:
:+1: Merge when green.

PR #4077, opened 2016-02-10
Upgrade nginx package

alexmuller:
This package was installed automatically by unattended-upgrades last night to patch CVEs CVE-2016-0742, CVE-2016-0746, CVE-2016-0747.

Puppet is failing to run correctly because it can't find the old version to install (and we wouldn't want it to anyway).

Paired with @MatMoore on this.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -880,7 +880,7 @@ monitoring::vpn_gateways::endpoints:
   vpn_gateway_router_dr:
     address: "%{hiera('environment_ip_prefix')}.9.1"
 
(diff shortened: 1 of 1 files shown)


rjw1:
do we want to just not specify anymore?
but merge this anyway. :+1:

mattbostock:
I agree, I don't think we should pin this anymore. We're using the official Ubuntu Apt repositories for this, and if nginx failed overnight because of a bad package we would failover to the mirrors, which I think is more desirable than running a vulnerable version.

alexmuller:
I would definitely entertain a pull request which does that :)

PR #4078, opened 2016-02-10
Allow mongodb secondaries to be rebooted overnight

MatMoore:
For now we don't allow the primaries, but the process is to
first step the primary down to secondary before rebooting.

The status check here is the same as the safe_reboot fabric command.

This replaces this PR, which was part reverted due to the status check not being complete https://github.com/alphagov/govuk-puppet/pull/3877

Paired with @alexmuller

### DIFF
--- a/hieradata/class/staging/api_mongo.yaml
+++ b/hieradata/class/staging/api_mongo.yaml
@@ -1,0 +1,3 @@
+---
+
+govuk::safe_to_reboot::can_reboot: 'overnight'
--- a/hieradata/class/staging/email_campaign_mongo.yaml
+++ b/hieradata/class/staging/email_campaign_mongo.yaml
@@ -1,0 +1,3 @@
+---
+
+govuk::safe_to_reboot::can_reboot: 'overnight'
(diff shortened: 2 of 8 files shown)


surminus:
This is great! :+1: I was going to ask about testing it but I guess the fabric task is a pretty good indicator that it works as required, though shame we can't test it on Integration, really. 

Due to the nature of this work I'd like someone else to review as well.

deanwilson:
This looks good both in code and as work that very much needed doing. Nice one.

alexmuller:
> shame we can't test it on Integration

We could roll this out on one cluster in integration to start with if that would make everyone happier :)

surminus:
It's the more that Integration machines turn off every night anyway, unless we updated the time Integration machines did updates & reboots to the middle of the day, showing full faith that it won't cause problems... :wink:

alexmuller:
Haha I forgot about that, good point!

mattbostock:
We could enable it in Integration and Staging first.

MatMoore:
I've moved the hieradata to staging so we can test this first. No point in adding it for integration since it will be off.

mattbostock:
Merge when green :+1:

PR #4079, opened 2016-02-10
Remove service-manual-publisher DISABLE_PUBLISHING feature flag

AndrewVos:
We're planning on going live with our guide content, so this feature
flag needs to be removed.

Related: https://github.com/alphagov/service-manual-publisher/pull/149

https://trello.com/c/3VCkqD3P/166-allow-publishing-guides-in-production

This was originally introduced here: https://github.com/alphagov/govuk-puppet/commit/42a45d8211a2a4f12f57d909ec260c2dd1196c1c


bradwright:
@AndrewVos just to confirm - when we merge this and Puppet goes live, the service will restart automatically and come up with publishing enabled. Are you okay with that, or did you want to deploy the app first?

AndrewVos:
@bradwright Yes that's perfectly fine (just confirmed with my team).

bradwright:
In that case. :+1: when you're ready.

PR #4080, opened 2016-02-10
Add env vars for travel-advice-publisher.

danielroseman:
This defines MONGODB_URI, using the existing mongodb_uri class with data
supplied in backend hieradata, and SECRET_KEY_BASE, which uses data in the
deployment hieradata.

### DIFF
--- a/hieradata/class/backend.yaml
+++ b/hieradata/class/backend.yaml
@@ -15,6 +15,12 @@ govuk::apps::share_sale_publisher::mongodb_nodes:
   - 'mongo-2.backend'
   - 'mongo-3.backend'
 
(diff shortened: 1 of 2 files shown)


hieradata/class/backend.yaml
```diff
@@ -15,6 +15,12 @@ govuk::apps::share_sale_publisher::mongodb_nodes:
   - 'mongo-2.backend'
   - 'mongo-3.backend'
 
+govuk::apps:travel_advice_publisher::mongodb_name: 'govuk_content_production'
```
bradwright: Missing `:` here between `apps` and `travel_advice_publisher`, which is causing a build failure.

mattbostock:
:+1: Merge when ready.

PR #4081, opened 2016-02-10
Allow overnight reboot of elasticsearch nodes

MatMoore:
The check is the same as the fabric task, but without the disable_reallocation part.
We think it will be simpler this way, although it might take longer for the cluster
to go green again after a single node is rebooted.

The other way we could do this is to disable reallocation in the /etc/unattended_reboot check, and then add something else to renable it after the machine boots up again.

https://trello.com/c/JD2ZFRcJ/13-write-script-to-check-if-elasticsearch-logs-elasticsearch-machines-can-be-rebooted

@rboulton can you review?

### DIFF
--- a/hieradata/class/staging/api_elasticsearch.yaml
+++ b/hieradata/class/staging/api_elasticsearch.yaml
@@ -1,0 +1,3 @@
+---
+
+govuk::safe_to_reboot::can_reboot: 'overnight'
--- a/hieradata/class/staging/elasticsearch.yaml
+++ b/hieradata/class/staging/elasticsearch.yaml
@@ -1,0 +1,3 @@
+---
+
+govuk::safe_to_reboot::can_reboot: 'overnight'
--- a/hieradata/class/staging/logs_elasticsearch.yaml
+++ b/hieradata/class/staging/logs_elasticsearch.yaml
@@ -1,0 +1,3 @@
+---
+
+govuk::safe_to_reboot::can_reboot: 'overnight'
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -408,6 +408,7 @@ govuk_sudo::sudo_conf:
 
 govuk_unattended_reboot::enabled: true
 govuk_unattended_reboot::mongodb::enabled: true
+govuk_unattended_reboot::elasticsearch::enabled: true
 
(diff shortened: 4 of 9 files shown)


surminus:
So pleased to see these PRs come in! :balloon:

mattbostock:
LGTM :+1:

As discussed in person it'd be nice to retain `disable_reallocation`, but I think this has value as-is.

MatMoore:
Similar to the mongo PR, I've restricted this to changing first so we can test it overnight.

MatMoore:
This may need to be rebased on top of PR #4101 and a flag added so this runs properly on development.

mattbostock:
Is there much value in the flag that determines whether the check script is installed? It wouldn't have any impact on functionality (since unattended reboots are disabled in development); managing the file in a dev environment seems like we're adding extra complexity with little benefit.

MatMoore:
@mattbostock this is a bit confusing, but the reason this came up was due to an implicit dependency between

* govuk_unattended_reboot
* and govuk_unattended_reboot::mongodb/elasticsearch

which are included from different places, because the latter two are only relevant for certain classes of hosts, so are included from the relevant module.

Without the flag on this, we always installed the checks on all environments, which caused the dev VMs to break. I think the behaviour here *should* be tied to govuk_unattended_reboot::enabled, and originally for #4101 we tried to make this explicit using a hieradata lookup, which unfortunately didn't work as we expected.

I'm sure there is a better way of structuring this whole thing, but I'm not sure what it is yet. If you'd prefer to rework this I would suggest doing it in a follow up PR as it affects both this and the mongodb one, and I expect somebody else would take over the ticket.

For now I've just rebased this on top of master and just brought this in line with the other PR so we don't cause more dev issues if this is merged.

rboulton:
This looks good to me. :+1:

The loss of `disable_reallocation` is a bit of a shame, but the complexity needed to implement it is best avoided, I think.  In mitigation, once we upgrade elasticsearch to version `1.7` or later, we gain a feature native to elasticsearch which allows us to delay the start of reallocation for a few minutes after node failure, which would be a good replacement for this: https://www.elastic.co/guide/en/elasticsearch/reference/1.7/delayed-allocation.html

PR #4082, opened 2016-02-10
Prompt for username, then password

mattbostock:

Prompt for the Fastly password after the username, consistent with
convention used by most login forms on the Internet.


surminus:
:+1: merge quando verde

mattbostock:
Muy bien, gracias.

PR #4083, opened 2016-02-10
Jenkins CDN jobs: Don't delete .git directory

mattbostock:

Configure the Jenkins GIT plugin not to delete the `.git` directory on
every build so that deploying the CDN configs or updating the CDN
dictionaries complete faster.

`wipe-workspace` is set to true by default. Instead, run `git-clean`
after `git-clone` to ensure that the workspace is clean.

See:
http://docs.openstack.org/infra/jenkins-job-builder/scm.html?highlight=force%20clone#scm.git


surminus:
:+1: merge when green

PR #4084, opened 2016-02-10
Keep Bouncer CDN logs for 30 days only

mattbostock:
These logs are rotated hourly, so 30 * 24 = 720 hours.

We changed our policy to retain other CDN logs for 30 days only in
edfe535 and this change makes the configuration for Bouncer the same in
that respect.


surminus:
:+1: merge when ready

Issue #1487, opened 2015-12-16
Add support for user expiry


alexmuller:
This pull request was [originally opened on GitHub Enterprise](https://github.gds/gds/puppet/pull/1487) on Wednesday 05 March 2014 at 11:52 GMT before GOV.UK's Puppet code was migrated to a public GitHub repository.

Issue #1402, opened 2015-12-16
Remove Ben Firshman


alexmuller:
This pull request was [originally opened on GitHub Enterprise](https://github.gds/gds/puppet/pull/1402) on Wednesday 12 February 2014 at 09:16 GMT before GOV.UK's Puppet code was migrated to a public GitHub repository.

PR #4088, opened 2016-02-11
Add support for user expiry

mattbostock:
After comments on #1402, adds support for user expiry so we can set a
user's expiry date when we learn their leaving date, rather than having
to push a release out on the day.

Credit goes to @philandstuff who originally raised this as #1487. I've added some validation for the date format.

### DIFF
--- a/modules/govuk/manifests/user.pp
+++ b/modules/govuk/manifests/user.pp
@@ -24,6 +24,11 @@
 # [*shell*]
 #   The user's login shell. Default: "/bin/bash"
 #
+# [*expiry*]
+#   The expiry date for this user. Must be provided in a zero-padded YYYY-MM-DD
+#   format - e.g. 2010-02-19. If you want to make sure the user account never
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/user.pp
```diff
@@ -68,9 +74,12 @@
     $membership = 'minimum'
   }
 
+  validate_re($expiry, '^(absent|\d{4}-\d{2}-\d{2})$')
```
surminus: Not a biggie, but it's a shame Puppet doesn't seem to have a function to validate a date specifically. I am not sure what the behaviour would be if an invalid date is inputted (like 2016-02-30).

mattbostock: It'd be nice, but complex to implement ourselves and I don't think there's anything readily available that we could use.

rjw1:
useful addition. merge when ready

philandstuff:
:whale:

PR #4090, opened 2016-02-11
Extract govuk::scripts into its own module

mattbostock:
In a bid to breakup the monolithic `govuk` module.


surminus:
:+1: looks good

PR #4096, opened 2016-02-11
Move files to their appropriate modules

mattbostock:
There's a few files in the `govuk` module that don't belong in that module. Move them to their rightful place.


surminus:
:+1: I think all of this make sense

PR #4097, opened 2016-02-11
Add unattended reboot check for lockfiles

MatMoore:
This is a first stab at https://trello.com/c/Z5T7xEcX/72-add-lockfile-functionality-to-nightly-reboots-have-important-jobs-write-release-this-lock after discussing with @alexmuller 

Checks that other jobs don't have active lockfiles before the unattended reboot can go ahead.

This assumes the jobs are already using setlock to acquire locks.

For now, this check is just going to immediately release the lock after acquiring it, to keep things simple. There's a short period of time when jobs could still be started after this check runs, but we think this is still an improvement.


surminus:
Really like the logic on this, easy to add and I think it'll work effectively to check across a bunch of different jobs :+1: 

I was wondering if the `setlock -n` check will work if the backup job is just `touch`ing an empty file rather than using `setlock` to create one as per https://github.com/alphagov/govuk-puppet/blob/321d0449b2e582252c080649c93650d2d1146a2f/modules/backup/files/etc/automysqlbackup/prebackup#L4

I think `setlock` should always be used regardless so perhaps just easiest to update the pre/post-command options.

MatMoore:
@surminus ah... yes, this won't work properly for the mysql backups because just touching the file is not enough (setlock -n will exit with 0 in that case). The man page says:

>        Here's the complete story: child is given a descriptor for a locked ofile pointing to the disk file named fn.  The lock disappears when this ofile is
>
>       o      closed by all the processes that have descriptors for it or
>
>       o      explicitly unlocked.

I think setlock should be used for this as well, but I'm not sure exactly how to implement it yet because automysqlbackup is an external module, and we just create the pre/post scripts. I'll see if I can take another look at it later today.

MatMoore:
Looks like we can customise the cron template via a parameter to automysqlbackup::backup, so that might be the best place to put a setlock in.
https://github.com/nextrevision/puppet-automysqlbackup/blob/master/manifests/backup.pp

mattbostock:
@MatMoore: I've changed the backup to use `setlock` but the mechanism is a hack. I've not tested it yet but would be interested in your opinion.

surminus:
Could we use `lockfile-create` and `lockfile-check` instead of `setlock`? http://linux.die.net/man/1/lockfile-create

MatMoore:
@surminus are those commands interoperable with setlock? If it works that would be preferable to using setlock in this way I think.

What I had in mind originally was customising the cron script generated by that module to use a setlock command rather than defining pre/postbackup scripts. 

```
$cron_template        = 'automysqlbackup/automysqlbackup.cron.erb',
```

I didn't look into it any further though - any idea if that would work?

mattbostock:
On second thoughts, I wonder if we can emulate the behaviour of setlock in bash by having bash open the lockfile for writing.

surminus:
@MatMoore not sure if they would work together, I think both sets of commands just put a write-lock on the specified file, could be worth testing.

mattbostock:
I wonder if `flock` would do what we want, might look into that.

surminus:
:bird:

tijmenb:
@mattbostock do you think this can be merged?

mattbostock:
@tijmenb My understand is that it won't work as-is and I'm afraid I didn't get chance to look into an alternative lock mechanism. This is definitely a change we want though.

tijmenb:
Allright, I'll close this until alternative lock mechanism has been looked at. Is there a ticket on your backlog?

mattbostock:
@tijmenb It's being tracked in the Second Line backlog: https://trello.com/c/Z5T7xEcX/72-add-lockfile-functionality-to-nightly-reboots-have-important-jobs-write-release-this-lock

PR #4099, opened 2016-02-12
Pin Puppet gems to correct versions

alexmuller:
In `puppet::package` we install specific versions of our dependencies. We should install the same versions in development as gems.

This commit downgrades facter from 2.x to 1.x to ensure that our tests run the same code as our other environments.


rjw1:
:+1: i think. its slightly odd that they had drifted
merge when ready.

mattbostock:
:eggplant:

PR #4100, opened 2016-02-15
Remove fail2ban package

surminus:
fail2ban puppet class was removed in c11b00ad231e0784c64c15f56f2272d7a442a356, but we did not ensure that the package would be stopped and removed.


rjw1:
merge when green :+1:

PR #4101, opened 2016-02-15
Include unattended reboot on dev but disable it

MatMoore:
There is an assumed dependency between unattended_reboot::mongo on unattended_reboot. Previously puppet was not running on development as we only included the mongo one from there. We include this automatically wherever we use mongo. 

This PR solves this by including unattended_reboot everywhere and disabling both classes on development.

update-notifier-common is now required everywhere, not just for monitoring clients.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -115,6 +115,7 @@ base::packages::packages:
   - 'tcpdump'
   - 'tmux'
   - 'tree'
+  - 'update-notifier-common'
   - 'unzip'
   - 'vim-nox'
   - 'xz-utils'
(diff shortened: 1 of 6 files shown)


surminus:
Were we going to disable this in Production for testing? So maybe put `govuk_unattended_reboot::mongodb::enabled: false` in `hieradata/production.yaml`?

alexmuller:
I think we're ok to keep it as enabled in all environments because we're rolling this out based on the `govuk::safe_to_reboot` value in the class hieradata, so production won't be affected by this yet.

tijmenb:
FWIW, this indeed solves `Error: Could not find dependency File[/etc/unattended-reboot/check] for File[/etc/unattended-reboot/check/02_mongodb.py] at /var/govuk/govuk-puppet/modules/govuk_unattended_reboot/manifests/mongodb.pp:18` on the development VM.

MatMoore:
@surminus I think this is equivalent to the previous behaviour for prod/staging.

In the original PR we left the safe to reboot variable unchanged on production, so for mongodb hosts the unattended reboot puppet class is still enabled but the checks will prevent it from actually running. On staging it's set to "overnight" and we'll propagate that setting to production when we're happy with it.

surminus:
:eggplant: Cool, go forth and merge!

PR #4102, opened 2016-02-15
Add wildcard for nagios sudoers rule

alexmuller:
In 0ac9c6a42cd3d56a732e064ebed218d2419cfbb8 we made this rule more restrictive, but it turns out that if you pass in one command-line argument to the sudoers file you need a wildcard to allow all arguments under that.


surminus:
:+1: merge when lime

PR #4103, opened 2016-02-15
Make integration work one hour earlier

alexmuller:
The content team are using integration for training and they need to have it working in the morning.

Specialist publisher currently has an issue where the database is in an inconsistent state in integration between the times when:

- The database is upgraded
- The data sync complete job is run

@stephaniekd has agreed to having integration switched on a little bit earlier.


surminus:
:+1: doing this might fix the change I pushed [here](https://github.com/alphagov/govuk-puppet/commit/8830cda2abedb8014a3b8c3ff2d25738d8967f86) too

merge when ready

PR #4105, opened 2016-02-15
Fix script to run tagging migration verifier

tijmenb:
Since we now have a dependency on http, we need to bundle and _also_ run it with `bundle exec`. As opposed to https://github.com/alphagov/govuk-puppet/pull/4098, I've now actually tried this change in jenkins first, so I'm pretty sure this actually works.

Trello: https://trello.com/c/q5jUopvT


alexmuller:
Looks good to me @tijmenb, merge whenever you want :+1:

PR #4106, opened 2016-02-15
Enable caching of vendor/{bundle,modules} on travis

rjw1:
To speed up travis runs.

Also changes the bundle install command to match the one used by travis normally


alexmuller:
Travis seems happy so this looks good to me :+1:

PR #4107, opened 2016-02-15
Run tagging migration in production & staging

tijmenb:
Panopticons procfileworker is not turned on currently in production and staging (but by default is in integration). The procfile worker starts the message queue consumer which copies tags from publishing-api to the contentapi.


alexmuller:
Merge away @tijmenb :+1:

PR #4108, opened 2016-02-15
Define icinga checks for every rabbitmq consumer

MatMoore:
Currently we define these checks in the app modules for a few consumers,
meaning they get run from the backend, and make requests against the
first rabbitmq node.

We get duplicate checks when services are clustered, and we don't have
checks defined for all consumers.

This changes the check to run on the rabbitmq nodes, against localhost,
and we define it in the same place we create service users, so the checks
are extended to every consumer.

There is still some duplicated checks because clustered nodes will return
the same result, but its easier to keep the configuration the same for
all of them.

### DIFF
--- a/modules/govuk/manifests/apps/content_register.pp
+++ b/modules/govuk/manifests/apps/content_register.pp
@@ -51,7 +51,4 @@ class govuk::apps::content_register(
   govuk::procfile::worker {'content-register':
     enable_service => $enable_procfile_worker,
   }
-  govuk_rabbitmq::monitor_consumers {'content-register_rabbitmq-consumers':
-    rabbitmq_queue => 'content_register',
-  }
 }
--- a/modules/govuk/manifests/apps/email_alert_service.pp
+++ b/modules/govuk/manifests/apps/email_alert_service.pp
@@ -31,9 +31,6 @@ class govuk::apps::email_alert_service(
     enable_nginx_vhost => false,
     command            => './bin/email_alert_service',
   }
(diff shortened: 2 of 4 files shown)


alexmuller:
I think this is the correct thing to do, merge away :+1:

This also fixes the case where the `rummager` app runs on the search machines, but the `search-n.api` machines can't connect to the `rabbitmq-1.backend` API over HTTPS because of our firewalling.

mattbostock:
@MatMoore I believe this has created unknown alerts in Icinga, e.g. [example](https://alert.publishing.service.gov.uk/cgi-bin/icinga/extinfo.cgi?type=2&host=rabbitmq-1.backend.publishing.service.gov.uk&service=Check+that+there+is+at+least+one+non-idle+consumer+for+rabbitmq+queue+rummager_to_be_indexed). I've raised a [story](https://trello.com/c/SYI3Isuz/127-unknown-alerts-for-rabbitmq-consumer) in Trello to fix.

alexmuller:
I think @tijmenb is aware that those checks are failing and they are because something hasn't been enabled in production yet.

tijmenb:
Correct - the rummager message queue consumer is not running on staging/production. Can we disable these checks for now?

We won't turn it on in staging & production until https://trello.com/c/afa0Mfa4 is finished (currently in product review).

MatMoore:
Yes that looks like what's going on. I think this should be protected by an "$ensure == present" check or something similar.

alexmuller:
#4149

PR #4109, opened 2016-02-16
Add Mapit to api load balancer

sihugh:
This is to enable us to use the new version of Mapit (initially for testing
with dependant apps in Integration).

This part sets up the configuration so that both versions are available (ie
via the backend or api loadbalancers).  The new Mapit server type is not
available in staging or production yet.

### DIFF
--- a/modules/govuk/manifests/node/s_api_lb.pp
+++ b/modules/govuk/manifests/node/s_api_lb.pp
@@ -4,6 +4,7 @@ class govuk::node::s_api_lb (
   $content_store_servers,
   $draft_content_store_servers,
   $email_campaign_api_servers,
+  $mapit_servers,
   $search_servers,
 ) {
   include govuk::node::s_base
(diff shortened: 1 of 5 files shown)


alexmuller:
This logic all looks correct to me and I think this change is fine! Are you happy with it?

I'd probably put `govuk::node::s_api_lb::mapit_servers: [] # doesn't exist yet` into the staging and production YAML files rather than having an empty string as the server name.

sihugh:
Cool, thanks, will amend.  Wasn't sure how to represent the absence in an appropriate manner, but that sounds better.

modules/govuk/manifests/node/s_api_lb.pp
```diff
@@ -60,6 +61,14 @@
     internal_only => true,
   }
 
+  if !empty($mapit_servers) {
+    loadbalancer::balance { 'mapit':
+      servers       => $mapit_servers,
+      internal_only => true,
+      https_only    => false, # FIXME: Remove for #51136581
```
alexmuller: Can we remove this line? If there's nothing that will be making requests to mapit on port 80 it's not required :)

sihugh: I'd like to, but given that we're testing the apps for compatibility with new MapIt, I wanted to keep the other variables to a minimum.  I'd envisaged coming back to this later on.  (Not that I've got any idea what 51136581 refers to)

alexmuller: Makes sense to me!

I think that ID is a story in the backlog that infrastructure had a long long time ago, so I'll go through separately and tidy that up :)

alexmuller:
Looks good to me, I think you can merge this whenever you want with absolutely no side-effects :+1:

sihugh:
Thanks!

PR #4111, opened 2016-02-16
Enable travel advice email alerts

bradwright:
Enable email alerts in integration and development.

https://github.com/alphagov/travel-advice-publisher/blob/cd0fdf68fa00c66b841e9a8ee263f20a69df62b0/config/environments/production.rb#L71 shows the use of the feature flag.


alexmuller:
Looks like correct Puppet which sets an environment variable to me :+1:

PR #4113, opened 2016-02-17
Remove prototype taxonomy task in Rummager

mobaig:
This was an initial spike by Finding Things to get early-state taxonomy
data into the search API for convenient retrieval. It is no longer needed
as the approach for taxon data has since changed.

Trello: https://trello.com/c/OYzkJzxZ


tijmenb:
:+1: once green.

PR #4114, opened 2016-02-17
Rotate jackscotti key

jackscotti:
Changed my laptop therefore updating the key.


mattbostock:
:+1: Merge when ready.

PR #4115, opened 2016-02-17
Disable plaintext HTTP for Errbit and Kibana

alexmuller:
I don't think there's any reason for these services to listen on port 80.

Both the HTTP and HTTPS requests log to the same place so I don't think there's any way to see if anything is using port 80 right now.

Tests failed on Travis but Jenkins is fine.


deanwilson:
Code change looks fine. Be nice to understand why the travis checks are failing before you merge though.

alexmuller:
I think Travis was just having a bad day, I've pushed the commit and it's passing now.

rjw1:
:+1: merge when ready

PR #4117, opened 2016-02-18
Make upstream apps aware of secure status

boffbowsh:
Sends the `X-Forwarded-Proto` header to unicorn et al so that apps can
make decisions based on whether a page is being served over http or
https.

Will allow us to set `config.force_ssl` in Signon.

/cc @alexmuller @mattbostock

### DIFF
--- a/modules/loadbalancer/templates/nginx_balance.conf.erb
+++ b/modules/loadbalancer/templates/nginx_balance.conf.erb
@@ -25,6 +25,7 @@ server {
   proxy_set_header X-Real-IP $remote_addr;
   proxy_set_header X-Forwarded-Server $host;
   proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+  proxy_set_header X-Forwarded-Proto $scheme;
   proxy_redirect off;
 
   proxy_connect_timeout 1s;


mattbostock:
This looks good but I'm not sure it'll work as-is because the upstream load balancer does not set the same header, and the load balancer always connects to port 443 on the proxy (app) server.

I'm also not sure how nginx behaves it it's proxying a request that already has that header set, presumably it'd retain the original header but I think it's worth testing.

boffbowsh:
Ok, will close this then.

boffbowsh:
Updated, @mattbostock @alexmuller

boffbowsh:
Tested this on integration, results in requests like the following:
```
GET / HTTP/1.0
X-Forwarded-Ssl: on
X-Forwarded-Proto: https
<snip>
...

mattbostock:
There's now a standard 'Fowarded' header in RFC 7239:
https://tools.ietf.org/html/rfc7239#section-4

Can we make use of that and will Rails/Rack recognise it?

boffbowsh:
We could add this too for future usage, but Rack will ignore it. Maybe a future improvement?

mattbostock:
:+1: Merge when ready.

mattbostock:
@boffbowsh I'd be grateful for any feedback on https://github.com/rack/rack/pull/1016 (/cc @binaryberry)

PR #4118, opened 2016-02-18
Extract govuk::testing_tools to own module

mattbostock:
To help breakup the monolithic `govuk` module.


surminus:
:+1: merge when ready

PR #4119, opened 2016-02-18
Extract govuk::user into its own module

mattbostock:

I tested this in Vagrant.


mattbostock:
Rebased against recent changes in master.

rjw1:
merge when ready

PR #4121, opened 2016-02-19
Use the new MapIt server in integration

sihugh:
We want to test the new MapIt server in integration against applications that depend on it.  

To do this we need to point mapit.integration.publishing.service.gov.uk to the api load balancer while leaving the other environments as is.


alexmuller:
This change is mostly good but there's a little weirdness in how hiera works which means it probably won't do what you expect.

Hiera won't do merging of arrays across environments, so I think the more correct thing to do in this case is to leave `common.yaml` is it is on master and set `hosts::production::backend::app_hostnames` in `integration.yaml` to be the array of apps not including `mapit`. The change in `integration.yaml` to add `hosts::production::api::app_hostnames` will therefore also need to include all of the apps in that list in `common.yaml`.

You then won't need to change `staging.yaml` or `production.yaml` either so the change is much more safe.

sihugh:
Thanks, that's exactly what I was thinking!

sihugh:
Integration now has its own lists of apps for `hosts::production::api::app_hostnames` and `hosts::production::backend::app_hostnames`.

We'll remove this when testing is complete.

alexmuller:
Thanks! This change will move mapit from the backend to the api in integration, merge whenever you like :+1:

PR #4122, opened 2016-02-19
Remove dependency on `/usr/share/nginx`

alexmuller:
This is only in the catalog on the router machines so it will need some more work to add a dependency for this resource.


mattbostock:
:+1: Merge when green.

PR #4126, opened 2016-02-22
Rename 'name' to 'named' to avoid puppet 4 errors.

deanwilson:
In puppet 4/future parser this code will cause
"The parameter $name redefines a built in parameter in the 'define'
expression" to be logged and rspec tests to fail.


mattbostock:
:+1: Merge when ready

PR #4127, opened 2016-02-22
Make port use an integer value, not implicit string conversion

deanwilson:
This is an early step to support future/parser puppet 4. Our rspec
tests currently specify (correctly) that port is an int. Our manifest
should reflect that choice of type.

We also shouldn't be running our apps on priviledged ports so we set
the lowest allowed port to be 1025.

This will require apps using strings to be converted over to ints
when we start to run puppet 4. Those will be done in a seperate PR.

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -269,14 +269,20 @@ define govuk::app (
   if ! ($app_type in ['procfile', 'rack', 'bare']) {
     fail 'Invalid argument $app_type to govuk::app! Must be one of "procfile", "rack", or "bare"'
   }
+
   if ($app_type in ['procfile', 'rack']) and ($port == 0) {
     fail 'Must pass port when $app_type is "procfile" or "rack"'
   }
+
   if ($app_type == 'bare') and !($command) {
     fail 'Invalid $command parameter'
   }
+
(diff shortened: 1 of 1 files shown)


deanwilson:
if the validate_integer was moved in to the if would this still cover all cases where we want to use port?

rjw1:
I wonder if the port should be mandatory?

rjw1:
:+1:
 The intent is correct. 
I do wonder if the fix should go in here as well so the PR doesnt break deploys. 

What would happen if we set the default to null.

modules/govuk/manifests/app.pp
```diff
@@ -269,14 +269,20 @@
   if ! ($app_type in ['procfile', 'rack', 'bare']) {
     fail 'Invalid argument $app_type to govuk::app! Must be one of "procfile", "rack", or "bare"'
   }
-  if ($app_type in ['procfile', 'rack']) and ($port == 'NOTSET') {
+
+  if ($app_type in ['procfile', 'rack']) and ($port == 0) {
```
mattbostock: Just realised that this conditional is redundant as any ports outside of 1025-65535 will be caught by `validate_integer` below.

rjw1: It is a more useful error message though.

PR #4128, opened 2016-02-22
Postgres futureparser specs

deanwilson:
These spec changes make the postgres module a lot more future parser friendly and allow most of the specs to now pass under Puppet 3.8 on OSX.


mattbostock:
:+1: Good, descriptive, commit messages. Merge when ready.

PR #4130, opened 2016-02-22
Add a vdc fact to our test runs to make them pass

deanwilson:
When testing under the future parser we see issues in
icinga/manifests/client.pp under
`if ($::vdc == 'licensify') or ($::vdc == 'efg') or ($::vdc =~ /^*_dr$/) {`
if the vdc is undefined.

Adding this fact doesn't fix the tests but it does move the problem along from
'Left match operand must result in a String value. Got an Undef Value.'
to Could not find template '' which is an issue from an upstream module we'll
deal with in isolation.

### DIFF
--- a/modules/govuk/spec/classes/govuk__apt_spec.rb
+++ b/modules/govuk/spec/classes/govuk__apt_spec.rb
@@ -3,6 +3,7 @@ require_relative '../../../../spec_helper'
 describe 'govuk::node::s_apt' do
   let(:facts) {{
     :concat_basedir => '/var/lib/puppet/concat/',
+    :vdc            => 'fake_vdc',
   }}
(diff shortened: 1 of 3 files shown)


mattbostock:
What's the significance of the cache and whitehall_backend machines with relation to `icinga/manifests/client.pp`? Are other nodes also affected?

deanwilson:
Those issues will come up in integration testing. This is just for the unit testing phase.

mattbostock:
@deanwilson: I meant that you made changes to the spec tests for those machines but not the other machines. Why are those machines affected but not others?

deanwilson:
Those are the only two machines that had failing tests. Either the other machines define their environments differently or they have less test coverage.

PR #4131, opened 2016-02-22
Add the unicorn_port param to the package.pp file.

deanwilson:
The unicorn_port setting is used in package.pp ->
puppet/etc/init/puppetmaster.conf.erb and so needs to be
plumbed in. The test fails under futuer parser without this change.

### DIFF
--- a/modules/puppet/manifests/master/package.pp
+++ b/modules/puppet/manifests/master/package.pp
@@ -4,13 +4,21 @@
 #
 # === Parameters
 #
+# [*hiera_eyaml_gpg_version*]
+#  Specify the version of hiera_eyaml_gpg to install.
+#
 # [*puppetdb_version*]
 #   Specify the version of puppetdb-terminus to install which should match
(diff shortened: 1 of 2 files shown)


modules/puppet/manifests/master/package.pp
```diff
@@ -4,13 +4,21 @@
...
 #   the puppetdb installation. Passed in by parent class.
 #
+# [*unicorn_port*]
+#   Specify the port on which unicorn (and hence the puppetmaster) should
+#   listen.
+#
 class puppet::master::package(
   $hiera_eyaml_gpg_version = 'latest',
   $puppetdb_version = 'present',
+  $unicorn_port = '9090',
```
alexmuller: Should this default to a string or an integer? :trollface:

deanwilson:
A string for now to keep the change to a minimum and then an integer when we move everything else over too. It's consistently wrong as a string at the moment so it does work.

alexmuller:
This change is correct :+1:

PR #4132, opened 2016-02-22
Disable access on port 80 for backend apps

alexmuller:
This commit ensures that connections on port 80 are redirected to HTTPS for all of our public backend apps.


rjw1:
merge when ready.

mattbostock:
'Disable access' is a bit misleading, as we still allow connections to port 80.

alexmuller:
I don't think it's too misleading. Previously it was possible to get a response from those apps over port 80 but it's not any more.

PR #4133, opened 2016-02-23
Add asset manager token into share sale environment

bradwright:
To be merged after deployment#907


alexmuller:
Looks right to me :+1:

PR #4135, opened 2016-02-23
Mark the "original_url" field as not analysed

rboulton:
The purpose of recording this is to be able to find the list of pages
which a particular app is involved in rendering, or the list of apps
involved in rendering a particular page.  If the field is
analysed we can't do this, because a facet query will instead return the
list of components of urls.


tijmenb:
:+1: when :green_apple:.

PR #4137, opened 2016-02-23
Make hmrc-manuals-api error on plaintext HTTP

alexmuller:
Easier to review commit-by-commit.

/cc @jennyd

### DIFF
--- a/modules/loadbalancer/templates/nginx_balance.conf.erb
+++ b/modules/loadbalancer/templates/nginx_balance.conf.erb
@@ -4,7 +4,13 @@ upstream <%= @vhost %>-upstream {
 <%- end -%>
 }
 
-<%- if @https_redirect -%>
+<%- if @error_on_http -%>
+server {
+  server_name <%= @vhost_real %> <%= @aliases.join(" ") unless @aliases.empty? %>;
+  listen 80;
+  return 426 "Request cannot be completed using a plaintext HTTP connection. Retry over HTTPS.\n";
+}
+<%- elsif @https_redirect -%>
 server {
   server_name <%= @vhost_real %> <%= @aliases.join(" ") unless @aliases.empty? %>;
   listen 80;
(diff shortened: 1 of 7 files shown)


mattbostock:
:+1: to renaming `https_only`.

Can we stop serving connections on port 80 for APIs, rather than making them error? Even if we make them error, there's still the potential for someone to send data in plaintext to port 80.

alexmuller:
I believe it's impossible to do that because port 80 has to resolve to the backend loadbalancer so that user-facing apps can be redirected, so it will always be possible to establish a connection (it will just go to the wrong vhost).

mattbostock:
@alexmuller We could do it by splitting the user-facing and API backends to use separate IP addresses, but that's a lot of work and shouldn't block this pull request.

mattbostock:
@alexmuller: Can you please clarify the commit message in ea0a2c56 where `true` is repeated twice?

```
- error_on_http true: https_redirect is ignored, errors are served
- error_on_http true: https_redirect decides whether to redirect or serve the app
```

mattbostock:
Nitpick with ea0a2c5 for the curious: RFC 7231 updates RFC 2817 with regards to `426 Upgrade Required`: https://tools.ietf.org/html/rfc7231#section-6.5.15

It doesn't affect how you'd use it in this case.

modules/loadbalancer/templates/nginx_balance.conf.erb
```diff
@@ -4,7 +4,13 @@ upstream <%= @vhost %>-upstream {
 <%- end -%>
 }
 
-<%- if @https_only -%>
+<%- if @error_on_http -%>
+server {
+  server_name <%= @vhost_real %> <%= @aliases.join(" ") unless @aliases.empty? %>;
+  listen 80;
+  return 426 "Request cannot be completed using a plaintext HTTP connection\n";
```
mattbostock: This could specify that HTTPS must be used (seems obvious to us but might not be immediately obvious to everyone).

mattbostock:
Overall this looks good to me.

alexmuller:
Updated, thanks!

mattbostock:
:+1: Merge when ready.

jennyd:
Thanks @alexmuller!

PR #4140, opened 2016-02-24
Fix Icinga hostname for offsite backup machine

mattbostock:
Without this change, the Icinga hostname was defaulting to the `fqdn`
facter fact, which made alerts for the offsite backup machine appear as
`monitoring-1.management`.

### DIFF
--- a/modules/backup/manifests/offsite/monitoring.pp
+++ b/modules/backup/manifests/offsite/monitoring.pp
@@ -16,8 +16,9 @@ class backup::offsite::monitoring(
 ){
 
   icinga::host { $offsite_fqdn:
-    hostalias => $offsite_fqdn,
-    address   => $offsite_fqdn,
+    hostalias    => $offsite_fqdn,
+    address      => $offsite_fqdn,
+    display_name => $offsite_hostname,
   }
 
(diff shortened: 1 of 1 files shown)


deanwilson:
Is 'address => $offsite_fqdn,' deliberate? I'd normally expect address to be an ip address. Esp if we want monitoring to continue if DNS is unavailable.

mattbostock:
@deanwilson It's consistent with what we do [elsewhere](https://github.com/alphagov/govuk-puppet/blob/f47cec44a1577c7b31577ef69554ab2ebf02074e/modules/monitoring/manifests/checks/mirror.pp#L24-L29).

For internal DNS it's not currently an issue as we use `/etc/hosts`, though the offsite backup machines may rely on external DNS.

mattbostock:
_bump_

alexmuller:
Where does the `host_name` parameter to `icinga::host` actually get used?

mattbostock:
@alexmuller I've removed it, I think that's something we need to clean up.

alexmuller:
Thanks, opened #4235 to tidy up.

alexmuller:
Looks good to me :+1:

deanwilson:
Merge away.

PR #4141, opened 2016-02-24
Remove unused dir for processed CDN logs

mattbostock:
We stopped using this directory as of 1ec6a44 and I've confirmed that
there's no data in this directory in Production. Remove it so as to
avoid confusion.


deanwilson:
Looks mergeable. Curious as to why the `force` though.

mattbostock:
According to the [Puppet docs](https://docs.puppetlabs.com/puppet/3.6/reference/type.html#file-attribute-force), `force` is required to remove a directory. It's also necessary to purge the empty subdirectory inside that directory.

mattbostock:
_bump_

deanwilson:
Merge away.

mattbostock:
@deanwilson I've amended and force-pushed as I realised I wasn't removing the directory one level up which would be left empty - can you please review?

deanwilson:
Merge at will.

PR #4143, opened 2016-02-24
Add jenkins job to run rake tasks

tijmenb:
Running rake tasks from jenkins has a couple of advantages:

- We get a history of commands that have been run
- We save the output to see what happens
- We don't have to SSH into machines
- Long-running tasks won't be interrupted when connection gets lost
- We can link to the rake task with filled-in variables. This is nice for in Trello tickets.

screen shot 2016-02-25 at 14 42 51


modules/govuk_jenkins/templates/jobs/run_rake_task.yaml.erb
```diff
@@ -0,0 +1,86 @@
+---
+- job:
+    name: run-rake-task
+    display-name: Run take task
```
alexmuller: Nitpick: this should be "rake task" :)

alexmuller:
It would be nice if we didn't have to duplicate the list of applications but I can't think of an easy way to get around that. I wonder whether hiera lookups for lists work yet. If not we could probably make it a multiline string and look that up although that's a bit terrible.

tijmenb:
That's a good idea, I'll try that.

tijmenb:
@alexmuller extracted the choices, updated PR. Is that what you meant?

tijmenb:
I've tested the shell script/parameters thing in Jenkins integration by creating a new task:

https://deploy.integration.publishing.service.gov.uk/job/test-rake-task

tijmenb:
I've added a not about *why* using `..app_choices: "%{hiera('app_choices')}"` actually works.

mattbostock:
I think we could improve the way the machines are selected as you say but I think this is a valuable addition as-is. Merge when ready :+1:

PR #4145, opened 2016-02-25
Revert "Merge pull request #4081 from alphagov/elasticsearch-unattend…

mattbostock:
…ed-reboot"

This reverts commit 494766a50168581ba50e8d8da1e3d84ca199ce64, reversing
changes made to 20baf06ea33a572662513ada57e2c7648157916f.

This change is not currently safe as the `02_elasticsearch.py` script
will never be executed:
https://github.com/gds-operations/puppet-unattended_reboot/pull/5

We can re-instate this once we're using the fixed version of the
unattended reboot module.


deanwilson:
Can we rename the script and change the file resource to not end in .py to fix it rather than revert all the work?

mattbostock:
@deanwilson: We could but I think I prefer to revert as there's a [fix](https://github.com/gds-operations/puppet-unattended_reboot/pull/5) for the file extension issue and I'm not very confident that this will work as expected in Production without further testing. I'm about to deploy Puppet and won't have time to do that testing before then.

mattbostock:
On second thoughts, this change does not affect Production so changing the file extension is probably the most straight-forward solution.

mattbostock:
@deanwilson Please see: https://github.com/alphagov/govuk-puppet/pull/4147

PR #4146, opened 2016-02-25
Ensure that scripts with file extensions are run

mattbostock:
I'm raising this for discussion.

The current behaviour is surprising and explains why jobs like [`prune-whisper-files.sh`](https://github.com/alphagov/govuk-puppet/blob/f47cec44a1577c7b31577ef69554ab2ebf02074e/modules/govuk/templates/node/s_graphite/prune-whisper-files.sh.erb) are not running.

We could fix the issue by renaming the scripts that have file extensions, however it's likely that someone else might hit the same issue in future and it go unreported for a long time.

On the other hand, it's also surprising behaviour if we modify the default way that `cron.daily` jobs are run.

My feeling is that adding the regex is the most pragmatic solution.

* * *

Scripts with file extensions are not currently being executed by
cron.daily because `run-parts`, which is used to run `cron.daily` jobs,
is not executing scripts with file extensions (e.g. `.sh`):

http://manpages.ubuntu.com/manpages/trusty/man8/run-parts.8.html

> If neither the --lsbsysinit option nor the --regex option is given
> then the names must consist entirely of ASCII upper- and lower-case
> letters, ASCII digits, ASCII underscores, and ASCII minus-hyphens.

Passing a regex to run-parts means that it will execute all executable
scripts in `/etc/cron.daily`.

Ubuntu bug: https://bugs.launchpad.net/ubuntu/+source/debianutils/+bug/38022

The upstream Debian bug has been archived:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=472585

### DIFF
--- a/modules/cron/templates/etc/crontab.erb
+++ b/modules/cron/templates/etc/crontab.erb
@@ -8,8 +8,8 @@ SHELL=/bin/sh
 PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
 
 # m h dom mon dow user  command
-<%= @hourly_min %> *    * * *   root    cd / && run-parts --report /etc/cron.hourly
-<%= @daily_min %> <%= @daily_hour %>    * * *   root    test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.daily )
-<%= @weekly_min %> 6    * * <%= @weekly_dow %>   root    test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.weekly )
-<%= @monthly_min %> 6    1 * *   root    test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.monthly )
+<%= @hourly_min %> *    * * *   root    cd / && run-parts --report --regex '.*' /etc/cron.hourly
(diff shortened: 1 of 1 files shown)


alexmuller:
I'm not sure that the regex is the best solution but I also don't have anything I can volunteer which is better.

Regardless, nice sleuthing for a bug which has affected a load of different parts of our stuff :clap:

ajlanghorn:
Out of interest, did you consider just renaming the files that are run by `run-parts` to exclude the extension? That might stop you having to specify a regex, and complicating that further.

mattbostock:
@ajlanghorn: We did that in https://github.com/alphagov/govuk-puppet/pull/4147. As mentioned in the PR description, it's likely that someone else might hit the same issue in future as it's already surprised a few of us.

dcarley:
Makes sense. We also came to the same conclusion when we discovered this "feature" (not for cron and we later refactored it out anyway, but still).

ajlanghorn:
Nice - I hadn't seen #4147. It is documented in the manpage, but nice to see it in a more human-readable format.

samjsharpe:
While I see what you are trying to do here, would it not be better to stick to the system's default behaviour and instead somehow monitor (maybe with nagios?) that there aren't any files with extensions in these directories?

mattbostock:
@samjsharpe I am hesitant to change the default behaviour. The idea of the Icinga check is a nice solution, I'll look at that. Thanks!

mattbostock:
Related: https://github.com/alphagov/govuk-puppet/pull/4154

PR #4147, opened 2016-02-25
Ensure reboot check scripts are executed

mattbostock:
The 02_mongodb.py and 02_elasticsearch.py scripts will never be executed
currently because they have a file extension, meaning that reboots will
go ahead without checking the Elasticsearch or MongoDB cluster health:
gds-operations/puppet-unattended_reboot#5

Removing the file extension fixes the issue.


deanwilson:
Looks fine. Merge when green.

PR #4149, opened 2016-02-25
Only monitor RabbitMQ consumers that we expect

alexmuller:
Pass `$ensure` from the consumer definition and make the monitoring conditional on it.


mattbostock:
:+1: Merge when ready

PR #4150, opened 2016-02-25
Use enabled param for Elasticsearch reboot

alexmuller:
We define this param already but we don't use it.


mattbostock:
The `govuk_unattended_reboot::elasticsearch::enabled` parameter feels misleading. It defaults to false, meaning that the MongoDB check script won't be present, meaning that reboots are more likely to go ahead (because there's no check script to prevent that) if we forget to otherwise disable them.

In other words, setting `enabled` to false doesn't disable the unattended reboots for Elasticsearch machines, which is surprising behaviour.

alexmuller:
I agree, but for now this is bugfix for provisioning the dev VM which maintains the existing behaviour :)

mattbostock:
@alexmuller Agreed, merge when ready.

PR #4151, opened 2016-02-25
Add new MapIt servers to staging and production

sihugh:
We're currently running with out of date MapIt data on production. We've recently built new MapIt servers in integration which have up to date software, data and the ability to import new data when it is released.

This is the next stage, which is to add the new machines to staging and production.  

When they've been tested and we're happy, we'll raise another PR to point mapit at the api load balancer (it's happily running against the old servers at present.)

[Trello card](https://trello.com/c/b3XBoEPI/29-get-new-mapit-running-in-staging-and-production-5)

Related PRs

- https://github.gds/gds/alphagov-deployment/pull/1149
- https://github.gds/gds/govuk-provisioning/pull/522

### DIFF
--- a/hieradata/production.yaml
+++ b/hieradata/production.yaml
@@ -131,7 +131,9 @@ govuk::node::s_api_lb::email_campaign_api_servers:
   - "email-campaign-api-1.api"
   - "email-campaign-api-2.api"
   - "email-campaign-api-3.api"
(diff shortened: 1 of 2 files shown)


alexmuller:
This is all correct :+1:

sihugh:
This should probably be the last of the PRs to merge?

alexmuller:
It needs to get deployed before the machines can be built but apart from that I don't think it matters :)

PR #4154, opened 2016-02-26
Rename script to prune whisper files

mattbostock:
I recently found that `run-parts`, used by cron.daily`, doesn't run
scripts that have a file extension:

http://manpages.ubuntu.com/manpages/trusty/man8/run-parts.8.html

> If neither the --lsbsysinit option nor the --regex option is given
> then the names must consist entirely of ASCII upper- and lower-case
> letters, ASCII digits, ASCII underscores, and ASCII minus-hyphens.

Rename the `prune-whisper-files.sh` file so that cron.daily will execute
it. Currently, it does not run.


rjw1:
Im not entirely upset that this wasnt working. I think we should revisit the decision behind this script in the first place.

mattbostock:
@rjw1: Why is that?

rjw1:
I'm just not sure pruning the whisper files is the right thing to do.

mattbostock:
@rjw1 In case they have valid data but just haven't been updated in a long time, for example if something breaks?

mattbostock:
I think this is worth discussing. For example, we currently have [stale metrics](https://graphite.staging.publishing.service.gov.uk/render?from=-2hours&until=now&width=400&height=250&target=api-elasticsearch-1_api_production.uptime.uptime&_uniq=0.6051619870122522&title=api-elasticsearch-1_api_production.uptime.uptime) in our Staging environment for machines with a `production` suffix, which I assume come from when the environment was first provisioned.

mattbostock:
This is ready for review.

After an internal discussion, we determined that pruning old metrics (which have not been modified in 90 days) is valuable, despite the small risk that we might purge metrics that we care about which are updated infrequently.

For example, we might have an outage where an app generates a lot of HTTP 5XX errors, after which it doesn't generate an error for 90 days, and we'd lose the data for the outage. It's unlikely that an app would not generate a HTTP 5XX response in 90 days, but that's the kind of risk.

rjw1:
:+1:

PR #4156, opened 2016-02-26
Allow Backend and API Elasticsearch clusters to reboot in Production overnight

mattbostock:
Story: https://trello.com/c/JD2ZFRcJ/13-write-script-to-check-if-elasticsearch-logs-elasticsearch-machines-can-be-rebooted

Allow the backend and API Elasticsearch clusters to reboot in Production
unattended overnight. They will only reboot if the cluster is healthy.

I haven't enabled this for the logs Elasticsearch cluster as the shard
reallocation, which is not disabled by the unattended reboots, takes
hours and I'd rather test it further in Staging first.

### DIFF
--- a/hieradata/class/api_elasticsearch.yaml
+++ b/hieradata/class/api_elasticsearch.yaml
@@ -1,7 +1,7 @@
 ---
 
-govuk::safe_to_reboot::can_reboot: 'careful'
-govuk::safe_to_reboot::reason: 'Reboot a single node at time and check that the cluster goes green between reboots'
+govuk::safe_to_reboot::can_reboot: 'overnight'
+govuk::safe_to_reboot::reason: 'Will reboot overnight if the cluster is healthy'
 
 govuk_elasticsearch::version: '1.4.4'
 
--- a/hieradata/class/elasticsearch.yaml
+++ b/hieradata/class/elasticsearch.yaml
@@ -1,7 +1,7 @@
 ---
 
-govuk::safe_to_reboot::can_reboot: 'careful'
-govuk::safe_to_reboot::reason: 'Reboot a single node at time and check that the cluster goes green between reboots'
+govuk::safe_to_reboot::can_reboot: 'overnight'
+govuk::safe_to_reboot::reason: 'Will reboot overnight if the cluster is healthy'
 
 govuk_elasticsearch::version: '0.90.12'
 
--- a/hieradata/class/staging/api_elasticsearch.yaml
+++ b/hieradata/class/staging/api_elasticsearch.yaml
@@ -1,3 +1,0 @@
----
-
-govuk::safe_to_reboot::can_reboot: 'overnight'
--- a/hieradata/class/staging/elasticsearch.yaml
+++ b/hieradata/class/staging/elasticsearch.yaml
@@ -1,3 +1,0 @@
----
-
-govuk::safe_to_reboot::can_reboot: 'overnight'


mattbostock:
_bump_

mattbostock:
Closing this as we had an incident specific to Elasticsearch in Staging this morning that may have been exacerbated the API elasticsearch nodes rebooting. I'll write up more details shortly.

mattbostock:
I've drawn up a [timeline](https://cdn.knightlab.com/libs/timeline3/latest/embed/index.html?source=12swaa8FhiAWt9hS0wdwwiCSBK-9oRqv3yD-sS0Z5TqU&font=Bitter-Raleway&lang=en&initial_zoom=2) of the events we saw in Staging (authorised users can edit the source [here](https://docs.google.com/a/digital.cabinet-office.gov.uk/spreadsheets/d/12swaa8FhiAWt9hS0wdwwiCSBK-9oRqv3yD-sS0Z5TqU/edit?usp=sharing)) following investigation by myself, @tijmenb, @sihugh and @dsingleton.

#### API Elasticsearch cluster in Production (normal behaviour):
screen shot 2016-03-03 at 10 25 54

#### API Elasticsearch cluster in Staging:

There are three issues that are notable in this screenshot:
- there are duplicate index aliases; one for the index from February 16th and the other for today's index (March 3rd)
- the indices for February 16th should have been removed by Rummager's [`clean`](https://github.com/alphagov/rummager/blob/8eda5fec3869ca0035fd0021ec961b7d6f052c06/lib/tasks/indices.rake#L65-L70) Rake task, which is [run in Production](https://github.com/alphagov/govuk-puppet/blob/fbd278d4fcb1b35971006fd003ecc61de05922b6/modules/govuk_jenkins/templates/jobs/search_fetch_analytics_data.yaml.erb#L30) nightly, after data is synchronised to Staging.
- the indices for February 16th have unallocated shards

screen shot 2016-03-03 at 10 26 08

The existing index aliases are supposed to replaced by a new alias pointing to today's index when we [restore the Elasticsearch dumps from Production](https://github.gds/gds/env-sync-and-backup/blob/aebe1e33dca19329613c4c90b8158b4806135ec8/scripts/elasticsearch-restore.sh#L30). According to the `es_dump_restore` tool we use, this should be an atomic operation:

>  If DESTINATION_ALIAS already exists, it will be atomically changed to point to the new location.

https://github.com/patientslikeme/es_dump_restore/tree/d8d51c26f9b6bd9bd406cf8cbe4b20fade882c49#usage

#### Resolution

We were able to restore cluster health by removing the old indices from February 16th and removing the extraneous aliases.

#### Conclusions

We came to the following conclusions:

- It's not clear whether the Elasticsearch nodes rebooting caused the cluster to become unhealthy, however it seems likely that it may have exacerbated any existing issue

- We're relatively confident that the [script that checks the cluster health](https://github.com/alphagov/govuk-puppet/blob/fbd278d4fcb1b35971006fd003ecc61de05922b6/modules/govuk_unattended_reboot/files/etc/unattended-reboot/check/02_mongodb) works correctly as we found logs on the cluster nodes where the reboot had been averted because the script returned a non-zero exit code, indicating that the cluster was not healthy. This assumption is also backed up by the Icinga alert history and Elasticsearch logs, which show no significant alerts or history before the cluster became unhealthy. The Elasticsearch logs are outputting at the `INFO` log level.

- We suspect there may be a bug in the 'atomic' operation that replaces the existing indices (that happens when we run `es_dump_restore`

- The unallocated shards were limited to the old indices from Februrary 16th only, for which we don't have an explanation

After talking to @rboulton, we determined that the mean time to recovery to recreate the Elasticsearch indices, should a similar but more serious issue occur in Production, would be longer than we're currently happy with. We're addressing this by creating Elasticsearch snapshots which should dramatically reduce our recovery time. Once these are implemented, we could review whether to introduce unattended reboots in Production for Elasticsearch.

@rboulton also recommended upgrading Elasticsearch. We're using a relatively old version and there's a good chance that these issues may have been fixed in a future release.

PR #4157, opened 2016-02-26
Move spec file for consistency

mattbostock:
Move the spec file for the etcd class to a `classes` subdirectory for
consistency with the directory schema we use for our other modules.


deanwilson:
Merge away.

Issue #3665, opened 2015-12-16
Add mongo-pp as option for data sync


alexmuller:
This pull request was [originally opened on GitHub Enterprise](https://github.gds/gds/puppet/pull/3665) on Monday 16 November 2015 at 11:03 GMT before GOV.UK's Puppet code was migrated to a public GitHub repository.

PR #4162, opened 2016-03-01
Move PARALLEL_JOBS env var to correct place

mattbostock:
The `PARALLEL_JOBS` environment variable in the data sync Jenkins job
was in the wrong place.

Fixing this will allow our Integration sync job to complete before
office hours by running the sync jobs in parallel; currently it has been
overrunning until after 9:30am. It has been taking 1.5 hours longer
since the `mongo-pp` job was added in alphagov/govuk-puppet#3665.

According to the contextual help in Jenkins, `script-content` is used as
follows (note the last line):

    Execute a script file aimed at setting an environment such as creating folders, copying files, and so on.
    Give the script file content.
    You can use the above properties variables.
    However, adding or overriding environment variables in the script doesn't have any impacts in the build job.

So that this variable is taken into account by the build job, move it to
`properties-content`, which is the correct place to set environment
variables for the build.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/copy_data_to_integration.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/copy_data_to_integration.yaml.erb
@@ -17,6 +17,9 @@
     properties:
         - github:
             url: https://github.gds/gds/env-sync-and-backup/
+        - inject:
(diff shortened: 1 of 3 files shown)


tijmenb:
:+1:

PR #4163, opened 2016-03-01
Log format is JSON for service-manual-publisher

thehenster:
This is needed so that logstream will read the JSON log and the request logs will appear in Kibana.


bradwright:
:+1: looks good

PR #4153, opened 2016-02-25
Don't link pg_config for PostgreSQL

alexmuller:
On Debian systems this change is destructive. It creates a symlink on top of the `pg_config` binary which exists.

This was caused by the recent upgrade of the [Puppet PostgreSQL module from 3.4.1 to 4.1.0][1]: "Link pg_config binary into /usr/bin"

[1]: https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/CHANGELOG.md#2014-09-03---supported-release-400

Subsequently [MODULES-1485][2] was opened which correctly identified that this change was very destructive on Debian systems.

[2]: https://tickets.puppetlabs.com/browse/MODULES-1485

This has caused our PostgreSQL clients (including the development VM) to stop being able to install the `pg` gem.

It can be fixed by removing the postgresql-common packages and allowing Puppet to re-run to reinstall them:

```
sudo rm /usr/bin/pg_config
sudo apt-get remove postgresql-client-9.3 postgresql-client-common
```

The change to the PostgreSQL module has been [fixed in 4.2.0][3]

[3]: https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/CHANGELOG.md#2015-03-10---supported-release-420

PR #4164, opened 2016-03-01
Ensure `libpq-dev` is present for signon backends

benlovell:
Since signon now supports either mysql or postgresql database adapters
we'll need to ensure `libpq-dev` is around so the `pg` gem can find the
necessary headers.

### DIFF
--- a/modules/govuk/manifests/apps/signon.pp
+++ b/modules/govuk/manifests/apps/signon.pp
@@ -68,4 +68,6 @@ class govuk::apps::signon(
       value   => $redis_url,
     }
   }
+
+  include govuk_postgresql::client #installs libpq-dev package needed for pg gem
 }


alexmuller:
This is absolutely the correct change to make if signon is going to support PostgreSQL, so merge when the tests have passed :+1:

But I'm also confused about why this isn't working for you at the moment, because the backend app servers should already include this class (eg for transition).

alexmuller:
Or was it just broken because I broke it?

h-lame:
Is this going to be the same root cause as https://github.com/alphagov/govuk-puppet/pull/4153 (e.g. we need pg_config to build the gem, but it's not there because puppetlabs puppet module symlinked it somewhere it shouldn't be).

benlovell:
> Or was it just broken because I broke it?

@alexmuller Maybe, but also I can't help feeling this is related to #4153 as @h-lame points out. :thought_balloon:

alexmuller:
Oh yep sorry, that's what I meant - it definitely is that. I thought I'd fixed the problem I caused but obviously not. If you notice this problem anywhere else please leave a shouty comment on #4153 and get 2nd line to help fix it.

PR #4167, opened 2016-03-01
Don't create app.pid automatically for unicornherder procfile apps

sihugh:
As part of deploying a new version of Mapit, we want to use Unicorn Herder to
help us with zero-downtime deploys.

We were experiencing problems with this, as the pid file was being created by
the Unicorn Herder process (running as the super user), which meant that the
gunicorn processes couldn't access it (as they run under the app user).

We know that other applications use the 'procfile' app type which has
'--make-pid' defined, however we want to skip this for procfile apps that use 
unicornherder so that the first gunicorn process creates it.

If there's a more appropriate way of tackling this, please let me know!

(This in conjunction with @brenetic )

https://trello.com/c/d82WTfeq/314-get-mapit-working-with-unicorn-herder-3-time-boxed-1-day

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -37,6 +37,18 @@
 # `GOVUK_APP_CMD` which is used by `govuk_spinup`.
 #
 #
+# [*create-pidfile*]
+# Determines whether a pidfile is created when a `procfile` app is started
+#
+# By default procfile apps will create pid files (via the --make-pid switch
+# in govuk_spinup).
+# If the create-pidfile value is set to false, then a pid file will not be
+# created.  This is appropriate when using unicornherder with gunicorn.
+# The value defaults to 'NOTSET' (as opposed to true) to try to flag up that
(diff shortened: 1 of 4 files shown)


mattbostock:
@sihugh Merge when ready. I think we need to review how we configure different application types, but for now this is :+1:

modules/govuk/manifests/app.pp
```diff
@@ -37,6 +37,18 @@
...
+# [*create-pidfile*]
+# Determines whether a pidfile is created when a `procfile` app is started
+#
+# By default procfile apps will create pid files (via the --make-pid switch
+# in govuk_spinup).
+# If the create-pidfile value is set to false, then a pid file will not be
+# created.  This is appropriate when using unicornherder with gunicorn.
+# The value defaults to 'NOTSET' (as opposed to true) to try to flag up that
+# this doesn't apply to all app types. (eg Rack apps don't create pid files
+# by default)
```
mattbostock: :+1: Great documentation.

PR #4168, opened 2016-03-01
Offsite postgresql backups using WAL-E

surminus:
This change would implement daily base backups and continuous archiving of [WAL segments](http://www.postgresql.org/docs/9.3/static/wal-intro.html) for the Bouncer application in Integration.

It has a couple of dependencies in that it requires a user and subsequent ID, KEY and S3 bucket to be created.

### DIFF
--- a/hieradata/vagrant.yaml
+++ b/hieradata/vagrant.yaml
@@ -157,6 +157,7 @@ govuk_postgresql::server::snakeoil_ssl_key: |
 govuk_postgresql::wal_e::backup::aws_access_key_id: '123456789'
 govuk_postgresql::wal_e::backup::aws_secret_access_key: 'thisisafakekey'
 govuk_postgresql::wal_e::backup::s3_bucket_url: 's3://foo/bar'
+govuk_postgresql::wal_e::backup::enabled: true
 
 icinga::config::enable_event_handlers: true
 icinga::nginx::enable_basic_auth: false
--- a/hieradata/vagrant.yaml
+++ b/hieradata/vagrant.yaml
@@ -154,6 +154,10 @@ govuk_postgresql::server::snakeoil_ssl_key: |
   vxBODhF5cvyO/S2jUuvApGU4dU2KULT5Dv3/Lh/Otg==
   -----END RSA PRIVATE KEY-----
 
+govuk_postgresql::wal_e::backup::aws_access_key_id: '123456789'
+govuk_postgresql::wal_e::backup::aws_secret_access_key: 'thisisafakekey'
+govuk_postgresql::wal_e::backup::s3_bucket_url: 's3://foo/bar'
+
 icinga::config::enable_event_handlers: true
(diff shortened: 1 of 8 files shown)


surminus:
Oh, I just realised I completely forgot to add encryption to this. Anyway, thoughts still welcome on the general gist!

hieradata/vagrant.yaml
```diff
@@ -154,6 +154,11 @@ govuk_postgresql::server::snakeoil_ssl_key: |
   vxBODhF5cvyO/S2jUuvApGU4dU2KULT5Dv3/Lh/Otg==
   -----END RSA PRIVATE KEY-----
 
+govuk_postgresql::wal_e::backup::aws_access_key_id: '123456789'
```
mattbostock: These credentials should be in `vagrant_credentials.yaml`.

modules/govuk_postgresql/manifests/wal_e/backup.pp
```diff
@@ -0,0 +1,94 @@
...
+    postgresql::server::config_entry {
+      'archive_mode':
+        value => 'on';
+      'archive_command':
+        value => 'envdir /etc/govuk/env.d /usr/local/bin/wal-e wal-push %p';
+      'archive_timeout':
+        value => '60';
+    }
+
+    govuk::envvar { 'AWS_SECRET_ACCESS_KEY':
```
mattbostock: Rather than make these environment variables globally available, what if we put them in the cron job itself? As the user should only be used for this backup job anyway.

modules/govuk_postgresql/manifests/backup.pp
```diff
@@ -34,4 +34,7 @@
         source => 'puppet:///modules/govuk_postgresql/etc/postgresql-backup-pre',
     }
 
+    # Include offsite backups to S3
+    include wal_e::backup
```
alexmuller: I think I'd find this a bit surprising if I was looking back at this code later. I'd look for this class in an external `wal_e` module or in a directory called `modules/wal_e/` rather than `modules/govuk_postgresql/manifests/wal_e/`. Is that just me?

surminus: Would it be better to explicitly state `govuk_postgresql::wal_e::backup` or actually just move it into it's entirely own `govuk_wale` module? I agree though, it's reads confusingly.

modules/govuk_postgresql/manifests/wal_e/backup.pp
surminus: I thought about putting them in the script but they're also used in the continuous archive command, plus they'd just be hardcoded in a script elsewhere so would be available to see anyway? Unless I'm misunderstanding.

surminus:
After a bit of consideration I thought this might be nicer as it's own module and a defined type, which takes into accounts doing multiple backups from the same server and also makes it easier to split off namespaces per backup.

PR #4169, opened 2016-03-01
Make draft-email-campaign-frontend available

deanwilson:
This vhost wasn't set, which means nothing is listening to this vhost.

@bradwright @mattbostock and I looked at the previous PR and this seems to fix the tests. Opened as a new PR in case it's wrong.

### DIFF
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -101,6 +101,7 @@ govuk::node::s_development::apps:
 govuk_app_enable_capistrano_layout: false
 govuk_app_enable_services: false
 
+govuk::apps::email_campaign_frontend::vhost: 'email-campaign-frontend'
 govuk::apps::asset_manager::enable_delayed_job_worker: false
 govuk::apps::asset_manager::mongodb_name: 'govuk_assets_development'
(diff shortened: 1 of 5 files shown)


hieradata/development.yaml
```diff
@@ -101,6 +101,7 @@ govuk::node::s_development::apps:
 govuk_app_enable_capistrano_layout: false
 govuk_app_enable_services: false
 
+govuk::apps::email_campaign_frontend::vhost: 'draft-email-campaign-frontend'
```
mattbostock: This is specific to draft, which might be confusing (I think non-draft would be better as a default).

bradwright: Updated and force pushed.

bradwright:
I know this pattern just follows the others, but I was wondering why we don't set the default `vhost` (eg `email-campaign-frontend` or `government-frontend`) in the Puppet class as a default argument value? Then we'd only need to override it on draft machines and not explicitly set it everywhere.

alexmuller:
I can't think of a good reason to not set a default. I think that would be the correct thing to do in this case.

bradwright:
Okay, so I'll do the default vhost refactor, but as a separate PR as this one seems functionally complete.

PR #4171, opened 2016-03-01
Ensure GPG key is re-imported when changed

mattbostock:
Story: https://trello.com/c/S8B99y46/120-asset-slave-backups-no-longer-working

Currently, if we change the GPG private key in Hiera data, the new key
is not imported.

This change ensures that the old key is deleted to allow the new key to
be imported.

Since `refreshonly` is set to `true`, this command should only run when
the private key file is changed and so I have removed the `unless`
option as it's unnecessary.

The GPG fingerprint must now be a full fingerprint for the `gpg --batch
--delete-secret-key` to work; it errors if I try to supply a full
fingerprint.

When deploying this, I'll need to:

- [x] Raise a corresponding PR in the deployment repo to change the
  GPG fingerprint to be full-length

- [ ] Remove the old secret key files using Fabric
  i.e. `/root/.gnupg/<SHORTFINGERPRINT>_secret_key.asc`

### DIFF
--- a/modules/backup/manifests/assets.pp
+++ b/modules/backup/manifests/assets.pp
@@ -50,6 +50,8 @@ class backup::assets(
   }
 
   if $backup_private_gpg_key and $backup_private_gpg_key_fingerprint {
+    validate_re($backup_private_gpg_key_fingerprint, '^[[:alnum:]]{40}$', 'Must supply full GPG fingerprint')
+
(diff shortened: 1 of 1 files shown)


surminus:
:+1: LGTM, merge when ready

mattbostock:
Related: https://github.gds/gds/deployment/pull/914

PR #4173, opened 2016-03-02
Refactor: use default vhost values

bradwright:
These were previously duplicated across config files, when they really only diverged from a standard value in the draft environments.

The only slightly weird refactor is the `contacts` one - more details in that commit message.

This follows up #4169


mattbostock:
:+1: Merge when ready.

PR #4174, opened 2016-03-02
Fix deploy job

bradwright:
Several deploys have failed as a result of the downstream CI process handing off an invalid app name to the deploy Jenkins.

The reason for this is that the app names need to exist *and* match the Capistrano deploy names we use.

Fix this by either repairing broken names or adding missing ones.


surminus:
:+1: lgtm

PR #4177, opened 2016-03-03
Allow jumpboxes to reboot unattended overnight

mattbostock:
The reason we had not previously allowed the jumpboxes to reboot
overnight was to avoid exacerbating an incident if someone was
investigating an issue out of hours (the time when the unattended
reboots occur) and was logged off mid-incident because the jumpbox was
rebooted.

This script checks if there are any users logged on so that the jumpbox
can reboot overnight.

Note that there's still the potential that the jumpbox never returns
from the reboot, or that it's still rebooting when someone needs to log
in, however I think that that scenario is rare enough that it's not
something we should worry about. We also have a second jumpbox.

* * *

I plan to refactor the way we're installing the check scripts to use a defined type rather than a new class for each check script but I'll do that in a later PR.

### DIFF
--- a/hieradata/class/jumpbox.yaml
+++ b/hieradata/class/jumpbox.yaml
@@ -1,4 +1,4 @@
 ---
 
-govuk::safe_to_reboot::can_reboot: 'careful'
-govuk::safe_to_reboot::reason: 'Warn connected users before rebooting'
+govuk::safe_to_reboot::can_reboot: 'overnight'
+govuk::safe_to_reboot::reason: 'Will only reboot if no users are connected'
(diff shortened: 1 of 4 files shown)


surminus:
Nice solution! I think also it was problematic because the jumpboxes are used heavily in env-syncs for data transfers? For a server like this it might even be worth pushing forward the time they do maintenance tasks (such as rebooting) to 11pm or something. 

I can just see them rebooting when the env-sync scripts try to connect which means the sync will probably timeout, abort and skip doing a sync completely that morning.

mattbostock:
@surminus: Good catch; I'll look at implementing a lock file (similar to #4097) to prevent that from happening.

PR #4179, opened 2016-03-04
Fix quoting in deploy Jenkins job

mattbostock:
Without the double quotes, the bash script appears in Jenkins as:

    export DEPLOY_TO=staging            export DEPLOY_TASK="$DEPLOY_TASK"

This is because the trailing newline is removed if the quotes are not
present.


surminus:
:+1: merge when ready

PR #4184, opened 2016-03-04
Add ikennaokpala user

ikennaokpala:
This PR adds Ikennaokpala user to GOV.UK integration environment.


benlovell:
I think that key might be a little weak.

ikennaokpala:
Cheers @benlovell I have updated it to use 2048 bits

PR #4185, opened 2016-03-04
Increase minimum SSH RSA key length to 4096 bits

mattbostock:
While 2048 bits likely provides sufficient security, using 4096 bits is
good practice and may provide better protection from cryptographic
attacks.

@ikennaokpala and @leelongmore will need to replace their existing keys before this can be merged.


benlovell:
Whoops, sorry for merging #4184 prematurely.

mattbostock:
@benlovell No need to apologise, it's something we've been moving closer towards and now feels like a good time to enforce 4096 bits as a minimum.

daibach:
Happy for this to be merged once @ikennaokpala and @leelongmore have updated their keys.

PR #4187, opened 2016-03-04
Enable email_campaign_api procfile worker

bradwright:
This should now be enabled everywhere.

Without this the worker keeps getting killed in production:

```(/Stage[main]/Govuk::Apps::Email_campaign_api/Govuk::Procfile::Worker[email-campaign-api]/Service[email-campaign-api-procfile-worker]/ensure) ensure changed 'running' to 'stopped'```


mattbostock:
:+1: Merge when green

PR #4190, opened 2016-03-07
Server uptime dashboard: Remove null metrics and make server names easier to read

mattbostock:
Please see individual commit messages.


surminus:
:+1: I trust this will make them better

PR #4191, opened 2016-03-07
Remove annotations from server uptime dashboard

mattbostock:
The annotations, which showed when deploys and application restarts
happen, made it more difficult to interpret the server uptime graphs.

Remove them.


surminus:
I agree with annotations don't add much or alternatively you could just disable as default in case people want them? I don't find it straight forward to interpret those graphs either way.

mattbostock:
@surminus: Hadn't thought of that, though in this case I think it's fine to remove them. It's also one less thing to maintain.

surminus:
I don't mind either way so go for it :+1:

PR #4192, opened 2016-03-07
Add environment variables for rummager S3 bucket

MatMoore:
Bucket name and region are defined per environment so we would prefer to
pass them to rummager rather than put them in config files.


tijmenb:
Looks good! :+1:

PR #4193, opened 2016-03-08
Re-add multipage-frontend to deployable apps...

steventux:
Multipage frontend was removed or omitted in a recent de-dupe / sanitization of deployable apps so re-add it.


mattbostock:
:+1: Merge when green.

PR #4194, opened 2016-03-08
Add theodorvararu user

tvararu:
:wave:

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -86,6 +86,7 @@ govuk_jenkins::config::admins:
   - stevelaing
   - stuartgale
   - tatianasoukiassian
+  - theodorvararu
   - tijmenbrommet
 
 govuk_jenkins::job_builder::jobs:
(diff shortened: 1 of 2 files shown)


hieradata/integration.yaml
```diff
@@ -87,6 +87,7 @@ govuk_jenkins::config::admins:
   - stuartgale
   - tatianasoukiassian
   - tijmenbrommet
+  - theodorvararu
```
benlovell: The alpha ordering here and below is a little out.

Sorry for the pedantry.

mattbostock:
Welcome @tvararu!

We're trying to increase the minimum SSH key length to 4096 bits (please see https://github.com/alphagov/govuk-puppet/pull/4185), which your key is currently below. Would you mind create a new key and raising a PR to add it?

The command you'll need is:
```
ssh-keygen -t rsa -b 4096
```

tvararu:
@mattbostock :ok_hand: On it!

PR #4195, opened 2016-03-08
Create the douglasroper user

dougdroper:
Add to integration


benlovell:
It looks as though that key may be a little weak, given the impending #4185

dougdroper:
@benlovell :+1: updated in #4199

PR #4196, opened 2016-03-08
Cron job for creating snapshots & jenkins job for pruning old snapshots

MatMoore:
https://trello.com/c/YAJY2zIl/359-implement-snapshot-restore-to-backup-elasticsearch-indexes

Only enabled on integration for the time being. I'd like to leave this running there for a few days and make sure its stable.

The frequency of the snapshotting is still open to discussion. It's currently every 10m. I don't think the frequency of snapshots will affect the storage used but it could influence the time other snapshot operations take to complete. The plan is to prune old snapshots overnight, but I'm not sure how long this will take to run yet. There is some discussion of the issue here: http://grokbase.com/t/gg/elasticsearch/14b7wv8a6p/what-is-the-best-practice-for-periodic-snapshotting-with-awc-cloud-s3

### DIFF
--- a/modules/govuk/templates/etc/cron.d/snapshot.erb
+++ b/modules/govuk/templates/etc/cron.d/snapshot.erb
@@ -1,1 +1,1 @@
-*/10 0-4,6-23 * * * <%= @snapshot_user %> /usr/local/bin/rummager-snapshot
+*/10 * * * * <%= @snapshot_user %> /usr/local/bin/rummager-snapshot
--- a/modules/govuk/templates/etc/cron.d/snapshot.erb
+++ b/modules/govuk/templates/etc/cron.d/snapshot.erb
@@ -1,0 +1,1 @@
+*/10 0-4,6-23 * * * <%= @snapshot_user %> /usr/local/bin/rummager-snapshot
--- a/modules/govuk_jenkins/manifests/job/prune_elasticsearch_snapshots.pp
+++ b/modules/govuk_jenkins/manifests/job/prune_elasticsearch_snapshots.pp
@@ -1,0 +1,26 @@
+# == Class: govuk_jenkins::job::prune_elasticsearch_snapshots
+#
+# Delete old Elasticsearch snapshots
+#
+class govuk_jenkins::job::prune_elasticsearch_snapshots (
+    $app_domain = hiera('app_domain'),
+) {
+
+  $check_name = 'prune_elasticsearch_snapshots'
+  $service_description = 'Prune Elasticsearch snapshots'
+  $job_url = "https://deploy.${app_domain}/job/prune_elasticsearch_snapshots"
+
+  file { '/etc/jenkins_jobs/jobs/prune_elasticsearch_snapshots.yaml':
+    ensure  => present,
+    content => template('govuk_jenkins/jobs/prune_elasticsearch_snapshots.yaml.erb'),
+    notify  => Exec['jenkins_jobs_update'],
+  }
+
+  $threshold_secs = 25 * 60 * 60
+
+  @@icinga::passive_check { "${check_name}_${::hostname}":
+    service_description => $service_description,
+    host_name           => $::fqdn,
+    freshness_threshold => $threshold_secs,
+  }
+}
--- a/modules/govuk_jenkins/manifests/job/prune_elasticsearch_snapshots.pp
+++ b/modules/govuk_jenkins/manifests/job/prune_elasticsearch_snapshots.pp
@@ -9,6 +9,7 @@ class govuk_jenkins::job::prune_elasticsearch_snapshots (
   $check_name = 'prune_elasticsearch_snapshots'
   $service_description = 'Prune Elasticsearch snapshots'
   $job_url = "https://deploy.${app_domain}/job/prune_elasticsearch_snapshots"
+  $snapshot_lock_path = '/var/run/elasticsearch-snapshot.lock'
 
   file { '/etc/jenkins_jobs/jobs/prune_elasticsearch_snapshots.yaml':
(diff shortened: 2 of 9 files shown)


MatMoore:
We've just added snapshot pruning to the rummager jenkins job that runs overnight. I'd like to test this manually on integration to make sure it runs in a reasonable amount of time with a days worth of snapshots.

Note that this is problematic for integration at the moment because integration will be down when this job runs. After we've tested this we can switch off snapshotting in integration for the time being.

I think we should follow up this work by syncing the S3 buckets from production -> integration when the environment is recreated. What do you think?

MatMoore:
I've tested this some more on integration today, and the cron and icinga monitoring seems to work.

I noticed that the snapshot deletion in elasticsearch doesn't like it when snapshots are being taken at the same time. I've worked around this by scheduling the cron to stop in the morning when the jenkins job runs.

modules/govuk_jenkins/manifests/job/prune_elasticsearch_snapshots.pp
```diff
@@ -0,0 +1,26 @@
...
+#
+# Delete old Elasticsearch snapshots
+#
+class govuk_jenkins::job::prune_elasticsearch_snapshots (
+    $app_domain = hiera('app_domain'),
+) {
+
+  $check_name = 'prune_elasticsearch_snapshots'
+  $service_description = 'Prune Elasticsearch snapshots'
+  $job_url = "https://deploy.${app_domain}/job/prune_elasticsearch_snapshots"
```
surminus: This variable isn't used anywhere, it needs to be added to the passive check as:
`action_url => $job_url`

jackscotti: Updated, thanks.

modules/govuk_jenkins/templates/jobs/prune_elasticsearch_snapshots.yaml.erb
```diff
@@ -0,0 +1,34 @@
...
+    name: prune-elasticsearch-snapshots
+    display-name: prune-elasticsearch-snapshots
+    project-type: freestyle
+    description: "<p>Delete old Elasticsearch snapshots</p>
+      <p>This is run to delete old (7 days) Elasticsearch snapshots.</p>"
+    builders:
+        - shell: |
+            ssh deploy@search-1.api '(cd /var/apps/rummager; RUMMAGER_INDEX=all govuk_setenv rummager bundle exec rake rummager:snapshot:clean)'
+    triggers:
+        - timed: '5 4 * * *' # to update
```
mattbostock: Does this line need updating?

modules/govuk/templates/etc/cron.d/snapshot.erb
```diff
@@ -0,0 +1 @@
+*/10 0-4,6-23 * * * <%= @snapshot_user %> /usr/local/bin/rummager-snapshot
```
mattbostock: Any reason not to run this 7 days a week? Are we avoiding 0-6am because of the data sync?

It'd be good to avoid times and use a lock if possible, otherwsie we end up with a lot of timing information that we need to keep in lockstep.

MatMoore: The intention here is to run every day of the week, those ranges should be the hour setting. The time I'm trying to avoid is when jenkins runs the pruning job, as these operations share resources in elasticsearch.

I'm not sure exactly how to do the locking, because the snapshotting job doesn't block at the moment - elasticsearch gives a response straight away and continues processing the snapshot in the background. It's possible to make that request hang until it's done via a request parameter though, which means we could put a setlock around the cron if you prefer. See https://www.elastic.co/guide/en/elasticsearch/reference/1.4/modules-snapshots.html#_snapshot 

It wouldn't be foolproof, because you could always issue a snapshot/delete some other way, that would make elasticsearch go unresponsible, but maybe its still better than what we have now. I don't the think the documentation says explicitly whether the DELETE operation is supposed to block until its done, but the testing I've done so far suggests this is the case. I'm also curious how long is this going to take when we have a backlog of snapshots built up.

modules/govuk_jenkins/templates/jobs/prune_elasticsearch_snapshots.yaml.erb
MatMoore: Possibly, although early morning seems like a reasonable time to run this. What do you think?

MatMoore:
I've just rebased this branch and removed some unwanted configuration from the pruning job.

modules/govuk/templates/etc/cron.d/snapshot.erb
mattbostock: Sorry, I misread the cron timing.

I found [this](http://crontab.guru/#*/10_0-4,6-23_*_*_*) helpful to interpret it. It runs at every hour except for at 5am.

So if the pruning and snapshotting runs at the same time, Elasticsearch becomes unresponsive? I think a lock would be a good idea as relying on timing is liable to someone changing times elsewhere and not understanding this dependency.

MatMoore: Not unresponsive, it will just return a 503 straight away.

I'll take a look at adding a setlock around both tasks. We'll need to change the rummager rake task first to pass the extra param.

For the snapshot cron we'll use `setlock -n` (give up if locked) - this way we can have the cron execute frequently and get multiple processes waiting on the lock.

For the pruning job we can use `setlock -N` (wait for lock).

MatMoore:
I'm going to close this PR for now as we're going to test against a newer version of elasticsearch first.

PR #4197, opened 2016-03-08
Expose HTTP_USERNAME and HTTP_PASSWORD to service-manual-publisher

AndrewVos:
This is so that we can hit www-origin servers from inside the app, with the
correct http basic auth:
https://github.com/alphagov/service-manual-publisher/pull/175

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -351,6 +351,9 @@ govuk::apps::publishing_api::db::backend_ip_range: "%{hiera('environment_ip_pref
 govuk::apps::rummager::nagios_memory_warning: 4600
 govuk::apps::rummager::nagios_memory_critical: 4900
 
+govuk::apps::service_manual_publisher::http_username: "%{hiera('http_username')}"
+govuk::apps::service_manual_publisher::http_password: "%{hiera('http_password')}"
+
 govuk::apps::service_manual_publisher::db::backend_ip_range: "%{hiera('environment_ip_prefix')}.3.0/24"
 
 govuk::apps::signon::redis_url: "redis://redis-1.backend:6379/0"
--- a/modules/govuk/manifests/apps/service_manual_publisher.pp
+++ b/modules/govuk/manifests/apps/service_manual_publisher.pp
@@ -40,6 +40,14 @@
 #   The bearer token to use when communicating with Asset Manager.
 #   Default: undef
 #
+# [*http_username*]
+#   The http basic auth username when running on integration.
+#   Default: undef
+#
+# [*http_password*]
+#   The http basic auth password when running on integration.
+#   Default: undef
+#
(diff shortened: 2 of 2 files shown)


modules/govuk/manifests/apps/service_manual_publisher.pp
```diff
@@ -86,6 +88,12 @@
     "${title}-ASSET_MANAGER_BEARER_TOKEN":
       varname => 'ASSET_MANAGER_BEARER_TOKEN',
       value   => $asset_manager_bearer_token;
+    "${title}-HTTP_USERNAME":
```
benlovell: Can you document these? The rest are well documented.

hieradata/common.yaml
```diff
@@ -351,6 +351,9 @@ govuk::apps::publishing_api::db::backend_ip_range: "%{hiera('environment_ip_pref
 govuk::apps::rummager::nagios_memory_warning: 4600
 govuk::apps::rummager::nagios_memory_critical: 4900
 
+govuk::apps::service_manual_publisher::http_username: "%{hiera('http_username')}"
+govuk::apps::service_manual_publisher::http_password: "%{hiera('http_password')}"
+
```
danielroseman: Don't see any need for these here. That's only useful if you're using the value from other hiera files and modifying/extending it somewhere here, which you're not.

modules/govuk/manifests/apps/service_manual_publisher.pp
```diff
@@ -52,6 +52,8 @@
   $secret_key_base = undef,
   $publishing_api_bearer_token = undef,
   $asset_manager_bearer_token = undef,
+  $http_username = undef,
```
danielroseman: need to add these new params to the documentation string above.

AndrewVos: Yep just did that.

hieradata/common.yaml
AndrewVos: Are you saying that removing them would make no difference? Confused

danielroseman: My mistake, I though these were going to be defined in separate hieradata files for integration/staging/prod. In this case since you're referencing things in the same file it makes sense.

AndrewVos: Oh cool! Was really confused for a bit there.

modules/govuk/manifests/apps/service_manual_publisher.pp
```diff
@@ -40,6 +40,14 @@
 #   The bearer token to use when communicating with Asset Manager.
 #   Default: undef
 #
+# [*http_username*]
+#   The http basic auth username when running on integration.
+#   Default: undef
+#
+# [*asset_manager_bearer_token*]
```
danielroseman: wrong param name

PR #4200, opened 2016-03-10
Update my SSH key length to 4096 bits

tvararu:
As requested by @mattbostock in https://github.com/alphagov/govuk-puppet/pull/4194.


mattbostock:
Thanks @tvararu!

PR #4202, opened 2016-03-11
Remove Alex Muller from GPG recipients

mattbostock:
@alexmuller's GPG key has expired, meaning that attempts to encrypt credentials
(phony credentials in the case of our Vagrant environment) fails with a
'general error'.


surminus:
:+1: LGTM, just need to remember to prompt @alexmuller into adding these back when he's back tomorrow. Will this also cause problems in the deployment repo?

mattbostock:
@surminus It will, see https://github.gds/gds/deployment/pull/917, for example.

On second thoughts, let's close this and wait until tomorrow.

PR #4204, opened 2016-03-11
Use dedicated disk for Duplicity cache

mattbostock:
Story: https://trello.com/c/S8B99y46/120-asset-slave-backups-no-longer-working

__Depends on: https://github.gds/gds/govuk-provisioning/pull/526__

Move the Duplicity cache onto its own disk, which will allow us to
monitor the disk space used by assets synced from the asset master
without the size of the cache directory interfering.

We perform offsite backups using Duplicity on the asset-slave-1
machines. Duplicity maintains a cache which is uses to know what is
already backed up when performing incremental backups. Otherwise, it
would have to retrieve the backups from the remote and decrypt them.

We recently started rsyncing /mnt/uploads from the asset-master-1 to
the asset-slaves and added an Icinga check (ee785db7) to ensure that the
free space on each machine corresponds (to confirm that the sync is
working correctly). Since the Duplicity cache directory exists on the
asset-slave but not the asset-master, this Icinga check is raising false
alarms.

### DIFF
--- a/hieradata/class/asset_slave.yaml
+++ b/hieradata/class/asset_slave.yaml
@@ -1,6 +1,6 @@
 ---
 
-backup::assets::archive_directory: '/mnt/uploads/.cache/duplicity/'
+backup::assets::archive_directory: '/mnt/duplicity-cache'
 
 govuk_safe_to_reboot::can_reboot: 'careful'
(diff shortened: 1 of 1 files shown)


surminus:
:+1: This looks fine, merge when ready.

jonauman:
:+1:

deanwilson:
Merge away.

PR #4206, opened 2016-03-14
Enable rummager message queue in production

tijmenb:
This commit enables the Rummager message queue consumer in production and staging.

Trello: https://trello.com/c/lRnxI2x5


mattbostock:
:+1: Merge when green.

PR #4205, opened 2016-03-14
Temporarily prevent application restarts

mattbostock:
...while we deploy the changes in fc8532d8. Having all the applications
restart at once is too disruptive.

Once this is deployed, this line can be uncommented again.

PR #4207, opened 2016-03-14
Remove deployed FIXMEs

mattbostock:
All of these FIXMEs have been deployed or the thing they rely upon has
been actioned.

This also re-enables application restarts, see: #4205.


surminus:
:+1: merge when green

PR #4208, opened 2016-03-14
Add mongodb settings for Publisher.

danielroseman:
Publisher now expects a single MONGODB_URI env var.


mattbostock:
:+1: Merge when ready.

PR #4210, opened 2016-03-14
Enable MapIt in staging

sihugh:
New MapIt is currently enabled in integration, however it requires extra
configuration to be able to work in staging.


mattbostock:
:+1: Merge when ready.

PR #4211, opened 2016-03-14
Remove unused Icinga service template

mattbostock:
I don't think that this service template is used.


alexmuller:
I agree, as of 3cb4aab4dc2a3d3b17ee477273fa70d462b4f9c4 :+1:

PR #4214, opened 2016-03-15
Add all our built rubies to the monitoring host

deanwilson:
Smokey needs newer rubies and developers assume we have the
latest and greatest inline with the versions applications use.
This should suffice until we agree and merge the refactored
hieradata/application approach.


AndrewVos:
Nice!

PR #4216, opened 2016-03-15
Deploy share_sale_publisher to all environments.

steventux:
It's now being released to all the environments.


surminus:
:+1: LGTM, merge when ready

PR #4217, opened 2016-03-15
Improve ClamAV freshness monitoring

alexmuller:
Instead of monitoring the age of the file (which is frequently broken for example because the file isn't updated) we should check the value that ClamAV publish over DNS and make sure it isn't too out-of-date.

Being a version behind should be ok.

### DIFF
--- a/modules/clamav/files/usr/local/bin/check_clamav_definitions
+++ b/modules/clamav/files/usr/local/bin/check_clamav_definitions
@@ -1,26 +1,15 @@
-#!/bin/bash
+#!/usr/bin/env ruby
 
-set -euo pipefail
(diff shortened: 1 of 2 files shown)


mattbostock:
@alexmuller has lint.

mattbostock:
Assuming that `dig` etc are on the `PATH`, merge when ready :+1:

alexmuller:
Good point, updated to use absolute paths to the binaries because I can't be bothered to think about `PATH` :)

mattbostock:
:+1:

deanwilson:
Can we wrap the DNS call to return a controlled "UNKNOWN" and message rather than a stack trace on dns failure?

PR #4218, opened 2016-03-16
Disable rummager message queue in production

tijmenb:
This PR reverts alphagov/govuk-puppet#4206.

We are experiencing difficulties on production with the message queue consumer. It seems the indexing of the messages is generating too much load on elasticsearch, causing other (more important) processes to error as well.

We'd like to disable the indexing until we've replicated the issue on integration, established a root cause and have proper monitoring in place.


mattbostock:
:+1: Merge when ready.

PR #4220, opened 2016-03-16
Enable new mapit for deployment to production

brenetic:
- Enable mapit in the backend api's section for deployment to production.
- Removed temporary overriding of the load balancer configuration in integration which was done in [this previous commit](https://github.com/alphagov/govuk-puppet/commit/73d06386725379d85708083ca08870c08394eaab).


alexmuller:
This will switch the loadbalanced address of mapit to point to the new machines rather than the old ones. Merge when you're ready :+1:

sihugh:
Hold off for a bit, we're going to do Staging first.

PR #4221, opened 2016-03-16
New frontend service-manual-frontend

thehenster:
We are adding a new frontend, service-manual-frontend, which mimics government-frontend.

https://github.com/alphagov/service-manual-frontend is a fork of government-frontend and we are in the process of stripping out the non-service manual related stuff.

### DIFF
--- a/modules/govuk/manifests/apps/service_manual_frontend.pp
+++ b/modules/govuk/manifests/apps/service_manual_frontend.pp
@@ -11,10 +11,25 @@
 # [*port*]
 #   What port should the app run on?
 #
+# [*enabled*]
+#   Is the app enabled
+#
+# [*errbit_api_key*]
+#   Errbit API key used by airbrake
+#   Default: ''
+#
+# [*secret_key_base*]
+#   Used to set the app ENV var SECRET_KEY_BASE which is used to configure
+#   rails 4.x signed cookie mechanism. If unset the app will be unable to
+#   start.
+#   Default: undef
+#
 class govuk::apps::service_manual_frontend(
   $vhost = 'service-manual-frontend',
   $port = 3122,
   $enabled = false,
+  $errbit_api_key = undef,
+  $secret_key_base = undef,
 ) {
   if $enabled {
     Govuk::App::Envvar {
@@ -30,6 +45,15 @@ class govuk::apps::service_manual_frontend(
       asset_pipeline        => true,
       asset_pipeline_prefix => 'service-manual-frontend',
       vhost                 => $vhost,
+    }
+
+    govuk::app::envvar {
+      "${title}-ERRBIT_API_KEY":
+        varname => 'ERRBIT_API_KEY',
+        value   => $errbit_api_key;
+      "${title}-SECRET_KEY_BASE":
+        varname => 'SECRET_KEY_BASE',
+        value   => $secret_key_base;
     }
   }
 }
--- a/modules/govuk/manifests/apps/service_manual_frontend.pp
+++ b/modules/govuk/manifests/apps/service_manual_frontend.pp
@@ -13,7 +13,7 @@
 #
 class govuk::apps::service_manual_frontend(
   $vhost = 'service-manual-frontend',
-  $port = '3090',
+  $port = 3122,
 ) {
   Govuk::App::Envvar {
     app => 'service-manual-frontend',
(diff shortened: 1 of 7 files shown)


modules/govuk/manifests/apps/service_manual_frontend.pp
```diff
@@ -0,0 +1,32 @@
...
+# [*vhost*]
+#   Virtual host used by the application.
+#   Default: 'service-manual-frontend'
+#
+# [*port*]
+#   What port should the app run on?
+#
+class govuk::apps::service_manual_frontend(
+  $vhost = 'service-manual-frontend',
+  $port = '3090',
```
thehenster: This is the government-frontend's port number. We need a new one.

alexmuller: Add a line in [this file](https://github.gds/gds/development/blob/master/Procfile)

thehenster: Just checking - it doesn't have to go anywhere else does it? This manifest feels production-y and that Procfile feels development-y. I guess the port isn't used in production?

alexmuller: Nope, the development Procfile is how we assign ports - you're right that it feels weird but that's how we do it :). The port is used in production for the app to spin up onto and it's the place nginx will proxy requests to.

thehenster: (it's been added to the Profile and Pinfile btw https://github.gds/gds/development/pull/302)

modules/govuk/manifests/apps/service_manual_frontend.pp
```diff
@@ -0,0 +1,32 @@
...
+# [*vhost*]
+#   Virtual host used by the application.
+#   Default: 'service-manual-frontend'
+#
+# [*port*]
+#   What port should the app run on?
+#
+class govuk::apps::service_manual_frontend(
+  $vhost = 'service-manual-frontend',
+  $port = '3122',
```
mattbostock: Can you please make this an integer?

modules/govuk/manifests/apps/service_manual_frontend.pp
```diff
@@ -0,0 +1,32 @@
...
+    app_type              => 'rack',
+    port                  => $port,
+    vhost_ssl_only        => true,
+    health_check_path     => '/healthcheck',
+    legacy_logging        => false,
+    asset_pipeline        => true,
+    asset_pipeline_prefix => 'service-manual-frontend',
+    vhost                 => $vhost,
+  }
+}
```
dsingleton: For new apps we should be using [12factor](http://12factor.net/) approach to configuration, where we use ENV vars to popular secret values ([multipage frontend](https://github.com/alphagov/govuk-puppet/blob/fd51ac0c86f9c481650130d8b21c155d041e5b8a/modules/govuk/manifests/apps/multipage_frontend.pp#L40-L47)'s puppet config is a good example, and [how it consumes the values](https://github.com/alphagov/multipage-frontend/blob/ee3bf30b4bdf6aa117fde8f40edeec70a6058b1c/config/initializers/airbrake.rb)). 

The values for the ENV vars come from a private configuration repo, which are injected in to the app when puppet is run - this is the private repo someone will need to help you change, as you won't have access.

`government-frontend` actually predates this approach, and these values are inserted by overwriting initialiser files - this isn't something we should copy for new apps. Could you update the puppet cdefinition to use the same pattern as `multipage-frontend`?

dsingleton: Actually, https://github.com/alphagov/govuk-puppet/pull/3884 is a good example of adding `ENV` vars to an existing application definition.

You might also want to add a [feature flag](https://github.com/alphagov/govuk-puppet/commit/75c0b78f822fa939c7b9ea15cc799349f7c669b0) so the app can be enabled in integration only, while setting the app up.

thehenster: Thanks for all of this. Coming up.

thehenster:
`enabled` flag added and we have added environment variables for errbit and the secret_key_base.

For the next step we might need to come and join someone to see what initializers are being overwritten in government-frontend. Perhaps there are more env vars we could extract.

boffbowsh:
👍 for the bits I didn't do

alexmuller:
This all looks sensible to me :+1:

dsingleton:
👍  from me too

alexmuller:
Oop, I think you'll need to enable this in the integration hieradata.

boffbowsh:
Thanks, updated.

PR #4222, opened 2016-03-17
Rework transition manifest to be more modern

h-lame:
Mostly this is so we get the secret_key_base into env to match up with
how rails 4.1+ apps expect secrets to be delivered.  We also enable
asset_pipeline because it seems that other backend apps with UI (e.g.
publisher) have this.

This is required for https://github.com/alphagov/transition/pull/508 and will require deployment changes to support it (https://github.gds/gds/deployment/pull/934).


alexmuller:
This looks correct to me, merge whenever you like :+1:

PR #4226, opened 2016-03-18
Remove departmental projects publish job

ssatish:
as github publishing is not in use anymore.
Share-sale-publisher is used to publish content to
email-campaign-frontend.


mattbostock:
Confirmed that this will be [purged automatically](https://github.com/alphagov/govuk-puppet/blob/3c6e2d7428ba0fcfdba41c7de9f903be797460ec/modules/govuk_jenkins/manifests/job_builder.pp#L56).

PR #4231, opened 2016-03-21
Create Jenkins job for terraform deployments

surminus:
We require a Jenkins job so we can keep track of deployments for Terraform. This commit adds the job manifest and template, and also adds in the hieradata required, including the list of projects available to
deploy.

Story: https://trello.com/c/FOG8jmb5/53-create-jenkins-job-to-deploy-aws-configuration

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_terraform_project.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_terraform_project.yaml.erb
@@ -1,0 +1,47 @@
+---
+- scm:
+    name: govuk-terraform-provisioning_Deploy_Terraform_Project
+    scm:
+        - git:
+            url: git@github.com:alphagov/govuk-terraform-provisioning.git
+            branches:
+                - master
+- job:
+    name: Deploy_Terraform_Project
+    display-name: Deploy_Terraform_Project
+    project-type: freestyle
+    description: |
+      Deploy a specific Terraform project in <%= @environment -%>.
+    properties:
+        - github:
+            url: https://github.com/alphagov/govuk-terraform-provisioning/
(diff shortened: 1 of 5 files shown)


mattbostock:
Merge when ready :+1:

surminus:
I actually had a thought about this on the way home yesterday, instead of doing an `export` in the builder shell, I think we can actually do an inject of parameters somewhere? That would be a lot cleaner.

mattbostock:
@surminus Which `export` are you thinking of?

modules/govuk_jenkins/templates/jobs/deploy_terraform_project.yaml.erb
```diff
@@ -0,0 +1,47 @@
...
+              - master
+- job:
+    name: Deploy_Terraform_Project
+    display-name: Deploy_Terraform_Project
+    project-type: freestyle
+    description: |
+      Deploy a specific Terraform project in <%= @environment -%>.
+    properties:
+        - github:
+          url: https://github.com/alphagov/govuk-terraform-provisioning/
```
alexmuller: I think this needs a bit of indentation of the `url` to be valid YAML.

surminus:
@mattbostock I updated the export bit. It originally did `export DEPLOY_ENV=foo` in the builder shell, but I updated it [here](https://github.com/alphagov/govuk-puppet/blob/terraform-jenkins-job/modules/govuk_jenkins/templates/jobs/deploy_terraform_project.yaml.erb#L18-L21)

surminus:
@alexmuller thanks, amended the yaml so hopefully it's correct now

mattbostock:
:+1:

PR #4233, opened 2016-03-21
Upgrade puppetdb to 2.0.0

alexmuller:
This commit upgrades puppetdb on precise machines to 2.0.0.

2.0.0 requires a newer version of the JDK than 6. 6 seems to be provided by the old puppetdb package so I've uninstalled it and manually installed 7.

I've also modified the upstart file to allow puppetdb to start - the change I've made here is similar to the change the package made to its own init file (which we purge).

### DIFF
--- a/hieradata/common.precise.yaml
+++ b/hieradata/common.precise.yaml
@@ -9,6 +9,4 @@ nginx::package::version: '1.4.4-1~precise0'
 
 postgresql::globals::version: '9.1'
 
-puppet::master::puppetdb_version: '1.3.2-1puppetlabs1'
-
 resolvconf::options: []
--- a/modules/puppetdb/manifests/package.pp
+++ b/modules/puppetdb/manifests/package.pp
@@ -8,4 +8,12 @@ class puppetdb::package($package_ensure) {
     require => Class['puppet::package'],
   }
 
+  # FIXME: Remove once deployed to production
+  package { 'openjdk-6-jre-headless':
+    ensure => 'absent',
+  }
+  package { 'openjdk-6-jre-lib':
+    ensure => 'absent',
+  }
+
 }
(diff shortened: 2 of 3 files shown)


mattbostock:
I wonder if now's a good time for us to go back to using the SysV init script provided by the PuppetDB package, instead of providing our own Upstart script. It'd be less for us to maintain and we use SysV for a lot of other services.

mattbostock:
LGTM, merge when ready :+1:

alexmuller:
Actually I lied just now when I said it doesn't install a new JDK:

```
vagrant@puppetmaster-1:~$ apt-cache policy puppetdb | grep Installed
  Installed: 2.0.0-1puppetlabs1
vagrant@puppetmaster-1:~$ apt-cache depends puppetdb
puppetdb
 |Depends: <java7-runtime-headless>
    openjdk-7-jre-headless
    oracle-java7-installer
  Depends: <j2re1.7>
  Depends: adduser
 |Depends: puppet
  Depends: <puppet-agent>
  Suggests: postgresql
```

Still ok or shall we remove the explicit JDK installation?

mattbostock:
Probably good to be explicit though I'm ambivalent.

surminus:
If the package pulls the dependency I'd say it's better to not be explicit as it's another thing to potentially change in the future. Otherwise looks good :+1:

deanwilson:
Are we moving to 2.0 as it's the last version compatible with Puppet 3.* ?

alexmuller:
I didn't know that was the case, but picked 2.0 because it's a version with packages available for both 12.04 and 14.04 so we can upgrade the operating system independent of the package.

alexmuller:
Going to close this until I've had a chance to test it more, I think it's breaking sometimes.

PR #4234, opened 2016-03-22
Bugfix Service Description

surminus:
For some reason Puppet created the passive check details as having the incorrect service description ("AutoPostgreSQL backup"). This may be because the class and parameter both share the same name. I assumed this wouldn't cause a problem, but I've made it unique to test and ensure this fixes it.


deanwilson:
Merge when green.

PR #4241, opened 2016-03-23
Encrypted incremental MySQL backups

surminus:
This strives to create incremental backups to an S3 bucket which
are no more than 15 minutes out of date.

Functionally it uses Percona Xtrabackup (https://www.percona.com/doc/percona-xtrabackup/2.4/index.html), using the innobackupex wrapper (https://www.percona.com/doc/percona-xtrabackup/2.2/innobackupex/innobackupex_script.html) and gof3r (https://github.com/rlmcpherson/s3gof3r) to take the stream output and push to an S3 bucket.

It's recommend use is to take a base backup and then incremental backups on top of that. It also compresses and encrypts on the fly.

Within the bucket, it creates a new folder for each set of backups, and does this by the base backup echoing a date into a file, which the incremental backup then references. This is so it splits up the daily
backups and hopefully makes it easier to restore.

To test this in Integration it requires an [S3 bucket and user creating](https://github.com/alphagov/govuk-terraform-provisioning/pull/30).


modules/govuk_mysql/manifests/xtrabackup/packages.pp
```diff
@@ -0,0 +1,21 @@
+# == Class: Govuk_mysql::Xtrabackup::Packages
+#
+class govuk_mysql::xtrabackup::packages (
+  $s3gof3r_latest  = 'https://github.com/rlmcpherson/s3gof3r/releases/download/v0.5.0/gof3r_0.5.0_linux_amd64.tar.gz',
```
mattbostock: I think we should package this as otherwise the owner of that repo could replace that version without our knowledge.

Perhaps it's a good opportunity to try something like FPM for packaging instead of our debhelper scripts?

modules/govuk_mysql/manifests/xtrabackup/backup.pp
```diff
@@ -0,0 +1,113 @@
...
+#
+# [*s3_bucket_name*]
+#   The unique name of the S3 bucket.
+#
+# [*encryption_key*]
+#   The encryption key used to encrypt the backups.
+#
+# [*base_backup_cron_minute*]
+#   The minute(s) that the daily base backup(s) run(s).
+#   Default: 10
```
mattbostock: I think we could remove the defaults from the documentation since they're likely to get out of date quickly.

modules/govuk_mysql/manifests/xtrabackup/backup.pp
```diff
@@ -0,0 +1,113 @@
...
+  $incremental_backup_cron_minute = '*/15',
+  $enabled = false,
+) {
+  if $enabled {
+    include govuk_mysql::xtrabackup::packages
+
+    file {['/etc/mysql/xtrabackup', '/etc/mysql/xtrabackup/env.d']:
+      ensure => 'directory',
+      mode   => '0755',
+      owner  => 'root',
```
mattbostock: We set the `owner` and `group` to `root` [by default](https://github.com/alphagov/govuk-puppet/blob/d41593b133bfc1a2984cb8b4870f2fefd79e7377/manifests/site.pp#L3-L7).

modules/govuk_mysql/manifests/xtrabackup/backup.pp
```diff
@@ -0,0 +1,113 @@
...
+      mode    => '0755',
+    }
+
+    @@icinga::passive_check { "check_mysql_xtrabackup_daily_base_push-${::hostname}":
+        service_description => $base_service_desc,
+        freshness_threshold => $base_threshold_secs,
+        host_name           => $::fqdn,
+      }
+
+    $incremental_threshold_secs = 20 * 60
```
mattbostock: Should we make this value lower, since we don't want to be more than 15 minutes behind the most current data?

What happens if we run the incremental script more often, and what prevents it from running two incremental backups at the same time if the old one hasn't finished?

surminus: Overlooked the threshold, thanks.

The [cronjob](https://github.com/alphagov/govuk-puppet/blob/encrypted-incremental-mysql-backups/modules/govuk_mysql/templates/etc/cron.d/xtrabackup_s3) deals with any overlapping work, using `setlock -N` and `setlock -n` respectively.

modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_base
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+# Takes an argument of either 'base' or 'incremental'
```
mattbostock: Is this comment still correct?

modules/govuk_mysql/templates/etc/cron.d/xtrabackup_s3
```diff
@@ -0,0 +1,3 @@
+MAILTO=""
+<% @base_backup_cron_minute %> <% @base_backup_cron_hour %> * * * setlock -N /var/run/xtrabackup /usr/local/bin/xtrabackup_s3_base
```
mattbostock: Should we include `mysql` in the lock name to make it clear what `xtrabackup` is used for?

modules/govuk_mysql/manifests/xtrabackup/backup.pp
mattbostock: Now that this threshold has been lowered, we'd need to change the default on `incremental_backup_cron_minute` so that we don't raise false alerts.

What would the impact be do you think in terms of bandwidth, disk space and recovery time if we ran the job every 5 minutes and relied on `setlock` to avoid backups from running into each other?

modules/govuk_mysql/manifests/xtrabackup/backup.pp
```diff
@@ -0,0 +1,102 @@
...
+      ensure  => 'present',
+      content => template('govuk_mysql/usr/local/bin/xtrabackup_s3_base'),
+      mode    => '0755',
+    }
+
+    @@icinga::passive_check { "check_mysql_xtrabackup_daily_base_push-${::hostname}":
+        service_description => $base_service_desc,
+        freshness_threshold => $base_threshold_secs,
+        host_name           => $::fqdn,
+      }
```
deanwilson: Incorrect indentation

modules/govuk_mysql/manifests/xtrabackup/backup.pp
```diff
@@ -0,0 +1,102 @@
...
+      ensure  => 'present',
+      content => template('govuk_mysql/usr/local/bin/xtrabackup_s3_incremental'),
+      mode    => '0755',
+    }
+
+    @@icinga::passive_check { "check_mysql_xtrabackup_incremental_push-${::hostname}":
+        service_description => $incremental_service_desc,
+        freshness_threshold => $incremental_threshold_secs,
+        host_name           => $::fqdn,
+      }
```
deanwilson: Incorrect indentation

modules/govuk_mysql/manifests/xtrabackup/backup.pp
```diff
@@ -0,0 +1,102 @@
...
+# [*enabled*]
+#   If this is true then the packages are installed and backups are triggered.
+#
+class govuk_mysql::xtrabackup::backup (
+  $aws_access_key_id = undef,
+  $aws_secret_access_key = undef,
+  $aws_region = 'eu-west-1',
+  $s3_bucket_name = undef,
+  $encryption_key = undef,
+  $base_backup_cron_minute = '10',
```
deanwilson: Minute and hour should be ints, not strings.

modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_incremental
```diff
@@ -0,0 +1,36 @@
...
+if [ $? == 0 ]
+  then
+    STATUS=0
+  else
+    STATUS=1
+fi
+
+if [ $STATUS -eq 0 ]; then
+  NAGIOS_CODE=0
+  NAGIOS_MESSAGE="OK: MySQL Xtrabackup daily incremental to S3 succeeded"
```
deanwilson: This isn't a daily thing in cron so the message may be incorrect.

deanwilson:
As a general approach I'd rather have the `if` wrap the include than every resource inside the class itself. But that might be a personal thing and we can discuss it in the team meeting.

rjw1:
So, is there way we could use gpg encryption. AES is good enough but we probably just need to check with Paul Colman before we backup any sensitive data. /cc @daibach

rjw1:
Do we want this on the master instead of the backup slave?

surminus:
Addressed comments. @rjw1 see the [encryption options for the innobackupex tool](https://www.percona.com/doc/percona-xtrabackup/2.2/innobackupex/encrypted_backups_innobackupex.html). Using GPG might be complicated because of how this script pipes into the `gof3r` tool. In theory we could pipe into GPG... but not sure how that would then pipe in `gof3r` to upload to S3 in the current mechanism, and they might need to be written to disk, encrypted and then uploaded, which is less efficient. 

tl;dr I'll check the options

surminus:
The gof3r package now gets installed as a .deb package.

mattbostock:
@rjw1 What features of GPG is that we'd take advantage of? For example, you can ask GPG to use AES.

surminus:
@rjw1 also probably better on a slave, yes.

jonauman:
Additionally, we can encrypt the backups in S3 using the '-m x-amz-server-side-encryption:AES256' switch with the gof3r command.

mattbostock:
@rjw1 FWIW, innobackupex uses libgcrypt [1][], which is based on GPG [2][].

[1]: https://www.percona.com/doc/percona-xtrabackup/2.2/innobackupex/encrypted_backups_innobackupex.html
[2]: https://www.gnu.org/software/libgcrypt/

modules/govuk_mysql/manifests/xtrabackup/packages.pp
```diff
@@ -0,0 +1,20 @@
+# == Class: Govuk_mysql::Xtrabackup::Packages
+#
+# === Parameters
+#
+# [*apt_mirror_hostname*]
+#   The hostname of an APT mirror
+#
+class govuk_mysql::xtrabackup::packages (
+  $apt_mirror_hostname  = '',
```
mattbostock: Where is this class parameter used?

rjw1:
So I have no real objection to the encryption so Im happy to  have this merged.

surminus:
I've updated some stuff and added a test, so as per @rjw1's comment I'll merge soon

rjw1:
I think the commits may need squashing a bit.

PR #4242, opened 2016-03-23
Fix ordering of variable

surminus:
I always forget that a variable must be declared before creating a template that uses that variable.


mattbostock:
:+1:

PR #4244, opened 2016-03-23
Use new boxes for Virtualbox in Vagrant

alexmuller:
- The boxes from 21 March did not have Puppet installed properly
- The Vagrantfile for bringing up specific nodes was not updated to use those 21 March boxes but should have been


mattbostock:
:+1: Merge when ready.

PR #4245, opened 2016-03-23
Add aptly mirror for elasticsearch 1.7

MatMoore:
This creates an aptly mirror so we can start testing 1.7. We plan to update api elasticsearch to this soon.

The key is the fingerprint documented at https://www.elastic.co/guide/en/elasticsearch/reference/1.7/setup-repositories.html


alexmuller:
Nice, thank you!

PR #4246, opened 2016-03-26
More accurate puppet run reporting

boffbowsh:
Now reports how long since you ran puppet, for added shaming.

/cc @tuzz


tuzz:
:+1:

PR #4247, opened 2016-03-29
Use the puppet from $PATH

boffbowsh:
We’re now installing puppet from the gem, which installs the binary in
`/usr/lib/rbenv/shims/puppet`. We should respect this if it’s been
installed, rather than using the binary installed by the Ubuntu package.

/cc @alexmuller


jackscotti:
This fixed my machine. :+1:

alexmuller:
Hm, I'm not sure about this change. I don't think we want to be using Puppet from the gem on the dev VM - I think it's installed by accident because it's required for the tests.

PR #4250, opened 2016-03-29
Update properties for Terraform Jenkins job

alexmuller:
- `ENVIRONMENT` isn't set as an environment variable
- `PROJECT_NAME` doesn't need to be set because it is already


surminus:
:+1: LGTM

rjw1:
:+1:

PR #4252, opened 2016-03-29
Update development readme vagrant DNS section.

tvararu:
Add instructions on how to purge the vagrant DNS config.

For when `vagrant-dns` is running fine but your hosts still don't seem to resolve.


tvararu:
@boffbowsh

PR #4255, opened 2016-03-30
Revert "Use the puppet from $PATH"

benhyland:
Reverts alphagov/govuk-puppet#4247

Breaks the puppet run for the dev vm with the error
"Error: Could not find a suitable provider for augeas"

### DIFF
--- a/tools/puppet-apply
+++ b/tools/puppet-apply
@@ -9,7 +9,7 @@ sed -e "s;^  :datadir: .*$;  :datadir: '/var/govuk/govuk-puppet/hieradata';g" /v
 # Filter warnings about 'storeconfigs' not being set
 exec sudo \
   RBENV_VERSION=1.9.3 \
-  puppet apply /var/govuk/govuk-puppet/manifests/site.pp \
+  /usr/bin/puppet apply /var/govuk/govuk-puppet/manifests/site.pp \
   --trusted_node_data \
   --environment ${ENVIRONMENT:-development} \
   --modulepath /var/govuk/govuk-puppet/modules:/var/govuk/govuk-puppet/vendor/modules \


Davidslv:
\cc @boffbowsh

boffbowsh:
@alexmuller what should we be using instead? At present a new VM isn't bootstrappable from start it seems, as the original `/usr/bin/puppet` doesn't exist.

alexmuller:
Hm, I'm a little confused because I can't reproduce that behaviour. If I destroy my VM and provision a new one, Puppet is installed because the `puppet-common` package is on the base box as part of the packer build process:

```
vagrant@development:~$ apt-cache policy puppet-common | grep Installed
  Installed: 3.8.5-1puppetlabs1
vagrant@development:~$ dpkg -L puppet-common | grep /usr/bin/puppet
/usr/bin/puppet
```

Which means that the initial Puppet run is fine.

alexmuller:
There was a bug which I fixed in 801e82889b3b0c0c6eb75fbe181bd19990aa8e92 where some boxes didn't have Puppet installed, maybe that's the problem you're seeing?

boffbowsh:
Oh, I've not experienced this myself, it was @tvararu that had the issue. If it's not an issue normally, I'll merge this revert.

tvararu:
I had issues provisioning my machine with `/usr/bin/puppet` as the path.

Slack ate my messages, but it was along the lines of (recounted from memory):

```
mac$ vagrant destroy
mac$ vagrant up --provision
ERROR: /usr/bin/puppet does not exist
mac$ vagrant ssh
dev$ which puppet
/usr/local/bin/puppet

# replace path with /usr/local/bin/puppet in tools/puppet-apply

mac$ vagrant provision
Works fine, once. (timed out on some Go stuff, let's run it again)
mac$ vagrant provision
ERROR: Could not find a suitable provider for augeas
mac$ vagrant ssh
dev$ which puppet
/usr/lib/rbenv/shims/puppet

# replace path with /usr/lib/rbenv/shims/puppet in tools/puppet-apply

mac$ vagrant provision
I can't remember if this worked or not.
```

So yeah, it would seem that the box didn't have `puppet` installed correctly to begin with.

PR #4256, opened 2016-03-30
Add email alert api bearer token for specialist publisher phase 2

christophwong:
This PR is made for setting up email alerts in the new specialist publisher.
[Email Alerts integration with specialist publisher](https://github.com/alphagov/specialist-publisher/pull/687)

### DIFF
--- a/modules/govuk/manifests/apps/specialist_publisher.pp
+++ b/modules/govuk/manifests/apps/specialist_publisher.pp
@@ -24,6 +24,10 @@
 #   The bearer token to use when communicating with Publishing API.
 #   Default: undef
 #
+# [*email_alert_api_bearer_token*]
+#   The bearer token to use when communicating with Email Alert API.
+#   Default: undef
+#
 # [*nagios_memory_warning*]
 #   Memory use at which Nagios should generate a warning.
 #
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/specialist_publisher.pp
```diff
@@ -73,5 +74,10 @@
       varname => 'PUBLISHING_API_BEARER_TOKEN',
       value   => $publishing_api_bearer_token,
     }
+
+    govuk::app::envvar { "${title}-EMAIL_ALERT_API_BEARER_TOKEN":
```
danielroseman: Rather than repeating the class, you should add it into the previous one; change the comma in line 75 to a semicolon and add it after that. See for example https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/travel_advice_publisher.pp#L73

modules/govuk/manifests/apps/specialist_publisher.pp
```diff
@@ -36,6 +36,7 @@
   $enable_procfile_worker = true,
   $publish_pre_production_finders = false,
   $publishing_api_bearer_token = undef,
+  $email_alert_api_bearer_token = undef,
```
danielroseman: Needs adding to the documentation comments above.

christophwong: Thanks for catching that. I've updated it.

PR #4261, opened 2016-03-30
Set higher memory warning levels for publishing api...

tuzz:
Previously, it was just using the defaults which are
700 / 800 MB. Publishing API is taking on more and more
content and so it probably makes sense to up this a bit.

Here is a graph of the last two weeks of memory usage:
https://graphite.publishing.service.gov.uk/render/?width=1400&height=800&colorList=red,orange,blue,green,purple,brown&target=alias(dashed(constantLine(800000000)),%22critical%22)&target=alias(dashed(constantLine(700000000)),%22warning%22)&target=backend-*_backend.processes-app-publishing-api.ps_rss&from=-14d

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -353,6 +353,8 @@ govuk::apps::publishing_api::db_hostname: "postgresql-primary-1.backend"
 govuk::apps::publishing_api::draft_content_store: "https://draft-content-store.%{hiera('app_domain')}"
 govuk::apps::publishing_api::redis_host: 'redis-1.backend'
 govuk::apps::publishing_api::db::backend_ip_range: "%{hiera('environment_ip_prefix')}.3.0/24"
+govuk::apps::publishing_api::nagios_memory_warning: 1400
+govuk::apps::publishing_api::nagios_memory_critical: 1600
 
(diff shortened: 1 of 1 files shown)


jennyd:
[Looking back further](https://graphite.publishing.service.gov.uk/render/?width=1400&height=800&colorList=red,orange,blue,green,purple,brown&target=alias(dashed(constantLine(800000000)),%22critical%22)&target=alias(dashed(constantLine(700000000)),%22warning%22)&target=backend-*_backend.processes-app-publishing-api.ps_rss&from=-90d), its memory usage increased significantly from ~400MB to ~700MB on March 18th, when [these changes](https://github.com/alphagov/publishing-api/compare/release_235...release_251) to the publishing API were deployed:

screen shot 2016-03-30 at 17 39 03

The default thresholds for these alerts were [reduced fairly recently](https://github.com/alphagov/govuk-puppet/pull/4037) and some apps were given higher thresholds then based on their recent usage. Publishing API still uses much less memory than some of those apps, and it may be that this is the right time to increase the thresholds for this app, but I think it's worth investigating this recent jump.

mattbostock:
Top :mag: @jennyd!

tuzz:
I'll bump this over to my esteemed colleagues for investigation.

PR #4264, opened 2016-03-31
Update elasticsearch to 1.7.5

MatMoore:
Only update integration for now.

We'd like to do some review of the search results before deploying to staging and production.

See https://trello.com/c/EYXNHkyr/524-upgrade-elasticsearch-to-1-7-5-m for the deploy steps.

I've disabled puppet on the api-elasticsearch machines so will do the upgrade once this is merged.


tijmenb:
:+1: good to go!

PR #4267, opened 2016-03-31
Add Jenkins job to build an fpm package

alexmuller:
This runs a build script in the packager repo which will make a Debian package.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/build_fpm_package.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/build_fpm_package.yaml.erb
@@ -1,0 +1,26 @@
+---
+- scm:
+    name: build_fpm_package
+    scm:
+        - git:
+            url: git@github.com:alphagov/packager.git
+            branches:
+              - master
+
+- job:
+    name: build_fpm_package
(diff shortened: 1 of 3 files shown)


modules/govuk_jenkins/templates/jobs/build_fpm_package.yaml.erb
```diff
@@ -0,0 +1,26 @@
+---
+- scm:
+    name: build_fpm_package
```
surminus: In other jobs the scm name and job name are different, I'm not sure if this is a requirement or just an oddity?

surminus:
Other than my comment (which is incredibly minor), looks good, merge away :+1:

deanwilson:
Looks good. Merge away.

modules/govuk_jenkins/templates/jobs/build_fpm_package.yaml.erb
alexmuller: I think that's just the way that we've done stuff in the past but I think we should probably refactor the whole way we do SCM for Jenkins jobs at some point.

PR #4271, opened 2016-04-01
Add local-links-manager to development apps

brenetic:
- Added local-links-manager to list of development apps


sihugh:
You missed '-o-tron' from the end, but aside from that, this looks fine to me!
:+1:

PR #4273, opened 2016-04-01
Prepare for api-elasticsearch upgrade to 1.7

MatMoore:
Please check puppet is disabled on api-elasticsearch hosts before merging, as we need to do a rolling deploy.


alexmuller:
Puppet is disabled on the 3 api-elasticsearch machines in staging, so if you're happy then go ahead :+1:

PR #4274, opened 2016-04-01
Change asset_root in production

alexmuller:
This will change `/etc/govuk/env.d/GOVUK_ASSET_HOST` and `/etc/govuk/env.d/GOVUK_ASSET_ROOT` on every machine.

~~I don't think this will restart every application but it may do, so it requires care when deploying.~~

These 2 hostnames are served by the same backend so changing them will make no difference at all.


surminus:
:+1: go go go

alexmuller:
Going to deploy this on Wednesday morning when I'm on 2nd line so will close it for now. Keep the comments coming if you have feedback.

alexmuller:
Waiting for when it's a bit quieter.

PR #4275, opened 2016-04-01
[DO NOT MERGE] Enable SSL in development

alexmuller:
I'd appreciate feedback from application developers about how much harder this will make your lives.

You will need to add this SSL certificate to your Keychain on OS X (or similar system on other operating systems) in order to make using apps in a browser work properly.


boffbowsh:
This private key will be public, allowing people to serve content using it which would appear to us developers as trusted and secure. I think there might be possible issues there.

alexmuller:
That's true, I meant to mention that - thanks!

I think that risk is very small. For people with this certificate installed you'd have to be using a service hosted on `*.dev.gov.uk` (as far as I know that doesn't exist) or set up DNS for that hostname pointing to an attacker's network and would then have to transmit sensitive information.

boffbowsh:
Overall, :+1:. I think it'll need some documentation changes in https://github.gds/gds/development/blob/master/README.md, or possibly have it added automatically by https://github.com/alphagov/govuk-puppet/blob/master/development/bootstrap.sh?

alexmuller:
I've learnt some scary things about OS X's Keychain and apparently it is possible to install it automatically. I've pushed a change that does that but I don't know if it's something we want.

alexmuller:
Also worth noting that this doesn't set the certificate to be trusted (I haven't been able to make that work) so maybe it's not worth doing.

deanwilson:
Can we discuss this in the team meeting and either agree to finish it up or close the PR? The longer it's out for the more drift we'll see.

deanwilson:
We discussed this in the team meeting and we're not going to deploy it at this time so we'll close the PR for now.

PR #4276, opened 2016-04-01
Remove nginx_enable_basic_auth param

alexmuller:
This is only disabled in development and it only affects one app in that environment, whitehall-admin.


mattbostock:
I'm not sure I understand the benefit of this change.

alexmuller:
I've restructured the commits a little bit to make this make more sense so that somebody can review it. The benefit is that we're making our Puppet repo a tiny bit smaller and removing some complexity that we don't need.

deanwilson:
The code looks fine without having much context as to why we'd want this. One for the team meeting?

PR #4278, opened 2016-04-01
Reduce nginx basic auth for whitehall-admin

alexmuller:
## Todo

- [x] Get sign-off
- [ ] Verify in integration


elliotcm:
We've got to keep this on fact check requests so

alexmuller:
Updated this to reduce the amount of basic auth rather than removing it entirely. Needs a bit of careful testing in integration.

elliotcm:
Awesome.  Needs a rebase.

Should we be deploying this branch to integration then?  Presumably merging it could block production deploys otherwise.

alexmuller:
I've just tested this locally and I'm happy with the way it behaves. If somebody thinks this looks ok I think we should be alright to merge this.

modules/govuk/manifests/app/nginx_vhost.pp
```diff
@@ -4,6 +4,7 @@
   $app_port,
   $aliases = [],
   $protected = false,
+  $protected_location = '/',
```
rjw1: puppetdocs?

modules/nginx/manifests/config/vhost/proxy.pp
```diff
@@ -29,6 +29,12 @@
 # [*alert_5xx_critical_rate*]
 #   The error percentage that triggers a critical alert
 #
+# [*protected*]
+#   Boolean, whether or not the vhost should be protected with basic auth
+#
+# [*protected_location*]
+#   Prefix path to protect with basic auth, defaults to everything
+#
 # TODO: More docs!
```
rjw1: since your editing this file how about doing the todo?

alexmuller: You're killing me bob, I have added all the docs.

rjw1:
:+1:

PR #4280, opened 2016-04-01
Remove basic auth for monitoring

alexmuller:
Graphite and Icinga both have basic authentication enabled in integration but they are also restricted to only be available from the GDS office, so there's no need for this basic auth at all.


surminus:
:+1:

mattbostock:
:+1:

PR #4281, opened 2016-04-04
Enable elasticsearch 1.7 on production

MatMoore:
See https://trello.com/c/EYXNHkyr/524-upgrade-elasticsearch-to-1-7-5-m

Puppet has been temporarily disabled on api-elasticsearch hosts as we need to do a rolling deploy.


rjw1:
:+1:
Merge when green and you are ready.

PR #4282, opened 2016-04-04
Increase memory alert levels for content store

tuzz:
The last two weeks of memory usage:

screen shot 2016-04-04 at 10 19 13

It looks like it increased some time around March 10th. I have [added a ticket](https://trello.com/c/eFdamRxD/662-investigate-why-content-store-memory-usage-has-spiked) to the Publishing Platform team's backlog. Until then, this gives content store a bit more breathing room.

The last 1 month of memory usage:

screen shot 2016-04-04 at 10 26 48

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -252,8 +252,8 @@ govuk::apps::content_register::rabbitmq_hosts:
   - rabbitmq-2.backend
   - rabbitmq-3.backend
 
-govuk::apps::content_store::nagios_memory_warning: 800
-govuk::apps::content_store::nagios_memory_critical: 900
+govuk::apps::content_store::nagios_memory_warning: 900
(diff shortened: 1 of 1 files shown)


hieradata/common.yaml
```diff
@@ -252,8 +252,8 @@ govuk::apps::content_register::rabbitmq_hosts:
   - rabbitmq-2.backend
   - rabbitmq-3.backend
 
-govuk::apps::content_store::nagios_memory_warning: 800
-govuk::apps::content_store::nagios_memory_critical: 900
+govuk::apps::content_store::nagios_memory_warning: 1000
```
mattbostock: Should these be lower, e.g. 900 and 1000 respectively, given the graph shown in the PR description?

tuzz: :+1: will update

PR #4283, opened 2016-04-04
Add parameter for Terraform's AWS account ID

alexmuller:
We've made this a required environment variable in our Terraform project.


rjw1:
merge when ready

PR #4285, opened 2016-04-04
Enable event handlers in all environments

rjw1:
Restarting services which use to much memory by 'hand' is tedious.
Change the default value of enable_event_handlers to true so that all
our environments will restart apps which are above their memory
thresholds.

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -283,7 +283,6 @@ hosts::production::licensify_hosts:
     ip: '31.210.245.117'
 
 icinga::client::check_cputype::cputype: 'amd'
-icinga::config::enable_event_handlers: true
 
 licensify::apps::licensing_web_forms::enabled: true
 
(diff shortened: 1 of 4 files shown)


mattbostock:
:+1: though I think we should improve on the way we report these events back to teams. My initial thoughts were to send an email to the tech lead mailing list and the #govuk-developers Slack channel.

@deanwilson Did you get to the bottom of why you were having to manually restart some apps in Staging?

rjw1:
@mattbostock possibly 2ndline as well.

deanwilson:
When you say "all environments" should this be enabled in prod too?

mattbostock:
@rjw1: I don't see any need to notify 2ndline; they can see the service restart annotations in Grafana and I'd like to push as much of this down to application teams.

alexmuller:
I agree that these should be noisier but I'm not sure that sending them to the tech lead mailing list is the right way to go, I think that might just create noise in the wrong place.

deanwilson:
We've added a new trello story to discuss who, how and when to surface these restarts.

mattbostock:
Automation FTW :tada:

PR #4290, opened 2016-04-05
adds user gov.uk infrastruture user jonauman to integration

jonauman:


mattbostock:
:+1: Would you mind squashing 3984cd1 into the previous commit?

PR #4291, opened 2016-04-05
Archive Debian packages

alexmuller:
Jenkins is going to generate Debian packages and we want to archive them permanently (ish).


mattbostock:
:+1: Merge when ready.

PR #4296, opened 2016-04-07
removes paging alerts for third party hosts

jonauman:
We don't want to be paged when there are issues contacting third
party hosts like the VPN. Whereas the service alerts do not page
us, the host alerts currently do as that is the default for the
'generic-host' template


rjw1:
Im not entirely sure the host template config file should be called third-party-host_nagios2.cfg. The nagios2 doesn't need to be there. Admittedly this is only a minor quibble apart from that I think is good to merge when green

alexmuller:
I think this change will work, but a more simple way of achieving the same thing might be to set the `notification_period` parameter to `never` (which is an `icinga::timeperiod` resource that we define).

mattbostock:
@alexmuller: I'd be in favour of that as it avoids duplicating config.

jonauman:
@alexmuller Fair enough, but the original story ("Stop sending pages when VPN goes down") was written to squelch the pager notification only. Your suggestion would squelch all notifications.

@rjw1 We can rename the file. It is a copy of the oddly named file ''generic-host_nagios2.cfg"

surminus:
I think `notification_period` in this instance does only refer to stuff that is pushed to Pagerduty, we'd still see the alerts pop up in Icinga. I guess the thing to consider is that perhaps we should notify people if it's been down for more than an hour, for example, because then we should probably get $vendor on the case.

rjw1:
I think the file name should certainly change and then this is possibly ready for merging. Since I dont want to get called again.

jonauman:
@rjw1 File has been renamed

rjw1:
Squashed commits in  #4335

PR #4297, opened 2016-04-07
Alert on-call people for travel advice email problems

tijmenb:
The icinga check verifies that travel advice emails have been sent to some test-email accounts.

This commit ups the priority to alert an on-call person. Previously this had the default "high_priority" priority which doesn't escalate to pagerduty.


tijmenb:
cc @alexmuller @danielroseman

alexmuller:
Worth noting that this is the first passive check we have which will phone people.

PR #4298, opened 2016-04-07
Fix protected locations for nginx

alexmuller:
Nested location blocks don't work as I expect. This commit introduces more code but is more explicit about what's going on.

In the `location /` block, if basic auth is enabled and the protected location is `/` then set it up.

If basic auth is enabled and the location isn't `/`, create a new block (which nginx will match to if it's there because the prefix URL is longer).


rjw1:
Merge when ready.

PR #4299, opened 2016-04-08
Add `service_template` param to passive checks

alexmuller:
We want to make some passive checks more important than just regular service.


deanwilson:
The code looks fine. Merge when green.

rjw1:
`service_template` may be a better variable name than `use` but thats bikeshedding. 
Good to merge otherwise

deanwilson:
Actually I think @rjw1 has a good point. Changing the name will make life easier when we return to this code.

alexmuller:
Sure, updated!

deanwilson:
Looks much nicer. Thank you.

PR #4300, opened 2016-04-08
Add per-queue stats for rabbitmq.

benhyland:
We achieve this by replacing the existing python script with a plugin packaged by NYTimes which does per-queue stats out of the box.

Note the lack of testing - I've not even proved this works manually so it probably should not be merged for now.

I'm not sure how to examine the collectd output in a vm, nor whether I am doing the right thing with puppet.

Does anything (e.g. alerting) rely on the existing stat names?

I think we have decided it is OK to have access to the internet from our production machines. Thus, pip will work - at least it seems we expect this in other places in govuk-puppet. Is this correct, or do we need to ensure the package is available in some private repository before live environments can install it?

Advice welcome!

@benhyland and @tijmenb


alexmuller:
> Does anything (e.g. alerting) rely on the existing stat names?

Instances of `icinga::check::graphite` in this repo take a `target` parameter which is the name of a Graphite stat which might be affected by this.

> I think we have decided it is OK to have access to the internet from our production machines. Thus, pip will work - at least it seems we expect this in other places in govuk-puppet. Is this correct, or do we need to ensure the package is available in some private repository before live environments can install it?

That's right, production has fairly unrestricted internet access. For some stuff (eg Debian packages) we prefer to cache them using [aptly](https://www.aptly.info/) but we don't have anything like that for pip packages at the moment.

modules/collectd/manifests/plugin/rabbitmq.pp
```diff
@@ -1,31 +1,23 @@
 # == Class: collectd::plugin::rabbitmq
 #
-# Sets up a collectd plugin to monitor RabbitMQ.
+# Sets up the NYTimes collectd-rabbitmq collectd plugin to monitor RabbitMQ.
```
Davidslv: I would suggest to add the url for the project ( https://github.com/NYTimes/collectd-rabbitmq ? )

modules/collectd/manifests/plugin/rabbitmq.pp
```diff
@@ -1,31 +1,23 @@
...
-    source => 'puppet:///modules/collectd/usr/lib/collectd/python/rabbitmq.py',
-    tag    => 'collectd::plugin',
-    notify => File['/etc/collectd/conf.d/rabbitmq.conf'],
-  }
-
-  @file { '/usr/lib/collectd/python/rabbitmq.pyc':
-    ensure => undef,
-    tag    => 'collectd::plugin',
+  package { 'collectd-rabbitmq':
+    ensure   => present,
```
alexmuller: Could you specify a version of the package here (eg `ensure => '1.2.3'`). This prevents the case where we build a new machine which installs the latest version of the package rather than a version we know is good.

benhyland:
Thanks for the comments.

I can't find any `icinga::check::graphite`s that reference rabbitmq, so I think we're safe there.
I've pushed the other improvements.

alexmuller:
I agree, I can't see any either. If there is any monitoring based on this we'll notice because it will start alerting, so I think this is relatively safe to merge and test in integration when you're ready :+1:

PR #4301, opened 2016-04-08
Fix icinga::passive_check param

alexmuller:
While discussing 2a7b8d272a057903c653a5758f0e9e6a0a4e606e we agreed to rename this from `use` to `service_template`.


deanwilson:
Looks good. merge away.

PR #4302, opened 2016-04-08
Revert "Add per-queue stats for rabbitmq"

benhyland:
Reverts alphagov/govuk-puppet#4300

We discovered some authorisation issues in integration, so we're backing this out until we have corrected them.


benhyland:
Also we need to add the custom types for the collectd plugin and adjust the config to load them.

tijmenb:
:+1:

PR #4303, opened 2016-04-11
Give rabbitmq monitoring access to the stagecraft and backdrop_write vhosts

benhyland:
We need this so that our new collectd stats gathering plugin (https://github.com/alphagov/govuk-puppet/pull/4304) will work on all queues.

@benhyland and @matmoore

Trello: https://trello.com/c/Ipl9UVsA

PR #4304, opened 2016-04-11
Add per-queue stats for rabbitmq

benhyland:
We achieve this by replacing the existing python script with a plugin packaged by NYTimes which does per-queue stats out of the box.

We have adjusted the monitoring user permissions in a separate PR (#4303), which should be merged first.

We also correct the plugin config from our previous attempt.

This reverts commit 4fdc8ccb850f956e6c618a58739aa0e5c5f2d3e9.

@benhyland and @matmoore

Trello: https://trello.com/c/Ipl9UVsA


tijmenb:
This looks good.

The only question I have relates to https://github.com/alphagov/govuk-puppet/pull/4303. What if someone introduces a new consumer and doesn't set the permissions for `monitoring`, would our process here be silently failing?

benhyland:
@tijmenb valid concern, but I think we shouldn't worry so long as we go ahead with the cleanup we discussed soonish.

There are two ways to introduce a consumer at the moment:

- stagecraft/backdrop way, which I think is probably rare and only used for custom vhosts.
- normal way, which hardcodes the `/` vhost.

The monitoring user is preconfigured with access to the `/` vhost.
We should make all consumers configured in the same way,
and provide the monitoring user config automatically.
Until that is done, the stagecraft/backdrop way could break the collectd-rabbitmq
stats collection if it is introduced anywhere else without also configuring the monitoring user.

rjw1:
looks fine to me. merge when ready.

tijmenb:
Updated description with link to Trello.

PR #4305, opened 2016-04-11
Add projects to Jenkins job

surminus:
There are more projects that have been added to the Terraform repo. A programatic way of updating and listing the current projects based upon the dir structure would be nice, but for now just add the new ones.


deanwilson:
Looks fine. Merge when green.

PR #4306, opened 2016-04-11
Prevent accidental activation of rummager worker.

benhyland:
Recently the capistrano deployment job ran a task which erroneously enabled
the rummager indexer worker process. This should not be possible, and
could possibly have led to production issues (we got away with it this time).

Tie the presence of the upstart conf file for the worker to the feature flag -
when the flag is off the file is missing so it is not possible to accidentally
start the worker.

 Also tie the appropriate rabbitmq queue config to the same feature.

 @matmoore and @benhyland

See https://trello.com/c/hUWLNllj/583-fix-rogue-message-queue-worker-on-production

### DIFF
--- a/modules/govuk/manifests/apps/rummager.pp
+++ b/modules/govuk/manifests/apps/rummager.pp
@@ -104,7 +104,13 @@ class govuk::apps::rummager(
     enable_service => $enable_procfile_worker,
   }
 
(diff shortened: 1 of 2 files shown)


rjw1:
Looks good to me. merge when ready.

PR #4307, opened 2016-04-11
Add specialist-publisher-rebuild app for development

christophwong:
This is the first commit for setting up specialist-publisher-rebuild to
run on development.
Compared with specialist-publisher we have dropped the env vars for
nagios configs, preproduction finders, and sidekiq procfile workers
because the phase-2 rebuild doesn't yet use those configs.

Integration and further environments will come later.

cc: @benilovj @brenetic


elliotcm:
LGTM

elliotcm:
(hit close and comment by mistake)

PR #4309, opened 2016-04-12
Revert "Add per-queue stats for rabbitmq"

benhyland:
Reverts alphagov/govuk-puppet#4304

This has puppet class ordering issues on the dev and rabbit vms.

E.g.

```
Error: Could not find dependent File[/etc/collectd/conf.d/rabbitmq.conf] for File[/usr/share/collectd/types.db.rabbitmq] at /var/govuk/govuk-puppet/modules/collectd/manifests/plugin/rabbitmq.pp:25
```


Davidslv:
👍 thanks

PR #4311, opened 2016-04-12
Redo "Add per-queue stats for rabbitmq"

MatMoore:
This reverts commit b3b46fa595619293dd0ca314b309e493edf986be.

- "Revert the revert of "Add per-queue stats for rabbitmq""
- Add missing dependency that was breaking the dev VM

The original PR is https://github.com/alphagov/govuk-puppet/pull/4304
Reverted by https://github.com/alphagov/govuk-puppet/pull/4309
Part of: https://trello.com/c/Ipl9UVsA/578-monitor-rummager-s-current-queue-size-with-statsd-graphite-m

The dev VM was breaking with this error: `Error: Could not find dependent File[/etc/collectd/conf.d/rabbitmq.conf] for File[/usr/share/collectd/types.db.rabbitmq] at /var/govuk/govuk-puppet/modules/collectd/manifests/plugin/rabbitmq.pp:25`

This is because the file we create requires a directory owned by the collectd package, so we need an explicit dependency on this.

The only difference from the previous pull request should be `+    require => Package['collectd-rabbitmq'],`.


MatMoore:
Seems like this is still running into problems :sob: 

`Error: Could not find dependency Package[collectd-core] for File[/usr/share/collectd/types.db.rabbitmq] at /var/govuk/govuk-puppet/modules/collectd/manifests/plugin/rabbitmq.pp:27`

MatMoore:
This is now working on expected on the dev VM, as well as a freshly created rabbitmq-1.backend VM.

The previous attempt at this caused some problems in development
because we were running collectd plugins on nodes where we didn't
install the collectd package.

On dev, we disable a lot of monitoring functionality, so that's why the problem only occurred there.

We now use an explicit `require` dependency and use virtual resources so that the plugins are only set up when collected into the collectd::plugins class.

@tijmenb @benhyland

tijmenb:
Heroes! :+1: when tests are green.

PR #4312, opened 2016-04-12
Incorrect syntax

surminus:
Caused the cronjob to be wrong.


deanwilson:
Looks fine. Merge when green.

PR #4313, opened 2016-04-12
Crawler bugfixes

alexmuller:


deanwilson:
Can you add a little context please? Code looks ok but I'd like to know why too.

surminus:
@deanwilson doing a git blame and checking the commit when the blacklists were originally added leads to this issue https://github.com/alphagov/govuk_crawler_worker/issues/97

surminus:
In light of that I'd :+1: go forth and merge, would be good to get this deployed today.

PR #4315, opened 2016-04-12
Configure queues and queue bindings for rabbitmq from puppet

benhyland:
We are going to stop handling queue creation and binding in the consumer
processes, since it is too easy to mess up and leave a unintended binding
around.

Instead, we will configure everything in puppet and the consumer process
will error out if the queue it expects is missing.

This is the first step.
Next, we will modify the consumer gem and provide a way to remove
unwanted bindings via puppet (perhaps automatically).

We deliberately omit stagecraft and backdrop_write config, since they do
not use the consumer gem and are likely to change as the performance
platform exits alpha.

See https://trello.com/c/eP2CdnrC/590-fix-rabbitmq-queue-binding

@matmoore and @benhyland


modules/govuk/manifests/apps/panopticon/rabbitmq.pp
```diff
@@ -24,13 +24,15 @@
     amqp_pass     => $password,
     amqp_exchange => 'published_documents',
     amqp_queue    => 'panopticon',
+    routing_key   => '#',
```
tijmenb: Panopticon actually only needs links updates, like rummager.

tijmenb:
Once we have a way to automatically remove the bindings, we should:

- Make panopticon only listen to `*.links`
- Remove the inconsistency between rummager `*.links` and email-alert-service `*.major.#` (the trailing `#`) - I believe there's no reason for them to be different.

benhyland:
Good comments Tijmen, I'll put them on the trello ticket for later.

benhyland:
The latest commits update this PR to include the classes to be used for manual removal of queues and bindings. I've done some manual testing in a local vm, but as usual I am not sure how best to automate testing of puppet - advice welcome.
This could be merged after review from @matmoore, @tijmenb, and infrastructure.

modules/govuk_rabbitmq/manifests/consumer.pp
```diff
@@ -48,9 +61,9 @@
       ensure => $ensure,
       type   => $exchange_type,
     }
-    $write_permission = "^(amq\\.gen.*|${amqp_queue}|${amqp_exchange})$"
+    $write_permission = "^(amq\\.gen.*|${amqp_queue}|${amqp_exchange})\$"
```
MatMoore: whats the extra slash for here? Does this tell rabbitmq to match a literal $ in a name?

modules/govuk_rabbitmq/manifests/consumer.pp
```diff
@@ -60,33 +73,35 @@
...
       write_permission     => $write_permission,
-      read_permission      => "^(amq\\.gen.*|${amqp_queue}|${amqp_exchange})$",
+      read_permission      => "^(amq\\.gen.*|${amqp_queue}|${amqp_exchange})\$",
     }
+
+    rabbitmq_queue { "${amqp_queue}@/":
+      ensure      => present,
+      user        => 'root',
+      password    => $::govuk_rabbitmq::root_password,
+      durable     => true,
```
MatMoore: maybe we could parameterise these things and just default them to the current values?

modules/govuk_rabbitmq/manifests/remove_bindings.pp
```diff
@@ -0,0 +1,37 @@
...
+# [*amqp_vhost*]
+#   The RabbitMQ vhost from which to delete bindings.
+#
+# [*amqp_exchange*]
+#   The RabbitMQ exchange from which to delete bindings.
+#
+# [*amqp_queue*]
+#   The RabbitMQ queue from which to delete bindings.
+#
+define govuk_rabbitmq::remove_bindings (
```
MatMoore: I think this should be non pluralised since it only does one routing key at a time.

modules/govuk_rabbitmq/manifests/consumer.pp
benhyland: https://docs.puppet.com/puppet/3.8/reference/future_lang_data_string.html#escape-sequences

It's to explicitly include the $ at the end in the permissions regex. It worked before, but $ should be escaped if not being used for interpolation AIUI. My IDE inspections agree...

modules/govuk_rabbitmq/manifests/consumer.pp
benhyland: I don't think tidying up the consumer class is in scope for this story. If we changed this, I'd want to think about all the params and improve the usability of the whole class.

MatMoore:
Nice, the stuff for purging queues & bindings looks quite straightforward to use now. Can you add something to the rummager/rabbitmq manifest to use the remove_bindings class on the duplicate binding of rummager_to_be_indexed, so that we ensure it gets removed from all environments as this gets rolled out?

modules/govuk_rabbitmq/manifests/remove_bindings.pp
benhyland: Actually, it can handle an array of keys - that's what the `$routing_key = $title` achieves.

Many thanks for all the comments btw :)

benhyland:
I intended to actually do the extra binding removal in a separate PR. Happy to add it here if you prefer.

Puppetfile
```diff
@@ -61,7 +61,8 @@ mod 'KyleAnderson/etcd',
 mod 'puppetlabs/lvm',
   :git => 'git://github.com/alphagov/puppetlabs-lvm.git'
 mod 'puppetlabs/rabbitmq',
-  :git => 'git://github.com/puppetlabs/puppetlabs-rabbitmq.git'
+  :git => 'git://github.com/puppetlabs/puppetlabs-rabbitmq.git',
+  :ref => '5.3.1'
```
rjw1: Since this bumps us to a released version we should probably stop using the git version and just use the [forge version](https://forge.puppet.com/puppetlabs/rabbitmq).

modules/govuk_rabbitmq/manifests/consumer.pp
```diff
@@ -33,11 +46,13 @@
   $amqp_pass,
   $amqp_exchange,
   $amqp_queue,
+  $routing_key,
```
rjw1: Could we add puppetdocs for this new variable

MatMoore:
@benhyland either is fine with me :+1:

benhyland:
Thanks all for comments.
I've addressed bob's in the latest push.
I'll merge this on Mon unless anyone wants me to hold off.

tijmenb:
@rjw1 are you okay with the changes here? Can we get a thumbs up from you on this one?

tijmenb:
👍

PR #4316, opened 2016-04-12
Mongo backups to S3

jonauman:


modules/mongodb/templates/mongodb-backup-s3-wrapper.erb
```diff
@@ -0,0 +1,34 @@
...
+exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
+
+# The default Icinga passive alert assumes that the script failed
+NAGIOS_CODE=2
+NAGIOS_MESSAGE="CRITICAL: Mongodb  backup push to S3 failed"
+
+# Triggered whenever this script exits, successful or otherwise. The values
+# of CODE/MESSAGE will be taken from that point in time.
+function nagios_passive () {
+  printf "<%= @ipaddress %>\t<%= @service_desc_wale %>\t${NAGIOS_CODE}\t${NAGIOS_MESSAGE}\n" | /usr/sbin/send_nsca -H alert.cluster >/dev/null
```
surminus: The Icinga passive check needs to be added to the Puppet manifest ([check here for an example](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_postgresql/manifests/wal_e/backup.pp#L170-L174)). Required variables will also need adding and amending in this template.

modules/mongodb/templates/mongodb-backup-s3-wrapper.erb
```diff
@@ -0,0 +1,34 @@
...
+if [ $? == 0 ]
+  then
+    STATUS=0
+  else
+    STATUS=1
+fi
+
+if [ $STATUS -eq 0 ]; then
+  NAGIOS_CODE=0
+  NAGIOS_MESSAGE="OK: Mongodb  backup push to S3 succeeded"
```
surminus: Nitpick time! Unneeded space.

surminus:
* Some scheduling (cron?) needs to be added
* How do we ensure it's only run on a single Mongo server? I know the script itself has logic to only run against a secondary, but how do we ensure it's only run on one of the cluster as we don't want all three to try running at the same time?

rjw1:
So, I think we could just enable this on the mongo node in DR since it will always be a secondary.

The commits need squashing.

surminus:
@rjw1 does that mean we're relying on a DR solution for a DR solution?

modules/mongodb/manifests/s3backup/backup.pp
```diff
@@ -0,0 +1,173 @@
...
+
+  validate_re($private_gpg_key_fingerprint, '^[[:alnum:]]{40}$', 'Must supply full GPG fingerprint')
+
+# create user
+  user { $user:
+    ensure     => 'present',
+    managehome => true,
+  }
+
+# push env files
```
rjw1: this probably should use govuk::app::envar

spec/fixtures/hieradata/common.yaml
```diff
@@ -36,6 +36,8 @@ hosts::production::management::hosts:
   jumpbox-1:
     ip: '10.1.0.100'
 
+mongodb::s3backup::backup::private_gpg_key_fingerprint: 'CB77872D51ADD27CF75BD63CB60B50E6DBE2EAFF'
```
surminus: Should this be here?

jonauman:
@surminus 
- cron has been added
- Backup will be initiated only on $replicaset_members[0], as we do for normal backups. I'm thinking at some point we will disable the nightly backups in favor of the every 15 minute backups.
- 'mongodb::s3backup::backup::private_gpg_key_fingerprint' must be defined in the spec fixtures or else the spec tests fail. I think there are some comments in the commit message

jonauman:
@rjw1 
- I believe you've squashed the commits. Thanks.
- I originally planned on using govuk::app::envar but upon further inspection, I saw it was very app specific and setup the server accordingly. I wanted the mongo backups to work without the dependency of govuk::app::envar because normally we would not be running that module on mongo nodes

surminus:
I think this looks good, but can we ensure this is only enabled on Integration to begin with? Like with Postgres and MySQL, once we've tested restoring the data we should then enable on Staging and Production (with a view to replace offsite backups).

jonauman:
@surminus Happy for me to merge after commit e195aed?

modules/mongodb/manifests/s3backup/backup.pp
```diff
@@ -0,0 +1,173 @@
...
+# [*server*]
+#   Defines the server to connect to. The backup
+#   script will pick a secondary to backup, unless
+#   'standalone' is 'True'
+#
+# [*standalone*]
+#   If true, will backup localhost instead of a 
+#   Secondary
+#
+# [*threshold_secs*]
```
surminus: We don't need to document this if it isn't set in the parameters list

surminus:
Other than my nitpick I think this is looking good to test :+1: just waiting on the credentials to be merged

surminus:
:+1: Credentials merged, bucket deployed, PR looking good for testing in Integration! Merge when green

PR #4317, opened 2016-04-12
Need to specify username in cron job

surminus:
Means that the cronjobs do not run with the following error:

```
cron[1081]: (*system*xtrabackup_s3) RELOAD (/etc/cron.d/xtrabackup_s3)
cron[1081]: Error: bad username; while reading /etc/cron.d/xtrabackup_s3
cron[1081]: (*system*xtrabackup_s3) ERROR (Syntax error, this crontab file will be
ignored)
```


deanwilson:
Merge away.

PR #4319, opened 2016-04-13
Revert "Redo "Add per-queue stats for rabbitmq""

MatMoore:
Reverts alphagov/govuk-puppet#4311

This PR added a new collectd plugin to record queue level metrics from RabbitMQ.

Unfortunately there is a problem with this PR that has the potential to cause some inconvenience if deployed to production.

Currently we deploy the types db from the new plugin only to servers where the plugin itself is enabled, and not to the collectd server hosted on the graphite cluster, which metrics are forwarded to via the network plugin. This means the server cannot interpret incoming messages, and they get dropped. This logs a lot of noise to the syslog, and as a result we are wasting a lot of disk space. We're going to back this out and raise another PR that corrects this problem.


tijmenb:
:+1:

tijmenb:
Follow up: https://github.com/alphagov/govuk-puppet/pull/4323

PR #4320, opened 2016-04-13
Move GOVUK-Request-Id from varnish to nginx.

elliotcm:
DO NOT MERGE YET

This gives us tracing in our admin systems.

If we find that we need tracing back between varnish
and the router on our frontend systems we can add
a "GOVUK-Cache-Request-Id" or similar.

If anyone has any changes they'd like to make to the "unique enough" ID, please let me know.

I've confirmed via vagrant that `backend-lb-1` and `backend-1` will build and boot nginx happily, but had trouble trying to form convincing requests for it.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -126,6 +126,3 @@ DEPENDENCIES
   rspec-puppet
   sshkey (= 1.7.0)
   webmock (~> 1.20.0)
-
-BUNDLED WITH
-   1.10.5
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -126,6 +126,3 @@ DEPENDENCIES
   rspec-puppet
   sshkey (= 1.7.0)
   webmock (~> 1.20.0)
-
-BUNDLED WITH
-   1.10.5
```
rjw1: this probably shouldnt be here. :)

elliotcm: The removal shouldn't or the BUNDLED WITH shouldn't?  (it's largely been backed out from bundler)

elliotcm:
Updated with config fix, shown working here: https://kibana.integration.publishing.service.gov.uk/kibana#dashboard/temp/AVQPwvUHAWArxCU_s3s8

surminus:
:+1:

PR #4321, opened 2016-04-13
Change gof3r path to full

surminus:
Continuing my series of annoying little oversights, I got the following error on the cron run:

`envdir: fatal: unable to run gof3r: file does not exist`

Changed from just "gof3r" to "/usr/local/bin/gof3r" in the command.


deanwilson:
The code for this looks ok but we might want to discuss adding `/usr/local/bin` to out default paths in the team meeting. I'd also probably not specify the whole command in the spec file assert. I try to only match as little as possible to keep it from being too brittle.

But merge away.

surminus:
@deanwilson thanks, I actually assumed `/usr/local/bin` was in the default path because it normally is, but that's what I get for assuming.

rjw1:
probably not in cron's default path though.

PR #4323, opened 2016-04-13
Add collectd rabbitmq stats for queues

benhyland:
Another attempt at introducing the NYTimes collectd-rabbitmq plugin.

Our previous attempts at this ran into problems:
- didn't understand puppet ordering and virtual resources
- didn't understand collectd custom types
- didn't know that collectd also runs as a server next to graphite

Original was here:
https://github.com/alphagov/govuk-puppet/pull/4300

See https://trello.com/c/Ipl9UVsA/578-monitor-rummager-s-current-queue-size-with-statsd-graphite-m

Added to our earlier fixes, we now introduce a config file which lists the
types files required and moves the typesdb file for the rabbitmq plugin
so they are available to the server as well as all clients.
Clients which do not install the collectd-rabbitmq plugin have no use for
the additional types but suffer no harm, and now there is one true home
for any future custom types files.

@matmoore and @benhyland

Review by @tijmenb (among others) please :)


tijmenb:
I'm good with the changes, but I also was with the previous ones that were reverted. So I'd really like someone at infra to look at this too. Also, what's the approach taken to test this locally? Should we deploy this branch to integration before merging?

MatMoore:
We've tested:

* A newly created rabbitmq-1.backend node provisions correctly, and no collectd errors are visible in the syslog 
* The same for a graphite-1.management node
* The config files and type dbs exist in the places we expect

What we haven't tested (and plan to use integration for) is that the metrics now propagate through the collectd server into graphite/carbon. In our previous deploy we verified that the collectd server was receiving messages but was not able to handle them.

If there's a good way to test this before integration let us know - I'm not sure what the best way to handle these kinds of changes is.

tijmenb:
👍 good to merge

PR #4325, opened 2016-04-13
Add Russell Garner

rgarner:
"We may not accept contributions if they aren't useful for GOV.UK or if they introduce operational complexity."

I realise this may be a blocker.


Davidslv:
`SSH key for russellgarner is only 2048 bits and must be stronger`

benilovj:
Closing in favour of https://github.com/alphagov/govuk-puppet/pull/4327.

PR #4334, opened 2016-04-15
Change rabbitmq monitoring user tag to 'monitoring'

benhyland:
Tag change to 'monitoring' from 'administrator'.

See the discussion of permissions at https://www.rabbitmq.com/management.html
This should have no production impact since noone should be using the monitoring
user to view users or permissions, close connections, or create vhosts.

**Merge only after https://github.com/alphagov/govuk-puppet/pull/4315 is merged, since we need to update the puppetlabs/rabbitmq module for this to work.**

See https://trello.com/c/eP2CdnrC/590-fix-rabbitmq-queue-binding

### DIFF
--- a/modules/govuk_rabbitmq/manifests/init.pp
+++ b/modules/govuk_rabbitmq/manifests/init.pp
@@ -29,11 +29,9 @@ class govuk_rabbitmq (
     write_permission     => '.*',
   }
 
-  # FIXME: Consider using a normal user with a "monitoring" user tag.  Not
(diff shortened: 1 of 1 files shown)


tijmenb:
👍 once https://github.com/alphagov/govuk-puppet/pull/4315 is merged.

Offline discussion: this changes the permissions of the monitoring user to not be admin. Which is made possible by the upgrade of the puppetlabs/rabbitmq-module.

PR #4336, opened 2016-04-16
Sync Whitehall attachments to S3

surminus:
This adds functionality to also backup Whitehall attachments to S3 by adding `s3cmd` to the loop in which attachments are copied over to the asset-slaves if they are detected as being free from viruses.

It is uploaded to S3 in the same directory structure as to the asset-slaves to help with consistency and make it easier to restore the directory structure if required. The nightly sync which also happens (to
ensure that the master and slaves are in sync is also updated.

The process which creates the `process-uploaded-attachments.sh` and `copy-attachments.sh` scripts are moved to asset-master only due to the way that they had to incorporate s3_bucket credentials.

### DIFF
--- a/modules/govuk/templates/node/s_asset_base/copy-attachments.sh
+++ b/modules/govuk/templates/node/s_asset_base/copy-attachments.sh
@@ -27,3 +27,11 @@ for NODE in $ASSET_SLAVE_NODES; do
     logger -t copy_attachments "Attachments errored while copying to $NODE"
   fi
 done
+
+<% if @s3_bucket %>
+  if envdir /etc/govuk/aws/env.d /usr/local/bin/s3cmd --server-side-encryption sync --skip-existing --deleted-removed "$DIRECTORY_TO_COPY/" "s3://<%= @s3_bucket -%>$DIRECTORY_TO_COPY/"; then
+    logger -t process_uploaded_attachment "Attachments copied to S3 (<%= @s3_bucket -%>) successfully"
+  else
+    logger -t process_uploaded_attachment "Attachments errored while copying to S3 (<%= @s3_bucket -%>)"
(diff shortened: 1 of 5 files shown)


modules/govuk/templates/node/s_asset_base/copy-attachments.sh
```diff
@@ -21,9 +21,19 @@ fi
...
     logger -t copy_attachments "Attachments copied to $NODE successfully"
   else
     logger -t copy_attachments "Attachments errored while copying to $NODE"
   fi
 done
+
+<% if @s3_bucket %>
+    logger -t copy_attachments "Starting copying attachments to S3 (<%= @s3_bucket -%>)"
+  if envdir /etc/govuk/aws/env.d /usr/local/bin/s3cmd --server-side-encryption sync --skip-existing --deleted-removed "$DIRECTORY_TO_COPY/" "s3://<%= @s3_bucket -%>$DIRECTORY_TO_COPY/"; then
+    logger -t copy_attachments "Attachments copied to S3 (<%- @s3_bucket -%>) successfully"
```
deanwilson: Shouldn't `<%- @s3_bucket -%>` be `<%= @s3_bucket -%>` (= not - at the start)

deanwilson:
Other than the one erb comment it looks good.

jonauman:
Looks good to me

surminus:
Hesitant about merging this because it may sync 500GB into S3. Should check with @rjw1.

deanwilson:
150 USD a month for storage?

surminus:
Bearing in mind this amount of the _same_ data will also be synced in Staging and Production...

deanwilson:
Merge away.

PR #4338, opened 2016-04-18
Ensure rsync cron resources are purged

alexmuller:
These were removed in 1e13c72ea04979305716767b37cc86f5e6898dd4 but I didn't specify the user so the resources were never actually purged from production.


deanwilson:
Merge away.

PR #4339, opened 2016-04-18
Licence Finder's short name is licencefinder

h-lame:
A rogue hyphen snuck in as part of 0b6a732edbe48cab1353b65f530aebe83be16989 meaing we couldn't deploy Licence Finder.


alexmuller:
I think this hyphen came from the fact [the repo name](https://github.com/alphagov/licence-finder) is inconsistent with everything else, but I reckon this change is correct :+1:

PR #4341, opened 2016-04-19
Update duplicity module

alexmuller:
To keep things up to date. This includes a bugfix for backing up to S3 which is something we might want to do in the future.

https://github.com/alphagov/puppet-duplicity/compare/alphagov:74bc95b...alphagov:e238b67

---

We should probably move to a support duplicity module at some point.


deanwilson:
Merge away.

PR #4342, opened 2016-04-19
Clarify alert descriptions for 5xx rate alerts

alexmuller:
We have 2 alerts because the threshold to be phoned out of hours is higher. This commit makes the Icinga interface reflect that.

/cc @jennyd @dsingleton


modules/router/manifests/nginx.pp
```diff
@@ -159,7 +159,7 @@
     target              => "${::fqdn_metrics}.nginx.nginx_requests",
     warning             => $check_requests_warning,
     critical            => $check_requests_critical,
-    desc                => 'nginx requests - too low',
+    desc                => 'Minimum HTTP request rate for www-origin',
```
rjw1: Should this change leak into this commit :)

alexmuller: Stickler for the rules ;)

deanwilson:
Merge away.

PR #4343, opened 2016-04-19
Add DR machines to shutdown overnight

surminus:
We'll save some more money!

Requires https://github.gds/gds/deployment/pull/967

### DIFF
--- a/modules/govuk_jenkins/manifests/job/vapps.pp
+++ b/modules/govuk_jenkins/manifests/job/vapps.pp
@@ -14,6 +14,7 @@
 #
 class govuk_jenkins::job::vapps (
   $vcloud_properties = {},
+  $vcloud_properties_dr = {},
 ) {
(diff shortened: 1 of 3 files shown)


boffbowsh:
I thought the idea behind the DR machines is that they're warmed and ready to go as data sources for our infrastructure should we lose our primary DC? If we shut them off overnight and we lost the primary DC while the DR machines are off, we'll need to restore them from backup first before they're usable. Plus we'll need to actually turn them on manually as we'll have lost the production Jenkins.

In addition, when they get turned on in the morning they may not be able to catch up with replication - they might be too far behind.

If this has all been considered and discussed, I'll shut up :)

surminus:
@boffbowsh it's just for Integration. I probably should have been clearer! They just need to be turned off like we do for the rest of the machines in Integration.

boffbowsh:
That makes more sense :)

deanwilson:
Code looks good. Merge away.

PR #4344, opened 2016-04-19
Update app deployment job

alexmuller:
This change allows us to use Capistrano recipes which are in the app
repo (in public). This is good because it means more people can access
stuff and see how it works.

- We need to wipe the workspace because we're making clones inside clones
  (it only takes about a minute longer)
- For existing apps the process should be exactly the same
- For apps that aren't defined in alphagov-deployment, this change will
  use the Capistrano recipe from the app repo
- We have to remove the Gemfile so that it uses the bundler bundle from
  the main alphagov-deployment (not the app's bundle)

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
@@ -6,7 +6,7 @@
             url: git@github.gds:gds/alphagov-deployment.git
             branches:
               - <%= @alphagov_deployment_branch %>
-            wipe-workspace: false
+            wipe-workspace: true
 
 - job:
     name: Deploy_App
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
```diff
@@ -29,15 +29,24 @@
...
+                echo "---> No deploy.sh found, running 'bundle exec cap \"${DEPLOY_TASK}\"'" >&2
+                exec bundle exec cap "$DEPLOY_TASK"
+              fi
             else
-              echo "---> No deploy.sh found, running 'bundle exec cap \"${DEPLOY_TASK}\"'" >&2
-              exec bundle exec cap "$DEPLOY_TASK"
+              git clone "git@github.com:alphagov/${TARGET_APPLICATION}.git"
+              cd "$TARGET_APPLICATION/"
+              git checkout $TAG
+              rm -f Gemfile
```
boffbowsh: You can specify the Gemfile to use with `BUNDLE_GEMFILE`

```shell
BUNDLE_GEMFILE=${WORKSPACE}/Gemfile bundle exec cap "$DEPLOY_TASK"
```

boffbowsh:
One small comment, otherwise 👍

modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
```diff
@@ -29,15 +29,24 @@
...
+                echo "---> Found deploy.sh, running 'sh -e deploy.sh'" >&2
+                exec sh -e deploy.sh
+              else
+                echo "---> No deploy.sh found, running 'bundle exec cap \"${DEPLOY_TASK}\"'" >&2
+                exec bundle exec cap "$DEPLOY_TASK"
+              fi
             else
-              echo "---> No deploy.sh found, running 'bundle exec cap \"${DEPLOY_TASK}\"'" >&2
-              exec bundle exec cap "$DEPLOY_TASK"
+              git clone "git@github.com:alphagov/${TARGET_APPLICATION}.git"
```
jennyd: Are there any non-alphagov apps we would want to use this approach with?

alexmuller:
Superceded by #4350 which we think is better!

PR #4345, opened 2016-04-20
Allow deploy user to restart postgres on integration and staging.

elliotcm:
This is used by the data sync job to clean up
connections after swapping out the postgres data.


boffbowsh:
👍 for intent

alexmuller:
This looks good to me and I think will work. Maybe you could also add `reload` or `restart` to the end of the line to limit it further.

PR #4347, opened 2016-04-20
Add script for running app consoles.

elliotcm:
Supports Rails apps and apps with a `console`
binary in the root (such as rummager).


boffbowsh:
😍 merge when green 👍

tijmenb:
❤️

tuzz:
💋

alexmuller:
I'm not sure this will actually end up on the box, it might need to be included in the `govuk_scripts` class.

elliotcm:
@alexmuller spoil sport: https://github.com/alphagov/govuk-puppet/pull/4352

PR #4348, opened 2016-04-20
Update project to string rather than choice

surminus:
With the amount of projects rapidly expanding it's a pain to have to add them to hieradata for the sake of a dropdown choice. Dynamically generating the projects in Jenkins is probably possible but isn't worth
the effort. Let's just get the user to type in the project instead, and if it's mistyped it'll just fail anyway.


deanwilson:
Sensible choice and the code looks fine. LGTM MWG.

PR #4350, opened 2016-04-20
Change how all apps deploy

brenetic:
**This cannot be merged until all the apps have been migrated to work this way.**

- This is so everyone is able to see how all the Capistrano config works
for each app.
- This is using the new [public
  repo](https://github.com/alphagov/govuk-app-deployment).
- The Capistrano recipes will all be the above public repo.
- The secrets will remain in alphagov-deployment.
- Also we are now using the master branch of govuk-app-deployment

@alexmuller @brenetic @emmabeynon @klssmith


alexmuller:
Can be merged after https://github.com/alphagov/govuk-app-deployment/pull/6

PR #4351, opened 2016-04-20
moves govuk_crawler_worker rabbitmq to 'mirrorer'

jonauman:
The crawler has been known to crash the production rabbitmq.
It has beed decided to move the rabbitmq process away from
the rabbitmq cluster and to create a local rabbitmq instance
on the mirrorer app server. This will effectively make the
govuk_crawler_worker app self contained and should lighten
the load on the rabbitmq cluster

### DIFF
--- a/modules/govuk/manifests/apps/govuk_crawler_worker.pp
+++ b/modules/govuk/manifests/apps/govuk_crawler_worker.pp
@@ -36,6 +36,7 @@ class govuk::apps::govuk_crawler_worker (
   $airbrake_api_key = '',
   $airbrake_endpoint = '',
   $airbrake_env = '',
+  $amqp_host = 'localhost',
   $amqp_pass = 'guest',
   $blacklist_paths = [],
   $enabled   = false,
@@ -58,7 +59,7 @@ class govuk::apps::govuk_crawler_worker (
       'AIRBRAKE_ENV':
         value => $airbrake_env;
       'AMQP_ADDRESS':
-        value => "amqp://govuk_crawler_worker:${amqp_pass}@rabbitmq-1:5672/";
+        value => "amqp://govuk_crawler_worker:${amqp_pass}@${amqp_host}:5672/";
       'AMQP_EXCHANGE':
         value => 'govuk_crawler_exchange';
       'AMQP_MESSAGE_QUEUE':
@@ -93,5 +94,7 @@ class govuk::apps::govuk_crawler_worker (
       command            => './govuk_crawler_worker -json',
       health_check_path  => '/healthcheck',
     }
+
+    include govuk::apps::govuk_crawler_worker::rabbitmq
   }
 }
--- a/modules/govuk/manifests/node/s_mirrorer.pp
+++ b/modules/govuk/manifests/node/s_mirrorer.pp
@@ -1,5 +1,6 @@
 # Node defintion for mirrorer-?.management boxen
 class govuk::node::s_mirrorer inherits govuk::node::s_base {
   include govuk_crawler
+  include govuk_rabbitmq
   include nginx
 }
--- a/hieradata/class/mirrorer.yaml
+++ b/hieradata/class/mirrorer.yaml
@@ -1,6 +1,8 @@
 ---
 govuk_crawler::mirror_root: '/mnt/crawler_worker'
 
+govuk::apps::govuk_crawler_worker::amqp_host: "%{hiera('govuk_crawler::amqp_host')}"
+
 govuk::node::s_base::apps:
   - govuk_crawler_worker
 
--- a/hieradata/class/rabbitmq.yaml
+++ b/hieradata/class/rabbitmq.yaml
@@ -4,7 +4,6 @@ govuk::node::s_rabbitmq::apps_using_rabbitmq:
   - backdrop_write
   - content_register
   - email_alert_service
-  - govuk_crawler_worker
   - panopticon
   - publishing_api
   - rummager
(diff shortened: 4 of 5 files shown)


modules/govuk/manifests/apps/govuk_crawler_worker.pp
```diff
@@ -58,7 +59,7 @@
       'AIRBRAKE_ENV':
         value => $airbrake_env;
       'AMQP_ADDRESS':
-        value => "amqp://govuk_crawler_worker:${amqp_pass}@rabbitmq-1:5672/";
+        value => "amqp://${amqp_host}:${amqp_pass}@rabbitmq-1:5672/";
```
boffbowsh: This is changing the username, not the host.

modules/govuk/manifests/node/s_mirrorer.pp
```diff
@@ -1,5 +1,11 @@
 # Node defintion for mirrorer-?.management boxen
 class govuk::node::s_mirrorer inherits govuk::node::s_base {
+
   include govuk_crawler
   include nginx
+
+  include govuk_rabbitmq
```
boffbowsh: You'll need to configure a RabbitMQ user for the crawler worker to connect on.

alexmuller: I think that can be achieved by doing `include govuk::apps::govuk_crawler_worker::rabbitmq` inside `govuk::apps::govuk_crawler_worker`.

modules/govuk/manifests/apps/govuk_crawler_worker.pp
jonauman: Good catch. Fixed it in latest commit

alexmuller:
On second thoughts I wonder whether we shouldn't parameterise the hostname at all? We're very unlikely to need to change it in the future and it defaulting to `rabbitmq-1` means that if our hieradata wasn't being applied correctly it would default to the publishing RabbitMQ.

alexmuller:
As we discovered yesterday, there's also a reference to `rabbitmq-1.backend` in one of the seed scripts - we should use a lookup in the hieradata to ensure that those 2 values are always set to the same thing.

alexmuller:
@jonauman: I've updated your commit with this diff, hope that's ok:

```diff
commit d08f293039c1ee3248ec5d843f5e5cb00f392192
Author: Alex Muller <alex.muller@digital.cabinet-office.gov.uk>
Date:   Tue Apr 26 17:46:32 2016 +0100

    fixup

diff --git a/hieradata/class/mirrorer.yaml b/hieradata/class/mirrorer.yaml
index 678f887..7fa621b 100644
--- a/hieradata/class/mirrorer.yaml
+++ b/hieradata/class/mirrorer.yaml
@@ -1,9 +1,10 @@
 ---
 govuk_crawler::mirror_root: '/mnt/crawler_worker'
 
+govuk::apps::govuk_crawler_worker::amqp_host: "%{hiera('govuk_crawler::amqp_host')}"
+
 govuk::node::s_base::apps:
   - govuk_crawler_worker
-govuk::apps::govuk_crawler_worker::amqp_host: 'localhost'
 
 icinga::client::checks::disk_time_warn: 150 # milliseconds
 icinga::client::checks::disk_time_critical: 250 # milliseconds
diff --git a/hieradata/common.yaml b/hieradata/common.yaml
index 590f8fc..688b9a1 100644
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -433,7 +433,7 @@ govuk_cdnlogs::service_port_map:
   assets: 6515
   bouncer: 6516
 
-govuk_crawler::amqp_host: 'rabbitmq-1.backend'
+govuk_crawler::amqp_host: 'localhost'
 govuk_crawler::site_root: 'https://www.gov.uk'
 
 govuk_elasticsearch::repo::apt_mirror_hostname: "%{hiera('apt_mirror_hostname')}"
diff --git a/modules/govuk/manifests/apps/govuk_crawler_worker.pp b/modules/govuk/manifests/apps/govuk_crawler_worker.pp
index b4be230..2c392f3 100644
--- a/modules/govuk/manifests/apps/govuk_crawler_worker.pp
+++ b/modules/govuk/manifests/apps/govuk_crawler_worker.pp
@@ -36,7 +36,7 @@ class govuk::apps::govuk_crawler_worker (
   $airbrake_api_key = '',
   $airbrake_endpoint = '',
   $airbrake_env = '',
-  $amqp_host = 'rabbitmq-1',
+  $amqp_host = 'localhost',
   $amqp_pass = 'guest',
   $blacklist_paths = [],
   $enabled   = false,
@@ -94,5 +94,7 @@ class govuk::apps::govuk_crawler_worker (
       command            => './govuk_crawler_worker -json',
       health_check_path  => '/healthcheck',
     }
+
+    include govuk::apps::govuk_crawler_worker::rabbitmq
   }
 }
diff --git a/modules/govuk/manifests/node/s_mirrorer.pp b/modules/govuk/manifests/node/s_mirrorer.pp
index 84c9407..03b6802 100644
--- a/modules/govuk/manifests/node/s_mirrorer.pp
+++ b/modules/govuk/manifests/node/s_mirrorer.pp
@@ -1,11 +1,6 @@
 # Node defintion for mirrorer-?.management boxen
 class govuk::node::s_mirrorer inherits govuk::node::s_base {
-
   include govuk_crawler
-  include nginx
-
   include govuk_rabbitmq
-
-  include govuk::apps::govuk_crawler_worker::rabbitmq
-
+  include nginx
 }
```

Would you mind reviewing the overall commit and making sure it's ok?

jonauman:
@alexmuller Yes looks good. I concur it is better to move the variable dependency for govuk_crawler::amqp_host from the module to hieradata.

PR #4353, opened 2016-04-21
Revert no-cache rule for front-end homepage

fofr:
The code to remove non-js tracking has been removed in:
https://github.com/alphagov/frontend/pull/944

Also remove the corresponding parts added in:
https://github.gds/gds/puppet/pull/978
https://github.com/alphagov/govuk-puppet/commit/003254aad44ce391b595220f707ccba7367c1adf
https://github.com/alphagov/govuk-puppet/commit/e10bc998cc9229308c321f00cc6a0129b846582e

### DIFF
--- a/modules/govuk/manifests/apps/frontend.pp
+++ b/modules/govuk/manifests/apps/frontend.pp
@@ -8,10 +8,6 @@
 #   The port that the app is served on.
 #   Default: 3005
 #
-# [*enable_homepage_nocache_location*]
-#   Adds nginx configuration to never cache the homepage/no-cache location.
-#   Default: true
-#
 # [*vhost_protected*]
 #   Should this vhost be protected with HTTP Basic auth?
 #   Default: undef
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/apps/frontend.pp
```diff
@@ -28,23 +24,12 @@
...
-  if ($enable_homepage_nocache_location) {
-    $nginx_extra_config = '
-  location ^~ /frontend/homepage/no-cache/ {
-    expires epoch;
-  }
-'
-  } else {
-    $nginx_extra_config = ''
-  }
+  $nginx_extra_config = ''
```
boffbowsh: You can remove this line and the parameter from the `govuk::app` declaration below, as the default is `''`

fofr: Updated in ea2604f

PR #4356, opened 2016-04-22
Stop GOVUK-Request-Id from being overwritten.

elliotcm:
Requests from backend apps to backend apps go back
through the relevant load balancers and so get their
ID reset.

This alternate method should only set the variable
if it is not already set.  For prior art see
https://github.com/alphagov/govuk-puppet/blob/ef5f4bc4f9ff011ce592098dd7f9f4c71e4648d8/modules/nginx/files/map-sts.conf


deanwilson:
Looks fine. Merge away.

PR #4359, opened 2016-04-25
Set GOVUK-Request-Id if not already set.

elliotcm:
Previously it was overwritten on downstream requests
that went back through the load balancers.

I have confirmed nginx on the backend loadbalancer will restart cleanly with this config.  Whether it actually does what we want is another matter.


rjw1:
Certianly doesnt break nginx starting

PR #4361, opened 2016-04-26
Add monitoring for archiving

surminus:
When we use WAL-E for backups we switch on the WAL continuous archiving option
(http://www.postgresql.org/docs/9.3/static/continuous-archiving.html). Everytime a WAL segment is archived ends up in S3, and the archive timeout option is set to 60 seconds, which means we will push something to S3 every 60 seconds.

This adds a wrapper around the command which includes a passive check set to a threshold of 15 minutes. If a WAL segment hasn't been archived for over 15 minutes then we'll be alerted (this might be a problem overnight in Integration when the machines turn off).

I've gone for this option for ease of use, though it may be better to gather metrics somehow on how far behind the archiving is, but ideally we just need to know if there's been a problem with the archiving process which this PR achieves. There might also be questions about how often we want to archive WAL segments, and if we want to make it less frequent than every minute (which could potentially save on disk and bandwidth charges to S3)?

### DIFF
--- a/modules/govuk_postgresql/manifests/wal_e/backup.pp
+++ b/modules/govuk_postgresql/manifests/wal_e/backup.pp
@@ -67,7 +67,7 @@ class govuk_postgresql::wal_e::backup (
       'archive_mode':
         value => 'on';
       'archive_command':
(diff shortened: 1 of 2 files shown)


jonauman:
Bob and I have a had a looks at this and it looks good. Merge when ready.

PR #4362, opened 2016-04-26
refactors python backup script and adds hieradata

jonauman:
The python script has been refactored to use functions for looping
through databases to create one file per database. Necessary hiearadata
has been added to the hieradata/class/%environment% location for
integration.


jonauman:
tested on vagrant box and works except that I've given it the wrong S3 bucket name (on purpose). The script fails on error as designed.

modules/mongodb/manifests/server.pp
```diff
@@ -28,6 +28,7 @@
   $oplog_size = undef,
   $replicaset_members = {},
   $development = false,
+  $s3_backups = false,
```
alexmuller: Is there any reason to define this parameter on this class?

hieradata/class/integration/mongo.yaml
```diff
@@ -0,0 +1,13 @@
...
+mongodb::s3backup::backup::standalone: 'True'
+mongodb::server::s3_backups: true
+mongodb::server::replicaset_members:
+  'mongo-1':
+
+
+# mongodb::s3backup::backup::aws_access_key_id: 
+# mongodb::s3backup::backup::aws_secret_access_key: 
+# mongodb::s3backup::backup::private_gpg_key_fingerprint:
+# mongodb::s3backup::backup::private_gpg_key: |
```
surminus: These have been merged now.

modules/mongodb/manifests/server.pp
jonauman: Yes, server.pp is the entry point for backup.pp so I thought it would also be a better place to include s3backup/* classes. I want to be able to include all the s3backup classes if this is turned on, regardless of whether regular backups are turned on.

hieradata/class/integration/mongo.yaml
```diff
@@ -0,0 +1,7 @@
+---
+
+mongodb::s3backup::backup::s3_bucket: govuk-mongodb-backup-s3-integration
+mongodb::s3backup::backup::standalone: 'True'
+mongodb::server::s3_backups: true
+mongodb::server::replicaset_members:
+  'mongo-1':
```
rjw1: is this correct?

rjw1: I dont think this should be here at all. It will actually break the mongo cluster.

hieradata/class/integration/mongo.yaml
```diff
@@ -0,0 +1,7 @@
+---
+
+mongodb::s3backup::backup::s3_bucket: govuk-mongodb-backup-s3-integration
+mongodb::s3backup::backup::standalone: 'True'
```
rjw1: not true in this case. should be false. 
also shouldn't be capitalised.

hieradata/class/vagrant/mongo.yaml
```diff
@@ -0,0 +1,49 @@
+---
+mongodb::backup::s3_backups: true
+mongodb::server::replicaset_members:
+  'mongo-1':
+
+mongodb::s3backup::backup::aws_access_key_id: 1234567890 
+mongodb::s3backup::backup::aws_secret_access_key: ABCDEFGHIJKLMNOP1234567890
+mongodb::s3backup::backup::s3_bucket: govuk-mongodb-backup-s3-vagrant
+mongodb::s3backup::backup::standalone: 'True'
```
rjw1: not actually true in vagrant either.

hieradata/class/vagrant/mongo.yaml
```diff
@@ -0,0 +1,49 @@
+---
+mongodb::backup::s3_backups: true
+mongodb::server::replicaset_members:
```
rjw1: Again this probably shoudnt be added.

modules/mongodb/manifests/s3backup/cron.pp
```diff
@@ -3,7 +3,7 @@
 
   cron { 'mongodb-s3backup':
     command => '/usr/local/bin/mongodb-backup-s3-wrapper',
-    user    => $mongodb::s3backup::user,
+    user    => $mongodb::s3backup::backup::user,
```
rjw1: does this actually work?

modules/mongodb/templates/mongodb-backup-s3.erb
```diff
@@ -17,65 +17,105 @@ date_format = '%Y-%m-%d:%H:%M:%S'
...
 s3_bucket = '<%= @s3_bucket %>'
-standalone = '<%= @standalone %>'
+standalone = <%= @standalone %>
 start = datetime.now().strftime(date_format)
 start_time = time.time()
-file = 'mongodump-<%= @fqdn %>-' + start + '.tar.gz.gpg'
+today = datetime.today().strftime('%Y-%m-%d')
 ############################################
 
 def logger(str):
```
rjw1: is this just going to stdout?

hieradata/class/integration/mongo.yaml
jonauman: You are correct, sorry copied from vagrant environment

hieradata/class/integration/mongo.yaml
jonauman: Correct again, it should be False. However, this is actually a string and not a boolean, because it is inserted into a python script as a python boolean. In Python, booleans are capitalised.

hieradata/class/vagrant/mongo.yaml
jonauman: I was having problem running puppet on a single vagrant mongo-1 instance until I added this. There is an exec that wants to contact all members in the replica set

modules/mongodb/manifests/s3backup/cron.pp
jonauman: Yes

modules/mongodb/templates/mongodb-backup-s3.erb
jonauman: Yes. The script is called by a wrapper script that sends stdout/err to syslog

hieradata/class/vagrant/mongo.yaml
alexmuller: I see that this is configured as a string because when it's interpolated into Python that's correct, but I think it would be less confusing to have a boolean `true/false` YAML value which is then converted into a Python `True/False` later.

surminus:
I think this is ready for testing again. One thing I came across in the PostgreSQL stuff was being able to define several different AWS credentials and buckets within the same environment, but this could probably be redefined later when this has been fully tested in Integration. 

So my opinion is :+1: and go merge.

modules/mongodb/manifests/s3backup/backup.pp
```diff
@@ -121,6 +121,13 @@
     require => User[$user],
   }
 
+  file { $backup_dir:
+    ensure => directory,
+    owner  => $user,
+    group  => $user,
+    mode   => '0650',
```
deanwilson: I assume this should be `0640` or `0750`

modules/mongodb/templates/mongodb-backup-s3.erb
```diff
@@ -17,65 +17,109 @@ date_format = '%Y-%m-%d:%H:%M:%S'
...
-delta = end_time - start_time
-
-print 'Script completed in     : %.1f seconds' % delta
+def get_mongo_nodes():
+    if not standalone:
+      logger('Find Secondary Mongo members')
+      try:
+          mongo_nodes = subprocess.check_output(["mongo",  mongo_server, "--quiet", "--eval", "printjson(rs.status().members.map(function(m) { return {'name':m.name, 'stateStr':m.stateStr} }))"])
+      except OSError:
+          sys.exit('Trouble querying mongo server')
```
deanwilson: Can we print the server it's trying to connect to in the failure please?

modules/mongodb/templates/mongodb-backup-s3.erb
```diff
@@ -17,65 +17,109 @@ date_format = '%Y-%m-%d:%H:%M:%S'
...
+        s3_key = str(today) + '/' + str(enc_file)
+        source = backup_dir + '/' + str(enc_file)
+        s3.Object(s3_bucket, s3_key).put(Body=open(source, 'rb'))
+        logger('End upload file %s to S3' % enc_file)
+    except Exception as e:
+        sys.exit(e)
+
+def cleanup_local():
+    logger('Start cleanup filesystem')
+#    delete_tree = backup_dir + '/*'
```
deanwilson: Might as well remove the dead code.

modules/mongodb/manifests/s3backup/backup.pp
jonauman: @deanwilson Yes, changed

modules/mongodb/templates/mongodb-backup-s3.erb
jonauman: @deanwilson Yes, changed

modules/mongodb/templates/mongodb-backup-s3.erb
jonauman: @deanwilson Yes, changed

modules/mongodb/templates/mongodb-backup-s3.erb
```diff
@@ -17,65 +17,109 @@ date_format = '%Y-%m-%d:%H:%M:%S'
 mongo_server = '<%= @server %>'
 key_fingerprint = '<%= @private_gpg_key_fingerprint %>'
 s3_bucket = '<%= @s3_bucket %>'
-standalone = '<%= @standalone %>'
+<% if standalone -%>
```
deanwilson: I really dislike this approach. I don't like the idea that the application code changes based upon puppet values. I'd much rather see this done as command line arguments to the script and have the invoking cronjob pass in the arguments that come from the hiera lookups.

jonauman: Sorry, I don't agree. I want the script to do exactly what I tell it to do based on my environment. Environment is handled by hiera. My approach is also consistent with the rest of the module:

```
DaveOps:templates jonauman$ git grep -A5 -E "if @" automongodbbackup
automongodbbackup:<% if @domonthly == false -%>
automongodbbackup-DOMONTHLY=00
automongodbbackup-<% else -%>
automongodbbackup-DOMONTHLY=01
automongodbbackup-<% end -%>
```

deanwilson: What you're doing is dynamically *re-writing* the script to be different based on your environment values. Telling it to do different things for each environment should be done via command line options which would be based on your environment. Environment would still be handled by hiera. In general do you re-write many apps when you change environment?

I'll agree with it being consistent but I don't think it's correct.

jonauman: Sorry, I still don't agree. I would argue that passing a config file is a better and more common practice (think init.d and upstart) than passing a bunch of command options. But still it's a matter of style which is subjective. For simplicity, I have opted to combine the config file and the code into the same file. The code has not changed, only the configuration variables.  

If anything, I would separate the config variables into a separate file and pass that as the only parameter. But why bother for a single purpose script when we've not done it elsewhere in the module?

rjw1: Can we not just use code to determine if its standalone? 
Given we already shell out to find the secondaries the logic there could determine if its stand alone or not.

jonauman:
@rjw1 Logic to select localhost server if no secondary mongo is found has been moved to python code as suggested.

hieradata/class/integration/mongo.yaml
```diff
@@ -0,0 +1,4 @@
+---
+
+mongodb::s3backup::backup::s3_bucket: govuk-mongodb-backup-s3-integration
+mongodb::server::s3_backups: true
```
alexmuller: I think this should probably be `mongodb::backup::s3_backups`

hieradata/class/vagrant/mongo.yaml
```diff
@@ -0,0 +1,47 @@
+---
+mongodb::backup::s3_backups: true
+
+mongodb::s3backup::backup::aws_access_key_id: '1234567890abcd' 
+mongodb::s3backup::backup::aws_secret_access_key: 'ABCDEFGHIJKLMNOP1234567890'
+mongodb::s3backup::backup::s3_bucket: govuk-mongodb-backup-s3-vagrant
+mongodb::s3backup::backup::standalone: true
```
alexmuller: This param was removed in the 2nd commit

hieradata/class/integration/mongo.yaml
jonauman: Yes, good catch.

hieradata/class/vagrant/mongo.yaml
jonauman: That's correct. It has been replaced by logic within the python script.

alexmuller: Yep, so I think it can be removed as part of 56b6088dc268e5cb7a280a4d554eab02c0979d7a to keep the hieradata tidy.

modules/mongodb/templates/mongodb-backup-s3.erb
```diff
@@ -17,65 +17,100 @@ date_format = '%Y-%m-%d:%H:%M:%S'
...
+        logger('End upload file %s to S3' % enc_file)
+    except Exception as e:
+        sys.exit(e)
+
+def cleanup_local():
+    logger('Start cleanup filesystem')
+    if backup_dir not in ['','/']:
+        rm_tree = backup_dir + '/'
+        try:
+            subprocess.call(['rm', '-rf', rm_tree])
```
alexmuller: This is pretty scary - it feels like there's quite a bit of potential for this to be misconfigured and remove something we care about.

Could we do something else instead? Given that this script is the same script that pushes the file to S3 would it be possible to remove the exact file we've pushed immediately after we push it?

modules/mongodb/manifests/s3backup/backup.pp
```diff
@@ -61,7 +57,6 @@
   $private_gpg_key_fingerprint,
   $s3_bucket  = 'govuk-mongodb-backup-s3',
   $server = 'localhost',
-  $standalone  = False,
   $user = 'govuk-backups'
```
alexmuller: The `backup::client` class creates a `govuk-backup` user already - could we reuse that instead of creating another user with almost the same name?

modules/mongodb/templates/mongodb-backup-s3.erb
alexmuller: I don't think this redirection is working - I'm getting stuff in `/var/mail/govuk-backups` when I run this in vagrant.

modules/mongodb/templates/mongodb-backup-s3.erb
```diff
@@ -17,65 +17,100 @@ date_format = '%Y-%m-%d:%H:%M:%S'
 mongo_server = '<%= @server %>'
 key_fingerprint = '<%= @private_gpg_key_fingerprint %>'
 s3_bucket = '<%= @s3_bucket %>'
-standalone = '<%= @standalone %>'
 start = datetime.now().strftime(date_format)
```
alexmuller: This variable `start` is no longer used as of this pull request so it can be removed.

rjw1:
Closed in favour of https://github.com/alphagov/govuk-puppet/pull/4453

PR #4363, opened 2016-04-26
Backend request tracing

elliotcm:
Use perl to set GOVUK-Request-Id if not set.

Tested as "installs nginx with perl and the config doesn't make nginx angry" on backend, frontend, api and licensify load balancer VMs.


rjw1:
yay! :dancers: :dancers: :party: :eggplant:

boffbowsh:
giphy

boffbowsh:
If we're going to add intelligence to nginx to provide a request id, maybe it should be a uuid rather than a pseudo-random string of identifiable request data. http://ryandlane.com/blog/2014/12/11/using-lua-in-nginx-for-unique-request-ids-and-millisecond-times-in-logs/

elliotcm:
@boffbowsh I wanted to avoid going down the path of adding perl or lua modules at this point, neither languages have UUIDs in their stdlib.

boffbowsh:
Fair enough 👍

elliotcm:
@rjw1 / @boffbowsh Are you OK with me merging this?

modules/loadbalancer/templates/nginx_balance.conf.erb
```diff
@@ -18,10 +18,25 @@ server {
...
+    my $r = shift;
+    my $current_header = $r->header_in("GOVUK-Request-Id");
+
+    if (defined $current_header && $current_header ne "") {
+      return $current_header;
+    } else {
+      my $pid = $r->variable("pid");
+      my $msec = $r->variable("msec");
+      my $remote_addr = $r->variable("remote_addr");
+      my $request_length = $r-variable("request_length");
```
steventux: shouldn't this be:
`my $request_length = $r->variable("request_length");`
assuming it's de-referencing `request_length` from `$r` and not some other tricksy Perl thing?

elliotcm: You are right, will fix

elliotcm:
@steventux updated

jabley:
And so we go full circle :sob: https://github.com/gds-attic/pp-puppet/pull/586

PR #4365, opened 2016-04-27
Remove content-register

danielroseman:
Once this is merged and puppet has removed everything from the servers, there will be a follow-up PR to remove the puppet classes themselves.

The db will need to be dropped manually.


elliotcm:
Just missing https://github.com/alphagov/govuk-puppet/blob/remove-content-register/hieradata/staging.yaml#L12

danielroseman:
@elliotcm added.

elliotcm:
👍

hieradata/class/backend.yaml
```diff
@@ -51,7 +51,6 @@ govuk::node::s_base::apps:
   - canary_backend
   - collections_publisher
   - contacts
-  - content_register
```
alexmuller: I don't think you want to do this just yet - it needs to stay in here for the `ensure => absent` to take effect.

danielroseman: @alexmuller removed.

PR #4366, opened 2016-04-27
Support old rails in govuk_app_console

elliotcm:
Before rails switched to the more standard `bin`
directory for storing the `rails` executable it
used the directory `script`, but it works the same.


MatMoore:
👍  Thanks! Merge when green

PR #4367, opened 2016-04-27
Add dashboard to hold graphs for publishing api queues

benhyland:
We introduce a graph for panopticon only, non-erb-templated, as a first pass.
We will tidy and add extra graphs soon, but want to merge as we go
because there is still incremental value here.

https://trello.com/c/Ipl9UVsA/578-monitor-rummager-s-current-queue-size-with-statsd-graphite-m

This graph shows all the stats that might be reported by the
govuk_message_queue_consumer gem.

It ensures that the height of the graph is always representing actual stat values,
by using summarize and consolidateBy(max) graphite functions.
See http://graphite.readthedocs.org/en/latest/functions.html
As a result, it is possible for the area under the graph to not accurately represent
the running total correctly - but this feature is not expected by dev users (i.e. us!).
The tradeoff is necessary due to the aggregation and bucketing done by statsd
and graphite, and by the possibility that there might be fewer pixels than buckets.

@jackscotti and @benhyland

### DIFF
--- a/modules/grafana/files/dashboards/publishing_queues.json
+++ b/modules/grafana/files/dashboards/publishing_queues.json
@@ -1,0 +1,129 @@
+{
+  "title": "Publishing Queues",
+  "style": "dark",
+  "timezone": "utc",
+  "editable": true,
+  "hideControls": false,
+  "sharedCrosshair": false,
+  "rows": [
+    {
+      "title": "Panopticon",
+      "height": "250px",
+      "editable": true,
+      "collapse": false,
+      "panels": [
+        {
+          "title": "Panopticon Consumers",
+          "error": false,
+          "span": 10,
+          "editable": true,
+          "type": "graph",
+          "id": 1,
+          "datasource": null,
+          "renderer": "flot",
+          "x-axis": true,
+          "y-axis": true,
+          "y_formats": [
+            "short",
+            "short"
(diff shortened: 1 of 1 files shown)


tijmenb:
Can you add a screenshot?

benhyland:
screen shot 2016-04-28 at 12 36 31

modules/grafana/files/dashboards/publishing_queues.json
```diff
@@ -0,0 +1,129 @@
...
+          },
+          "nullPointMode": "connected",
+          "steppedLine": false,
+          "tooltip": {
+            "value_type": "cumulative",
+            "shared": false
+          },
+          "targets": [
+            {
+              "target": "alias(consolidateBy(summarize(stats_counts.govuk.app.panopticon.panopticon.started, '1minute', 'sum'), 'max'), 'started')"
```
tijmenb: So:

```
started = ( acked + retried + discarded + uncaught_exception - currently processing)
```

Would it be better to split started into another thing, or remove it from this graph?

benhyland: I don't want to get into too much refinement until we have stats up on a board and start using them. We can figure out tweaks gradually as we come to understand what signals we value the most.

PR #4368, opened 2016-04-28
Update ssh key for Ikenna

ikennaokpala:
Updating my SSH Key as a result for re-recreating my dev VM


alexmuller:
:+1:

It might be worth putting a copy of the private key in `~/.ssh` on your host machine as well, to avoid having to change it here every time.

ikennaokpala:
@alexmuller lesson learnt.. cheers review 👍

PR #4370, opened 2016-04-28
Fixes for rebuilding service manual search index

thehenster:
After deploying designprincples we want to purge the service-manual search index and rebuild it so old pages no longer exist in search. It probably once worked but something changed at some point.


surminus:
:+1: LGTM, merge when ready!

PR #4372, opened 2016-04-28
Use new deployment repo for post-deploy repo backup job

jennyd:
All of the shared deployment code from `alphagov-deployment`, including the
`sync_to_ghe.sh` script which this job uses, is [already in the new repo](https://github.com/alphagov/govuk-app-deployment/blob/master/sync_to_ghe.sh).

The [`Deploy_App` job uses the new repo already](https://github.com/alphagov/govuk-puppet/pull/4350) so we should update this job too
for consistency. This change brings us nearer to getting all deployment code
out of `alphagov-deployment`.


jennyd:
Once this has been deployed we can remove the `sync_to_ghe.sh` script from alphagov-deployment: https://github.gds/gds/alphagov-deployment/pull/1187

PR #4374, opened 2016-04-29
Upgrade typesdb for collectd server

benhyland:
We need the new types for the collectd on the graphite box to
understand the new stats from https://github.com/alphagov/govuk-puppet/pull/4373


jackscotti:
👍

PR #4375, opened 2016-04-29
Change Jenkins job recipients away from webops

alexmuller:
The webops list is for the webops community and isn't GOV.UK specific.


rjw1:
We should probably have an opsmanual entry for the CDN ranges one. Also could we add a link for that into the mail. 
:+1: though. merge when ready if you ask me.

alexmuller:
I agree :)

- https://github.gds/gds/govuk-provisioning/pull/534
- https://github.gds/gds/opsmanual/pull/689
- https://github.com/alphagov/govuk-cdn-config/pull/5

PR #4376, opened 2016-04-29
Make govuk-cdn-logs-monitor deployable

brenetic:
- Added govuk-cdn-logs-monitor to deployable apps list


tijmenb:
👍  good to go!

PR #4378, opened 2016-04-29
Graph queue depths for rummager

benhyland:
Since our first pass at consumer results in
https://github.com/alphagov/govuk-puppet/pull/4367 we have decided what we
want in our first pass at graphs for rummager queue health.

This adds a stacked graph of rummager consumer outcomes, a line graph for
messages on the queue ready to be delivered, and another for the messages delivered
to a consumer but not yet acknowledged.


benhyland:
screen shot 2016-04-29 at 14 44 29
screen shot 2016-04-29 at 14 44 56

tijmenb:
👍  looks really nice, can't wait to turn on the consumer for rummager!

PR #4380, opened 2016-05-03
Completely remove content-register code.

danielroseman:
Now that "ensure => 'absent'" has run on all content-register classes, we can remove them completely.


elliotcm:
I think this repo is self-merge so 👍

PR #4382, opened 2016-05-04
Define imagemagick policies

MatMoore:
We use imagemagick to convert images in whitehall and backend publishers.

- Put existing ImageMagick policy file into puppet
-  Disable potentially dangerous imagemagick features


alexmuller:
This looks good to me but I'd like somebody else to review it briefly too.

surminus:
:+1: fftm. It'll be interesting to hear if it does the job

PR #4383, opened 2016-05-04
Create a READ ONLY role on Publishing API database

Davidslv:
The intention of this role is to allow the creation of a user that will be assigned this role, with the sole purpose of being used in [Finding Things Migration Checker](https://github.com/alphagov/finding-things-migration-checker)

**Please let me know what I missed**, I'm not used to Puppet at all, thank you.

For more information please [read here](https://trello.com/c/DYFgXNYr/621-automate-link-checker-tool#)

### DIFF
--- a/modules/govuk/manifests/apps/publishing_api/postgresql_role.pp
+++ b/modules/govuk/manifests/apps/publishing_api/postgresql_role.pp
@@ -1,0 +1,21 @@
+# == Class: govuk::apps::publishing_api::postgresql_role
+#
+# Creates a `migration_checker` role in publishing-api PostgreSQL database
+# with READ ONLY privileges
+class govuk::apps::publishing_api::postgresql_role ( $password = '' ) {
+  $role = 'migration_checker'
+  $db   = 'publishing_api_production'
+
+  postgresql::server::role { $role:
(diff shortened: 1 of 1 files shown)


MatMoore:
I think this should be named something other than `govuk::apps::publishing_api::postgresql_role`, otherwise it could be confused with the more generic govuk_postgresql module that sets up the user publishing api uses.

Maybe something like `govuk::apps::finding_things_migration_checker::publishing_api_role`?

You probably also need to set up hieradata to override $db per environment.

jennyd:
There are a few other things you'll need to add to make this work - perhaps not everything in the existing [bouncer role](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/bouncer/postgresql_role.pp) but much of it, and also a [`pg_hba.conf` rule](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/node/s_transition_postgresql_slave.pp#L24-L30) to allow this role to authenticate. I'm happy to talk through those examples with you to explain why each part is needed - I regret now not writing a fuller [commit message](https://github.com/alphagov/govuk-puppet/commit/00213f0e22617cbc5b29341e0c894df5034776a2) when I added it originally.

PR #4384, opened 2016-05-04
Refactor WAL-E backups class into defined type

surminus:
In order to pass in several different sets of AWS credentials into each environment the backups class for WAL-E has been refactored into a defined type, with each PostgreSQL primary with the required details.

Secrets have been added to vagrant_credentials.yaml for passing node spec tests.

The `archive_timeout` parameter has an number set as a string rather than an integer. This is because it's passed in to the `postgresql::server::config_entry` as a string (not sure if works as an integer /cc @deanwilson)


deanwilson:
Looks fine. Merge when green.

"The *number* of seconds"

PR #4385, opened 2016-05-05
Rename smokey-nagios to smokey-loop

alexmuller:
This confuses everyone who comes across it (including me). The fact that smokey runs in a loop is hardly anything to do with the fact it's consumed by Icinga (not Nagios).

### DIFF
--- a/modules/monitoring/manifests/checks/smokey.pp
+++ b/modules/monitoring/manifests/checks/smokey.pp
@@ -1,8 +1,8 @@
 # Class: monitoring::checks::smokey
 #
-# Nagios checks based on Smokey (cucumber HTTP tests). smokey-nagios runs
-# periodically and writes the results to a JSON file atomically. That output
-# is read by Nagios for each of the check_feature resources.
+# Monitoring checks based on Smokey (cucumber HTTP tests). smokey-loop runs
+# constantly and writes the results to a JSON file atomically. That output
+# is read by Icinga for each of the check_feature resources.
 #
 # === Parameters
 #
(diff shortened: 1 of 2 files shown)


surminus:
I also love it how [this](https://github.com/alphagov/govuk-puppet/blob/fd51ac0c86f9c481650130d8b21c155d041e5b8a/modules/monitoring/files/etc/init/smokey-nagios.conf#L9) is run as a service in a loop rather than in a cron job

alexmuller:
Yeah, it's a bit different. I think we do it like that because we want it running as frequently as possible. But it's been a long time.

surminus:
Yeah, I just meant the name of it is "cron" even though it's not used anywhere as a cronjob.

Anyway, :+1: fftm

alexmuller:
Oh yeah, I totally missed that. That's silly. Hold tight, let's fix it.

alexmuller:
https://github.com/alphagov/smokey/pull/178

modules/monitoring/manifests/checks/smokey.pp
```diff
@@ -14,16 +14,26 @@
...
+  service { 'smokey-nagios':
+    ensure => stopped,
+  } ->
   file { '/etc/init/smokey-nagios.conf':
+    ensure => absent,
+  }
+
+  $service_file = '/etc/init/smokey-loop.conf'
+
+  # TODO: Should this really run as root?
```
deanwilson: Probably not. Add an investigation to the ideas backlog?

deanwilson:
Merge away.

PR #4390, opened 2016-05-06
Add CDN backend graph to edge health dashboard

alexmuller:
To show where requests are being served from.


deanwilson:
Merge when ready.

PR #4392, opened 2016-05-09
Don't reboot while logrotate is running

alexmuller:
We have a theory that machines rebooting while logrotate is running causes logrotate to get confused and never rotate some logfiles, which causes the root partition to fill up, which is really annoying.

### DIFF
--- a/modules/govuk_unattended_reboot/templates/etc/unattended-reboot/check/00_safety.erb
+++ b/modules/govuk_unattended_reboot/templates/etc/unattended-reboot/check/00_safety.erb
@@ -3,3 +3,8 @@
 if ! [[ "$(/usr/local/bin/govuk_setenv default /usr/bin/printenv SAFE_TO_REBOOT)" =~ ^(yes|overnight)$ ]]; then
   exit 1
 fi
+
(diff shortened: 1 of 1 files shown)


surminus:
:+1: Super! Interested to see if the theory is more than just a theory. Also, why do I never think of `pgrep`? Much more complete than `pidof`, which I tend to think of.

modules/govuk_unattended_reboot/templates/etc/unattended-reboot/check/00_safety.erb
```diff
@@ -3,3 +3,8 @@
 if ! [[ "$(/usr/local/bin/govuk_setenv default /usr/bin/printenv SAFE_TO_REBOOT)" =~ ^(yes|overnight)$ ]]; then
   exit 1
 fi
+
+# Don't reboot while logrotate is running
+if pgrep "logrotate" > /dev/null; then
```
rjw1: full path?

alexmuller: Amended

PR #4393, opened 2016-05-09
Add recovery options

surminus:
This just creates the required recovery.conf file in the PostgreSQL config directory. It also updates a couple of parameters to use globally set parameters by the main postgresql class because I did not want to hardcode version numbers.


surminus:
Opened prematurely after testing

PR #4394, opened 2016-05-09
Configure kibana from the environment

alexmuller:
Instead of using alphagov-deployment.

### DIFF
--- a/modules/govuk/manifests/apps/kibana.pp
+++ b/modules/govuk/manifests/apps/kibana.pp
@@ -1,13 +1,63 @@
 # == Class: govuk::apps::kibana
 #
-# Set up the Kibana application
+# Set up the Kibana application behind Signon
+#
+# === Parameters
+#
+# [*elasticsearch_host*]
+#   Which logs-elasticsearch host Kibana should connect to, must
+#   be a single host: https://github.com/elasticsearch/kibana/issues/214
 #
 # [*port*]
-#  Port that Kibana listens on
+#   Port that the Kibana wrapper listens on
 #
-class govuk::apps::kibana($port = '3202') {
-  govuk::app { 'kibana':
+# [*secret_key*]
+#   A secret key for Rack's session cookies
+#
+# [*signon_client_id*]
+#   A client ID for GOV.UK's Signon
+#
+# [*signon_client_secret*]
+#   The secret which matches signon_client_id
+#
+# [*signon_root*]
+#   The HTTP location of Signon
+#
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/kibana.pp
```diff
@@ -1,13 +1,63 @@
...
+) {
+  $app_name = 'kibana'
+
+  govuk::app { $app_name:
     app_type => 'rack',
     port     => $port,
   }
+
+  Govuk::App::Envvar {
+    app => $app_name,
```
deanwilson: Is there a reason this is a variable when it's not possible to change it when calling the class?

modules/govuk/manifests/apps/kibana.pp
```diff
@@ -1,13 +1,63 @@
...
-  govuk::app { 'kibana':
+# [*signon_client_secret*]
+#   The secret which matches signon_client_id
+#
+# [*signon_root*]
+#   The HTTP location of Signon
+#
+class govuk::apps::kibana (
+  $elasticsearch_host = 'logs-elasticsearch-1.management',
+  $port = '3202',
```
deanwilson: Ports are numbers.

deanwilson:
A couple of small queries but it looks fine to merge.

modules/govuk/manifests/apps/kibana.pp
alexmuller: Mostly just because it's duplicated on line 36 but no other reason!

deanwilson: Fair enough.

deanwilson:
Looks good. Merge away.

modules/govuk/manifests/apps/kibana.pp
```diff
@@ -1,13 +1,63 @@
 # == Class: govuk::apps::kibana
 #
-# Set up the Kibana application
+# Set up the Kibana application behind Signon
+#
+# === Parameters
+#
+# [*elasticsearch_host*]
+#   Which logs-elasticsearch host Kibana should connect to, must
+#   be a single host: https://github.com/elasticsearch/kibana/issues/214
```
rjw1: sigh

rjw1:
:+1:

modules/govuk/manifests/apps/kibana.pp
alexmuller: Good to see that issue hasn't changed in a couple of years.

PR #4397, opened 2016-05-10
Include the standard postgres tuning

MatMoore:
This class sets the effective_cache_size and shared_buffers to values
recommended by the postgres documentation.

Some of our postgres servers include this already but not this one,
which is used by the publishing api database.


surminus:
:+1: merge when green

PR #4398, opened 2016-05-10
Only use one etcd node.

deanwilson:
We've decided that we're going to run locksmithctl, the only thing
that uses our etcd cluster, against a single node in each environment
and trade lower resilience for cost savings.

By convention we'll continue to run etcd on the etcd-1 node in each
environment.


surminus:
:+1: mwg

PR #4399, opened 2016-05-10
Exclude various atom feeds from govuk_crawler

daibach:
These atom feeds should never end up on the mirrors.

Statistics can be filtered using a query string which the crawler currently ignores - this can lead to an atom feed containing unrelated content being served to users when we fail to the mirror. This issue recently led to people subscribed to the Land Registry feed receiving content from Ministry of Defence in their feed readers.

Drug alerts & Travel advice feeds are safety related and so should only serve current content. We can’t guarantee that atom feeds on the mirror are up-to-date, so we should not serve those feeds at all.

Note: it's likely that we should exclude all atom feeds from the mirrors, but these are the most important to fix right now.


boffbowsh:
👍  Merge when green

PR #4400, opened 2016-05-11
These classes are not used on the same nodes so use normal packages

deanwilson:
We seem to have hit an issue with stdlib so we're reverting this change
out. 4.12.0 didn't work with it on Puppet 3 but it does on puppet 4.3


surminus:
:+1: mergey mergey

PR #4402, opened 2016-05-11
adds Elasticsearch index snapshots to S3

jonauman:
This commit adds the ES plugin to AWS which allows us to snapshot
individual indices to an S3 bucket. S3 credentials can be passed
a variety of ways, but adding them to elasticsearch.yml provides
the simplest solution. Passing the creds as environment variables
requires changes to ES user and the ES init.d script and overrides
to the vendor module, and only works if these requirements are met.

### DIFF
--- a/hieradata/class/vagrant/logs_elasticsearch.yaml
+++ b/hieradata/class/vagrant/logs_elasticsearch.yaml
@@ -1,0 +1,9 @@
+---
+# should set to real values
+govuk_elasticsearch::aws_access_key: 'ABCDEFGHIJKLMNOPQRST'
+govuk_elasticsearch::aws_secret_key: 'n4AQffsnBJ5rcQK9DBn5eYHVuDqhT5DVCYa7ZU9tt'
+
+govuk_elasticsearch::snapshot_backups:
+   'kibana-int':
+     s3_bucket: 'govuk-elasticsearch-vagrant'
+     s3_path: 'test'
--- a/modules/govuk/manifests/node/s_logs_elasticsearch.pp
+++ b/modules/govuk/manifests/node/s_logs_elasticsearch.pp
@@ -35,6 +35,24 @@ class govuk::node::s_logs_elasticsearch(
     ],
   }
 
+  case $govuk_elasticsearch::version {
+    /^0.9/: { $cloud_aws_version = '1.16.0' }
+    /^1.0/: { $cloud_aws_version = '2.0.0' }
+    /^1.1/: { $cloud_aws_version = '2.1.1' }
+    /^1.2/: { $cloud_aws_version = '2.2.0' }
+    /^1.3/: { $cloud_aws_version = '2.3.0' }
+    /^1.4/: { $cloud_aws_version = '2.4.2' }
+    /^1.5/: { $cloud_aws_version = '2.5.1' }
+    /^1.6/: { $cloud_aws_version = '2.6.1' }
+    /^1.7/: { $cloud_aws_version = '2.7.1' }
+    default: { fail('Plugin version must match elasticsearch version, see README at https://github.com/elastic/elasticsearch-cloud-aws') }
+  }
+
+  elasticsearch::plugin { "elasticsearch/elasticsearch-cloud-aws/${cloud_aws_version}":
+    module_dir => 'cloud-aws',
+    instances  => $::fqdn,
+  }
+
   elasticsearch::plugin { 'mobz/elasticsearch-head':
     module_dir => 'head',
     instances  => $::fqdn,
--- a/modules/govuk_elasticsearch/manifests/init.pp
+++ b/modules/govuk_elasticsearch/manifests/init.pp
@@ -6,10 +6,13 @@
 # === Parameters
 #
 # Lots missing!
+# [*aws_access_key*]
+#   The AWS access key IDfor the IAM user with access to the S3 bucket where
+#   index snapshots will go
 #
-# [*version*]
-#   The version of elasticsearch to install.  This must specify an exact
-#   version (eg 1.4.2)
+# [*aws_secret_key*]
+#   The AWS secret access key for the IAM user with access to the S3 bucket where
+#   index snapshots will go
 #
 # [*manage_repo*]
 #   Whether to configure an apt source for our mirror of the elasticsearch
@@ -22,6 +25,10 @@
 #   specific machines.
 #   Default: true
 #
+# [*version*]
+#   The version of elasticsearch to install.  This must specify an exact
+#   version (eg 1.4.2)
+#
 class govuk_elasticsearch (
   $version,
   $cluster_hosts = ['localhost'],
@@ -29,13 +36,16 @@ class govuk_elasticsearch (
   $heap_size = '512m',
   $number_of_shards = '3',
   $number_of_replicas = '1',
-  $minimum_master_nodes = '1',
+  $minimum_master_nodes = '2',
   $refresh_interval = '1s',
   $host = 'localhost',
   $log_index_type_count = {},
   $disable_gc_alerts = false,
   $manage_repo = true,
   $open_firewall_from_all = true,
+  $snapshot_backups = {},
+  $aws_access_key = '',
+  $aws_secret_key = '',
 ) {
 
   validate_re($version, '^\d+\.\d+\.\d+$', 'govuk_elasticsearch::version must be in the form x.y.z')
@@ -82,11 +92,17 @@ class govuk_elasticsearch (
     'http.port'                => $http_port,
     'discovery'                => {
       'zen' => {
-        'minimum_master_nodes' => $minimum_master_nodes,
+        'minimum_master_nodes' => 2,
         'ping'                 => {
           'multicast.enabled' => false,
           'unicast.hosts'     => $cluster_hosts,
         },
+      },
+    },
+    'cloud'                    => {
+      'aws' => {
+        'access_key' => $aws_access_key,
+        'secret_key' => $aws_secret_key,
       },
     },
   }
@@ -111,6 +127,12 @@ class govuk_elasticsearch (
     init_defaults => {
       'ES_HEAP_SIZE' => $heap_size,
     },
+  }
+
+  exec { 'secure_es_yaml':
+    command     => "chmod 0640 /etc/elasticsearch/${::fqdn}/elasticsearch.yml",
+    subscribe   => Elasticsearch::Instance[$::fqdn],
+    refreshonly => true,
   }
 
   Class['elasticsearch'] -> Elasticsearch::Instance[$::fqdn] -> Anchor['govuk_elasticsearch::end']
@@ -142,4 +164,7 @@ class govuk_elasticsearch (
   include govuk_elasticsearch::estools
 
   anchor { 'govuk_elasticsearch::end': }
+
+  create_resources(govuk_elasticsearch::snapshot, $govuk_elasticsearch::snapshot_backups)
+
 }
--- a/modules/govuk_elasticsearch/manifests/snapshot.pp
+++ b/modules/govuk_elasticsearch/manifests/snapshot.pp
@@ -1,0 +1,61 @@
+# define govuk_elasticsearch::snapshot
+#
+# [*s3_bucket]
+#   The S3 bucket where the snapshots will be stored
+#
+# [*s3_region]
+#   The AWS region for the S3 bucket. Defaults to eu-west-1.
+#
+# [*s3_path]
+#   The folder path within the S3 bucket where the snapshots
+#   will be stored
+#
+# [*index]
+#   The ES index to backup
+#
+# [*http_port]
+#   The port that ES runs on
+#
+# [*title*]
+#   The $title or $name of the hiera hash should be the index name
+#
+# Hiera example:
+# --------------
+# govuk_elasticsearch::snapshot::indices:
+#   kibana-int:
+#     s3_bucket: 'jonauman-test2'
+#     s3_path: 'elasticsearch'
+
+define govuk_elasticsearch::snapshot (
+  $s3_bucket = undef,
+  $s3_region = 'eu-west',
+  $s3_path = undef,
+  $index = $title,
+  $http_port = 9200,
+  ){
+
+  File {
+    mode => '0755',
+  }
+
+  file { "/usr/local/bin/elasticsearch-create-repo-${index}":
+    ensure  => present,
+    content => template('govuk_elasticsearch/create_repo.sh.erb'),
+  }
+
+  file { "/usr/local/bin/elasticsearch-backup-${index}":
+    ensure  => present,
+    content => template('govuk_elasticsearch/backup_index.sh.erb'),
+  }
+
+  exec { "/usr/local/bin/elasticsearch-create-repo-${index}":
+    require => File["/usr/local/bin/elasticsearch-create-repo-${index}"],
+  }
+
+  cron { "elasticsearch-backup-${index}":
+    command => "/usr/local/bin/elasticsearch-backup-${index}",
+    special => daily,
+    require => File["/usr/local/bin/elasticsearch-backup-${index}"],
+  }
+
+}
--- a/modules/govuk_elasticsearch/templates/backup_index.sh.erb
+++ b/modules/govuk_elasticsearch/templates/backup_index.sh.erb
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -e
+
+repo=<%= @s3_bucket %>
+timestamp=$(date +%Y-%m-%dt%H:%M:%S)
+snapshot=${timestamp}-<%= @title %>
+
+curl --fail -XPUT "http://localhost:<%= @http_port %>/_snapshot/${repo}/${snapshot}?wait_for_completion=true" -d '{
+  "indices": "<%= @title %>",
+  "ignore_unavailable": "true",
+  "include_global_state": false
+}'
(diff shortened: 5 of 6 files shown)


modules/govuk_elasticsearch/manifests/init.pp
```diff
@@ -89,6 +98,12 @@
         },
       },
     },
+    'cloud'                    => {
+      'aws' => {
+        'access_key' => $aws_access_key,
+        'secret_key' => $aws_secret_key,
+      },
+    },
```
surminus: I think this is fine to pass into config, assuming that only the elasticsearch user has access to it.

modules/govuk_elasticsearch/manifests/init.pp
```diff
@@ -106,13 +121,18 @@
...
     config        => $instance_config_real,
     datadir       => '/mnt/elasticsearch',
     init_defaults => {
       'ES_HEAP_SIZE' => $heap_size,
     },
   }
 
+  exec { 'secure_es_yaml':
+    command => "chmod 0640 /etc/elasticsearch/${::fqdn}/elasticsearch.yml",
+  }
```
surminus: Is this the only way to can ensure permissions on the config file? Seems untidy to me, is there a way we can set a default file attribute for that user?

modules/govuk_elasticsearch/manifests/snapshot.pp
```diff
@@ -0,0 +1,71 @@
...
+#
+# Hiera example:
+# --------------
+# govuk_elasticsearch::snapshot::indices:
+#   kibana-int:
+#     s3_bucket: 'jonauman-test2'
+#     s3_path: 'elasticsearch'
+
+define govuk_elasticsearch::snapshot
+  ($s3_bucket = undef,
```
surminus: Nitpick but can we move the `(` to the line above?

modules/govuk_elasticsearch/manifests/snapshot.pp
```diff
@@ -0,0 +1,71 @@
...
+  $s3_region = 'eu-west',
+  $s3_path = '',
+  $index = $title,
+  $http_port = '9200',
```
surminus: Should probably be a integer rather than a string.

modules/govuk_elasticsearch/manifests/snapshot.pp
```diff
@@ -0,0 +1,71 @@
...
+# --------------
+# govuk_elasticsearch::snapshot::indices:
+#   kibana-int:
+#     s3_bucket: 'jonauman-test2'
+#     s3_path: 'elasticsearch'
+
+define govuk_elasticsearch::snapshot
+  ($s3_bucket = undef,
+  $s3_region = 'eu-west',
+  $s3_path = '',
```
surminus: I _think_ an empty string default would cause the `<% if @s3_path %>` statement below to trigger anyway, while an `undef` would not (although I'm not 100%).

modules/govuk_elasticsearch/manifests/init.pp
jonauman: I would rather not use exec either. The problem is that this resource is controlled in the vendor elasticsearch module, which I'd rather not modify without pushing the change upstream. We could submit a PR to the vendor and update this code if they accept?

modules/govuk_elasticsearch/manifests/snapshot.pp
```diff
@@ -0,0 +1,71 @@
...
+  $index = $title,
+  $http_port = '9200',
+  $hour = 3,
+  $minute = 20,
+  ){
+
+  File {
+    ensure => present,
+    owner  => root,
+    group  => root,
```
surminus: I think the owner by default is root.

modules/govuk_elasticsearch/manifests/snapshot.pp
```diff
@@ -0,0 +1,66 @@
...
+    ensure  => present,
+    content => template('govuk_elasticsearch/create_repo.sh.erb'),
+  }
+
+  file { "/usr/local/bin/elasticsearch-backup-${index}":
+    ensure  => present,
+    content => template('govuk_elasticsearch/backup_index.sh.erb'),
+  }
+
+  exec { "/usr/local/bin/elasticsearch-create-repo-${index}":
```
surminus: How come we are relying on puppet to run this rather than wrapping it in a script?

modules/govuk/manifests/node/s_elasticsearch.pp
```diff
@@ -28,6 +31,24 @@
...
+    /^1.0/: { $cloud_aws_version = '2.0.0' }
+    /^1.1/: { $cloud_aws_version = '2.1.1' }
+    /^1.2/: { $cloud_aws_version = '2.2.0' }
+    /^1.3/: { $cloud_aws_version = '2.3.0' }
+    /^1.4/: { $cloud_aws_version = '2.4.2' }
+    /^1.5/: { $cloud_aws_version = '2.5.1' }
+    /^1.6/: { $cloud_aws_version = '2.6.1' }
+    /^1.7/: { $cloud_aws_version = '2.7.1' }
+    default: { $cloud_aws_version = '2.7.1' }
+  }
```
surminus: How come we loop through all these versions?

jonauman: It's in the readme for the AWS plugin: https://github.com/elastic/elasticsearch-cloud-aws
You have to match the plugin version with the version of ES, and since we are currently running multiple versions of ES...

modules/govuk_elasticsearch/manifests/snapshot.pp
jonauman: This script has to be run at least once to setup the S3 repository, though it doesn't hurt to run it with each puppet run. So cron didn't seem to be like a good way to get it executed.

I suppose I could add it to the backup script. Thoughts?

surminus: Will it overwrite anything?

modules/govuk_elasticsearch/manifests/init.pp
alexmuller: The permissions on the file this ends up in are 644, so I think anyone on the box will be able to see it. Not sure how big of a deal that is though.

alexmuller: Ah, I've just seen the comments further down.

hieradata/class/vagrant/elasticsearch.yaml
```diff
@@ -0,0 +1,26 @@
...
+  data:
+    pv: '/dev/sdb1'
+    vg: 'elasticsearch'
+
+mount:
+  /mnt/elasticsearch:
+    disk: '/dev/mapper/elasticsearch-data'
+    govuk_lvm: 'data'
+    mountoptions: 'defaults'
+    #FIXME: Adjust these values when elasticsearch is upgraded to 1.4
```
deanwilson: Considering we specify the version as 1.7.5 on line 3 can we either adjust the values or remove the fixme?

hieradata/class/vagrant/elasticsearch.yaml
```diff
@@ -0,0 +1,26 @@
...
+    percent_threshold_warning: 31
+    percent_threshold_critical: 16
+
+govuk_elasticsearch::snapshot::indices:
+   '.kibana':
+     s3_bucket: 'jonauman-test2'
```
deanwilson: Should this be renamed something more generic?

modules/govuk_elasticsearch/manifests/snapshot.pp
jonauman: No. It works like this:
If the bucket does not exist, it will create the bucket and tell ES to use it as a destination repo.
If the bucket does exist, it doesn't change anything in S3, and simply tells US to use it as a destination repo. 

So running this script once or 10 times has the same effect.

hieradata/class/vagrant/elasticsearch.yaml
jonauman: Yeah, I'll remove the fixme as I copied the config from the main class hieradata

modules/govuk_elasticsearch/manifests/snapshot.pp
surminus: Cool OK

hieradata/class/vagrant/elasticsearch.yaml
jonauman: Oops. Yes, that shouldn't be there. Also note that kibana 3 uses 'kibana-int' for the index whereas kibana4 uses '.kibana'. Do we want to assume 3 or 4 for vagrant purposes?

surminus:
As a general comment I think this is generic to be able to use to back up other Elasticsearch indexes. We actually sync from Production to Staging every night so it'll be good to be able to eventually use this as yet another replacement for the evil env-sync-and-backup.

modules/govuk/manifests/node/s_elasticsearch.pp
```diff
@@ -14,11 +17,11 @@
   }
 
   class { 'govuk_elasticsearch':
-    cluster_hosts        => ['elasticsearch-1.backend:9300', 'elasticsearch-2.backend:9300', 'elasticsearch-3.backend:9300'],
+    cluster_hosts        => $cluster_hosts,
     cluster_name         => 'govuk-production',
     heap_size            => "${es_heap_size}m",
     number_of_replicas   => '1',
-    minimum_master_nodes => '2',
+    minimum_master_nodes => $minimum_master_nodes,
```
deanwilson: Does this mean we'll change the minimum number of nodes in production from 2 to 1?

alexmuller:
Let's go through this on Monday - I think it's making some changes that it doesn't need to (for example upgrading Elasticsearch from 0.90.12 to 1.7.5 on the main Elasticsearch cluster in vagrant and changing the minimum number of master nodes which isn't something we should do).

modules/govuk_elasticsearch/manifests/init.pp
```diff
@@ -22,20 +25,26 @@
...
+#
 class govuk_elasticsearch (
   $version,
   $cluster_hosts = ['localhost'],
   $cluster_name = 'elasticsearch',
   $heap_size = '512m',
   $number_of_shards = '3',
   $number_of_replicas = '1',
-  $minimum_master_nodes = '1',
+  $minimum_master_nodes = '2',
```
alexmuller: I think changing this to 2 is correct but could we do it in a separate commit? It's unrelated to the other changes going on here.

We could actually remove this parameter entirely because we only ever set it to the same value in each of the 3 places its called from.

modules/govuk/manifests/node/s_logs_elasticsearch.pp
```diff
@@ -35,6 +35,24 @@
...
+    /^1.0/: { $cloud_aws_version = '2.0.0' }
+    /^1.1/: { $cloud_aws_version = '2.1.1' }
+    /^1.2/: { $cloud_aws_version = '2.2.0' }
+    /^1.3/: { $cloud_aws_version = '2.3.0' }
+    /^1.4/: { $cloud_aws_version = '2.4.2' }
+    /^1.5/: { $cloud_aws_version = '2.5.1' }
+    /^1.6/: { $cloud_aws_version = '2.6.1' }
+    /^1.7/: { $cloud_aws_version = '2.7.1' }
+    default: { $cloud_aws_version = '2.7.1' }
+  }
```
alexmuller: There's no need to set this for values below 1.4 because we'll never run those versions of Elasticsearch on our logs cluster. We could also set the default case to call `fail()` so that we don't install a potentially breaking version of the plugin if we upgrade our cluster.

modules/govuk_elasticsearch/manifests/init.pp
```diff
@@ -142,4 +162,8 @@
   include govuk_elasticsearch::estools
 
   anchor { 'govuk_elasticsearch::end': }
+
+  $snapshots = hiera_hash('govuk_elasticsearch::snapshot::indices', {})
+  create_resources(govuk_elasticsearch::snapshot, $snapshots)
```
alexmuller: Could we set this as a parameter of this class, like `$snapshot_indices`? Then we can avoid the call to `hiera_hash()` which helps to separate our data from our code and we're not using the name of a non-existent parameter on the `govuk_elasticsearch::snapshot` defined type.

modules/govuk_elasticsearch/manifests/snapshot.pp
```diff
@@ -0,0 +1,66 @@
...
+
+  file { "/usr/local/bin/elasticsearch-create-repo-${index}":
+    ensure  => present,
+    content => template('govuk_elasticsearch/create_repo.sh.erb'),
+  }
+
+  file { "/usr/local/bin/elasticsearch-backup-${index}":
+    ensure  => present,
+    content => template('govuk_elasticsearch/backup_index.sh.erb'),
+  }
```
alexmuller: I think this file resource and the one above it both need the executable bit set otherwise the Puppet run fails:

```
Error: '/usr/local/bin/elasticsearch-create-repo-kibana-int' is not executable
```

modules/govuk/manifests/node/s_logs_elasticsearch.pp
```diff
@@ -35,6 +35,24 @@
...
+  
+elasticsearch::plugin { "elasticsearch/elasticsearch-cloud-aws/${cloud_aws_version}":
```
alexmuller: Nitpick: there's a bit of trailing whitespace on line 50 which should probably be the indentation for line 51.

hieradata/class/vagrant/logs_elasticsearch.yaml
```diff
@@ -0,0 +1,20 @@
...
+lv:
+  data:
+    pv: '/dev/sdb1'
+    vg: 'elasticsearch'
+
+mount:
+  /mnt/elasticsearch:
+    disk: '/dev/mapper/elasticsearch-data'
+    govuk_lvm: 'data'
+    mountoptions: 'defaults'
```
alexmuller: I don't think we need to set either the `lv` or the `mount` properties in this file - they'll fall back to the production config which should work.

modules/govuk_elasticsearch/manifests/snapshot.pp
```diff
@@ -0,0 +1,66 @@
...
+  exec { "/usr/local/bin/elasticsearch-create-repo-${index}":
+    require => File["/usr/local/bin/elasticsearch-create-repo-${index}"],
+  }
+
+  cron { "elasticsearch-backup-${index}":
+    command => "/usr/local/bin/elasticsearch-backup-${index}",
+    hour    => $hour,
+    minute  => $minute,
+    require => File["/usr/local/bin/elasticsearch-backup-${index}"],
+  }
```
alexmuller: Could this run as part of `cron.daily`? Then we don't have problems for example with integration when it's switched off overnight.

modules/govuk_elasticsearch/manifests/init.pp
alexmuller: This is very noisy in the Puppet run output. On every run the elasticsearch module changes the mode from 640 to 644 and then this exec changes it back again.

modules/govuk/manifests/node/s_logs_elasticsearch.pp
jonauman: I don't think we should limit this to our current use case. The defined type is generic enough that you can backup any index on any cluster. If you upgrade the cluster, you must upgrade the plugin, as stated in the plugin documentation.

modules/govuk_elasticsearch/manifests/init.pp
alexmuller: I'm also not sure it works - the state of the system when Puppet has finished running is this:

```
-rw-r--r-- 1 elasticsearch elasticsearch 845 May 16 12:13 /etc/elasticsearch/logs-elasticsearch-1.management.dev.gov.uk/elasticsearch.yml
```

modules/govuk_elasticsearch/manifests/init.pp
jonauman: I don't see the harm of adding the parameter if it has use in an environment such as vagrant. I'm not a big fan of hardcoding values when most of the other values have already been set as parameters. That is, as long as the change does not affect the current default behviour.

alexmuller:
The 2 shell scripts that make HTTP requests against Elasticsearch should probably have `--fail` added to the curl command so that they exit non-zero if the request fails. We could also add `set -e` to the top of the script.

modules/govuk/manifests/node/s_logs_elasticsearch.pp
alexmuller: I think that's true, but until we reuse this code it's inside `govuk::node::s_logs_elasticsearch` and references versions of Elasticsearch that will never be running on those machines which might be confusing.

I understand that the documentation says that the plugin must be upgraded whenever Elasticsearch is, but the reality is that people forget things like that so we should write code which guards against those cases and fails early in the Puppet run to prevent issues.

jonauman: I think the original point about setting a default fail is valid.

modules/govuk_elasticsearch/manifests/snapshot.pp
jonauman: The file mode was inadvertently removed in a commit

jonauman:
@alexmuller Have added --fail and set -e to scripts. Note that the --fail will suppress the output of the curl  request, which will make troubleshooting harder.

PR #4403, opened 2016-05-12
Add Local Links Manager service interactions importer Jenkins job

brenetic:
- Local Links Manager requires all the services and interactions for each local authority to be imported at regular intervals from the [Local Government Association](standards.esd.org.uk) site. The import script is already part of the [local-links-manager repo](https://github.com/alphagov/local-links-manager). This commit adds a Jenkins job to run the rake task on the first day of each month.

[Trello card](https://trello.com/c/r8yv6gjk/368-import-all-services-and-interactions-into-local-links-manager-3)

### DIFF
--- a/modules/govuk_jenkins/manifests/job/local_links_manager_import_service_interactions.pp
+++ b/modules/govuk_jenkins/manifests/job/local_links_manager_import_service_interactions.pp
@@ -1,0 +1,25 @@
+# == Class: govuk_jenkins::job::local_links_manager_import_service_interactions
+#
+# Import services and interactions and the mapped service interactions into local-links-manager
+#
+class govuk_jenkins::job::local_links_manager_import_service_interactions (
+  $app_domain = hiera('app_domain'),
+) {
+
+  $check_name = 'local-links-manager-import-service-interactions'
+  $service_description = 'Import services and interactions into local-links-manager'
+  $job_url = "https://deploy.${app_domain}/job/local-links-manager-import-service-interactions/"
+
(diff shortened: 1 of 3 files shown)


surminus:
You'll need to add it to hieradata to enable it (https://github.com/alphagov/govuk-puppet/blob/master/hieradata/production.yaml#L68). 

Do we need to be alerted if the job fails? We can set success/failure downstream jobs to let us know if that's the case.

brenetic:
Thanks @surminus 

Have added it to production.yaml and also added the success/failure icinga notification.

modules/govuk_jenkins/manifests/job/local_links_manager_import_service_interactions.pp
```diff
@@ -0,0 +1,25 @@
...
+  file { '/etc/jenkins_jobs/jobs/local_links_manager_import_service_interactions.yaml':
+    ensure  => present,
+    content => template('govuk_jenkins/jobs/local_links_manager_import_service_interactions.yaml.erb'),
+    notify  => Exec['jenkins_jobs_update'],
+  }
+
+  @@icinga::passive_check { "${check_name}_${::hostname}":
+    service_description => $service_description,
+    host_name           => $::fqdn,
+    freshness_threshold => 3600,
```
surminus: If this only runs but once a month this will need to be higher (whatever a month is in seconds).

brenetic:
@surminus Have set it for 32 days in seconds

surminus:
:+1: Great, looks good, feel free to merge

PR #4406, opened 2016-05-13
Further bugfix for Whitehall attachments

surminus:
We should exclude "lost+found" because it throws an error when it does an rsync due to permissions, and it shouldn't ever be synced.


deanwilson:
Looks fine. Merge away.

PR #4407, opened 2016-05-13
Add logic for attachment syncs

surminus:
If an environment is syncing attachments from another environment, it shouldn't also backup (considering the size and number of attachments we're dealing with). This adds the logic to only enable backups on
environments that are not syncing, and simply adds the tool to allow environment syncs. As in line with other work of this nature, just add the tool to allow an environment sync rather than any other scheduling.


modules/govuk/manifests/node/s_asset_master.pp
```diff
@@ -55,12 +71,21 @@
     file {
     [ '/etc/govuk/aws', '/etc/govuk/aws/env.d']:
       ensure => directory;
+      owner   => 'assets',
+      group   => 'assets',
+      mode    => '750',
```
deanwilson: All 3 of these modes should be '0foo'. Puppet-lint should pick those up.

modules/govuk/templates/node/s_asset_base/attachments-s3-env-sync.sh
```diff
@@ -0,0 +1,21 @@
...
+
+usage() {
+  echo "USAGE: $0 <directory_to_sync>"
+  echo
+  echo "syncs attachments in S3 to local filesystem"
+  echo
+  exit 1
+}
+
+if [ ! $DIRECTORY_TO_SYNC ]; then
```
deanwilson: Probably double quote this variable usage.

modules/govuk/manifests/node/s_asset_master.pp
```diff
@@ -11,12 +11,28 @@
 # [*asset_slave_ip_ranges*]
 #   A hash of IP addresses that the asset slave machines run on.
 #
+# [*s3_bucket*]
```
rjw1: :+1: adding pupperdocs

rjw1:
im happy with this. :+1: merge conflicts though.

surminus:
Thanks @rjw1, realised I'd left some important bits out like adding the script to the server and disabling full syncs in Integration and Staging (which is just wasting money).

deanwilson:
This looks ok. Merge when ready.

alexmuller:
I might be missing something here, but you say:

> If an environment is syncing attachments from another environment, it shouldn't also backup (considering the size and number of attachments we're dealing with).

But I can't see any code that does that. Is that what you meant by:

> left some important bits out like ... disabling full syncs in Integration and Staging

surminus:
@alexmuller that is indeed what I meant. 

If [this](https://github.com/alphagov/govuk-puppet/blob/add-logic-for-attachment-syncs/modules/govuk/manifests/node/s_asset_master.pp#L35) is set to true, then the syncing script [gets created here](https://github.com/alphagov/govuk-puppet/blob/add-logic-for-attachment-syncs/modules/govuk/manifests/node/s_asset_master.pp#L91-L95), and actually there should be a `&& !@s3_env_sync_enabled` added to the conditionals [here](https://github.com/alphagov/govuk-puppet/blob/add-logic-for-attachment-syncs/modules/govuk/templates/node/s_asset_base/process-uploaded-attachments.sh.erb#L44) and [here](https://github.com/alphagov/govuk-puppet/blob/add-logic-for-attachment-syncs/modules/govuk/templates/node/s_asset_base/copy-attachments.sh.erb#L43), great spot, thanks!

surminus:
@alexmuller updated logic, please let me know if it now looks OK! Secrets, buckets and things still need to be added and deployed.

alexmuller:
This looks good to me :+1:

There's a small commit that's snuck into this branch that could be removed (1c911fb311101bc64be4fb73b1c36515f31062ea).

And then I think all this needs is the hieradata with the AWS access creds and it's good to go.

surminus:
Thanks! Removed the commit and rebased. AWS credentials should now be in so will talk about merging and deploying tomorrow - then hopefully the full Production sync can happen over the weekend.

PR #4408, opened 2016-05-13
Add cache-file for s3 syncs

surminus:
The cache-file flag compiles a list of source MD5s for the files that have been synced to the destination. This means that any files which haven't changed are skipped in the run, but if the file has changed
then they'll be included, which should hopefully make the syncs a lot quicker.


deanwilson:
Looks fine. Merge away.

PR #4411, opened 2016-05-13
Remove pruning Graphite whisper files

alexmuller:


deanwilson:
Looks fine. Merge when green.

PR #4412, opened 2016-05-13
Send the basic icinga server logs to the ElasticSearch cluster

deanwilson:
As requested in https://trello.com/c/845Fz7i8


surminus:
What's the use case? `/var/log/icinga/icinga.log` normally just shows shutdown/startups so I assume it's so we can see how often it restarts? 

Otherwise I think it looks good :+1: merge when ready

deanwilson:
Agreed. I think this is just to show exceptional events that happen to the server as check results are in logs we already ship.

PR #4413, opened 2016-05-16
Specialist publisher Rebuild: tweak flag default

benilovj:
Change the 'publish_pre_production_finders' to default to 'false',
rather than 'undef', so it's exactly the same as in Specialist Publisher v1.

/cc @deanwilson


deanwilson:
Looks fine. Merge when green.

PR #4416, opened 2016-05-16
Add filehandles to collectd

deanwilson:


modules/collectd/templates/etc/collectd/conf.d/file_handles.conf.erb
```diff
@@ -0,0 +1,3 @@
+<Plugin exec>
```
alexmuller: This needs a `LoadPlugin exec` at the top.

alexmuller:
This looks correct to me :+1:

PR #4417, opened 2016-05-17
Add a job to run DFID research outputs import

benilovj:
* Scheduled on the 49th minute of every hour (selected by fair random
  dice roll)
* Uses a Publishing API bearer token
* Calls `rake load:outputs` and stuffs some DFID research outputs into
  publishing_api and rummager

Paired with @rgarner on this.
/cc @alexmuller @tuzz


rgarner:
@benilovj, Do we have a paired PR in gds/deployment? Asking for @danielroseman

PR #4418, opened 2016-05-17
Set `$GOPATH` in the dev VM environment

gpeng:
This is part of the work underway to upgrade metadata-api to the latest go version.

Currently the dev VM has no `$GOPATH` which requires the apps try and build one within their own directory. This seems increasingly fragile with the most recent >1.6 versions of go.

This commit sets the `$GOPATH` to `/var/govuk/gopath` which will require our go apps to be located at
`/var/govuk/gopath/src/github.com/alphagov/<reponame>` at which point all will be well in the world.

I've opened another [PR](https://github.gds/gds/development/pull/319) to update the `Procfile` (currently just for metadata-api)

[Trello](https://trello.com/c/TGGWiuRB/399-upgrade-metadata-api-to-newest-go-version-large)

### DIFF
--- a/modules/golang/files/etc/profile.d/gopath.sh
+++ b/modules/golang/files/etc/profile.d/gopath.sh
@@ -1,0 +1,1 @@
+export GOPATH=/var/govuk/gopath
--- a/modules/golang/manifests/init.pp
+++ b/modules/golang/manifests/init.pp
@@ -19,4 +19,11 @@ class golang {
 
   # Ensure that scm tools used by `go get` are present.
   ensure_packages(['bzr', 'git', 'mercurial'])
+
+  if $::domain == 'development'{
+    file { '/etc/profile.d/gopath.sh':
+      ensure => present,
+      source => 'puppet:///modules/golang/etc/profile.d/gopath.sh',
+    }
+  }
 }
--- a/modules/golang/spec/classes/golang_spec.rb
+++ b/modules/golang/spec/classes/golang_spec.rb
@@ -1,0 +1,19 @@
+require_relative '../../../../spec_helper'
+
+describe 'golang', :type => :class do
+  let(:file_path) { '/etc/profile.d/gopath.sh' }
+  let(:go_path){ 'export GOPATH=/var/govuk/gopath' }
+
+  context 'in development' do
+    let(:facts) {{domain: 'development'}}
+    it {
+      is_expected.to contain_file(file_path).
+        with_source('puppet:///modules/golang/etc/profile.d/gopath.sh')
+    }
+  end
+
+  context 'not in development' do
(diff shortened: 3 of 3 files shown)


boffbowsh:
Could we have this in `~vagrant/src/github.com/alphagov/metadata-api`, which is a symlink to `/var/govuk/metadata-api`? Having a `src` directory in everyone's projects root feels a bit invasive.

gpeng:
There was a typo in the description. It's actually `gopath` which will be in the projects directory. I tried a few symlinking approaches (admittedly not quite the exact one you're suggesting) and they didn't work so well as just having the repo in a proper directory structure. I thought an explicit `gopath` directory seemed to be a reasonable approach. There is quite a lot of 'stuff' around complaining about and warning against using symlinks in the `$GOPATH`.

That said, it may have been other things that I've got rid of (like `gom`) that were causing the problems so I'll give it a try and see.

gpeng:
I've tried it and you get weird behaviour. If you `cd` in through the `$GOPATH` and run `make` it works but if you `cd` into `/var/govuk/metadata-api` it doesn't work. On top of that, even if it did work it would require people to know to find the `$GOPATH` and symlink the app into it. I think a convention that go apps live inside `gopath` is easier to communicate and use (and it works)

modules/govuk_envsys/templates/environment.erb
```diff
@@ -1,3 +1,9 @@
 PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games"
 LC_ALL=en_GB.UTF-8
 TZ=UTC
+
+<%- if @development -%>
+
+  GOPATH=/var/govuk/gopath/
```
boffbowsh: I'd leave this line unindented so the resulting file doesn't look weird.

modules/govuk_envsys/manifests/init.pp
```diff
@@ -4,8 +4,10 @@
 # LC_ALL.
 #
 class govuk_envsys {
+  $development = $::domain == 'development'
+
   file { '/etc/environment':
-    ensure => present,
-    source => 'puppet:///modules/govuk_envsys/etc/environment',
+    ensure  => present,
+    content => template('govuk_envsys/environment.erb'),
```
boffbowsh: I wonder if this GOPATH change would be better in a file in `/etc/profile.d` which is only created for dev vm?

boffbowsh:
Ok, having `gopath` in the projects directory sounds fine to me.

modules/govuk_envsys/manifests/init.pp
gpeng: Fixed in 19e99c7

PR #4423, opened 2016-05-19
[do not merge] Rename secret_key_base to secret_token to match v1 naming

brenetic:
This change makes it so that v1 and v2 use the same naming
for the session token, which is more consistent.

Worked on by @Rosa-Fox, @brenetic, @tuzz

Related: https://github.gds/gds/deployment/pull/995

### DIFF
--- a/modules/govuk/manifests/apps/specialist_publisher_rebuild.pp
+++ b/modules/govuk/manifests/apps/specialist_publisher_rebuild.pp
@@ -47,10 +47,11 @@
 #   The bearer token to use when communicating with Publishing API.
 #   Default: undef
 #
-# [*secret_key_base*]
-#   Used to set the app ENV var SECRET_KEY_BASE which is used to configure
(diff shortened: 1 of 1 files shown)


h-lame:
It would be good to update the commit message and comment in the manifest. `secret_key_base` is the value expected to be used for rails 4.x apps and `secret_token` is the value used for rails 3.x apps.  The reason we're using `secret_token` is to allow this rails 4.x app to share signed cookies with the rails 3.x app it is replacing.  So it's not just about naming consistency.

h-lame:
Oh, actually, above I mean 4.1+ where I've said 4.x and 3.x - 4.0.x where I've said 3.x above.  Rails versions are hard.

brenetic:
We've updated our commit message and added a comment line. Please could you take a look @h-lame?

h-lame:
👍 future code spelunkers will thank you!

PR #4426, opened 2016-05-20
Add hieradata for promoting staging to production

alexmuller:
- Change app domain (this will change machine hostnames)
- Stop machines rebooting
- Enable email delivery from us-east-1
- Set up monitoring

### DIFF
--- a/hiera.yml
+++ b/hiera.yml
@@ -1,5 +1,6 @@
 ---
 :hierarchy:
+  # - '%{::environment}_disaster_recovery' # For disaster recovery (DR), when staging needs to become production
   - 'node/%{::fqdn}'
   - 'class/%{::environment}/%{::govuk_node_class}'
   - 'class/%{::govuk_node_class}'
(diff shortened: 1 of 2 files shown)


jennyd:
It's probably a good idea to also disable gor - I don't know whether those requests would be disappearing into the ether or whether we'd be replaying against the same machines, but either isn't much use :)

surminus:
This is what is needed to add to staging.yaml for it to become production-like in the event that the Production datacentre has gone away, and I think gor is specifically enabled in Production? 

Either way here's hoping we'll never ever have to think about using this.

rjw1:
:+1:

alexmuller:
@jennyd got me thinking about something interesting.

There are probably additional changes that we'll need to make to the credentials files. Specifically, the environment variables that we set for Jenkins include `ORGANISATION` set to `staging` and `production`.

If we change that to `production` when staging is promoted, we'll start using the production hieradata on all machines. And if we don't do that we'll get the staging config for all of the apps.

Not sure what the correct thing to do is.

PR #4427, opened 2016-05-20
Add deployment job for Lambda apps

surminus:
The idea behind it is to grab the source from a new repository specifically for holding lambda
app info, push the content to an S3 bucket specifically created for Lambda apps, and trigger a build of a Terraform job which is designed to deploy the Lambda function in Amazon.

I think there's still a fair bit of work to do, including adding in pulling the object version of the uploaded object into the S3 bucket ([related](https://github.com/alphagov/govuk-terraform-provisioning/pull/46/commits/2d35845d8433661e03732c7cc21edb32cde4f1fc)). 

Another question is if we want to allow automatic deploys without passing in AWS credentials.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_lambda_app.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_lambda_app.yaml.erb
@@ -1,0 +1,54 @@
+---
+- scm:
+    name: govuk-lambda-app-deployment_Deploy_Lambda_App
+    scm:
+        - git:
+            url: git@github.com:alphagov/govuk-lambda-app-deployment.git
+            branches:
+              - master
+            wipe-workspace: true
+
+- job:
+    name: Deploy_Lambda_App
+    display-name: Deploy_Lambda_App
+    project-type: freestyle
+    description: "This job uploads an AWS Lambda app to an S3 bucket and kicks off the relevant Terraform job"
+    properties:
+        - github:
+            url: https://github.com/alphagov/govuk-lambda-app-deployment
+        - inject:
+            properties-content: |
+              ENVIRONMENT=<%= @environment %>
+    scm:
+      - govuk-lambda-app-deployment_Deploy_Lambda_App
+    builders:
+      - shell: |
+          cd $TARGET_APPLICATION
+          source ./jenkins.sh # this job should create the source ready for uploading to an S3 bucket
+          s3cmd sync $FILE_TO_UPLOAD s3://govuk-lambda-applications-$ENVIRONMENT/${TARGET_APPLICATION}/
+    publishers:
+        - trigger-parameterized-builds:
+            - project: Deploy_Terraform_Project
+              predefined-parameters: |
+                AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+                AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+                ACTION=apply
+                PROJECT_NAME=${TARGET_APPLICATION}
+              condition: 'UNSTABLE_OR_BETTER'
+    wrappers:
+        - ansicolor:
+            colormap: xterm
+    parameters:
+        - choice:
+            name: TARGET_APPLICATION
+            description: Lambda application to deploy.
+            choices: <%= ['-- Choose an app'] + @lambda_apps %>
+    parameters:
+        - string:
+            name: AWS_ACCESS_KEY_ID
+            description: Your AWS access key ID
+            default: false
+        - password:
+            name: AWS_SECRET_ACCESS_KEY
+            description: Your AWS secret access key
+            default: false
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -483,6 +483,9 @@ govuk_jenkins::job::smokey::signon_password: "%{hiera('smokey_signon_password')}
 govuk_jenkins::job::run_rake_task::applications: *deployable_applications
 govuk_jenkins::job::deploy_app::applications: *deployable_applications
 
+govuk_jenkins::job::deploy_lambda_app::lambda_apps:
+  - 'email_alert_notifications'
+
 govuk_mysql::server::expire_log_days: 3
 govuk_mysql::server::monitoring::master::plaintext_mysql_password: "%{hiera('mysql_nagios')}"
 govuk_mysql::server::monitoring::slave::plaintext_mysql_password: "%{hiera('mysql_nagios')}"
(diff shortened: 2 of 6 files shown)


rjw1:
Looks good so far.
Lets go for supplying AWS creds. 
We can always add them in later if need be.

modules/govuk_jenkins/templates/jobs/deploy_lambda_app.yaml.erb
```diff
@@ -0,0 +1,51 @@
...
+          cd $TARGET_APPLICATION
+          ./jenkins.sh # this job should create the source ready for uploading to an S3 bucket
+          s3cmd sync . s3://govuk-lambda-applications-$ENVIRONMENT/$TARGET_APPLICATION/
+    publishers:
+        - trigger-parameterized-builds:
+            - project: Deploy_Terraform_Project
+              predefined-parameters: |
+                AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID
+                AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY
+                ACTION=apply
```
gpeng: The lambda resource needs the S3 version to be passed in. Would this need to happen here? In the email alert project we can pass it in as `TF_VAR_ rename_email_files_with_request_id_version` which I thought could be the makings of a reasonably convention.

modules/govuk_jenkins/templates/jobs/deploy_lambda_app.yaml.erb
```diff
@@ -0,0 +1,51 @@
...
+    properties:
+        - github:
+            url: https://github.com/alphagov/govuk-lambda-app-deployment
+    scm:
+      - govuk-lambda-app-deployment_Deploy_Lambda_App
+    builders:
+      - shell: |
+          cd $TARGET_APPLICATION
+          ./jenkins.sh # this job should create the source ready for uploading to an S3 bucket
+          s3cmd sync . s3://govuk-lambda-applications-$ENVIRONMENT/$TARGET_APPLICATION/
```
gpeng: If we use `aws s3api put-object` here it returns the version id. It's in JSON though... e.g.

```
{
    "VersionId": "HdrY9L9wuhSA1z0d.Xc1h05F3zeigeYM",
    "ETag": "\"814c08483064b7d0da54ce1512a5d33a\""
}
```

modules/govuk_jenkins/templates/jobs/deploy_lambda_app.yaml.erb
surminus: This will happen here, just once I've worked on getting the appropriate version to pass in out of the script.

modules/govuk_jenkins/templates/jobs/deploy_lambda_app.yaml.erb
surminus: Yeah, sadly s3cmd appears to have no support for versioning. Will look into an alternative method.

surminus:
I've used awscli instead of s3cmd in 0228899833cb832cdf44c46333f04caae6d88380 but can't think of a nice way to cleanly export this over to the triggered parametrized build. We could export into a file and read the file in the build shell of the Terraform job, or add an optional parameter.

gpeng:
Rather than exporting a file here could we just get all of the files from the appropriate env's lambda bucket in the terraform rake task and read all the latest versions into env variables before running `apply`|`plan`? Maybe only add the env variable if it doesn't exist and then we could override if there's a need to deploy a previous version for some reason. Would that work?

surminus:
Exporting a file is not very nice and I think this approach is good because it means we can run this project independently from the deployment job. Thanks!

surminus:
As per [this comment](https://github.com/alphagov/govuk-terraform-provisioning/pull/46/files#r64203473) if the parameter is not set it pulls the latest by default, which would mean we wouldn't have to add any further dependencies as Terraform can't get the version and we wouldn't have to rely on more things in Rakefile. 

It _would_ be good to have a conditional if we wanted to optionally set a specific version.

surminus:
My latest commits have added in variables which are used to [trigger the job for Terraform projects](https://github.com/alphagov/govuk-terraform-provisioning/pull/46), one of which is gathered from the [script found in the govuk-lambda-app-deployment repo](https://github.com/alphagov/govuk-lambda-app-deployment/pull/2).

modules/govuk_jenkins/templates/jobs/deploy_lambda_app.yaml.erb
```diff
@@ -0,0 +1,55 @@
...
+        - inject:
+            properties-content: |
+                ENVIRONMENT=<%= @environment %>
+    scm:
+      - govuk-lambda-app-deployment_Deploy_Lambda_App
+    builders:
+      - shell: |
+          cd ${TARGET_APPLICATION}
+          ./jenkins.sh # this job should create the source ready for uploading to an S3 bucket
+          /usr/local/bin/s3cmd sync . s3://govuk-lambda-applications-${ENVIRONMENT}/${TARGET_APPLICATION}
```
alexmuller: I think this will include everything in the `TARGET_APPLICATION` directory which includes the `jenkins.sh` script but that probably isn't too much of a problem? We might be able to pass in `*.zip` here if that's cleaner?

surminus: Oh yes, `jenkins.sh` should have actually exported the `FILE_TO_UPLOAD` variable which should be used here.

surminus:
Addressed comments.

modules/govuk_jenkins/templates/jobs/deploy_lambda_app.yaml.erb
```diff
@@ -0,0 +1,54 @@
...
+        - inject:
+            properties-content: |
+              ENVIRONMENT=<%= @environment %>
+    scm:
+      - govuk-lambda-app-deployment_Deploy_Lambda_App
+    builders:
+      - shell: |
+          cd $TARGET_APPLICATION
+          ./jenkins.sh # this job should create the source ready for uploading to an S3 bucket
+          s3cmd sync $FILE_TO_UPLOAD s3://govuk-lambda-applications-$ENVIRONMENT/${TARGET_APPLICATION}/
```
alexmuller: I think this might have to `source ./jenkins.sh` in order to get the exported `$FILE_TO_UPLOAD` to be available here.

surminus: You are correct. Amended.

deanwilson:
Code looks fine. Merge away.

PR #4429, opened 2016-05-20
Update Nginx config to support specialist publisher rewrite

brenetic:
We plan on releasing specialist publisher one document_type at a
time. In order to facilitate this we need the capability of
routing specific paths to the 'rebuild' application whilst serving
the rest from the existing v1 application.

This updates the loadbalancer to direct sp-rebuild/assets to the specialist-publisher-rebuild app.
It also sets the Host HTTP header to specialist-publisher for the
specialist-publisher-rebuild app.


boffbowsh:
This looks good 👍 . Why is it Do Not Merge?

brenetic:
We still have one small change to make to a mongo database url so that v1 and v2 can share signon users. Once that's updated and tested, we'll ping you. :-)

tijmenb:
👍

PR #4431, opened 2016-05-23
Enable specialist publisher rebuild in all environments

tuzz:
Part of: https://trello.com/c/h2N84bHb/122-per-format-path-to-live-large

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -380,6 +380,8 @@ govuk::apps::specialist_publisher::enabled: true
 govuk::apps::specialist_publisher::nagios_memory_warning: 800
 govuk::apps::specialist_publisher::nagios_memory_critical: 900
 
+govuk::apps::specialist_publisher_rebuild::enabled: true
+
 govuk::apps::spotlight::enabled: true
 govuk::apps::spotlight::alert_5xx_warning_rate: 0.2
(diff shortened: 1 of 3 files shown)


hieradata/common.yaml
```diff
@@ -380,6 +380,9 @@ govuk::apps::specialist_publisher::enabled: true
 govuk::apps::specialist_publisher::nagios_memory_warning: 800
 govuk::apps::specialist_publisher::nagios_memory_critical: 900
 
+govuk::apps::specialist_publisher_rebuild::enabled: true
+govuk::apps::specialist_publisher_rebuild::publish_pre_production_finders: true
```
jennyd: Setting this to `true` without overriding it to `false` for Production and Staging will mean that pre-production finders will be published in all environments, because common hieradata is used as the base for all environments (I think...). The param on the class [already defaults to `false`](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/specialist_publisher_rebuild.pp#L74), so omitting this line and not removing the `true` override for dev and Integration would do what you want.

tuzz: Good spot. 🔬

We've updated this pull request to only hoist the 'enabled' setting up.

Thanks

PR #4434, opened 2016-05-23
Rate limit paths under /performance

alexmuller:
Set up a new Nginx request limit zone for the Performance Platform and apply that zone to the whole `/performance` location block.

We're doing this because the Performance Platform frontend is sometimes quite slow to serve responses and we'd rather not phone on-call. Serving a 429 rate limit response is better.


surminus:
:+1:

PR #4435, opened 2016-05-24
Make jenkins run links import daily for local links manager

h-lame:
Local Links Manager needs to import all the links that each local
authority provides to services and interactions.  These links are a csv
dataset provided on local.directgov.uk and we should import it daily.  The
import script is already part of the [local-links-manager
repo](https://github.com/alphagov/local-links-manager) and this commit
adds a jenkins job to run that import script configured to run at 5am (ish)
every day.

For: https://trello.com/c/Q0fMMggq/370-import-local-links-into-local-links-manager-5

The script we run is in https://github.com/alphagov/local-links-manager/pull/15 which is due for a deploy today.

### DIFF
--- a/hieradata/production.yaml
+++ b/hieradata/production.yaml
@@ -87,6 +87,7 @@ govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::job::govuk_cdn_nightly_2xx_status_collection
   - govuk_jenkins::job::launch_vms
   - govuk_jenkins::job::local_links_manager_import_service_interactions
+  - govuk_jenkins::job::local_links_manager_import_links
   - govuk_jenkins::job::mirror_network_config
   - govuk_jenkins::job::network_config_deploy
   - govuk_jenkins::job::passive_checks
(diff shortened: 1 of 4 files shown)


alexmuller:
Could this job trigger the job we have [to run rake tasks](https://deploy.publishing.service.gov.uk/job/run-rake-task/)?

h-lame:
@alexmuller oh, it certainly could if that's better - I was blindly following the pattern used in #4403, but I could update both to trigger the other job instead.

h-lame:
@alexmuller worked a treat - see new commit, thanks!

alexmuller:
This is great, thanks! Sorry to make more work for you but in 57f33b5cd1d46a30b8d2ba0e02c1e27d63c8a77d I've just changed `MACHINE` so it'll now be `MACHINE_CLASS` with a value of `backend` in both of those files.

h-lame:
No worries, thanks for pointing it out @alexmuller - rebased too.

alexmuller:
:+1: :+1:

PR #4436, opened 2016-05-24
Use a random node for running rake tasks

alexmuller:
This will help us avoid single points of failure and flush out configuration problems.


surminus:
:+1: fftm

PR #4437, opened 2016-05-24
Bugfix `/performance` rate limiting

alexmuller:
Proving once again that I have no idea how Nginx location blocks work.


surminus:
Well I don't either because I reviewed the last change so I'll give you a :+1: in the assumption that you know what you're doing this time

PR #4439, opened 2016-05-25
Move s3 asset sync to slave

surminus:
The asset master in Production was set up to do a full sync of attachments to an S3 bucket. Unfortunately this caused the machine to run out of memory and OOM killer to kill the process before it had completed checking all the files on the source machine (completed about 40% before it ran OOM). We made the decision to move this process to an asset slave instead where we can upgrade memory, perform maintenance and will have less of an impact overall than the asset master.

However, because there will now be two machines sharing AWS credentials it made sense to move the logic to s_asset_base, set enabling parameters and specifically pick nodes to run the required function on. By default we should have nothing that copies to S3 until we specifically enable it
in the node specific hieradata. Once enabled it should pick up the same credentials so we don't have to repeat them in our secrets store.

### DIFF
--- a/modules/govuk/manifests/node/s_asset_base.pp
+++ b/modules/govuk/manifests/node/s_asset_base.pp
@@ -7,8 +7,37 @@
 # [*firewall_allow_ip_range*]
 #   The IP range that is allowed to access asset machines
 #
+# [*s3_bucket*]
+#   The URL of an S3 bucket. This can be used to either push backups or
+#   sync from an environment. Format should be the full URL:
+#   e.g s3://foo-bucket/
+#
+# [*aws_access_key_id*]
+#   The AWS key ID for the bucket you are accessing.
+#
+# [*aws_secret_access_key*]
+#   The secret part of the keypair.
+#
+# [*process_uploaded_attachments_to_s3*]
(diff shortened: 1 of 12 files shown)


modules/govuk/manifests/node/s_asset_base.pp
```diff
@@ -99,4 +128,71 @@
...
+      owner   => 'assets',
+      group   => 'assets',
+      mode    => '0750';
+    '/etc/govuk/aws/env.d/AWS_ACCESS_KEY_ID':
+      ensure  => present,
+      content => $aws_access_key_id,
+      owner   => 'assets',
+      group   => 'assets',
+      mode    => '0640';
+    '/etc/govuk/aws/env.d/AWS_SECRET_ACCESS_KEY':
```
deanwilson: Do we do this explicitly with `file` and not `govuk_envvar` so we can limit the permissions?

deanwilson:
Looks fine. Merge away.

modules/govuk/manifests/node/s_asset_base.pp
surminus: Correct

PR #4440, opened 2016-05-25
Disable full asset sync to s3

surminus:
There is a larger change proposed in e194f1847a866c8e0d32ddc6a417d62b61456b3f but in the meantime we need to ensure that the full asset sync doesn't run so the asset master doesn't
run out of memory again.


rjw1:
merge when green

PR #4441, opened 2016-05-25
Only restart apps on critical high memory alerts

steventux:
Applications should only be restarted when a `CRITICAL` service state is encountered by the high memory event handler.


alexmuller:
My only concern with this was what happens when the `case` statement is called with a value that's not listed, but it seems like bash just doesn't execute anything (which is the behaviour we want) :+1:

steventux:
I did think to log something on `WARNING`, wasn't convinced it was all that useful. Happy to add this though.

PR #4443, opened 2016-05-25
Add logstash filter for logrus logging

alexmuller:
Apps that use Go's [logrus][1] (all of our Go apps) output a `msg` field which contains the log message. Logstash/Kibana uses an `@message` field.

This commit adds a rename filter to Logstash so that we have more consistent naming of that field.

[1]: https://github.com/Sirupsen/logrus

I haven't tested this in Vagrant (yet).

### DIFF
--- a/modules/govuk/manifests/node/s_logging.pp
+++ b/modules/govuk/manifests/node/s_logging.pp
@@ -78,6 +78,11 @@ class govuk::node::s_logging (
                 'MMM  d HH:mm:ss' ],
     order => '13',
   }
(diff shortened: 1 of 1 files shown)


surminus:
TBH, the easiest way to test this is probably just in integration, so I'd vote to merge with a flag that only enables it in that environment, or just YOLO?

deanwilson:
I agree with the "push to integration" and test approach as we're using very old versions of the module and it's a pain to work through.

PR #4444, opened 2016-05-26
Document Vagrant 1.8.x hostname provisioning solution

tvararu:
This is a temporary fix to an issue that appears when using Vagrant 1.8.x.


Davidslv:
LGTM 👍

PR #4446, opened 2016-05-26
WAL-E env sync bugfixes

surminus:
In testing we found a few bugs which meant that this didn't work off the bat.

One addition is to disable and then re-enable and re-run Puppet: because we're restoring from a Production DB the passwords get updated, so running Puppet ensures the passwords are correct. It will also ensure that the gpg.conf file is updated and will remove the added passphrase.

### DIFF
--- a/modules/govuk_postgresql/templates/usr/local/bin/wal-e_env_sync
+++ b/modules/govuk_postgresql/templates/usr/local/bin/wal-e_env_sync
@@ -13,20 +13,35 @@ set -e
 # Redirect stdout and stderr to syslog
 exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
 
-# Stop the PostgreSQL service
-sudo service postgresql stop
+# Check if Puppet is enabled
+if sudo test -f /var/lib/puppet/state/agent_disabled.lock; then
+  # If it isn't then quit
+  echo "Puppet agent disabled! If Puppet cannot run then we will not be able"
+  echo "to resync passwords and things will break. Quitting"
+  exit 1
+else
+  echo "PostgreSQL environment sync starting"
 
-# Push the passphrase to a file so we can decrypt without prompt
-sudo -iu postgres envdir <%= @env_sync_envdir %> sh -c  'echo "passphrase $GPG_PASSPHRASE" >> /var/lib/postgresql/.gnupg/gpg.conf'
+  # Otherwise go ahead and disable Puppet to start the sync
+  govuk_puppet --disable "Disabling Puppet to sync data (by $0)"
 
-# Start the restore process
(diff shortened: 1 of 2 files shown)


modules/govuk_postgresql/templates/usr/local/bin/wal-e_env_sync
```diff
@@ -13,20 +13,23 @@ set -e
 # Redirect stdout and stderr to syslog
 exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
 
+# Disable Puppet so PostgreSQL isn't automatically restarted
+govuk_puppet --disable
```
rjw1: adding a reason would be good.
Also possibly checking if its already disabled?

deanwilson: Agreed. If we already have puppet disabled on the host then this job will reenable it no matter the previous reason.

alexmuller: Hm, the `/var/lib/puppet` directory is owned by `puppet:puppet` and isn't world-readable so I'm not sure how the deploy or postgres user could test whether Puppet is disabled on the host. The file that gets created is `/var/lib/puppet/state/agent_disabled.lock`.

alexmuller: Added a reason though.

modules/govuk_postgresql/templates/usr/local/bin/wal-e_env_sync
```diff
@@ -13,20 +13,35 @@ set -e
...
 
-# Add the recovery.conf configuration (this is renamed to recovery.done after PostgreSQL starts)
-sudo -iu postgres echo "restore_command = \'envdir /etc/wal-e/env.d /usr/local/bin/wal-e wal-fetch \"%f\" \"%p\"\'" > $RECOVERY_FILE
+  # Start the restore process
+  sudo -iu postgres envdir <%= @env_sync_envdir %> /usr/local/bin/wal-e backup-fetch --blind-restore <%= @datadir -%> LATEST
 
-# Start the PostgreSQL service
-sudo service postgresql start
+  # Add the recovery.conf configuration (this is renamed to recovery.done after PostgreSQL starts)
+  sudo -iu postgres echo "restore_command = 'envdir /etc/wal-e/env_sync/env.d /usr/local/bin/wal-e wal-fetch \"%f\" \"%p\"'" > $RECOVERY_FILE
```
alexmuller: I think the envdir here could use the `@env_sync_envdir` variable

alexmuller:
There's a couple of mentions of "MySQL" in there which I think should be PostgreSQL :)

alexmuller:
I updated this as Laura's on 2nd line.

surminus:
Thanks!

PR #4447, opened 2016-05-26
Add access for postgresql standbys to S3

surminus:
We are able to successfully restore the data from the Production S3 bucket to the PostgreSQL master, but when we do this the slaves get extremely unhappy. We would like to test pulling all the data from S3
onto the slaves as well, and while keeping hold of some files and configuration, try to restore replication at the same time as the master so we don't need to do a full resync.

This change is mainly for testing purposes, but would also be kept if we went along with this approach in the future.


deanwilson:
The code looks fine. We should discuss the approach before we decide.

deanwilson:
Merge away.

PR #4452, opened 2016-05-31
Disable sidekiq workers in development

boffbowsh:
All Sidekiq workers are normally run manually
as needed in the development VM. This changes
email-campaign-api and travel-advice-publisher
to do the same.


alexmuller:
:+1: as long as the tests pass, which they will.

PR #4455, opened 2016-06-01
Run proper bundle install before running tagging sync check

tijmenb:
This was previously done with `bin/run_automated_checks`, but that
didn't account for the fact that jenkins can't install gems to the
default directory. This caused the jenkins job to fail:

https://deploy.integration.publishing.service.gov.uk/job/tagging-sync-check/1/console

https://trello.com/c/InnHPkm5/645-add-jenkins-job-for-tagging-migration-checker


benhyland:
👍 
once green

PR #4457, opened 2016-06-02
Add publishing-api bearer token to sync check

tijmenb:
The sync check needs `PUBLISHING_API_BEARER_TOKEN` set so that it can communicate with the publishing-api.

A corresponding change in the encrypted hieradata: https://github.gds/gds/deployment/pull/1012


MatMoore:
👍

PR #4458, opened 2016-06-02
Bugfix attachment syncs

surminus:
A couple of things managed to sneak past in my very large set of changes last time:

 - The cronjob was incorrectly set up
 - The bucket name format was incorrect


rjw1:
Looks good to me. merge when ready

PR #4461, opened 2016-06-03
Bug fixes for Terraform and Lambda deployments

alexmuller:


surminus:
:+1: That's weird, I thought I'd had set it up to explictly set `TF_VAR_LAMBDA_FILENAME` so it correctly triggered the Terraform build, but I guess I never committed it. Merge away.

alexmuller:
Oh yeah, does the deploy Lambda app job need to set `TF_VAR_LAMBDA_FILENAME` when it calls the downstream job?

surminus:
Yes as there is a conditional in the Rakefile, but it should automagically be set by the `jenkins.sh` task in the _other_ repo.

alexmuller:
Sorry, I'm getting confused. Is it [this thing](https://github.com/alphagov/govuk-lambda-app-deployment/blob/0fea2136e16080faf04e6910203550d77e067fff/email_alert_notifications/jenkins.sh) that should be setting `TF_VAR_LAMBDA_FILENAME`?

alexmuller:
Pushed some more commits. We also need https://github.com/alphagov/packager/pull/113.

surminus:
:+1: Probably a more sane way to do it.

PR #4463, opened 2016-06-03
Create defined crondotdee

surminus:
This creates a way to add stuff to `/etc/cron.d` which can be a preference from `/etc/cron.daily` and a shared crontab. In this example we'll add rkhunter as it tends to alert throughout the day when the cron.daily job doesn't run for whatever reason.

### DIFF
--- a/modules/cron/manifests/crondotdee.pp
+++ b/modules/cron/manifests/crondotdee.pp
@@ -1,0 +1,53 @@
+# == Define: Cron::Crondotdee
+#
+# Create a file in /etc/cron.d rather than in cron.daily or a shared crontab.
+# Default behaviour is to set a command, minute and hour and to run as root.
+#
+# Note: set hour/minute/day/month/weekday as a string rather than an integer as
+# they could potentially contain non-integer values, so keep it consisent.
+#
+# === Parameters:
+#
+# [*command*]
+#   The command to schedule. Required.
+#
+# [*hour*]
+#   The hour(s) to run. Default is 09:00.
+#
+# [*minute*]
+#   The minute(s) to run. Default is on the hour.
+#
+# [*day*]
+#   The day to run. Default is everyday.
+#
+# [*month*]
+#   The month to run. Default is every month.
+#
+# [*weekday*]
+#   The weekday to run. Default is every weekday.
+#
+# [*user*]
+#   Which user to run as. This can be used if you wish to run a script without
+#   invoking "sudo -u $user". Default is root, as per the standard cron job.
+#
+# [*mailto*]
+#   Specifically set where the cron should attempt to email output to.
+#   Example would be a user or an email address, or use an empty string to
+#   suppress mailing any output.
+define cron::crondotdee (
+  $command,
+  $hour = '9',
+  $minute = '0',
+  $day = '*',
+  $month = '*',
+  $weekday = '*',
+  $user = 'root',
+  $mailto = undef,
+) {
+  file { "/etc/cron.d/${title}":
+    ensure  => present,
+    content => template('cron/etc/cron.d/crondotdee.erb'),
+    require => File['/etc/cron.d'],
+  }
+}
+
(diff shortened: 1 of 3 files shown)


rjw1:
:+1: although I do wonder if we should use an existing [module](https://forge.puppet.com/torrancew/cron), although that may be out of scope of this PR.

surminus:
I did think about that but thought it seemed easier to write my own than install another module. Would the module you linked to clash with our own module? Would that need to be renamed or refactored?

modules/rkhunter/manifests/monitoring.pp
```diff
@@ -44,11 +44,19 @@
...
+  file { '/usr/local/bin/rkhunter-passive-check':
     ensure  => present,
     mode    => '0755',
     content => template('rkhunter/rkhunter-passive-check.erb'),
   }
+
+  # Run it twice a day to stop superfluous alerting
+  cron::crondotdee { 'rkhunter-passive-check':
+    command => '/usr/local/bin/rkhunter-passive-check',
+    hour    => '6, 14',
```
deanwilson: Does a space between the two hour specifiers do what you'd expect? I didn't think you could have whitespace.

modules/rkhunter/manifests/monitoring.pp
```diff
@@ -44,11 +44,19 @@
...
+    mailto  => '',
```
deanwilson: Should the default `mailto` be undef or an empty string? It seems odd to use both when i _think_ you just want an empty `MAILTO=`.

modules/cron/manifests/crondotdee.pp
```diff
@@ -0,0 +1,53 @@
+# == Define: Cron::Crondotdee
+#
+# Create a file in /etc/cron.d rather than in cron.daily or a shared crontab.
+# Default behaviour is to set a command, minute and hour and to run as root.
+#
+# Note: set hour/minute/day/month/weekday as a string rather than an integer as
+# they could potentially contain non-integer values, so keep it consisent.
```
deanwilson: Consistent.

deanwilson:
Do we want the ability to specify the file mode for cronjobs created via this? `0644` by default but with the option of `0640` for things with creds in them (which I _think_ we have).

deanwilson:
If we decide to go with our own code this looks fine. A spec or two wouldn't hurt though.

deanwilson:
I've made a few tiny changes to this, and we've agreed to use our own module, so please re-review with the understanding we will be using this code. @rjw1 @alexmuller

modules/cron/manifests/crondotdee.pp
```diff
@@ -0,0 +1,56 @@
...
+#   Which user to run as. This can be used if you wish to run a script without
+#   invoking "sudo -u $user". Default is root, as per the standard cron job.
+#
+# [*mailto*]
+#   Specifically set where the cron should attempt to email output to.
+#   Example would be a user or an email address, or use an empty string to
+#   suppress mailing any output.
+define cron::crondotdee (
+  $command,
+  $hour = '9',
```
rjw1: I do wonder if these should be required parameters

modules/cron/templates/etc/cron.d/crondotdee.erb
```diff
@@ -0,0 +1,4 @@
+<% if @mailto %>
```
rjw1: does this need `-`s to not generate whitespace?

rjw1:
couple of minor questions which to be honest shouldn't stop this from being merged.

modules/cron/templates/etc/cron.d/crondotdee.erb
deanwilson: "We want whitespace to add a visual delimiter between the jobs" he bluffs.

rjw1:
merge when ready :+1:

PR #4464, opened 2016-06-03
This commit adds the script /usr/local/bin/{mongodb-restore-s3,mongod…

djsd123:
…b-restore-s3-wrapper}. All the script does is downloads the most recent backup, decrypts it, uncompresses it and restores the backup into mongo. We're able to get the latest backup with the script because amazon s3 stores files in alphabetical,chronological order. Note: this only works when the filenames are timestamped in 'yyyy-mm-dd' format.


djsd123:
[DO NOT MERGE] Mongo auto restore

PR #4465, opened 2016-06-06
Increase memory limits

tuzz:
1) Set smartanswers critical alert at 4,500MB not 45,000MB

2) Increase the memory thresholds for specialist publisher…
> This app looks to be routinely using more than this amount of memory. It’s going to be killed soon so I think it’s reasonable to increase the limits until that time.

3) Increase content-store memory limits…
> Dependency resolution has increased load on content store as of 25th May, 2016. This gives it a bit more breathing room.


alexmuller:
Excellent first commit 👏

surminus:
:+1: merge when ready.

alexmuller:
Improved the validation so that it would have caught this in #4466.

PR #4466, opened 2016-06-06
Improve validation for app memory threshold

alexmuller:
Use `validate_integer` instead of validating a regex because it's much easier to figure out what's going on.


alexmuller:
Tests are failing because of an actual problem that @tuzz has fixed in #4465.

tuzz:
Merge when ready.

PR #4469, opened 2016-06-06
Update key for David Basalla

davidbasalla:
I received a new laptop today, so adding my new ssh key.


tijmenb:
Don't forget to set your git email so that commits get connected to your GitHub account (it's default new laptop now: https://github.com/alphagov/govuk-puppet/commit/2731f6a35e25b7e9a60600025891a8ada7f7b1c7.patch)

davidbasalla:
Thanks @tijmenb for pointing that out, should be sorted now

PR #4470, opened 2016-06-07
Bugfix attachment syncs

surminus:
The attachment sync wasn't able to gain a lock so failed to run.

The cron daily entry still exists and while doesn't do anything, shouldn't be there.


deanwilson:
Is the directory creation a deliberate addition to this PR? @alexmuller can we chat about this tomorrow please?

PR #4471, opened 2016-06-07
Add DR nodes to the allowed list of reboots

surminus:
The DR nodes are safe to reboot at anytime, so we should allow automatic reboots overnight.


rjw1:
:+1: merge when ready

PR #4472, opened 2016-06-07
Add timestamper plugin

benhyland:
This plugin add timestamps to console output.
It is useful for longer running or multi-step jobs, especially
when they are not doing what they are supposed to.

See https://wiki.jenkins-ci.org/display/JENKINS/Timestamper


surminus:
:+1: Looks useful

PR #4473, opened 2016-06-08
Mongodb s3 Restore

djsd123:
This commit adds the script /usr/local/bin/{mongodb-restore-s3,mongodb-restore-s3-wrapper}. All the script does is downloads the most recent backup, decrypts it, uncompresses it and restores the backup into mongo. We're able to get the latest backup because amazon s3 stores files in alphabetical,chronological order. Note: this only works when the filenames are timestamped in 'yyyy-mm-dd' format.

Added quiet flags to commands.  Referenced User variable from Backup.


modules/mongodb/templates/mongodb-restore-s3.erb
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -e
+
+AWS_ACCESS_KEY_ID=<%= @aws_access_key_id %>
+AWS_SECRET_ACCESS_KEY=<%= @aws_secret_access_key %>
```
surminus: This can be removed as we're passing in `envdir`

modules/mongodb/templates/mongodb-restore-s3.erb
```diff
@@ -0,0 +1,30 @@
...
+S3_BUCKET=<%= @s3_bucket %>
+BACKUP_DIR=<%= @backup_dir %>
+
+cd $BACKUP_DIR
+
+
+TIME="$(date +%s)"
+
+s3cmd get --force `s3cmd ls s3://${S3_BUCKET}/mongodump* | tail -1 | awk '{print $4}'`| \
```
surminus: Listing the contents of the bucket each time might become a slow operation when the bucket gets larger. An alternative method would be to know the name of the file you're GETing before you've even gone into the bucket i.e. "mongodump-latest.tar.gz"

surminus:
We shouldn't schedule or add a passive check because we're going to schedule this in another way (i.e. Jenkins). Otherwise the tool itself looks fine!

surminus:
This should also include: 

 - Which machine it's going to run on (add the class include the to the node manifest?)
 - How do we choose the right machine in the cluster? 
 - What happens to the slave - do we perform a full resync, too?

modules/mongodb/manifests/s3backup/restore.pp
```diff
@@ -0,0 +1,66 @@
...
+  $backup_dir = '/var/lib/s3backup',
+  $cron = false
+){
+  include backup::client
+  $backup_user = 'govuk-backup'
+
+  File {
+    owner => $backup_user,
+    group => $backup_user,
+    mode  => '0660',
```
deanwilson: This line of code is never directly used and so can be removed.

djsd123:
In regards to listing bucket contents.  I could not find a good way to stat objects (that aren't buckets) on s3 from the command line.  A possible solution could be to make use of S3's versioning and lifecycle rules to housekeep.

djsd123:
This branch diverged a little too much from master in that tests were failing.

PR #4474, opened 2016-06-08
Remove Jake Benilov

boffbowsh:
Jake has now left GDS

Jake was the only recipient of emails about the success or failure of
the `run_govuk_complaint_rate_report` job in production. This job has
been running successfully for a while, so I’m removing the email
publisher.

@benilovj were these emails useful? Should someone else pick them up?


surminus:
:+1: when @benilovj confirms about the job feel free to merge

benilovj:
👍  from me.

PR #4477, opened 2016-06-08
Add param to deploy apps from GitHub Enterprise

alexmuller:
Boolean parameter to deploy from GHE.


jennyd:
Tested on integration and it works 👍

PR #4486, opened 2016-06-13
Rename Rummager indexing process

tijmenb:
`enable_publishing_api_document_indexer` isn't correct because we don't use the publishing-api message queue to index documents. Rather, we listen to changes and trigger re-indexes. This updates the naming to avoid confusion.

Remove `enable_publishing_api_document_indexer` from integration.yml because it duplicates the default.

Related: https://github.com/alphagov/rummager/pull/650


benhyland:
👍 looks fine, merge when green

PR #4488, opened 2016-06-13
asset-slave-1 rebooted and rearranged its disks

boffbowsh:
Now pv is trying to move its volume groups around. This
commit updates the expected disk layout.

```
paulbowsher@production-asset-slave-1:~$ sudo fdisk -l 2>/dev/null | grep 'Disk /dev/sd'
Disk /dev/sda: 48.3 GB, 48318382080 bytes
Disk /dev/sdb: 53.7 GB, 53687091200 bytes
Disk /dev/sdd: 274.9 GB, 274877906944 bytes
Disk /dev/sdc: 549.8 GB, 549755813888 bytes
```


rjw1:
merge when ready.

PR #4489, opened 2016-06-14
Alerting is currently slightly off due to summer time changes.

deanwilson:
As a quick fix we're temporarily changing the valid alerting hours by hand to ensure
people only get called when they should be. We'll have to change it back by hand later.

https://trello.com/c/NLSxzODq


djsd123:
FFTM

PR #4491, opened 2016-06-14
mongodb s3 backup daily

djsd123:
A mongo dump will be copied to a different bucket daily.  Added; statement which uploads to daily bucket between midnight and ten past, cron job to execute script at midnight.

### DIFF
--- a/modules/mongodb/manifests/s3backup/backup.pp
+++ b/modules/mongodb/manifests/s3backup/backup.pp
@@ -56,6 +56,7 @@ class mongodb::s3backup::backup(
   $private_gpg_key = undef,
   $private_gpg_key_fingerprint,
   $s3_bucket  = 'govuk-mongodb-backup-s3',
+  $s3_bucket_daily = 'govuk-mongodb-backup-s3-daily',
   $user = 'govuk-backups'
 ){
 
--- a/modules/mongodb/manifests/s3backup/backup.pp
+++ b/modules/mongodb/manifests/s3backup/backup.pp
@@ -1,3 +1,4 @@
+# FIXME  Once deployed please remove user and file references that ensure absent
 # == Class: mongodb::s3backup::backup
 #
 # Backup a MongoDB server to AWS S3
@@ -54,17 +55,18 @@ class mongodb::s3backup::backup(
   $cron = true,
   $env_dir = '/etc/mongo_s3backup',
   $private_gpg_key = undef,
-  $private_gpg_key_fingerprint,
-  $s3_bucket  = 'govuk-mongodb-backup-s3',
-  $s3_bucket_daily = 'govuk-mongodb-backup-s3-daily',
-  $user = 'govuk-backups'
+  $private_gpg_key_fingerprint = undef,
+  $s3_bucket  = undef,
+  $s3_bucket_daily = undef,
+  $user = 'govuk-backup'
 ){
+
+  include backup::client
 
   validate_re($private_gpg_key_fingerprint, '^[[:alnum:]]{40}$', 'Must supply full GPG fingerprint')
 
-  user { $user:
-    ensure     => 'present',
-    managehome => true,
+  user { 'govuk-backups':
+    ensure => absent,
   }
 
   file { $backup_dir:
@@ -102,6 +104,13 @@ class mongodb::s3backup::backup(
     mode    => '0640',
   }
 
+  file { '/var/lib/s3backup':
+    ensure => directory,
+    owner  => $user,
+    group  => $user,
+    mode   => '0775',
+  }
+
   # push script
   file { '/usr/local/bin/mongodb-backup-s3':
     ensure  => present,
@@ -110,6 +119,10 @@ class mongodb::s3backup::backup(
     group   => $user,
     mode    => '0755',
     require => User[$user],
+  }
+
+  file { '/usr/local/bin/mongodb-backup-s3-wrapper':
+    ensure => absent,
   }
 
   # push gpg key
@@ -159,3 +172,4 @@ class mongodb::s3backup::backup(
   }
 
 }
+
--- a/modules/mongodb/manifests/s3backup/cron.pp
+++ b/modules/mongodb/manifests/s3backup/cron.pp
@@ -5,23 +5,30 @@
 # [*user*]
 #   The user to run the cronjob as.
 #
+# [*backup_dir*]
+#   The directory where backups are managed
+#
 class mongodb::s3backup::cron(
-  $user = mongodb::s3backup::backup::user
+  $user = 'govuk-backup',
 ) {
 
+  include backup::client
   require mongodb::s3backup::package
   require mongodb::s3backup::backup
 
   cron { 'mongodb-s3backup':
-    command => '/usr/local/bin/mongodb-backup-s3',
+    command => '/usr/bin/setlock -n /var/lock/mongodb-s3backup \
+    /usr/local/bin/mongodb-backup-s3',
     user    => $user,
     minute  => '*/15',
   }
 
   cron { 'mongodb-s3-night-backup':
-    command => '/usr/local/bin/mongodb-backup-s3',
+    command => '/usr/bin/setlock /var/lock/mongodb-s3backup \
+    /usr/local/bin/mongodb-backup-s3 daily',
     user    => $user,
     hour    => '0',
+    minute  => '0',
   }
 
 }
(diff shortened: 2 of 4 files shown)


modules/mongodb/templates/mongodb-backup-s3.erb
```diff
@@ -25,7 +25,7 @@ trap nagios_passive EXIT
 
 # Backup mongodb
 TIME="$(date +%s)"
-/usr/bin/mongodump -h $BACKUP_NODE -o $BACKUP_DIR --quiet
+/usr/bin/mongodump -h $BACKUP_NODE -o $BACKUP_DIR &> /dev/null
```
deanwilson: Why are we suppressing this? I imagine it'd be handy to have processed output sent to the normal logs address.

surminus:
Can we have some context on why we want a separate daily full dump when the "incremental" dump is actually just a full dump every 15 minutes?

djsd123:
Correct.  Mongo dump only does full exports of the database.  A mongosnapshot however does support the concept of incremental backups with a replay mechanism to restore.

djsd123:
I've modified the commit message to explain why we are taking a backup at midnight.  The backups that go to the other bucket will be for longterm storage.

modules/mongodb/manifests/s3backup/backup.pp
```diff
@@ -46,20 +46,22 @@
...
+  include backup::client
+
   validate_re($private_gpg_key_fingerprint, '^[[:alnum:]]{40}$', 'Must supply full GPG fingerprint')
 
-  # create user
-  user { $user:
-    ensure     => 'present',
-    managehome => true,
+  user { 'govuk-backups':
+    ensure => absent,
```
surminus: If we can put a FIXME comment on top of this so we know to remove it when it's deployed to Production that'd be grand

modules/mongodb/manifests/s3backup/backup.pp
```diff
@@ -101,6 +110,10 @@
     require => User[$user],
   }
 
+  file { '/usr/local/bin/mongodb-backup-s3-wrapper':
+    ensure => absent,
```
surminus: FIXME, same as above

modules/mongodb/manifests/s3backup/cron.pp
```diff
@@ -5,13 +5,24 @@
...
-    user    => $mongodb::s3backup::backup::user,
+    user    => $user,
     minute  => '*/15',
-    require => [Class['mongodb::s3backup::package'],Class['mongodb::s3backup::backup']],
+  }
+
+  cron { 'mongodb-s3-night-backup':
+    command => '/usr/local/bin/mongodb-backup-s3',
+    user    => $user,
+    hour    => '0',
```
surminus: I can't find the answer in Puppet docs, but if minute isn't specified what does it default to? i.e. I want to make sure it doesn't run every minute at midnight

modules/mongodb/templates/mongodb-backup-s3.erb
```diff
@@ -24,19 +25,25 @@ trap nagios_passive EXIT
...
-(/bin/tar cvz $BACKUP_DIR | /usr/bin/gpg -e -o $BACKUP_DIR/$BACKUP_FILE -r $KEY_FINGERPRINT) &> /dev/null
+(/bin/tar cvz $BACKUP_DIR | /usr/bin/gpg -e -o $BACKUP_DIR/$BACKUP_FILE -r $KEY_FINGERPRINT)
 TIME="$(($(date +%s)-TIME))"
 printf "COMPRESSION AND ENCRYPTION FINISHED IN: ${TIME}s\n"
 
 # Upload backup to s3 bucket
 TIME="$(date +%s)"
-/usr/bin/envdir <%= @env_dir %>/env.d /usr/local/bin/s3cmd put $BACKUP_DIR/mongodump-* s3://$S3_BUCKET --quiet
+CURRTIME=$(date +%H%M)
+if [ $CURRTIME -ge 0000 -a $CURRTIME -le 0010 ]; then
```
surminus: I like the intent but not a fan of hardcoding the time into the code. It might be better to pass in an argument in the cron job, for example `/usr/local/bin/mongodb-backup-s3 daily`, and then do a conditional around the argument: `if [ $1 == "daily" ]; then ...`

modules/mongodb/manifests/s3backup/cron.pp
surminus: Thinking about it this also clashes with the other cron job because it's running the same command

modules/mongodb/manifests/s3backup/backup.pp
```diff
@@ -46,20 +46,22 @@
   $aws_secret_access_key = undef,
   $aws_region = 'eu-west-1',
   $backup_dir = '/var/lib/s3backup',
+  $backup_node = 'localhost',
   $cron = true,
   $env_dir = '/etc/mongo_s3backup',
   $private_gpg_key = undef,
   $private_gpg_key_fingerprint,
```
surminus: Should this be `undef` if `$private_gpg_key` is also `undef`?

modules/mongodb/templates/mongodb-backup-s3.erb
djsd123: Good suggestion. Done!

modules/mongodb/templates/mongodb-backup-s3.erb
```diff
@@ -24,19 +25,24 @@ trap nagios_passive EXIT
...
 TIME="$(date +%s)"
-(/bin/tar cvz $BACKUP_DIR | /usr/bin/gpg -e -o $BACKUP_DIR/$BACKUP_FILE -r $KEY_FINGERPRINT) &> /dev/null
+(/bin/tar cvz $BACKUP_DIR | /usr/bin/gpg -e -o $BACKUP_DIR/$BACKUP_FILE -r $KEY_FINGERPRINT)
 TIME="$(($(date +%s)-TIME))"
 printf "COMPRESSION AND ENCRYPTION FINISHED IN: ${TIME}s\n"
 
 # Upload backup to s3 bucket
 TIME="$(date +%s)"
-/usr/bin/envdir <%= @env_dir %>/env.d /usr/local/bin/s3cmd put $BACKUP_DIR/mongodump-* s3://$S3_BUCKET --quiet
+if [ $1 == "daily ]; then
```
surminus: This is missing another ending `"`

modules/mongodb/manifests/s3backup/backup.pp
```diff
@@ -46,11 +46,13 @@
...
   $s3_bucket  = 'govuk-mongodb-backup-s3',
+  $s3_bucket_daily = 'govuk-mongodb-backup-s3-daily',
```
surminus: Add environment suffix (bucket also needs creating).

surminus:
Please note that this will need to be rebased against master as we did some work to remove the errors currently in Integration.

modules/mongodb/manifests/s3backup/cron.pp
```diff
@@ -5,13 +5,24 @@
...
-    user    => $mongodb::s3backup::backup::user,
+    user    => $user,
     minute  => '*/15',
-    require => [Class['mongodb::s3backup::package'],Class['mongodb::s3backup::backup']],
+  }
+
+  cron { 'mongodb-s3-night-backup':
+    command => '/usr/local/bin/mongodb-backup-s3 daily',
+    user    => $user,
+    hour    => '0',
```
surminus: `minute => 0` still needs to be added. I'm also wondering whether we need to set some sort of lock as this will kick off two mongodump processes at the same time - will that work or will one error out? 

If you use `setlock -N /var/run/foo` for the daily backup that will wait until a lock is available, and `setlock -n` on the regular backup will just exit if it can't get the lock.

surminus:
Once the cron comment is addressed I think this is good to go and test in Integration :+1:

surminus:
Can we squash these commits into one?

djsd123:
I've squashed one, so now we have two commits. The 2nd commit represents work done post merge.

modules/mongodb/manifests/s3backup/cron.pp
```diff
@@ -5,13 +5,31 @@
...
+) {
+
+  include backup::client
+  require mongodb::s3backup::package
+  require mongodb::s3backup::backup
 
   cron { 'mongodb-s3backup':
-    command => '/usr/local/bin/mongodb-backup-s3',
-    user    => $mongodb::s3backup::backup::user,
+    command => '/usr/bin/setlock -n $backup_dir \
```
surminus: I think you need to specify a file for setlock rather than a directory.

surminus:
Mob review addressed a couple of comments. Ensure the terraform stuff is done and this is ready for merge.

surminus:
:+1: Merge away! 🎉

PR #4492, opened 2016-06-14
Elasticsearch backup to s3

djsd123:
Cron runs a bash script to push indexes to s3.  Further indexes can be added in hiera.


djsd123:
Could not get tests to run so cerated a fresh branch elasticsearch-index-backup

PR #4495, opened 2016-06-14
Set disk layout for asset-slave-1 in integration

boffbowsh:
Not gonna lie, I'm a bit bored of these PRs now.

The disk layout has changed recently, so pv is trying
to make the wrong disks members of the volume groups.


rjw1:
merge when green

PR #4496, opened 2016-06-14
Add timestamps to tagging-sync-check build

benhyland:
A day may come when this sort of thing is templated into all jobs.
But it is not this day.

See https://github.com/alphagov/govuk-puppet/pull/4472


tijmenb:
👍  when green

PR #4498, opened 2016-06-15
Fix number to string for whitehall port

alexmuller:


deanwilson:
This feels very wrong. Ports are not strings.

But I understand your pain so merge away and we can deal with it if we move to Puppet 4.

PR #4499, opened 2016-06-15
`/etc/cron.d` isn't a managed resource so require the cron class intead

deanwilson:
We should also ensure our cron package is installed if we're going
to manage the service.


djsd123:
Looks spot on to me

PR #4500, opened 2016-06-15
Elasticsearch index backup

djsd123:
The script backs up and uploads the specified indices.  At the moment I have the named the elasticsearch repo to kibana-dashboards but we could name it something more generic in case we wanted to use this script to backup other indexes on other hosts in future. The fact that a specified index may not exist will not cause the command to fail and it should just continue.

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -39,6 +39,10 @@ govuk_cdnlogs::monitoring_enabled: false
 govuk_crawler::seed_enable: true
 
 govuk_elasticsearch::dump::run_es_dump_hour: '9'
+govuk_elasticsearch::backup::s3_bucket: 'kibana-dashboards'
+govuk_elasticsearch::backup::es_repo: 'kibana-dashboards'
+govuk_elasticsearch::backup::es_indices:
+  - kibana_int
 
 govuk_jenkins::config::banner_colour_background: '#ffbf47'
 govuk_jenkins::config::banner_colour_text: 'black'
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -39,10 +39,6 @@ govuk_cdnlogs::monitoring_enabled: false
 govuk_crawler::seed_enable: true
 
 govuk_elasticsearch::dump::run_es_dump_hour: '9'
-govuk_elasticsearch::backup::s3_bucket: 'kibana-dashboards'
-govuk_elasticsearch::backup::es_repo: 'kibana-dashboards'
-govuk_elasticsearch::backup::es_indices:
-  - kibana_int
 
 govuk_jenkins::config::banner_colour_background: '#ffbf47'
 govuk_jenkins::config::banner_colour_text: 'black'
--- a/modules/govuk_elasticsearch/manifests/backup.pp
+++ b/modules/govuk_elasticsearch/manifests/backup.pp
@@ -40,14 +40,14 @@ class govuk_elasticsearch::backup(
   $env_dir = '/etc/es_s3backup',
   $user = backup::client::govuk-backup,
   $es_repo = undef,
-  $enabled = fasle,
+  $housekeeping_enabled = fasle,
   $es_indices = [
     join(regsubst($govuk_elasticsearch::backup::es_indices, '.*', '"\1"'), ',')]
   ){
 
   include backup::client
 
-  if $enabled {
+  if $housekeeping_enabled {
     include govuk_elasticsearch::housekeeping
   }
 
--- a/modules/govuk_elasticsearch/manifests/backup.pp
+++ b/modules/govuk_elasticsearch/manifests/backup.pp
@@ -1,0 +1,92 @@
+# == Class: govuk_elasticsearch::backup
+#
+# GOV.UK specific class to backup elasticsearch indexes.
+#
+# === Parameters
+#
+# [*aws_access_key_id*]
+#
+# [*aws_secret_access_key*]
+#
+# [*aws_region*]
+#   AWS region for the S3 bucket
+#
+# [*env_dir*]
+#   Defines directory for the environment
+#   variables
+#
+# [*s3_bucket*]
+#   Defines the AWS S3 bucket where the backups
+#   will be uploaded. It should be created by the
+#   user
+#
+# [*user*]
+#   Defines the system user that will be created
+#   to run the backups
+#
+# [*es_repo*]
+#   The repository within elasticsearch
+#   where backups will be stored
+#
+# [*es_indices*]
+#   Elaticsearch indexes
+#
+class govuk_elasticsearch::backup(
+  $aws_access_key_id = undef,
+  $aws_secret_access_key = undef,
+  $aws_region = 'eu-west-1',
+  $s3_bucket = undef,
+  $cron = true,
+  $env_dir = '/etc/es_s3backup',
+  $user = backup::client::govuk-backup,
+  $es_repo = undef,
+  $es_indices = [
+    join(regsubst($govuk_elasticsearch::backup::es_indices, '.*', '"\1"'), ',')]
+  ){
+
+  include backup::client
+
+  # push env files
+  file { [$env_dir,"${env_dir}/env.d"]:
+    ensure => directory,
+    owner  => $user,
+    group  => $user,
+    mode   => '0770',
+  }
+
+  file { "${env_dir}/env.d/AWS_SECRET_ACCESS_KEY":
+    content => $aws_secret_access_key,
+    owner   => $user,
+    group   => $user,
+    mode    => '0640',
+  }
+
+  file { "${env_dir}/env.d/AWS_ACCESS_KEY_ID":
+    content => $aws_access_key_id,
+    owner   => $user,
+    group   => $user,
+    mode    => '0640',
+  }
+
+  file { "${env_dir}/env.d/AWS_REGION":
+    content => $aws_region,
+    owner   => $user,
+    group   => $user,
+    mode    => '0640',
+  }
+
+  file { 'es-backup-s3':
+    path    => '/usr/local/bin',
+    owner   => $user,
+    group   => $user,
+    mode    => '0550',
+    content => template('govuk_elasticsearch/es-backup-s3.erb'),
+  }
+
+  cron { 's3-elasticsearch-indexes':
+    command => '/usr/local/bin/es-backup-s3',
+    user    => $user,
+    hour    => 0,
+  }
+
+}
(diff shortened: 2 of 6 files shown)


tijmenb:
@MatMoore can we use this ^ for search indices?

modules/govuk_elasticsearch/spec/classes/govuk_elasticsearch_spec.rb
```diff
@@ -23,11 +23,11 @@
 
   describe '#manage_repo' do
     let(:params) {{
-      :version => '1.7.0',
+      :version => '1.4.3',
```
rboulton: Why is this needed?  Will this downgrade our cluster to using elasticsearch 1.4.3?

MatMoore:
@tijmenb in theory yes, this PR is doing is the same thing as our rummager code.

For more info about what we did, see the [rummager implentation](https://github.com/alphagov/rummager/tree/master/lib/snapshot) and the [previous PR against puppet](https://github.com/alphagov/govuk-puppet/pull/4196)

The complications we ran into were around pruning old snapshots, and ensuring that multiple snapshot operations (create and delete) don't run concurrently.

If you don't prune the snapshots, then I think the time taken degrades over time. Some discussion of this here: https://groups.google.com/forum/#!msg/elasticsearch/cXl9UrqgwBI/et56xPUqwGkJ

The things that will differ for rummager:

- There is currently nginx server between rummager and the elasticsearch cluster - I think this introduces some issues with request timeouts, since the api requests can take a long time to complete
- The code doesn't exist yet, but if we add tools for restoring snapshots, the name of the index matters even if its aliased

Will the configuration of the ES repository be handled by puppet as well?

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -0,0 +1,41 @@
...
+# indices: option allows us to specify indexes in a comma separated list rather than snap-shooting everything. However we can set
+# the $ES_REPO variable to '_all' to snapshot all indices.
+#
+# ignore_unavailable: option when true will cause indices that do not exist to be ignored during snapshot
+#
+# include_global_state: option will prevent the cluster global state from being stored as part of the snapshot.
+# The entire snapshot will fail if one or more indices participating in the snapshot don’t have all primary shards available
+
+
+/usr/bin/envdir <%= @env_dir %>/env.d /usr/bin/curl --connect-timeout 10 -sS -XPUT "http://127.0.0.1:9200/_snapshot/"${ES_REPO}"?wait_for_completion=true&pretty" -d '{
```
MatMoore: Is this conflating the "create repository" call with the "create snapshot" call? I think this just creates the initial repository if it doesn't exist.

I expected the snapshot one to look like `/_snapshot/$ES_REPO/$SNAPSHOT_NAME`

https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html#_snapshot

modules/govuk_elasticsearch/spec/classes/govuk_elasticsearch_spec.rb
deanwilson: These are test values for rspec and are only used when running spec tests, they are not visible, and will not be used, on actual puppet runs.

modules/govuk_elasticsearch/templates/es-backup-s3.erb
djsd123: Hi Matt,
Yes I somehow left out the command to create the snapshot.  Thanks for the heads up.

I think we need to talk about the best way to implement some sort of housekeeping job to ensure we don't experience slow down when the repo gets too large.  I was told this was specifically for kibana dashboard indices but hoped it could be used more widely.

modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -0,0 +1,92 @@
+# == Class: govuk_elasticsearch::backup
+#
+# GOV.UK specific class to backup elasticsearch indexes.
+#
+# === Parameters
+#
+# [*aws_access_key_id*]
```
surminus: Empty description.

modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -0,0 +1,92 @@
...
+    owner   => $user,
+    group   => $user,
+    mode    => '0550',
+    content => template('govuk_elasticsearch/es-backup-s3.erb'),
+  }
+
+  cron { 's3-elasticsearch-indexes':
+    command => '/usr/local/bin/es-backup-s3',
+    user    => $user,
+    hour    => 0,
```
surminus: Need to specify minute.

surminus:
I think the code looks good.

modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -40,12 +40,17 @@
   $env_dir = '/etc/es_s3backup',
   $user = backup::client::govuk-backup,
   $es_repo = undef,
+  $enabled = fasle,
```
MatMoore: you have a typo here

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -0,0 +1,55 @@
...
+  "bucket": "'"${S3_BUCKET}"'",
+  "region": "'"${AWS_REGION}"'",
+  "access_key": "'"${AWS_ACCESS_KEY_ID}"'",
+  "secret_key": "'"${AWS_SECRET_ACCESS_KEY}"'"
+  },
+  "compress": "true"
+  }'
+
+
+# indices: option allows us to specify indexes in a comma separated list rather than snap-shooting everything. However we can set
```
MatMoore: this should be "snapshotting" rather than "snap-shooting"

rjbate: "snap-shooting" could be the opposite of "snapshotting"

hieradata/integration.yaml
```diff
@@ -39,6 +39,10 @@ govuk_cdnlogs::monitoring_enabled: false
 govuk_crawler::seed_enable: true
 
 govuk_elasticsearch::dump::run_es_dump_hour: '9'
+govuk_elasticsearch::backup::s3_bucket: 'kibana-dashboards'
```
MatMoore: on naming: maybe we could name these things after the cluster itself, rather than the index?

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -0,0 +1,55 @@
...
+# ignore_unavailable: option when true will cause indices that do not exist to be ignored during snapshot
+#
+# include_global_state: option will prevent the cluster global state from being stored as part of the snapshot.
+# The entire snapshot will fail if one or more indices participating in the snapshot don’t have all primary shards available
+
+if [ $? -eq 0 ]; then
+
+  /usr/bin/envdir <%= @env_dir %>/env.d /usr/bin/curl --connect-timeout 10 -sS -XPUT "http://127.0.0.1:9200/_snapshot/"${ES_REPO}"/"${ES_SNAPSHOT}"?wait_for_completion=true&pretty" -d '{
+    "type": "s3",
+    "settings": {
```
MatMoore: I think you can take `type` and `settings` out of this request - the ES repo is configured in the previous one.

MatMoore:
Will the cron job have a passive icinga alert associated with it?

djsd123:
Yes, I can add the passive check.

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -0,0 +1,55 @@
...
+  "secret_key": "'"${AWS_SECRET_ACCESS_KEY}"'"
+  },
+  "compress": "true"
+  }'
+
+
+# indices: option allows us to specify indexes in a comma separated list rather than snap-shooting everything. However we can set
+# the $ES_REPO variable to '_all' to snapshot all indices.
+#
+# ignore_unavailable: option when true will cause indices that do not exist to be ignored during snapshot
```
djsd123: Agreed.  When I added it I was unsure of how and where the script would be invoked.

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+set -e
+
+# Redirect stdout and stderr to syslog
+exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
+
+# The default Icinga passive alert assumes that the script failed
+NAGIOS_CODE=2
+NAGIOS_MESSAGE="CRITICAL: Mongodb backup push to S3 failed"
```
alexmuller: There are 2 references to MongoDB in this file which should probably be Elasticsearch.

djsd123: Sorry. Sloppy copy&paste.  Fixed.

modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -0,0 +1,112 @@
...
+    file { 'es-backup-s3':
+      path    => '/usr/local/bin',
+      owner   => $user,
+      group   => $user,
+      mode    => '0550',
+      content => template('govuk_elasticsearch/es-backup-s3.erb'),
+    }
+
+  # Remove cron job if backup is not enabled
+    if $enabled {
```
surminus: Wrap this in the entire class

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -0,0 +1,71 @@
...
+
+# Triggered whenever this script exits, successful or otherwise. The values
+# of CODE/MESSAGE will be taken from that point in time.
+function nagios_passive () {
+printf "<%= @ipaddress %>\t<%= @service_desc %>\t${NAGIOS_CODE}\t${NAGIOS_MESSAGE}\n" | /usr/sbin/send_nsca -H alert.cluster >/dev/null
+}
+
+trap nagios_passive EXIT
+
+/usr/bin/envdir <%= @env_dir %>/env.d  /usr/bin/curl --connect-timeout 10 -sS -XPUT "http://127.0.0.1:9200/_snapshot/"${ES_REPO}"?wait_for_completion=true&pretty" -d '{
```
surminus: Can we test if this would work just in a single string?

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -0,0 +1,71 @@
...
+else
+STATUS=1
+fi
+
+if [ $STATUS -eq 0 ]; then
+NAGIOS_CODE=0
+NAGIOS_MESSAGE="OK: Elasticsearch backup push to S3 succeeded"
+fi
+
+exit $STATUS
```
surminus: Refactor status

surminus:
Couple of comments but other than that it looks OK. Probably merge when ready and test in Integration.

djsd123:
Retiring this one in favour of this https://github.com/alphagov/govuk-puppet/pull/4637

PR #4502, opened 2016-06-15
Enable message queue listener on integration

tijmenb:
This enables the message queue listener for rummager in integration.
 
Full story: https://trello.com/c/0FKSqAkK


benhyland:
This is fine.

PR #4505, opened 2016-06-16
Improve rummager dashboard some more

tijmenb:
## Before

screen shot 2016-06-16 at 13 31 22

## After

screen shot 2016-06-16 at 13 30 52


mobaig:
👍

PR #4508, opened 2016-06-16
Fix Wal-e catalog error

brenetic:
- `wal-e_env_sync` script requires the `govuk_postgresql::wal_e::package` in order to run a `wal-e` command


alexmuller:
:+1:

PR #4509, opened 2016-06-17
Fix creation of wal-e env sync directory

alexmuller:
The parent directory doesn't exist so Puppet refuses to run.


deanwilson:
Merge away.

PR #4511, opened 2016-06-17
Mailto must be set

deanwilson:


rjw1:
shouldn't the template not add the MAILTO line if its empty?

modules/cron/manifests/crondotdee.pp
```diff
@@ -44,6 +44,11 @@
   $user = 'root',
   $mailto = undef,
 ) {
+
+  if $mailto == false {
```
alexmuller: I wouldn't think this code would work as I'd expect, is it right?

```
irb(main):002:0> mailto = ''
irb(main):009:0> !!mailto
=> true
```

rjw1:
lgtm :+1: mwg

PR #4513, opened 2016-06-17
Sum sidekiq worker activity for rummager graph

benhyland:
We were looking at the maximum number of messages processed
in any 1min bucket of samples - what we want to see is the sum
of messages processed in each bucket.

@benhyland @davidslv


tijmenb:
Yes 👍

PR #4515, opened 2016-06-17
Improve graphs of rummager queue stats

benhyland:
1. Have the sidekiq activity graph display only AmendWorker activity.
This is nicer for comparing the consumer activity with sidekiq activity,
since the consumer only ever uses the AmendWorker.

2. Avoid `sumSeries()`. This function is unnecessary since we only need to
sum into buckets, not sum multiple `seriesList`s. Note that `summarize()`
is already capable of summing multiple `series`, and a `series` is not
the same as a `seriesList`. `sumSeries()` has the unfortunate side effect
of averaging `series` where the sample rates are different, which in our
case is always. It doesn't make sense to show that sideqick processed a
non-integer fraction of a message in some bucket.

3. Use `stat_counts` instead of `stats`. statsd emits values for all stats under
both top-level keys. For counter stats, `stats` contains per-second rate values,
while `stat_counts` contains bucketed integer counts, which is what we want.

4. Add 'Retry Set' to the sidekiq queue size graph - this is where jobs go to wait
for the backoff period to expire when they are retried.

We have some more detailed graphs which will be put on another dashboard,
since this one is supposed to be for a big-screen info radiator.
Unfortunately I can't add a link to the detailed view from here since Grafana
doesn't know how to link to file-based dashboards.

screen shot 2016-06-27 at 18 31 55

### DIFF
--- a/modules/grafana/files/dashboards/rummager_queues.json
+++ b/modules/grafana/files/dashboards/rummager_queues.json
@@ -52,8 +52,8 @@
           "leftYAxisLabel": "Messages per minute"
         },
         {
-          "id": 5,
-          "title": "Sidekiq worker activity (search indexing)",
+          "id": 2,
+          "title": "Sidekiq AmendWorker activity (search indexing)",
           "span": 6,
           "type": "graph",
           "lines": false,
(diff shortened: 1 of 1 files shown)


tijmenb:
Perhaps we could make a detailed dashboard that includes this? I can see why it would be useful, but maybe it's not a thing we should be concerned about day-to-day?

modules/grafana/files/dashboards/rummager_queues.json
```diff
@@ -75,22 +75,22 @@
...
     },
     {
-      "height": "400px",
+      "height": "200px",
       "editable": true,
       "collapse": false,
       "panels": [
         {
-          "title": "Publishing queue consumer queue size (RabbitMQ)",
+          "title": "Publishing queue depth (RabbitMQ)",
```
tijmenb: I think "queue size" is a more descriptive name understood by most developers.

Also, apparently RabbitMQ defines queue depth as:

> Sum of ready and unacknowledged messages (queue depth).

https://www.rabbitmq.com/man/rabbitmqctl.1.man.html

benhyland: I think queue size is more typically used to refer to the total bytes used by a queue.
It's true that the docs refer to depth as the sum of ready and acknowledged, but in practice the unack'd are in a different data structure in rabbit.

benhyland:
@tijmenb I want this displayed for a bit because we still don't understand the discrepancy between rabbit activity and sidekiq activity. Once we are happy that it is only due to docs not present in rummager we could remove the graph again?

Davidslv:
Can we merge this, or there's still some concerns?

benhyland:
@Davidslv I think @tijmenb was worried about graph names.

I think this change is a step in the right direction but we need to keep looking at these graphs as evenwith this applied there are still unexplained inconsistencies. I can explain in person if you like.

tijmenb:
Let's talk offline about what problem this PR is trying to solve, not sure if I have all the context.

tijmenb:
👍 thanks, good to merge!

PR #4516, opened 2016-06-20
Move Redis docs

alexmuller:


deanwilson:
Merge away. I do wonder if we should make it take a full url and change all the old calls to use that. As a non-intrusive change this works though.

PR #4517, opened 2016-06-20
Elasticsearch snapshot housekeeping

djsd123:
This commit adds a shell script to remove old snapshots from the local repository in elasticsearch.  Doing this will hopefully prevent degraded performance as a result of having a backup solution.

### DIFF
--- a/modules/govuk_elasticsearch/manifests/housekeeping.pp
+++ b/modules/govuk_elasticsearch/manifests/housekeeping.pp
@@ -1,0 +1,58 @@
+# == Class: govuk_elasticsearch::housekeeping
+#
+#  Delete old snapshots from elasticsearch
+#
+#
+# === Parameters:
+#
+# [*es_repos*]
+#   The local elasticsearch repositories where snapshots are stored
+#
+# [*es_snapshot_limit*]
+#   The number of snapshots we want to keep
+#
+# [*user*]
+#   User that invokes backup
+#
+# === Variables
+#
+# [*$repositories*]
+#   A string array of Elasticsearch repositories
+#
+class govuk_elasticsearch::housekeeping(
+  $es_repos,
+  $es_snapshot_limit = 5,
+  $user = 'govuk-backup',
+  ) {
+
+  if $es_repos == undef {
+    fail('No Elasticsearch repositories were set')
+  } else {
+    # Validates a maximum and minimum string length
+    validate_slength($es_repos, 120, 3)
+    $repositories = join(regsubst($es_repos,'(.*)','"\1"'), ' ')
+  }
+
+    # This is a CLI JSON processor which allows us to "awk" the data on the fly
+    package { 'jq':
+      ensure => installed,
+    }
+
+    file { 'es-prune-snapshots':
+      ensure  => file,
+      path    => '/usr/local/bin/es-prune-snapshots',
+      content => template('govuk_elasticsearch/es-prune-snapshots.erb'),
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0755',
+    }
+
+    cron { 'elasticsearch-remove-old-snapshots':
+      ensure  => present,
+      command => '/usr/local/bin/es-prune-snapshots',
+      user    => $user,
+      weekday => 3,
+      hour    => 4,
+      minute  => 0,
+    }
+}
--- a/modules/govuk_elasticsearch/spec/classes/govuk_elasticsearch_housekeeping_spec.rb
+++ b/modules/govuk_elasticsearch/spec/classes/govuk_elasticsearch_housekeeping_spec.rb
@@ -1,0 +1,19 @@
+require_relative '../../../../spec_helper'
+
+  describe 'govuk_elasticsearch::housekeeping', :type => :class do
+
+    let(:params) {{
+        :es_repos => [
+            'kibana-lint',
+            'kibana-mint',
+            'kibana_tint'
+        ]
+    }}
+
+    context 'Checking what repositories variable expands like' do
+
+      it { is_expected.to contain_file('es-prune-snapshots').with_path('/usr/local/bin/es-prune-snapshots').with_content(/"kibana-lint"\s"kibana-mint"\s"kibana_tint"/) }
+    end
+
+
+  end
--- a/modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
+++ b/modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -e
+
+# Redirect stdout and stderr to syslog
+exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
+
+ES_REPOS=(<%= @repositories %>)
+LIMIT=<%= @es_snapshot_limit %>
+
+for i in ${ES_REPOS[@]}; do
+  # This gets snapshots for the repo but leaves behind 'n' youngest snapshots, 'n' being the LIMIT set above
+  SNAPSHOTS=$(/usr/bin/curl --connect-timeout 10 -sS -XGET "127.0.0.1:9200/_snapshot/${ES_REPOS[$i]}/_all" | /usr/bin/jq --raw-output ".snapshots[:-${LIMIT}][].snapshot")
+
+  for SNAPSHOT in $SNAPSHOTS; do
+    /usr/bin/curl --connect-timeout 10 -s -XDELETE "127.0.0.1:9200/_snapshot/${ES_REPOS}/${SNAPSHOT}?pretty"
+  done
+done
+


modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -0,0 +1,43 @@
...
+#
+# === Parameters:
+#
+# [*es_repo*]
+#   The local elasticsearch repository here snapshots are stored
+#
+# [*es_snapshot_limit*]
+#   The number of snapshots we want to keep
+#
+# {*user*]
```
deanwilson: This should be a `[`

modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -0,0 +1,43 @@
+# == Class: govuk_elasticsearch::housekeeping
+#
+#  Delete old snapshots from elasticsearch
+#
+#
+# === Parameters:
+#
+# [*es_repo*]
+#   The local elasticsearch repository here snapshots are stored
```
deanwilson: "where"

modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -0,0 +1,43 @@
...
+
+
+  # This is a CLI JSON processor which allows us to "awk" the data on the fly
+  package { 'jq':
+    ensure => installed,
+  }
+
+  file { 'es-prune-snapshots':
+    ensure  => file,
+    path    => 'usr/local/bin/es-remove-old-snapshots',
```
deanwilson: Missing a `/` in the path. Also should it be prune to match everything else?

modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
```diff
@@ -0,0 +1,37 @@
...
+if [ $? == 0 ]
+then
+  STATUS=0
+else
+  STATUS=1
+fi
+
+if [ $STATUS -eq 0 ]; then
+  NAGIOS_CODE=0
+  NAGIOS_MESSAGE="OK: Mongodb backup push to S3 succeeded"
```
deanwilson: Where do these values get used?

modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -0,0 +1,43 @@
...
+    content => template('govuk_elasticsearch/es-prune-snapshots.erb'),
+    owner   => $user,
+    group   => $user,
+    mode    => '0550',
+  }
+
+  cron { 'elasticsearch-remove-old-snapshots':
+    command => '/usr/local/bin/es-prune-snapshots',
+    user    => $user,
+    weekday => 3,
```
MatMoore: This makes sense in general, but it's another area where rummager elasticsearch might be different. 😞 

We rebuild the indexes every day so we effectively lose the benefits of incremental snapshots if we keep them around longer than a day.

Maybe we could parameterise this?

modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
```diff
@@ -0,0 +1,30 @@
...
+
+# Redirect stdout and stderr to syslog
+exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
+
+ES_REPO=<%= @es_repo %>
+LIMIT=<%= @es_snapshot_limit %>
+
+# Get a list of snapshots in the specified repo
+
+SNAPSHOTS=`/usr/bin/curl -s -XGET "127.0.0.1:9200/_snapshot/"${ES_REPO}"/_all" | /usr/bin/jq -r ".snapshots[:-${LIMIT}][].snapshot"`
```
MatMoore: Not sure if it helps, but there is a set of APIs that return data in tabular format: https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-snapshots.html

Would that be easier than using jq?

modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
```diff
@@ -0,0 +1,30 @@
...
+
+
+# Loop through results deleting snapshots we do not want.  Note that "deleting" snapshots via the api will ensure
+# that only the oldest snapshots are deleted and that snapshots aren't left orphaned.
+
+for SNAPSHOT in $SNAPSHOTS; do
+
+    /usr/bin/curl -s -XDELETE "127.0.0.1:9200/_snapshot/"${ES_REPO}"/"${SNAPSHOT}"?pretty"
```
MatMoore: Should this have an explicit timeout?

MatMoore:
Added a few comments in line, but generally looks good.

What hosts will this run on? It'll need to only run on one node in each cluster, or they'll conflict with each other (the others will get a 500 response I think).

modules/govuk_elasticsearch/manifests/housekeeping.pp
djsd123: Actually parameterising makes a lot of sense.  

How frequently should this job run to be of any use to you?

modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
djsd123: Thanks, i'll have a look.

modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
djsd123: Yes it's good practice when running unattended to do so.  Will fix.

djsd123:
Originally it was intended to accompany the backup of kibana dashboards but it would be great if we could extend it's use case.

I will have to confer with colleagues to see if there is a way we could tie this job down to a single host per cluster.

modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
djsd123: After a quick parse It seems version 1.4.x doesn't support cat snapshots.  I'll look a little more just in case it's referenced differently.

surminus:
We can sit down and have a chat about which machine to schedule this on.

modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
```diff
@@ -0,0 +1,30 @@
...
+
+done
+
+if [ $? == 0 ]
+then
+  STATUS=0
+else
+  STATUS=1
+fi
```
surminus: If there's no passive check set up remove this bit

modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -0,0 +1,57 @@
...
+#   The number of snapshots we want to keep
+#
+# [*user*]
+#   User that invokes backup
+#
+# [*enabled*]
+#   Determines whether this task will run
+#
+class govuk_elasticsearch::housekeeping(
+  $es_repo = undef,
```
surminus: Validate string?

modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -0,0 +1,57 @@
...
+    path    => '/usr/local/bin/es-remove-old-snapshots',
+    content => template('govuk_elasticsearch/es-prune-snapshots.erb'),
+    owner   => $user,
+    group   => $user,
+    mode    => '0550',
+  }
+
+  # Remove cron job if $enabled set to false.  We could create a new variable
+  # and run housekeeping on a seperate node to the backup node
+  if $enabled {
```
surminus: This should be made redundant by being included in the backup conditional

modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -0,0 +1,57 @@
...
+    ensure => installed,
+  }
+
+  file { 'es-prune-snapshots':
+    ensure  => file,
+    path    => '/usr/local/bin/es-remove-old-snapshots',
+    content => template('govuk_elasticsearch/es-prune-snapshots.erb'),
+    owner   => $user,
+    group   => $user,
+    mode    => '0550',
```
surminus: Just own this by root and set perms to 0755

modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  # Remove cron job if $enabled set to false.  We could create a new variable
+  # and run housekeeping on a seperate node to the backup node
+  if $enabled {
+    cron { 'elasticsearch-remove-old-snapshots':
+      ensure  => present,
+      command => '/usr/local/bin/es-prune-snapshots',
+      user    => $user,
+      weekday => 3,
```
surminus: How often should this run? Specify a minute and hour

modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
```diff
@@ -0,0 +1,24 @@
...
+
+# Redirect stdout and stderr to syslog
+exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
+
+ES_REPO=<%= @es_repo %>
+LIMIT=<%= @es_snapshot_limit %>
+
+# Get a list of snapshots in the specified repo
+
+SNAPSHOTS=`/usr/bin/curl --connect-timeout 10 -s -XGET "127.0.0.1:9200/_snapshot/"${ES_REPO}"/_all" | /usr/bin/jq -r ".snapshots[:-${LIMIT}][].snapshot"`
```
surminus: Check the string thing

modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
```diff
@@ -0,0 +1,24 @@
...
+
+
+# Loop through results deleting snapshots we do not want.  Note that "deleting" snapshots via the api will ensure
+# that only the oldest snapshots are deleted and that snapshots aren't left orphaned.
+
+for SNAPSHOT in $SNAPSHOTS; do
+
+    /usr/bin/curl --connect-timeout 10 -s -XDELETE "127.0.0.1:9200/_snapshot/"${ES_REPO}"/"${SNAPSHOT}"?pretty"
```
surminus: Check the string thing

modules/govuk_elasticsearch/manifests/housekeeping.pp
djsd123: add hour minute.

modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -0,0 +1,64 @@
...
+      ensure => installed,
+    }
+
+    file { 'es-prune-snapshots':
+      ensure  => file,
+      path    => '/usr/local/bin/es-prune-snapshots',
+      content => template('govuk_elasticsearch/es-prune-snapshots.erb'),
+      owner   => $user,
+      group   => $user,
+      mode    => '0550',
```
surminus: These permissions could be owned by root and `0755` as per convention in /usr/local/bin, but again it's nitpicky :)

modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -0,0 +1,64 @@
...
+  $es_snapshot_limit = 5,
+  $user = 'govuk-backup',
+  $enabled = false
+) {
+
+  if $es_repos == undef {
+    fail('No Elasticsearch repositories were set')
+  } else {
+    # Validates a maximum and minimum string length
+    validate_slength($es_repos, 120, 3)
```
surminus: Curious as to why we do this. Sounds like a good thing, just interested.

djsd123: We used this method to validate/ensure we have a string or an array of strings. The rest is non essential but nice to have.

modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
```diff
@@ -0,0 +1,19 @@
...
+
+set -e
+
+# Redirect stdout and stderr to syslog
+exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
+
+ES_REPOS=(<%= @repositories %>)
+LIMIT=<%= @es_snapshot_limit %>
+
+for ((i=0; i < ${#ES_REPOS[@]}; i++)); do
```
deanwilson: You can probably simplify the loop with `for repo in ${ES_REPOS[@]}; do` and then use $repo later.

modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -0,0 +1,64 @@
...
+    $repositories = join(regsubst($es_repos,'(.*)','"\1"'), ' ')
+  }
+
+  if $enabled {
```
deanwilson: Can we move this gating to the calling class please and remove this conditional?

modules/govuk_elasticsearch/spec/classes/govuk_elasticsearch_housekeeping_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rspec'
```
deanwilson: You won't need this require. It'll come in via the spec_helper.spec file.

modules/govuk_elasticsearch/spec/classes/govuk_elasticsearch_housekeeping_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+require_relative '../../../../spec_helper'
+require 'rspec-puppet'
```
deanwilson: You won't need this require. It'll come in via the spec_helper.spec file.

modules/govuk_elasticsearch/spec/classes/govuk_elasticsearch_housekeeping_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+        :es_repos => [
+            'kibana-lint',
+            'kibana-mint',
+            'kibana_tint'
+        ]
+    }}
+
+    context 'Checking what repositories variable expands like' do
+
+      it { is_expected.to contain_file('es-prune-snapshots').with_path('/usr/local/bin/es-prune-snapshots').with_content(/("kibana-lint" "kibana-mint" "kibana_tint")/) }
```
deanwilson: You can remove the `/(    )/` as we're not using the captured value.

modules/govuk_elasticsearch/templates/es-prune-snapshots.erb
djsd123: @deanwilson Not sure I understand.  Could you show me later?

djsd123: @deanwilson Absolutely.  It looks like I am trying to code java or something.

modules/govuk_elasticsearch/spec/classes/govuk_elasticsearch_housekeeping_spec.rb
djsd123: Made the changes.  I had to keep the `/ /` because rspec doesn't respect word boundaries.

deanwilson:
Merge away. Let's get this in to integration.

PR #4518, opened 2016-06-20
Turn on rummager message queue listener in development

tijmenb:
Enabling the publishing listener tells puppet to create a RabbitMQ user
and queue for rummager. Without this it won't be possible to start
rummager after we've merged
https://github.gds/gds/development/pull/324.


deanwilson:
Merge away

PR #4519, opened 2016-06-20
Add watermark check script from https://github.com/nagios-plugins-rab…

rjbate:
…bitmq for rabbitmq memory and disk usage


surminus:
I think you still need to add the NRPE config, for example the puppet like [here](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_rabbitmq/manifests/monitoring.pp#L9-L12) and the template like [here](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_rabbitmq/templates/check_rabbitmq_network_partition.cfg.erb).

You also need to add the Puppet for actually creating the Icinga plugin on the server itself.

surminus:
I think this looks good :+1: merge when ready

PR #4521, opened 2016-06-20
Pin specific packages of wal-e dependencies

alexmuller:
We don't always want to upgrade these versions to whatever happens to be on PyPI, which is what `latest` will do.

Instead I've pinned the packages to the versions we currently have installed.


deanwilson:
Merge away.

PR #4522, opened 2016-06-21
Mongodb s3 Restore2

djsd123:
Restore script for mongodb s3 backups. It's still yet to be decided on how it will be invoked.

### DIFF
--- a/modules/mongodb/manifests/s3backup/restore.pp
+++ b/modules/mongodb/manifests/s3backup/restore.pp
@@ -5,8 +5,10 @@
 # === Parameters:
 #
 # [*aws_access_key_id*]
+#   Key used to sign programmatic requests in AWS
 #
 # [*aws_secret_access_key*]
+#   Key used to sign programmatic requests in AWS
 #
 # [*backup_dir*]
 #   Defines the directory to restore the backups
@@ -37,9 +39,9 @@ class mongodb::s3backup::restore(
   $aws_access_key_id = undef,
   $aws_secret_access_key = undef,
   $env_dir = '/etc/mongo_s3backup',
-  $s3_bucket  = 'govuk-mongodb-backup-s3',
+  $s3_bucket  = undef,
   $backup_dir = '/var/lib/s3backup',
-  $user = mongodb::s3backup::backup::user,
+  $user = 'govuk-backup',
   $cron = false
 ){
   include backup::client
--- a/modules/mongodb/manifests/s3backup/restore.pp
+++ b/modules/mongodb/manifests/s3backup/restore.pp
@@ -1,0 +1,55 @@
+# == Class: mongodb::s3backup::restore
+#
+# Restore a MongoDB backup to a server from s3
+#
+# === Parameters:
+#
+# [*aws_access_key_id*]
+#
+# [*aws_secret_access_key*]
+#
+# [*backup_dir*]
+#   Defines the directory to restore the backups
+#
+# [*env_dir*]
+#   Defines directory for the environment
+#   variables
+#
+# [*private_gpg_key*]
+#   Defines the ascii exported private gpg to
+#   use for decrypting backups. This key should
+#   be created by the user and encrypted with eyaml
+#
+# [*private_gpg_key_fingerprint*]
+#   Defines the fingerprint of the gpg private
+#   key to dencrypt the backups. The fingerprint
+#   should be 40 characters without spaces
+#
+# [*s3_bucket*]
+#   Defines the AWS S3 bucket where the backups
+#   will be downloaded from. It should be created by the
+#   user
+#
+# [*cron*]
+#   Defines whether to enable the cron job. Value
+#   should be true or false
+class mongodb::s3backup::restore(
+  $aws_access_key_id = undef,
+  $aws_secret_access_key = undef,
+  $env_dir = '/etc/mongo_s3backup',
+  $s3_bucket  = 'govuk-mongodb-backup-s3',
+  $backup_dir = '/var/lib/s3backup',
+  $user = mongodb::s3backup::backup::user,
+  $cron = false
+){
+  include backup::client
+
+  file { '/usr/local/bin/mongodb-restore-s3':
+    ensure  => file,
+    content => template('mongodb/mongodb-restore-s3.erb'),
+    mode    => '0770',
+    owner   => $user,
+    group   => $user,
+  }
+
+}
--- a/modules/mongodb/templates/mongodb-restore-s3.erb
+++ b/modules/mongodb/templates/mongodb-restore-s3.erb
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+set -e
+
+# Redirect stdout and stderr to syslog
+exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
+
+S3_BUCKET=<%= @s3_bucket %>
+BACKUP_DIR=<%= @backup_dir %>
+PRIMARY=$(/usr/bin/mongo --quiet --eval 'db.isMaster().ismaster')
+
+cd $BACKUP_DIR
+
+
+TIME="$(date +%s)"
+
+# Download and decrypt backup
+/usr/bin/envdir <%= @env_dir %>/env.d /usr/local/bin/s3cmd get --force `/usr/local/bin/s3cmd ls s3://${S3_BUCKET}/mongodump* | \
+tail -1 | /usr/bin/awk '{print $4}'`| /usr/bin/awk '{print $4}' | /usr/bin/tr -d "'" | \
+/usr/bin/xargs /usr/bin/gpg --yes --quiet --output mongodump.tar.gz --decrypt && /bin/tar xzf mongodump.tar.gz
+
+TIME="$(($(date +%s)-TIME))"
+/usr/bin/printf "DOWNLOAD AND DECRYPTION FINISHED IN: ${TIME}s\n"
+
+
+TIME="$(date +%s)"
+
+if [ $PRIMARY = "true" ]; then
+
+    # Restore mongo database
+    /usr/bin/mongorestore --quiet var/
+
+    # Tidy up
+    /bin/rm -rf var/ mongodump*
+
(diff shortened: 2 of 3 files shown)


modules/mongodb/templates/mongodb-restore-s3.erb
```diff
@@ -0,0 +1,66 @@
...
+PRIMARY=$(mongo --quiet --eval 'db.isMaster().ismaster')
+
+cd $BACKUP_DIR
+
+
+TIME="$(date +%s)"
+
+# Download and decrypt backup
+/usr/bin/envdir <%= @env_dir %>/env.d /usr/local/bin/s3cmd get --force `/usr/local/bin/s3cmd ls s3://${S3_BUCKET}/mongodump* | \
+tail -1 | awk '{print $4}'`| awk '{print $4}' | tr -d "'" | \
```
surminus: I still think this might take a while to process if the bucket gets large because we'll be waiting to list everything in the bucket... unless we do some housekeeping on the bucket every night to keep it a respectable size in light of 6df4ee45caae7c23eeb6edacdd98f5c44ef1db58

modules/mongodb/templates/mongodb-restore-s3.erb
```diff
@@ -0,0 +1,66 @@
...
+xargs /usr/bin/gpg --yes --quiet --output mongodump.tar.gz --decrypt && tar xzf mongodump.tar.gz
```
surminus: Does a specific key fingerprint need to be specified to ensure the right key is used to decrypt the backups?

modules/mongodb/manifests/s3backup/restore.pp
```diff
@@ -0,0 +1,55 @@
...
+# [*backup_dir*]
+#   Defines the directory to restore the backups
+#
+# [*env_dir*]
+#   Defines directory for the environment
+#   variables
+#
+# [*private_gpg_key*]
+#   Defines the ascii exported private gpg to
+#   use for encrypting backups. This key should
```
surminus: Technically this is used for decrypting

modules/mongodb/manifests/s3backup/restore.pp
```diff
@@ -0,0 +1,55 @@
...
+#   be created by the user and encrypted with eyaml
+#
+# [*private_gpg_key_fingerprint*]
+#   Defines the fingerprint of the gpg private
+#   key to encrypt the backups. The fingerprint
```
surminus: Technically this is used for decrypting

modules/mongodb/manifests/s3backup/restore.pp
```diff
@@ -0,0 +1,55 @@
...
+#   should be 40 characters without spaces
+#
+# [*s3_bucket*]
+#   Defines the AWS S3 bucket where the backups
+#   will be uploaded. It should be created by the
```
surminus: Downloaded?

modules/mongodb/templates/mongodb-restore-s3.erb
```diff
@@ -0,0 +1,66 @@
...
+    # Tidy up
+    /bin/rm -rf var/ mongodump*
+
+    TIME="$(($(date +%s)-TIME))"
+    printf "RESTORE COMPLETED IN: ${TIME}s\n"
+
+else
+
+    # Check if mongo is running
+    /bin/ps -ef | grep mongod | grep -v grep > /dev/null
```
surminus: I recently learned of the `pgrep` command, check it out (not entirely necessary to change but I thought it does the same job in a tidier fashion)

surminus:
I really like the way this sorts out the entire cluster all at the same time, do you have any ideas how we would schedule this to ensure the primary has the new data before the rest of the cluster catches up? Would we just run it across all the nodes at the same time?

As a quibble please can commit messages have a title and a separate paragraph of text as it makes it harder to read otherwise.

modules/mongodb/templates/mongodb-restore-s3.erb
djsd123: We will manage housekeeping within s3 using lifecycle rules. I expect there will be roughly 96 backups per day, so over 650 a week which should not affect listing dramatically as I've seen much larger buckets.

modules/mongodb/templates/mongodb-restore-s3.erb
djsd123: I assumed the backup user and his\her keys existed on all the mongo instances since they are built with the same class/modules??

modules/mongodb/manifests/s3backup/restore.pp
djsd123: True. Bad copy, paste on my part.  Will change.

modules/mongodb/manifests/s3backup/restore.pp
djsd123: True. Bad copy, paste on my part. Will change.

modules/mongodb/manifests/s3backup/restore.pp
djsd123: True. Bad copy, paste on my part. Will change.

modules/mongodb/templates/mongodb-restore-s3.erb
djsd123: Yep, I've used pgrep a lot in the past.  Just habit like using `kill <PID>` instead of `pkill <process name>`.

modules/mongodb/templates/mongodb-restore-s3.erb
surminus: This restore job will be pulling Production data to Staging and Integration machines, so they'll have different keys. We might need the Production keys on these machines, which isn't great, but it's better to be badly encrypted than not encrypted at all.

modules/mongodb/templates/mongodb-restore-s3.erb
surminus: Cool. I think Terraform has lifecycle rules for buckets so if you fancy adding [that to our terraform repo](https://github.com/alphagov/govuk-terraform-provisioning/) feel free ;)

surminus:
I think this is basically there, although when we come to test it we just need to ensure that it runs on the primary, and then decide how we're going to schedule it. Merge it after https://github.com/alphagov/govuk-puppet/pull/4491

modules/mongodb/manifests/s3backup/restore.pp
```diff
@@ -0,0 +1,55 @@
+# == Class: mongodb::s3backup::restore
+#
+# Restore a MongoDB backup to a server from s3
+#
+# === Parameters:
+#
+# [*aws_access_key_id*]
+#
```
deanwilson: These should be a non-empty description.

modules/mongodb/manifests/s3backup/restore.pp
```diff
@@ -0,0 +1,55 @@
...
+# [*aws_secret_access_key*]
```
deanwilson: These should be a non-empty description.

modules/mongodb/manifests/s3backup/restore.pp
```diff
@@ -0,0 +1,55 @@
...
+#   user
+#
+# [*cron*]
+#   Defines whether to enable the cron job. Value
+#   should be true or false
+class mongodb::s3backup::restore(
+  $aws_access_key_id = undef,
+  $aws_secret_access_key = undef,
+  $env_dir = '/etc/mongo_s3backup',
+  $s3_bucket  = 'govuk-mongodb-backup-s3',
```
deanwilson: This bucket name should include an environment specifier.

modules/mongodb/manifests/s3backup/restore.pp
```diff
@@ -0,0 +1,55 @@
...
+  $backup_dir = '/var/lib/s3backup',
+  $user = mongodb::s3backup::backup::user,
```
deanwilson: Can we please confirm this works? I suspect it needs a `$` at the front.

modules/mongodb/templates/mongodb-restore-s3.erb
```diff
@@ -0,0 +1,66 @@
...
+    # Tidy up
+    /bin/rm -rf var/ mongodump*
+
+    TIME="$(($(date +%s)-TIME))"
+    /usr/bin/printf "RESTORE COMPLETED IN: ${TIME}s\n"
+
+else
+
+    # Check if mongo is running
+    /usr/bin/pgrep mongod > /dev/null
```
deanwilson: Please use status to get a confirmed status, rather than a process name check.

modules/mongodb/templates/mongodb-restore-s3.erb
```diff
@@ -0,0 +1,66 @@
...
+TIME="$(($(date +%s)-TIME))"
+/usr/bin/printf "DOWNLOAD AND DECRYPTION FINISHED IN: ${TIME}s\n"
+
+
+TIME="$(date +%s)"
+
+if [ $PRIMARY = "true" ]; then
+
+    # Restore mongo database
+    /usr/bin/mongorestore --quiet var/
```
surminus: Is `--quiet` flag included on the version of Mongo we're using?

modules/mongodb/templates/mongodb-restore-s3.erb
```diff
@@ -0,0 +1,66 @@
...
+
+fi
+
+
+if [ $? == 0 ]
+then
+STATUS=0
+else
+STATUS=1
+fi
```
surminus: Probably don't need this if we're not doing a passive check

surminus:
We still need to work out how we're going to schedule this, but let's address the outstanding comments and push it through to test in Integration only.

modules/mongodb/templates/mongodb-restore-s3.erb
djsd123: No, good catch.

surminus:
As discussed in mob review trim this down to simply do the restore function and we'll worry about the slaves and selection of primary elsewhere.

surminus:
:+1: Looks good

PR #4524, opened 2016-06-21
Lowercase URLs if they are all capitals

gpeng:

CAPSLOCK would seem to account for 2.5% (13,311) of 404 errors in the sample analysed.

So...

```
```


modules/router/templates/router_include.conf.erb
```diff
@@ -26,6 +26,24 @@ access_log /var/log/nginx/lb-access.log timed_combined;
...
+        my $uri = $r->uri;
+        $uri = lc($uri);
+        return $uri;
+}';
+
+# If slug contains no lowercase letters then lowercase it
+# eg www.gov.uk/GOVERNMENT/GUIDANCE -> www.gov.uk/government/guidance
+# eg WWW.GOV.UK/GOVERNMENT/GUIDANCE -> www.gov.uk/government/guidance
+location ~ ^([A-Z]|\W|\d)+$ {
+  rewrite ^(.*)$ $scheme://$host:8080$uri_lowercase;
```
boffbowsh: Thinking about it, we should probably redirect with a 301 to help our Google juice

gpeng: Fixed in 6951b08

gpeng:
@alexmuller Could you confirm that this is the right place for what I'm trying to do?

modules/router/templates/router_include.conf.erb
```diff
@@ -26,6 +26,24 @@ access_log /var/log/nginx/lb-access.log timed_combined;
...
+        my $uri = $r->uri;
+        $uri = lc($uri);
+        return $uri;
+}';
+
+# If slug contains no lowercase letters then lowercase it
+# eg www.gov.uk/GOVERNMENT/GUIDANCE -> www.gov.uk/government/guidance
+# eg WWW.GOV.UK/GOVERNMENT/GUIDANCE -> www.gov.uk/government/guidance
+location ~ ^([A-Z]|\W|\d)+$ {
+  rewrite ^(.*)$ $scheme://$host:8080$uri_lowercase permanent;
```
rjw1: The port probably shouldn't be there.

gpeng: Thanks Bob. Fixed in ebdc9e7

alexmuller:
Can we add some tests to [smokey](https://github.com/alphagov/smokey/blob/master/features/gov_uk.feature) for this so that we'll be more likely to catch it if we break it?

gpeng:
@alexmuller added in https://github.com/alphagov/smokey/pull/186

boffbowsh:
This LGTM, but will let someone from infra merge

alexmuller:
Nominate @rjw1 because it involves Perl.

deanwilson:
Does this config require `perl_modules` if it's not using any extra modules?

rjw1:
:+1:

modules/router/templates/router_include.conf.erb
```diff
@@ -26,6 +26,24 @@ access_log /var/log/nginx/lb-access.log timed_combined;
 access_log /var/log/nginx/lb-json.event.access.log json_event;
 error_log  /var/log/nginx/lb-error.log;
 
+# Include the perl module
+perl_modules perl/lib;
+
+# Function to lowercase a uri
+perl_set $uri_lowercase 'sub {
+        my $r = shift;
+        my $uri = $r->uri;
```
Davidslv: use `lc` here instead of assigning $uri again on line 36.

gpeng: Fixed in 2366946

gpeng:
I've removed the unnecessary `perl_modules` @deanwilson and also condensed it based on @Davidslv 's suggestion.

deanwilson:
Looks good. I think the @rjw1 +1 still stands with those changes so merge away. Let's get this in to integration.

PR #4525, opened 2016-06-21
Only require 1 elasticsearch master node on the dev vm

h-lame:
In 1cfda1339ca819fdeb80f8c027173e565c6e38dd we changed the default for
`minimum_master_nodes` to be 2.  This makes sense everywhere except on the
dev vm where we still want it to be 1.  If it remains as 2 then the dev
vm cluster is unhealthy and reports:

   {"error":"MasterNotDiscoveredException[waited for [30s]]","status":503}

for `curl -XGET 'localhost:9200/_cluster/health?pretty'`.  This in turn
means we cannot run replication scripts.  Setting it to 1 makes things
work again.

### DIFF
--- a/modules/govuk/manifests/node/s_development.pp
+++ b/modules/govuk/manifests/node/s_development.pp
@@ -47,11 +47,12 @@ class govuk::node::s_development (
   }
 
   class { 'govuk_elasticsearch':
-    cluster_name       => 'govuk-development',
-    heap_size          => '1024m',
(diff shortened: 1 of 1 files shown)


tijmenb:
👍 great, thanks for fixing this.

mobaig:
👍  nice catch. By extension this fix is necessary to run Rummager tests, use the unified_search API in dev - anything relying on ES basically.

PR #4526, opened 2016-06-21
Move a directory we use in multiple places to a base class

deanwilson:
We need /etc/wal-e in multiple classes so we'll move it in to
the package class


surminus:
:+1: merge when green

PR #4527, opened 2016-06-22
Validate gpg key

deanwilson:


surminus:
:+1: Good call, thanks! fftm

PR #4528, opened 2016-06-22
WAL-E env sync bugfixes

surminus:
This fixes a couple of issues in the script which does an envioronment
sync.

- The first line incorrectly appended the passphrase to the end of another
line.
- The second line had permissions issues because of sudo and echo

### DIFF
--- a/modules/govuk_postgresql/templates/usr/local/bin/wal-e_env_sync
+++ b/modules/govuk_postgresql/templates/usr/local/bin/wal-e_env_sync
@@ -29,13 +29,13 @@ else
   sudo service postgresql stop
 
   # Push the passphrase to a file so we can decrypt without prompt
-  sudo envdir <%= @env_sync_envdir %> sh -c  'echo passphrase $GPG_PASSPHRASE >> /var/lib/postgresql/.gnupg/gpg.conf'
+  sudo envdir /etc/wal-e/env_sync/env.d sh -c  'echo -n "\npassphrase ${GPG_PASSPHRASE}\n" >> /var/lib/postgresql/.gnupg/gpg.conf'
 
   # Start the restore process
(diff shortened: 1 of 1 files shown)


modules/govuk_postgresql/templates/usr/local/bin/wal-e_env_sync
```diff
@@ -29,13 +29,13 @@ else
   sudo service postgresql stop
 
   # Push the passphrase to a file so we can decrypt without prompt
-  sudo envdir <%= @env_sync_envdir %> sh -c  'echo passphrase $GPG_PASSPHRASE >> /var/lib/postgresql/.gnupg/gpg.conf'
+  sudo envdir /etc/wal-e/env_sync/env.d sh -c  'echo -n "\npassphrase ${GPG_PASSPHRASE}\n" >> /var/lib/postgresql/.gnupg/gpg.conf'
```
rjw1: what clears up the passphrase?

rjw1:
I have a query about one line in the script but feel free to merge anyway since this change is the previous behaviour anyway.  :+1:

modules/govuk_postgresql/templates/usr/local/bin/wal-e_env_sync
surminus: When Puppet runs [this code](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_postgresql/manifests/wal_e/backup.pp#L132-L138) should clean up the passphrase file.

surminus: Having said that, this assumes you're running backups on the server, which may not be true.

surminus:
I've added a separate PR fixing up the passphrase thing https://github.com/alphagov/govuk-puppet/pull/4534

PR #4529, opened 2016-06-22
Remove this partial work until we're ready to focus on it

deanwilson:
This was the first part of a task we've changed priorities on for now
so remove it from the code base.


surminus:
:+1: lgtm fftm 🍆

PR #4534, opened 2016-06-27
Ensure passphrase in file is cleaned up

surminus:
This ensures we clean up the passphrase that gets added during the sync process.

The template file is also renamed.


deanwilson:
Looks fine. Merge away

PR #4535, opened 2016-06-27
Fix freshness threshold on local services import

sihugh:
The freshness threshold for the job to import services and interactions into
Local Links Manager was set to just over 14 days, however the Jenkins job is
scheduled to run on the [1st of the month](https://deploy.publishing.service.gov.uk/job/Local_Links_Manager_Import_Service_Interactions/).

This amends the freshness threshold to be 32 days so that we don't get noise in
Icinga.


deanwilson:
Actually had to test that one. Code looks fine and the change makes sense.

Merge away.

surminus:
The simple maths actually confused me.

PR #4536, opened 2016-06-27
Improve search-analytics docs

tijmenb:
Adds links to external documentation and emphasises that this task shouldn't be run during the day.


benhyland:
👍  yes please

PR #4533, opened 2016-06-24
[Do not merge] Add Grafana 3 config...

steventux:
This PR relies on https://github.com/alphagov/govuk-puppet/pull/4531

Grafana 3 takes config from the grafana.ini file, we need to switch on loading
dashboard templates from the filesystem at /etc/grafana/dashboards, apart from
this setting the config is all defaults.

PR #4538, opened 2016-06-28
Grafana 3

cbaines:
Upgrades Grafana to 3.0.4.

Replaces: https://github.com/alphagov/govuk-puppet/pull/4533


cbaines:
I discussed this with @rjw1 yesterday, and once https://github.com/alphagov/govuk-puppet/pull/4531 has been deployed to production, the mirror can be manually setup, at which point, this can be deployed.

I have not got access to production at the moment, so will need some assistance with this.

modules/grafana/manifests/repo.pp
```diff
@@ -0,0 +1,17 @@
...
+#   Hostname to use for the APT mirror.
+#
+class grafana::repo (
+  $apt_mirror_hostname,
+) {
+  apt::source { 'grafana':
+    location     => "http://${apt_mirror_hostname}/grafana",
+    release      => 'main',
+    architecture => $::architecture,
+    key          => '418A7F2FB0E1E6E7EABF6FE8C2E73424D59097AB',
```
alexmuller: I think this key should be `3803E444EB0235822AA36A66EC5FE1A937E3ACBB` because that's the key for our apt mirror rather than for the original Grafana repo.

modules/grafana/manifests/init.pp
```diff
@@ -14,9 +14,9 @@
     ensure => directory,
   }
 
-  file { '/etc/grafana/config.js':
+  file { '/etc/grafana/grafana.ini':
     ensure => file,
-    source => 'puppet:///modules/grafana/config.js',
```
alexmuller: `modules/grafana/files/config.js` can be removed from this repo. You could also tidy up `/etc/grafana/config.js` on disk using `ensure => absent` so that people don't get confused when they go looking for Grafana config.

modules/grafana/manifests/repo.pp
cbaines: Makes sense, I've now changed it.

modules/grafana/manifests/init.pp
cbaines: All done in 71f4b74

alexmuller:
Looks good to me :+1:. I'm going to mirror the new Grafana repo shortly so let's leave this open until that's done.

PR #4539, opened 2016-06-28
Revert "Lowercase URLs if they are all capitals"

boffbowsh:
Reverts alphagov/govuk-puppet#4524

Causes an nginx syntax error.


gpeng:
:+1: when green

PR #4540, opened 2016-06-28
Lowercase URLs if they are all capitals

gpeng:

This should reduce 404s received by users.

So...

```
```

This is a second attempt as the first #4524 didn't work in integration due to the perl function being included within the wrong scope of the config.


boffbowsh:
Lets deploy the branch to integration to test this. Test it with:
`sudo nginx -t`

gpeng:
I've deployed this branch to integration and tested. All good.

deanwilson:
Merge away.

PR #4543, opened 2016-06-28
Bugfix MySQL restore stuff

surminus:
My restore stuff foolishly assumed that we would have the packages
already installed from backups, but actually in MySQL we backup from the
slave but restore to a master, so the packages class wasn't already
included and there was a require which threw this error:

``` Error: Could not retrieve catalog from remote server: Error 400 on
SERVER: Invalid relationship: File[/usr/local/bin/xtrabackup_s3_restore]
{ require => Class[Govuk_mysql::Xtrabackup::Packages] }, because
Class[Govuk_mysql::Xtrabackup::Packages] doesn't seem to be in the
catalog ```

This just includes it if it's not installed already.

### DIFF
--- a/modules/govuk_mysql/manifests/xtrabackup/restore.pp
+++ b/modules/govuk_mysql/manifests/xtrabackup/restore.pp
@@ -12,6 +12,8 @@ define govuk_mysql::xtrabackup::restore (
 ) {
   $env_sync_envdir = '/etc/mysql/xtrabackup/env_sync/env.d'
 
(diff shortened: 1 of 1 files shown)


deanwilson:
Merge when green.

surminus:
It's not turning green 😦

deanwilson:
Looks green to me. Merge away before it notices.

surminus:
I force pushed it again.

PR #4544, opened 2016-06-29
Change broken link report email

gpeng:
This PR updates the Jenkins job that calls the Whitehall broken link report rake task to pass an updated email address parameter as per [Trello](https://trello.com/c/RCdWcC28/429-broken-links-report-for-publishers-medium)

It also adds an Icinga passive check for success/failure of the job. I'm not familiar with how Icinga is set up here so have just followed how the other jobs seem to hook into it. Any suggestions/corrections would be graciously received...

### DIFF
--- a/modules/govuk_jenkins/manifests/job/whitehall_run_broken_link_checker.pp
+++ b/modules/govuk_jenkins/manifests/job/whitehall_run_broken_link_checker.pp
@@ -3,9 +3,22 @@
 # Create a file on disk that can be parsed by jenkins-job-builder
 #
 class govuk_jenkins::job::whitehall_run_broken_link_checker {
+  $app_domain = hiera('app_domain')
+
+  $service_description = 'Runs a rake task on Whitehall that generates a broken link report'
+  $job_slug = 'whitehall_run_broken_link_checker'
+  $job_url = "https://deploy.${app_domain}/job/${job_slug}"
+
   file { '/etc/jenkins_jobs/jobs/whitehall_run_broken_link_checker.yaml':
     ensure  => present,
     content => template('govuk_jenkins/jobs/whitehall_run_broken_link_checker.yaml.erb'),
     notify  => Exec['jenkins_jobs_update'],
+  }
+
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/manifests/job/whitehall_run_broken_link_checker.pp
```diff
@@ -3,9 +3,22 @@
...
   file { '/etc/jenkins_jobs/jobs/whitehall_run_broken_link_checker.yaml':
     ensure  => present,
     content => template('govuk_jenkins/jobs/whitehall_run_broken_link_checker.yaml.erb'),
     notify  => Exec['jenkins_jobs_update'],
   }
+
+  @@icinga::passive_check { "${job_slug}_${::hostname}":
+    service_description => $service_description,
+    host_name           => $::fqdn,
+    freshness_threshold => 115200,
```
boffbowsh: This should be 31 * 86400 as it runs only once a month

gpeng: Fixed in 6617f47

surminus:
Hmm, Puppet on jenkins-1 is throwing an error related to this change:

```
Notice: /Stage[main]/Govuk_jenkins::Job_builder/Exec[jenkins_jobs_update]/returns: yaml.scanner.ScannerError: mapping values are not allowed here
Notice: /Stage[main]/Govuk_jenkins::Job_builder/Exec[jenkins_jobs_update]/returns:   in "/etc/jenkins_jobs/jobs/whitehall_run_broken_link_checker.yaml", line 21, column 26
```

surminus:
Ah, suspect that [this](https://github.com/alphagov/govuk-puppet/blob/6617f4770a672bd8db3bd21492b4780bfe5de265/modules/govuk_jenkins/templates/jobs/whitehall_run_broken_link_checker.yaml.erb#L21) needs to be in caps. I'll put in a fix.

PR #4545, opened 2016-06-29
Update Chris Lee's ssh key due to losing his private key

tuzz:


surminus:
Have you checked in the last place you remember having it? 

:+1: merge when green

tuzz:
We've looked under the rug and behind the fridge. It's nowhere to be seen.

PR #4548, opened 2016-06-29
Only enable the env_sync on selected machines.

deanwilson:
This shouldn't be deployed everywhere and in this specific case
we're only testing in integration on the transition master.


surminus:
:+1: Great thanks, merge when green

PR #4549, opened 2016-06-29
Remove the old cronjob

deanwilson:
We now run this from cron.d. Once this file is absent everywhere
we'll remove this puppet resource.


surminus:
:+1: mwg

PR #4550, opened 2016-06-29
Default the host used for mongo backups to 'localhost'

deanwilson:
We won't need the hiera value anymore so we're removing that too.


surminus:
:+1: Hopefully this should remove the errors in Integration! mwg

PR #4551, opened 2016-06-29
Remove explictly pinned version for aptly

deanwilson:
We need some of the newer bug fixes so we're going to remove the pin
for testing in integration. Originally pinned in
c62115d364b7e9a3cedd39f6ed47ac54a6674d79


surminus:
:+1: merge when green. Let's test in Integration and then deploy to Puppet quickly after we've confirmed it's successful.

PR #4552, opened 2016-06-29
Remove mongo-pp from Integration data sync

surminus:
This can be reverted when Performance Platform are ready to enable data syncs
again.

/cc @leelongmore


sihugh:
Merge when green 👍

PR #4553, opened 2016-06-29
Lowercase all capital urls

gpeng:

This should reduce 404s received by users.

So...

```
```

The previous attempt at this matched `/` and caused a redirect loop. The regex used in this PR fixes that but introduces a new redirect loop at `/ / / /`. This isn't a valid URL and so probably doesn't matter but I've added a second commit with a possible 'fix' for this using a location that matches that url to redirect to root. I've included it for discussion as I've yet to come up with a better solution if we agree that this is a problem.

### DIFF
--- a/modules/router/templates/base.conf.erb
+++ b/modules/router/templates/base.conf.erb
@@ -2,6 +2,12 @@ upstream varnish {
   server localhost:7999;
 }
 
+# Function to lowercase a uri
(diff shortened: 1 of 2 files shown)


boffbowsh:
Deploy this on staging for a little while when the puppet deployment queue is clear 👍

PR #4554, opened 2016-06-29
Remove --quiet option

deanwilson:
Our version of the mongo command doesn't support this so remove it from the
script.


surminus:
:+1: merge away

PR #4555, opened 2016-06-29
Create the backups directory

deanwilson:
We're going to put all our mongodumps in this directory
and make it owned by the backups user.


surminus:
:+1: merge away

PR #4556, opened 2016-06-29
Add the source on the actual server

surminus:
The aptly mirror had been added but the source hadn't been added on the
servers to be able to find the package.


deanwilson:
Looks fine. Merge when green.

PR #4557, opened 2016-06-29
Bugfix

surminus:
Key should actually be our aptly mirror key.

apt-get uses http by default, not https


deanwilson:
Merge when green

PR #4560, opened 2016-06-30
Amended host from localhost to hostname to resolve failing check

rjbate:


surminus:
:+1: merge when green

PR #4562, opened 2016-06-30
Add link to ops manual explaining watermark check

rjbate:


surminus:
:+1: Cool, merge together with the opsmanual entry when that's ready

PR #4564, opened 2016-06-30
Revert "Remove this partial work until we're ready to focus on it"

surminus:
This reverts commit c57e764bcce06f57a11d0c04337032e70aed98b6, which was actually a revert of 51de76f67c9b8ebb8372e1d1bb670412a7cc6c22. 

We can try adding this again because aptly has now been upgraded (manually but also by [removing the pinning](https://github.com/alphagov/govuk-puppet/pull/4551/commits/dc29c37e998201a21324170e1c4c75fd8bb6aa3f)) which _should_ fix the error it was throwing when trying to add this package.


rjw1:
Merge when green!

deanwilson:
Merge away.

PR #4565, opened 2016-06-30
Add description of file to add the name of the manifest into

carolinegreen:
Add description of file to add the name of the manifest into, as it isn't clear currently which file in the hieradata directory this should go into.


tijmenb:
👍

PR #4572, opened 2016-07-01
Bugfix Jenkins Job Builder yaml

surminus:
Jenkins Job Builder gets upset:

```
Notice:
/Stage[main]/Govuk_jenkins::Job_builder/Exec[jenkins_jobs_update]/returns:
yaml.scanner.ScannerError: mapping values are not allowed here
Notice:
/Stage[main]/Govuk_jenkins::Job_builder/Exec[jenkins_jobs_update]/returns:
in "/etc/jenkins_jobs/jobs/whitehall_run_broken_link_checker.yaml", line
21, column 26
```

I believe it's because the condition should be in capitals.


rjw1:
merge when gree

PR #4574, opened 2016-07-01
Add recursive option to s3cmd mv

surminus:
This works between "directories" in the S3 bucket so we need the
recursive option.

This also started failing because of incorrect permissions which will [hopefully
be fixed here](https://github.com/alphagov/govuk-terraform-provisioning/pull/64)


rjbate:
👍

PR #4575, opened 2016-07-01
Update grafana aptly mirror

surminus:
aptly complains when it tries to get the mirror. The documentation for
[adding a mirror](https://packagecloud.io/grafana/stable/mirror)
suggests we need to use `ubuntu` rather than `debian`, and set the
release to `trusty`.


rjbate:
👍

PR #4578, opened 2016-07-04
Recommend using Vagrant 1.7.4

tijmenb:
Vagrant 1.8 doesn't work with vagrant-dns.

https://trello.com/c/S4ulV1Qr/4-vagrant-1-8-is-buggy-recommend-1-7-4


alexmuller:
Can we enforce this in code in the Vagrantfile?

tijmenb:
Yeah, but that would break vagrant for people who have got it working with 1.8 (by hacking the gem). I'd rather not do that.

alexmuller:
Makes sense to me :+1:

PR #4579, opened 2016-07-04
Local links manager link checker job

brenetic:
- Add link checker jenkins job for Local Links Manager. The link checker rake task should run daily after the importing of the links in order to check their status

- Change time that links are imported into Local Links Manager. The link checker rake task takes 19 hours approximately. This changes the importing of the links to run at 1am instead of 5am so that the link
checker task can begin an hour later.

[Trello card](https://trello.com/c/KBT9RC9q/406-check-the-status-code-of-links)

### DIFF
--- a/hieradata/production.yaml
+++ b/hieradata/production.yaml
@@ -89,6 +89,7 @@ govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::job::launch_vms
   - govuk_jenkins::job::local_links_manager_import_service_interactions
   - govuk_jenkins::job::local_links_manager_import_links
+  - govuk_jenkins::job::local_links_manager_check_links
   - govuk_jenkins::job::mirror_network_config
   - govuk_jenkins::job::network_config_deploy
   - govuk_jenkins::job::passive_checks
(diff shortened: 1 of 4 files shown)


surminus:
As discussed this running for 19 hours per day via Jenkins isn't ideal as it could block other work, though opinion from other WebOps would be appreciated!

Alternatively you could run it as a cron job from one machine, and add a passive check into the job to let us know if it errors or hasn't run for whatever reason. You'd need to add the [check into a manifest so it creates the check in Icinga](https://github.com/alphagov/govuk-puppet/blob/3be69889fe4347a027d942a6a79069f8b93fe2d8/modules/govuk/manifests/node/s_asset_master.pp#L67-L72), along with [the code that actually sends the passive check after the job has run](https://github.com/alphagov/govuk-puppet/blob/3be69889fe4347a027d942a6a79069f8b93fe2d8/modules/govuk/templates/node/s_asset_base/process-uploaded-attachments.sh.erb#L60-L61).

brenetic:
Closing this as we are going to take an approach similar to @surminus comment above.

PR #4581, opened 2016-07-05
Add new wildcard_publishing certificate type

alexmuller:
wildcard_alphagov is no longer accurate because we're not using the alphagov domain anymore.


surminus:
:+1: mob review approved! merge away

PR #4582, opened 2016-07-05
Define redis url for whitehall

mgrassotti:
https://trello.com/c/JQREOuvh/424-use-sidekiq-gem

Define environment constants for whitehall, in order to configure the
govuk_sidekiq gem properly.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -287,6 +287,9 @@ govuk::apps::frontend::nagios_memory_critical: 1400
 
 govuk::apps::govuk_cdn_logs_monitor::processed_data_dir: '/mnt/logs_cdn/data'
 
+govuk::apps::whitehall::redis_host: 'redis-1.backend'
+govuk::apps::whitehall::redis_port: '6379'
+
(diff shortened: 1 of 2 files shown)


mgrassotti:
this is related to https://github.com/alphagov/whitehall/pull/2633

alexmuller:
`redis_host` and `redis_port` will have to be added as parameters to the `govuk::apps::whitehall` class or they won't do anything, similar to the `govuk::apps::publishing_api` class.

mgrassotti:
👍 thanks I hope it's ok now

alexmuller:
I'm not sure this will work yet, the parameters are present but they're not being used to do anything.

mgrassotti:
Well we are using them inside the govuk_sidekiq gem, so here they should just be set. Does this makes sense?

alexmuller:
Sure, but this Puppet change isn't doing anything - the parameters on the Puppet class aren't being used to change any state on the whitehall machines. Do they need to set an environment variable?

alexmuller:
Ah cool, so in that case you'll need something similar to what's been done in #4584

mgrassotti:
👍 great thanks

PR #4583, opened 2016-07-05
Define redis url for publisher

mgrassotti:
Define environment constants for publisher, in order to configure the
govuk_sidekiq gem properly.

### DIFF
--- a/modules/govuk/manifests/apps/publisher.pp
+++ b/modules/govuk/manifests/apps/publisher.pp
@@ -38,6 +38,8 @@ class govuk::apps::publisher(
     $secret_key_base = undef,
     $mongodb_name = undef,
     $mongodb_nodes = undef,
+    $redis_host = 'redis-1.backend',
+    $redis_port = '6379',
   ) {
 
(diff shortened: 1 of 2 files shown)


mgrassotti:
This is related to https://github.com/alphagov/publisher/pull/483

alexmuller:
I think this is similar to the comment I added on #4582.

modules/govuk/manifests/apps/publisher.pp
```diff
@@ -106,5 +108,11 @@
       app     => 'publisher',
       varname => 'SECRET_KEY_BASE',
       value   => $secret_key_base;
+    "${title}-REDIS_HOST":
+      varname => 'REDIS_HOST',
```
boffbowsh: This is missing the `app` variable, which is why the tests are failing

mgrassotti: 👍 thanks

boffbowsh:
LGTM merge on Thursdasy 👍

PR #4585, opened 2016-07-05
Add the kevindew user

kevindew:


danielroseman:
This has failed because there's trailing spaces on several lines; the linter is unfortunately very strict about that.

PR #4587, opened 2016-07-05
Link checker nagios passive check

brenetic:
We are adding a passive check for Local Links Manager to Icinga. 

Also adding a generic wrapper for communicating. Please read the commits for extra info.

[Trello card](https://trello.com/c/KBT9RC9q/406-check-the-status-code-of-links)

[Related PR](https://github.com/alphagov/local-links-manager/pull/29)


alexmuller:
I think this is a good approach. It means that every machine will have a script at `/usr/local/bin/notify_passive_check` which can be used to notify Icinga of the result of a passive check.

Our `send_nsca` logic is duplicated all over this Puppet repo and we should refactor those uses to use this script as we touch those parts of the code.

:+1: from me.

PR #4588, opened 2016-07-05
Add .pgpass file

surminus:
I think this will be a controversial change. Essentially it means when we run
`pg_resync_slave` we won't have to enter a password. This would allow auto resync
of slaves after [running the environment sync from S3]
(https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_postgresql/manifests/wal_e/env_sync.pp).

This creates a pgpass file that reads in a password so batch that resync
data from a primary to a standby does not prompt for a password.

Ref: https://www.postgresql.org/docs/current/static/libpq-pgpass.html

### DIFF
--- a/modules/govuk_postgresql/manifests/server/standby.pp
+++ b/modules/govuk_postgresql/manifests/server/standby.pp
@@ -11,9 +11,15 @@
 # [*master_password*]
 #   Password to authenticate against the master.
 #
+# [*pgpassfile_enabled*]
+#   This creates a password file so you can automatically resync a slave from
+#   a master without prompt (to be used in batch jobs). Should be explictly
+#   allowed in an environment for security reasons.
+#
 class govuk_postgresql::server::standby (
   $master_host,
   $master_password,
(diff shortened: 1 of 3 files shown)


deanwilson:
I don't think you need the `\` before the `:`

modules/govuk_postgresql/manifests/server/standby.pp
```diff
@@ -52,6 +58,15 @@
     mode    => '0755',
     content => template('govuk_postgresql/usr/local/bin/pg_resync_slave.erb'),
   }
+  if $pgpassfile_enabled {
+    file { '/var/lib/postgresql/.pgpass':
+      ensure  => present,
+      mode    => '0600',
+      owner   => $pg_user,
+      group   => $pg_group,
+      content => "${master_host}:5432:*:${username}:${master_password}",
```
surminus: Add a dependency for Class['postgresql::server']

deanwilson:
Merge when green.

PR #4590, opened 2016-07-06
Set redis env variables for whitehall

mgrassotti:
Add the constants definition to the environment variables. 
They are used by the govuk_sidekiq gem.

It was missing in the previous PR.


boffbowsh:
👍  merge on Thursday

PR #4592, opened 2016-07-06
Correct indentation in jenkins job yaml

MatMoore:
Currently puppet can't apply this configuration because the yaml parser
throws an exception:

     yaml.scanner.ScannerError: mapping values are not allowed here


gpeng:
:+1: when 🍏

PR #4593, opened 2016-07-06
Fix Grafana apt repo package path and release name

steventux:
According to the docs, the debian path and release name should be used for Ubuntu:
http://docs.grafana.org/installation/debian/

### DIFF
--- a/modules/govuk/manifests/node/s_apt.pp
+++ b/modules/govuk/manifests/node/s_apt.pp
@@ -70,8 +70,8 @@ class govuk::node::s_apt (
       release  => 'stable',
       key      => '46095ACC8548582C1A2699A9D27D666CD88E42B4';
     'grafana':
-      location => 'https://packagecloud.io/grafana/stable/ubuntu',
-      release  => 'trusty',
+      location => 'https://packagecloud.io/grafana/stable/debian',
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/node/s_apt.pp
```diff
@@ -70,8 +70,8 @@
       release  => 'stable',
       key      => '46095ACC8548582C1A2699A9D27D666CD88E42B4';
     'grafana':
-      location => 'https://packagecloud.io/grafana/stable/ubuntu',
-      release  => 'trusty',
+      location => 'https://packagecloud.io/grafana/stable/debian',
+      release  => 'wheezy',
```
alexmuller: I think this should probably be `jessie`:

```
alexmuller@integration-apt-1:~$ cat /etc/debian_version 
jessie/sid
```

steventux: OK, agree - looks like packagecloud have that release https://packagecloud.io/app/grafana/stable/search?dist=debian%2Fjessie

steventux:
@alexmuller updated to `jessie`.

alexmuller:
Thanks, looks good to me :+1:

PR #4594, opened 2016-07-06
Migrate PaaS domains in Signon after staging data sync

alexmuller:


rjw1:
Good approach. :+1: i think the code looks fine.

PR #4598, opened 2016-07-07
[DO NOT MERGE] Use govuk_node_list to get a list of servers

alexmuller:
This avoids us having to specify all of our machines. It does increase the dependency on PuppetDB though (especially in the case of Vagrant) but I think the trade-off is worth it.

Opening for comment: I'd like some opinions on whether this is a good idea or not.

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -203,9 +203,6 @@ govuk::node::s_frontend_lb::performance_frontend_servers:
 govuk::node::s_frontend_lb::performance_frontend_apps:
   - 'spotlight'
   - 'performanceplatform-big-screen-view'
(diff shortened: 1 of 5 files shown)


alexmuller:
Tests have failed because the catalog compilation can't execute the function, which is probably something we can work around if people think this is worth pursuing.

surminus:
Me and @deanwilson talked about this and we're not sure what would happen if puppetdb went away? Would anything bad happen?

alexmuller:
Yeah, that's a good question. `govuk_node_list` will return a non-zero exit code which is correct and I think that will prevent anything nasty from happening. But it's worth testing, will have a look.

djsd123:
We will mob test this

deanwilson:
We should discuss this in person. Closing for now.

PR #4601, opened 2016-07-07
Fix YAML syntax

gpeng:
A formatting error in this job was causing it to fail.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/search_fetch_analytics_data.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/search_fetch_analytics_data.yaml.erb
@@ -21,10 +21,10 @@
         <li><a href='https://github.com/alphagov/rummager/blob/master/docs/popularity.md'>Docs on Rummager</a></li>
       </ul>
     scm:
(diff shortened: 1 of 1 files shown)


tijmenb:
👍 sorry about that.

This is the second YAML error this week (https://github.com/alphagov/govuk-puppet/pull/4592). @alexmuller do you know if it's possible to test these files in CI?

alexmuller:
I think this is covered by the functional testing @deanwilson has been thinking about recently.

This isn't a problem with compiling the catalog (which works fine and is something we test), it's a problem with applying jenkins-job-builder given our broken YAML.

gpeng:
Could we add a linter for it? https://github.com/adrienverge/yamllint type of thing?

alexmuller:
We could, but we need to render the ERB as YAML before linting (using the Puppet catalog compilation) which I think is where that gets tricky.

gpeng:
True although in both the recent cases the `erb` files linted as YAML would have failed when they had the issues in them and passed when they were fixed. Expanding the erb could still introduce invalid YAML but it would have helped still in these cases.

PR #4603, opened 2016-07-08
Add publishing api content item count to collectd...

steventux:

We'd like to have a collectd count to highlight content growth
in the publishing api. This is a db specific query so no need
to interpolate the database name.

### DIFF
--- a/modules/collectd/templates/etc/collectd/conf.d/postgresql.conf.erb
+++ b/modules/collectd/templates/etc/collectd/conf.d/postgresql.conf.erb
@@ -37,6 +37,16 @@ LoadPlugin postgresql
     </Result>
   </Query>
 
+  <Query publishing_api_content_item_count>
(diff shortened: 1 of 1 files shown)


modules/collectd/templates/etc/collectd/conf.d/postgresql.conf.erb
```diff
@@ -37,6 +37,16 @@ LoadPlugin postgresql
     </Result>
   </Query>
 
+  <Query publishing_api_content_item_count>
+    Statement "SELECT count(*) AS count \
+               FROM publishing_api_production.content_items;"
```
boffbowsh: Does this work? The connection will be to the `postgres` database. I don't think you can select across databases like this in postgres?
```
postgres=# SELECT count(*) AS count FROM publishing_api_production.content_items;
ERROR:  relation "publishing_api_production.content_items" does not exist
```

PR #4606, opened 2016-07-08
Fix the release name for grafana

steventux:
According to the puppet apt module docs:

```
release: Specifies a distribution of the Apt repository.
repos: Specifies a component of the Apt repository. Valid options: a string. Default: 'main'.
```

So release should be 'jessie'

### DIFF
--- a/modules/grafana/manifests/repo.pp
+++ b/modules/grafana/manifests/repo.pp
@@ -10,7 +10,7 @@ class grafana::repo (
 ) {
   apt::source { 'grafana':
     location     => "http://${apt_mirror_hostname}/grafana",
-    release      => 'main',
+    release      => 'jessie',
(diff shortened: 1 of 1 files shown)


alexmuller:
I'm a bit confused by this because I hadn't realised that we are specifying `release` here already. So I don't understand where the `trusty` part of the source we currently have is coming from.

steventux:
Yes I'm fairly confused about it too, but the current docs say:
```
release: Specifies a distribution of the Apt repository. Valid options: a string. Default: "$lsbdistcodename".
repos: Specifies a component of the Apt repository. Valid options: a string. Default: 'main'.
```

alexmuller:
Ah right now I see, I was looking at an old version of the code :+1:

PR #4608, opened 2016-07-08
Restore script bugfix

surminus:
Several bugfixes found after testing.

I also think it's a good idea to raise the disk as this restore requires downloading and copying a large amount of data and the current disk size means it gets it very fine. It's quicker than a mysqldump, but the drawback is disk usage.

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -215,7 +215,7 @@ govuk::node::s_licensify_lb::enable_feed_console: true
 govuk::node::s_licensify_mongo::licensify_mongo_encrypted: false
 govuk::node::s_logging::compress_srv_logs_hour: '9'
 govuk::node::s_monitoring::offsite_backups: false
-govuk::node::s_mysql_master::s3_bucket_name: 'govuk-mysql-xtrabackups-integration'
+govuk::node::s_mysql_master::s3_bucket_name: 'govuk-mysql-xtrabackups-production'
 govuk::node::s_whitehall_backend::sync_mirror: true
 
 govuk::node::s_transition_postgresql_slave::redirector_ip_range: '10.1.5.0/24'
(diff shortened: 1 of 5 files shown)


modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
```diff
@@ -40,17 +43,18 @@ else
...
     echo "MySQL still running, exiting"
     exit 1
   fi
 
-  # Move the current directory to a backup location
+  # Copy the current directory to a backup location; we cannot move this if
+  # the directory is a mount point so we need to copy
   echo "Backing up current dir"
-  mv /var/lib/mysql /var/lib/mysql-bak
+  cp -R /var/lib/mysql /var/lib/mysql-bak
```
rjw1: I do wonder if were being over cautious here. 
Also what happens if /var/lib/mysql-bak is doesnt exist or already does.

surminus: Hmm, true about the directory name, I can make that unique. With regards to backing up, I just want to ensure that if the script fails for whatever reason then we're not left with no database whatsoever.

hieradata/integration.yaml
```diff
@@ -215,7 +215,7 @@ govuk::node::s_licensify_lb::enable_feed_console: true
 govuk::node::s_licensify_mongo::licensify_mongo_encrypted: false
 govuk::node::s_logging::compress_srv_logs_hour: '9'
 govuk::node::s_monitoring::offsite_backups: false
-govuk::node::s_mysql_master::s3_bucket_name: 'govuk-mysql-xtrabackups-integration'
+govuk::node::s_mysql_master::s3_bucket_name: 'govuk-mysql-xtrabackups-production'
```
deanwilson: Should this be set in common? All envs other than prod sync from here and prod syncs to here so the value should be the same. I think.

deanwilson:
Code looks fine. Let's get it in to integration so we can check it (and the permissions created by the restore script)

hieradata/integration.yaml
surminus: In Production the backup takes place from a different machine (backup-1) so it's actually a slightly different variable.

surminus:
Sadly we can't really test in Integration because it'll break signon, so we need to test in Staging. We'd need to deploy to Prod anyway for the changes to the backup scripts. 

If you're happy with it let me know and I will merge, otherwise we should be more careful.

djsd123:
FFTM 👍

PR #4609, opened 2016-07-11
Remove www-origin-staging from nginx

alexmuller:
Gor doesn't send traffic to this hostname anymore so it's completely unused. There's no need to keep it in our nginx config.


deanwilson:
Code looks fine. Merge away.

PR #4611, opened 2016-07-11
Set redis env variables for collection publisher

mgrassotti:
Set redis environment variables for collection publisher

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -246,6 +246,9 @@ govuk::apps::bouncer::db_hostname: "transition-postgresql-slave-1.backend"
 govuk::apps::bouncer::nagios_memory_warning: 1400
 govuk::apps::bouncer::nagios_memory_critical: 1500
 
+govuk::apps::collections_publisher::redis_host: 'redis-1.backend'
+govuk::apps::collections_publisher::redis_port: '6379'
+
 govuk::apps::contentapi::nagios_memory_warning: 3800
 govuk::apps::contentapi::nagios_memory_critical: 4000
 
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/collections_publisher.pp
```diff
@@ -49,6 +51,14 @@
     "${title}-PUBLISHING_API_BEARER_TOKEN":
       varname => 'PUBLISHING_API_BEARER_TOKEN',
       value   => $publishing_api_bearer_token;
+    "${title}-REDIS_HOST":
+      app     => 'collections_publisher',
```
boffbowsh: Because the app name is set on line 44, it doesn't need setting here

mgrassotti: 👍

hieradata/common.yaml
```diff
@@ -246,6 +246,9 @@ govuk::apps::bouncer::db_hostname: "transition-postgresql-slave-1.backend"
 govuk::apps::bouncer::nagios_memory_warning: 1400
 govuk::apps::bouncer::nagios_memory_critical: 1500
 
+govuk::apps::collections_publisher::redis_host: 'redis-1.backend'
```
deanwilson: Should this be hard coded to a single redis when we have multiple vms running?

mgrassotti: Yes, this should be hard coded as we need to use redis-1 specifically. Other apps already specify redis-1 too.

deanwilson:
Merge when ready.

PR #4613, opened 2016-07-11
Proxy requests for grafana

steventux:

Grafana 3 is a service with a web UI configured to use port 3204.
Configure nginx vhost to proxy requests to localhost on this port.

### DIFF
--- a/modules/grafana/files/vhost.conf
+++ b/modules/grafana/files/vhost.conf
@@ -1,3 +1,7 @@
+upstream grafana.gov.uk-proxy {
+  server localhost:3204;
+}
+
 server {
   server_name grafana.*;
   root /usr/share/grafana;
@@ -8,5 +12,11 @@ server {
 
   location /app/dashboards/ {
     alias /etc/grafana/dashboards/;
+
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header Host $http_host;
+    proxy_redirect off;
+
+    proxy_pass http://grafana.gov.uk-proxy;
   }
 }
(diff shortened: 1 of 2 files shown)


modules/grafana/files/vhost.conf
```diff
@@ -8,5 +12,11 @@ server {
 
   location /app/dashboards/ {
     alias /etc/grafana/dashboards/;
+
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header Host $http_host;
+    proxy_redirect off;
+
+    proxy_pass http://grafana.dev.gov.uk-proxy;
```
alexmuller: Can we remove the reference to dev here and just call it `grafana-proxy` or something similar?

boffbowsh:
I think this will end up happening in production too, which is not what we want. In production, grafana is a set of static assets which is configured so that the browser talks directly to graphite (I believe? correct me if I'm wrong) so it doesn't have a server component in production. Or does this change with grafana 3?

steventux:
Grafana 3 is a Go application running on port 3000 and I believe this upgrade will be rolled out to all environments as we want to use elasticsearch logs as a data source for better monitoring. Not sure if this approach is correct, but the intention is to proxy requests for grafana to the service running on port 3000. Please advise if a different port number should be used. I know this one is popular :)

modules/grafana/files/vhost.conf
steventux: Willdo

boffbowsh:
Ok that makes sense. The standard is to select a port number and record it in the Procfile in the development repo.

steventux:
Ah right, wasn't sure if we did this for management/monitoring applications, I can see we do. Will add that and configure Grafana accordingly. Thanks @boffbowsh

steventux:
@alexmuller updated (now with added port config!)

PR #4615, opened 2016-07-11
Ensure mongo service name is correct

surminus:
In later versions of MongoDB (3.x), and in the package we're installing
from specifically, the upstart service name has changed to 'mongod' from
'mongodb'. This causes a few problems in Puppet for dependencies, so as
a plaster before we upgrade fully to mongo 3 just specify the service
name.

### DIFF
--- a/hieradata/class/licensing_mongo.yaml
+++ b/hieradata/class/licensing_mongo.yaml
@@ -37,5 +37,4 @@ mongodb::server::replicaset_members:
 # Disable monthly backups to limit the retention of IL3 data.
 mongodb::backup::domonthly: false
 
-mongodb::server::package_name: 'mongodb-org'
 mongodb::server::version: '3.2.7'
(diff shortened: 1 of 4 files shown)


alexmuller:
Instead of having to set the service name manually could we do it conditionally based on the version, maybe using the `versioncmp()` function like we do for some things in `modules/govuk_elasticsearch/`?

deanwilson:
Possibly something more like 

```
if versioncmp($version, '3.0.0') >= 0 {
  $service_name = 'mongod'
} else {
  $service_name = 'mongodb'
}
```

surminus:
I wrote a conditional to begin with then decided it felt like overkill and just as hard-coded as adding hiera, but will do that if it's preferable. 

Haven't heard of `versioncmp` so thanks!

alexmuller:
This looks good to me, I like having to configure less stuff by hand :+1:

djsd123:
FFTM 👍

deanwilson:
Merging as we need this in integration.

PR #4617, opened 2016-07-12
Add apt repo for gor

alexmuller:
We're going to start packaging gor ourselves with fpm and storing it in aptly.


deanwilson:
Merge away.

PR #4619, opened 2016-07-12
Amended script to handle multiple locks

rjbate:
Per https://trello.com/c/z5wIKSLJ/124-modify-govuk-puppet-to-allow-multiple-locks

### DIFF
--- a/modules/puppet/manifests/init.pp
+++ b/modules/puppet/manifests/init.pp
@@ -46,6 +46,13 @@ class puppet (
     content => template($govuk_puppet_template),
   }
 
+  file { '/var/run/lock/puppet':
+      ensure => 'directory',
+      owner  => 'root',
+      group  => 'root',
+      mode   => '0700',
+  }
+
   service { 'puppet': # we're using cron, so we don't want the daemonized puppet agent
     ensure   => stopped,
     provider => base,
--- a/modules/puppet/templates/govuk_puppet
+++ b/modules/puppet/templates/govuk_puppet
@@ -8,11 +8,218 @@ fi
 
 START_TIME=`date +%s%3N`
 
-# Allow return code 2 in case -v/--detailed-exitcodes is passed.
-RBENV_VERSION=1.9.3 /usr/local/bin/govuk_setenv default puppet agent --onetime --no-daemonize "$@" || [ $? -eq 2 ]
+remove_all_locks() {
 
-TIME_TOOK=$((`date +%s%3N`-START_TIME))
-echo "<%= @fqdn_metrics %>.puppet.run_duration:${TIME_TOOK}|ms" | nc -w 1 -u localhost 8125
+	if ls *.lock 1> /dev/null 2>&1; then #there are locks
 
-# We should only reach here if a last_run_summary was written.
-/usr/local/bin/puppet_passive_check_update >/dev/null
+                        for filename in *.lock
+                        do
+                                rm $filename
+                                echo "Removed: $filename"
+                        done
+	fi
+
+	shift #Pop off --remove-all-locks
+
+	do_puppet --enable "$@" #Re-enable puppet
+
+}
+
+listlocks() {
+
+	#list all current locks, date, name, reason, identifier
+        #simplify this structure. push the date and a (kind of) 
+	#uuid to the filename. Keep the reason, name and uuid as
+	#the lockfile content
+
+	#NB The last field (the lock identifier) can be grabbed with
+	#a simple awk '{print $NF}' in a script
+
+	if ls *.lock 1> /dev/null 2>&1; then
+
+        	for filename in *.lock
+               	 	do
+               	        	line=$(head -n 1 "$filename") #Grab file contents
+               	         	echo "$filename : $line"
+               	 	done
+
+	else
+		echo "No lock files present"
+	fi
+
+}
+
+do_puppet() {
+
+	# Allow return code 2 in case -v/--detailed-exitcodes is passed.
+	RBENV_VERSION=1.9.3 /usr/local/bin/govuk_setenv default puppet agent --onetime --no-daemonize "$@" || [ $? -eq 2 ]
+
+	TIME_TOOK=$((`date +%s%3N`-START_TIME))
+	echo "<%= @fqdn_metrics %>.puppet.run_duration:${TIME_TOOK}|ms" | nc -w 1 -u localhost 8125
+
+	# We should only reach here if a last_run_summary was written.
+	/usr/local/bin/puppet_passive_check_update >/dev/null
+	exit 0;
+
+}
+
+
+usage() {
+
+	programname=$0
+
+	echo "usage: $programname"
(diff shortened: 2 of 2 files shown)


surminus:
Could we add detail in the commit message about why we want to do this?

modules/puppet/templates/govuk_puppet
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -eu
+set -e
```
deanwilson: Should `-u` be removed here?

rjbate: set -eu is back.

surminus:
I'm going to test this in Vagrant

modules/puppet/templates/govuk_puppet
```diff
@@ -8,11 +8,223 @@ fi
 
 START_TIME=`date +%s%3N`
 
-# Allow return code 2 in case -v/--detailed-exitcodes is passed.
-RBENV_VERSION=1.9.3 /usr/local/bin/govuk_setenv default puppet agent --onetime --no-daemonize "$@" || [ $? -eq 2 ]
+remove_all_locks() {
 
-TIME_TOOK=$((`date +%s%3N`-START_TIME))
-echo "<%= @fqdn_metrics %>.puppet.run_duration:${TIME_TOOK}|ms" | nc -w 1 -u localhost 8125
+	if ls *.lock 1> /dev/null 2>&1; then #there are locks
```
surminus: You could just use `test -f *.lock` here

modules/puppet/templates/govuk_puppet
```diff
@@ -8,11 +8,223 @@ fi
...
+	echo "	--list-locks : show all current locks "
+	echo "	--remove-lock : remove one lock, takes numeric identifier as parameter "
+	echo "	--disable : add a lock "
+	echo "	--enable : remove all locks "
+
+	exit 1
+
+}
+
+PUPPET_STATE_DIR=`puppet config print statedir`
```
surminus: Is this user specific? Might be easier just to have it in `/var/run/puppet/` or something

modules/puppet/templates/govuk_puppet
```diff
@@ -8,11 +8,223 @@ fi
...
+GOVUK_STATE_DIR="$PUPPET_STATE_DIR/govuk"
+
+#make it if it isn't
+[ -d $GOVUK_STATE_DIR ] || mkdir "$GOVUK_STATE_DIR"
```
surminus: We could make Puppet ensure that the directory exists

modules/puppet/templates/govuk_puppet
```diff
@@ -8,11 +8,222 @@ fi
...
+
+#change working directory or die
+cd "$GOVUK_STATE_DIR" || { echo "Unable to access locks directory: '$GOVUK_STATE_DIR'" 1>&2 ; exit 1; }
```
surminus: Is there a specific reason we have to work inside the directory?

modules/puppet/templates/govuk_puppet
```diff
@@ -8,11 +8,222 @@ fi
...
+	else
+		echo "No lock with identifier: $MYID found. Exiting.."
+		exit 1;
+	fi
+
+	;;
+
+--enable)
+
+	#TODO - interactive vs non-interactive behaviour
```
surminus: I actually don't think we need interactive behaviour. We just need to be able to pass the required arguments, and throw an error if people don't pass them.

modules/puppet/manifests/init.pp
```diff
@@ -46,6 +46,13 @@
     content => template($govuk_puppet_template),
   }
 
+  file { '/var/run/lock/puppet':
+      ensure => 'directory',
+      owner  => 'root',
+      group  => 'root',
+      mode   => '0700',
```
surminus: Are these permissions correct? Will it only be root that needs to write there?

rjbate: The script's first action is to sudo itself.

surminus: Oh sorry, yes!

surminus:
I tested this on Vagrant to get a feel on how it works. In general I think it works well, but I have a few comments. 

1)
The option `--enable` just removes all locks. I think it would be better to use this option specifying the same string that was used to disable it. So we disable puppet by running:
`govuk_puppet --disable "foo"`

Enable it again by running:
`govuk_puppet --enable "foo"` 

I think this would make the operation more consistent. If we wish to explicitly remove all locks then that should be a separate option `--remove-all-locks`. There's a risk people will just use `--enable` by default, which is avoiding the behaviour we want.

2) 
When it is disabled and you run the script to do a puppet run (`govuk_puppet --test`) it correctly says it won't run because it's disabled, and gives the reason of the first time you add a lock:
`Notice: Skipping run of Puppet configuration client; administratively disabled (Reason: 'test')`

Even if you remove the lock called "test" that is the reason still given, and listing the locks doesn't display this lock. I think this could potentially be confusing. 

Maybe do a search through the directory which contains the locks and do not even get to the puppet bit if some exist, and list all the locks currently in place? This would also be a quicker operation than running Puppet.

3) 
The lock reference doesn't make it easy to remove a specific lock, and it doesn't tell you the reference when you add a lock, you have to explicitly list the locks, find the reference number and then apply it. If we're doing it programatically would there be an easier way to do this?

modules/puppet/templates/govuk_puppet
```diff
@@ -8,11 +8,186 @@ fi
...
+		echo "Reason required as second parameter (quote-wrap); exiting .."
+		exit 1;
+	else
+		REASON="$2"
+	fi
+	
+   	#Create the lock
+		MYID=`date +%s`
+		LOCKFILE=puppet_disable_lock.$LOCK_DATE.$MYID.lock
+		echo "Creating $LOCKFILE .."
```
surminus: I don't think we need the `..` in the comment.

modules/puppet/templates/govuk_puppet
```diff
@@ -8,11 +8,200 @@ fi
...
+	LOCK_DATE=`date +%Y-%m-%d`	
+
+	REASON=${2:-} 
+
+	if [ -z "$REASON" ] #Check if a reason has been passed
+	then 
+		#It wasn't
+		echo "Reason required as second parameter (quote-wrap); exiting."
+		exit 1;
+	else
```
deanwilson: I don't think you need this code. If $2 was set then $REASON already contains it.

deanwilson:
I think the lock name should be a predictable value passed in from the calling script, such as "postgres_backup" or "mysql_backup". If this is used as the lock filename then you can also see what has the lock via an `ls`. If the same lock is attempted twice then i think this script should fail and refuse to add a second lock. This will help capture backup scripts that are not cleaning up after themselves.

rjbate:
Yep - that's a great idea.

deanwilson:
I'm happy to get this in to integration for testing. @surminus ?

surminus:
Agree, let's get it into Integration, but we probably need to let people know.

PR #4622, opened 2016-07-12
Make RAIB live for Specialist Publisher Rebuild

tuzz:
This routes requests to /raib-reports to the rebuilt
application rather than being handled by the existing
V1 application.


tuzz:
Mobbed by @Rosa-Fox @Seraphiana @tuzz @Chrislee187.

PR #4629, opened 2016-07-13
Configure vCloud host for each vCloud organisation

alexmuller:
This value is dependent on whether the organisation is normal, disaster recovery or licensify, so it needs to be set differently for each.


surminus:
:+1: merge away

PR #4631, opened 2016-07-13
Fix name of Kibana app

alexmuller:
We deploy this app as kibana but the repo is called kibana-gds which is where the confusion comes from.


djsd123:
FFTM 👍

PR #4632, opened 2016-07-13
Add an explicit architecture.

deanwilson:
We only mirror the 64bit version and without this apt updates
will complain and break on the lack of i386 files.


alexmuller:
Looks good to me :+1:

PR #4633, opened 2016-07-13
Make mongo config a class param.

deanwilson:
There is a difference in configuration filename between mongo 2 and 3.
This change lets you specify the one you want, although they both get
populated from the same template for now.


surminus:
:+1: merge when green

PR #4634, opened 2016-07-13
Fix mongo s3 backup Cron

djsd123:
Two things:

* Puppet fails to create cron job due to line continuation in command.
* Deleting backup job from root's crontab


deanwilson:
Code looks fine. Merge when green.

surminus:
Can we change the commit title of the first commit? It's a bit confusing otherwise.

PR #4635, opened 2016-07-13
Remove the line continuation character

deanwilson:
Puppet doesn't like resources declared in this way. It throws a
```
"-":6: bad minute
errors in crontab file, can't install.

Error.
```


surminus:
:+1: merge when green

deanwilson:
Also fixed in https://github.com/alphagov/govuk-puppet/pull/4634 so closing mine.

PR #4636, opened 2016-07-13
Fix mongodb s3 backup Cron

djsd123:
Two things:

Puppet fails to create cron job due to line continuation in command.
Deleting backup job from root's crontab

- Created a cron resource and marked it as absent to remove previously managed jobs.
- Removed backslash for two of the cron jobs as these were causing things to fail.
- Added conditional for $STATUS variable to the backup script.

Removing absent resource declarations and FIXME message in backup.pp.


surminus:
:+1: LGTM

deanwilson:
Merge when green. In general I'd put FIXME above the resource that they impact incase the code gets altered and the absolute line numbers change.

PR #4637, opened 2016-07-13
Backup Elasticsearch indices

djsd123:
Please merge this after https://github.com/alphagov/govuk-puppet/pull/4517
This is due to the 'Include' in the backup class causing the spec to fail when it cannot find the housekeeping class.

This refers o this PR https://github.com/alphagov/govuk-puppet/pull/4500
From coments on the last PR I have wrppped the entire class in a 'if'
to restrict to one node in the cluster.

### DIFF
--- a/modules/govuk_elasticsearch/manifests/backup.pp
+++ b/modules/govuk_elasticsearch/manifests/backup.pp
@@ -1,0 +1,114 @@
+# == Class: govuk_elasticsearch::backup
+#
+# GOV.UK specific class to backup elasticsearch indexes.
+#
+# === Parameters
+#
+# [*aws_access_key_id*]
+#   AWS key to authenticate requests
+#
+# [*aws_secret_access_key*]
+#   AWS key to authenticate requests
+#
+# [*aws_region*]
+#   AWS region for the S3 bucket
+#
+# [*env_dir*]
+#   Defines directory for the environment
+#   variables
+#
+# [*s3_bucket*]
+#   Defines the AWS S3 bucket where the backups
+#   will be uploaded. It should be created by the
+#   user
+#
+# [*user*]
+#   Defines the system user that will be created
+#   to run the backups
+#
+# [*es_repos*]
+#   The repositories within elasticsearch
+#   where backups will be stored
+#
+# [*es_indices*]
+#   Elaticsearch indexes
+#
+# === Variables
+#
+# [*json_es_indices*]
+#   Formatted indexes to be backed up when script is invoked
+#
+class govuk_elasticsearch::backup(
+  $aws_access_key_id = undef,
+  $aws_secret_access_key = undef,
+  $aws_region = 'eu-west-1',
+  $s3_bucket = undef,
+  $env_dir = '/etc/es_s3backup',
+  $user = 'govuk-backup',
+  $es_repo = undef,
+  $es_indices = []
+){
+
+
+    $threshold_secs = 28 * 3600
+    $service_desc = 'Elasticsearch-Index_Backup'
+    $json_es_indices = join(regsubst($es_indices, '(.*)', '"\1"'), ',')
+
+    include ::backup::client
+#   FIXME This include will cause tests to fail until the PR for housekeeping is merged
+    #include govuk_elasticsearch::housekeeping
+
+
+    # push env files
+    file { [$env_dir,"${env_dir}/env.d"]:
+      ensure => directory,
+      owner  => $user,
+      group  => $user,
+      mode   => '0750',
+    }
+
+    file { "${env_dir}/env.d/AWS_SECRET_ACCESS_KEY":
+      content => $aws_secret_access_key,
+      owner   => $user,
+      group   => $user,
+      mode    => '0640',
+    }
+
+    file { "${env_dir}/env.d/AWS_ACCESS_KEY_ID":
+      content => $aws_access_key_id,
+      owner   => $user,
+      group   => $user,
+      mode    => '0640',
+    }
+
+    file { "${env_dir}/env.d/AWS_REGION":
+      content => $aws_region,
+      owner   => $user,
+      group   => $user,
+      mode    => '0640',
+    }
+
+    @@icinga::passive_check { "check_esindexbackup-${::hostname}":
+      service_description => $service_desc,
+      freshness_threshold => $threshold_secs,
+      host_name           => $::fqdn,
+    }
+
+    file { 'es-backup-s3':
+      path    => '/usr/local/bin/es-backup-s3',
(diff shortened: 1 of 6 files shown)


modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -0,0 +1,83 @@
...
+ES_INDICES=<%= @json_es_indices %>
+ES_SNAPSHOT=snapshot_$(date +%d_%m_%y)
+AWS_REGION=<%= @aws_region %>
+AWS_ACCESS_KEY_ID=<%= @aws_access_key_id %>
+AWS_SECRET_ACCESS_KEY=<%= @aws_secret_access_key %>
+
+# Triggered whenever this script exits, successful or otherwise. The values
+# of CODE/MESSAGE will be taken from that point in time.
+function nagios_passive () {
+printf "<%= @ipaddress %>\t<%= @service_desc %>\t${NAGIOS_CODE}\t${NAGIOS_MESSAGE}\n" | /usr/sbin/send_nsca -H alert.cluster >/dev/null
```
surminus: Ensure the `service_desc` variable is set somewhere in the manifest

modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -0,0 +1,112 @@
...
+      group   => $user,
+      mode    => '0640',
+    }
+
+    file { 'es-backup-s3':
+      path    => '/usr/local/bin',
+      owner   => $user,
+      group   => $user,
+      mode    => '0550',
+      content => template('govuk_elasticsearch/es-backup-s3.erb'),
```
surminus: I think this is fine being owned by root and having `0755` permissions as is convention in /usr/local/bin, but having said that it's a bit picky

surminus:
Other a couple of comments I think this looks OK :+1:

deanwilson:
We still need to look at how this class gets included on the nodes as it is currently not used anywhere.

modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -0,0 +1,120 @@
...
+  $aws_region = 'eu-west-1',
+  $s3_bucket = undef,
+  $env_dir = '/etc/es_s3backup',
+  $user = 'govuk-backup',
+  $es_repos = undef,
+  $enabled = false,
+  $es_indices = undef
+){
+
+  if $enabled {
```
deanwilson: This should be removed and the gating should happen where this class is included.

modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -0,0 +1,120 @@
...
+
+    $threshold_secs = 28 * 3600
+    $service_desc = 'Elasticsearch-Index_Backup'
+    $json_es_indices = join(regsubst($es_indices, '(.*)', '"\1"'), ',')
+
+    include backup::client
+    # This include will cause tests to fail until the PR for housekeeping is merged
```
deanwilson: This should be a FIXME:

modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -0,0 +1,120 @@
...
+  $es_indices = undef
+){
+
+  if $enabled {
+
+    $threshold_secs = 28 * 3600
+    $service_desc = 'Elasticsearch-Index_Backup'
+    $json_es_indices = join(regsubst($es_indices, '(.*)', '"\1"'), ',')
+
+    include backup::client
```
deanwilson: This should be `include ::backup::client` to indicate it's a top level, separate, module.

modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -0,0 +1,120 @@
...
+    # This include will cause tests to fail until the PR for housekeeping is merged
+    #include govuk_elasticsearch::housekeeping
+
+
+    # push env files
+    file { [$env_dir,"${env_dir}/env.d"]:
+      ensure => directory,
+      owner  => $user,
+      group  => $user,
+      mode   => '0770',
```
deanwilson: I think this should be `0750` instead. Only puppet should write to these directories

modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @@icinga::passive_check { "check_esindexbackup-${::hostname}":
+      service_description => $service_desc,
+      freshness_threshold => $threshold_secs,
+      host_name           => $::fqdn,
+    }
+
+    file { 'es-backup-s3':
+      path    => '/usr/local/bin',
```
deanwilson: This should be a fully qualified path, including the filename.

modules/govuk_elasticsearch/spec/classes/govuk_elasticsearch_backup_spec.rb
```diff
@@ -0,0 +1,21 @@
+require_relative '../../../../spec_helper'
+require 'rspec-puppet'
```
deanwilson: You can remove this line.

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -0,0 +1,83 @@
...
+{
+  "indices": "${ES_INDICES}",
+  "include_global_state": "false",
+  "compress": "true",
+  "server_side_encryption": "true"
+}
+EOD
+)
+
+if [ $? -eq 0 ]; then
```
deanwilson: This is in the wrong place. We should discuss where it's supposed to be.

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -0,0 +1,83 @@
...
+  echo "Snapshot Failed"
+fi
+
+
+if [ $? == 0 ]
+then
+STATUS=0
+else
+STATUS=1
+fi
```
deanwilson: I think we can remove this and set and exit with the nagios code. We should test this.

djsd123: I removed the `else` block but I think this `if` block makes sense and is easy to read.

modules/govuk_elasticsearch/templates/es-backup-s3.erb
djsd123: @deanwilson Agreed but I think it best to move the heredoc???

deanwilson:
Merge away.

PR #4638, opened 2016-07-13
Fixup mongo backups

deanwilson:


brenetic:
FFTM as long as no PR description is need 👍

PR #4639, opened 2016-07-13
[DO NOT MERGE] Add Rate-Limit-Token to govuk_crawler_worker

surminus:
This ensures that the crawler doesn't get rate limited when hitting the
site.

This entirely depends on https://github.com/alphagov/govuk_crawler_worker/pull/107
which is highly likely to be incorrect because I was basically flapping at Go

### DIFF
--- a/modules/govuk/manifests/apps/govuk_crawler_worker.pp
+++ b/modules/govuk/manifests/apps/govuk_crawler_worker.pp
@@ -32,6 +32,10 @@
 # [*root_urls*]
 #   A list of hostnames that the crawler should crawl
 #
+# [*rate_limit_token*]
+#   Sets the header "Rate-Limit-Token" which ensures that the crawler is
+#   whitelisted by the rate limiting function (receiving 429 status)
+#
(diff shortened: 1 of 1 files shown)


rjw1:
I think this can get merged before the https://github.com/alphagov/govuk_crawler_worker/pull/107 since having another environment variable which we arnt using wont effect the current code. Will cause a restart but thats fine for crawler worker.

modules/govuk/manifests/apps/govuk_crawler_worker.pp
```diff
@@ -32,6 +32,10 @@
 # [*root_urls*]
 #   A list of hostnames that the crawler should crawl
 #
+# [*rate_limit_token*]
+#   Sets the header "Rate-Limit-Token" which ensures that the crawler is
+#   whitelisted the rate limiting function (receiving 429 status)
```
deanwilson: "by" the rate limiting function?

deanwilson:
Merge it Laura, merge it!

PR #4640, opened 2016-07-14
Tidy up scheduled maintenance functionality

alexmuller:
This fixes an old FIXME where we had a huge chunk of HTML stored in this repo. The new way to do this is to download a scheduled maintenance view from the static repo (like we do for our other error pages) and then use that instead.


modules/loadbalancer/manifests/maintenance.pp
```diff
@@ -0,0 +1,18 @@
...
+# Sets up a machine so that the loadbalancer can be put
+# into maintenance mode.
+#
+class loadbalancer::maintenance {
+  file { '/usr/share/nginx/www':
+    ensure  => directory,
+    mode    => '0755',
+    owner   => 'deploy',
+    group   => 'deploy',
+    require => Package['nginx'],
```
deanwilson: That package resource actually ensures nginx is absent. Class['nginx::package'] might be a better fit

alexmuller: Oops, thanks for catching that. Fixed.

modules/loadbalancer/manifests/maintenance.pp
```diff
@@ -0,0 +1,18 @@
+# == Class: loadbalancer::maintenance
+#
+# Sets up a machine so that the loadbalancer can be put
+# into maintenance mode.
+#
+class loadbalancer::maintenance {
+  file { '/usr/share/nginx/www':
+    ensure  => directory,
+    mode    => '0755',
+    owner   => 'deploy',
```
deanwilson: Does this need to be owned by deploy to allow fabric pushing of new files?

deanwilson:
Looks good. Merging.

PR #4648, opened 2016-07-14
Add publishing api overview dashboard to grafana

steventux:

Makes the Publishing API dashboard template available to Grafana 3

@deanwilson This adds a specific dashboard to the ever growing list, as discussed this could be done with some clever puppetry, happy to make it so with some advice.


deanwilson:
Looks fine. Merge away.

PR #4649, opened 2016-07-15
Enable Gor on cache-1 Staging only

surminus:
This will allow us to test a newer version of Gor without impacting on Deployments. Enabling on only a single cache machine shouldn't provide enough traffic to impact on Integration.

This is dependent on https://github.com/alphagov/puppet-gor/pull/8


surminus:
I guess the tests might fail until the new version of puppet-gor is pulled in.

alexmuller:
The new `.tar.gz` file needs to be committed as well (and the old one deleted).

surminus:
Oooooooh

modules/govuk_gor/manifests/init.pp
```diff
@@ -16,11 +16,19 @@
...
+#
+# [*envvars*]
+#   Set any environment variables that will be loaded into the Gor service.
+#
 class govuk_gor(
   $apt_mirror_hostname = '',
   $args = {},
   $enable = false,
   $version = '0.9.4~ppa1',
+  $binary_path = undef,
```
alexmuller: One thing I'm a little unsure of is what this will `undef` will do when it's passed through to the gor module, will it override the `/usr/bin/gor` from that module?

alexmuller:
This looks good to me :+1:. Added one comment about something I'm unsure of.

deanwilson:
Once the default file path is the packaged one rather than undef merge away.

PR #4650, opened 2016-07-15
Mongo restore s3 bucket path

djsd123:
Path in S3 bucket where backups will be downloaded from.
The paths for backups will vary by fully qualified hostname inside the S3 bucket.
This commit makes use of the ruby tag 'fqdn' to restore a backup for the machine
that invoked the restore.


surminus:
:+1: merge when green

PR #4653, opened 2016-07-15
Add timeout to processing command

surminus:
There might be an issue where it stalls for whatever reason; despite
there being an rsync timeout, this only relates to being able to connect
via SSH to the node. We've seen an issue where SSH was able to connect
on the port but not establish a connection, causing timeouts and forcing
the number of queued items in the incoming queue to back up. This
ensures that we process the items in a timely manner, even if they're
not copied an asset slave.

As an assurance we do a full sync every morning of the assets to ensure
the master and slaves are in sync.


deanwilson:
Merge away.

PR #4654, opened 2016-07-15
You need to run `vagrant reload` on host change

andrewgarner:
Updates ‘Developing on GDS Virtual Machines’ documentation to explain that `vagrant reload` should be run when making changes to the memory or CPU configuration of the host machine.


djsd123:
I agree.  Looks good too merge.

PR #4655, opened 2016-07-18
Mongodb 3.2 config file

djsd123:
Mongodb's config file format has changed since version 2.6 and will fail to start with without
the correct configuration file.
The parameters are mostly the same between versions so we are using puppet
'versioncmp' method as the conditional argument to apply the right file for
the version.
We added a 2nd template for mongodb version 3.2

### DIFF
--- a/modules/mongodb/manifests/config.pp
+++ b/modules/mongodb/manifests/config.pp
@@ -8,6 +8,7 @@
 #   The config file mongo should use. This varies between mongo 2 and 3.
 #
 # [*dbpath*]
+#   Path to database on filesystem
 #
 # [*development*]
 #   Disable journalling and enable query profiling.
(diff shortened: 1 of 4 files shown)


deanwilson:
Without pooping all over your work... I'd rather a literal value for the template to use be defined in the server class in the current per version puppet code and pass that in as a value to the config class. That keeps all the version differences in one place and isolates the version using code to a single conditional.

So a new value called `$template_name` or similar in https://github.com/alphagov/govuk-puppet/blob/663239b32823bfaaa0ac7aa1869e5215e30cc525/modules/mongodb/manifests/server.pp#L32 and then pass that along.

surminus:
I agree with Dean's sentiment. Otherwise config file looks good!

modules/mongodb/manifests/config.pp
```diff
@@ -4,10 +4,12 @@
 #
 # === Parameters:
 #
+#
```
surminus: This is probably the most nitpicky comment EVER but do we need the extra hash? We already have bare hash

surminus:
:+1: LGTM

djsd123:
FYI, I removed the extra hash :)

PR #4656, opened 2016-07-18
Add unboxed "support" dev team to integration.

elliotcm:
Needed for data dumps etc.


deanwilson:
Merge when green.

jackscotti:
`Failure/Error: expect(key_strength).to be >= 4096, "SSH key for #{user[:name]} is only #{key_strength} bits and must be stronger"`

PR #4657, opened 2016-07-18
Update 2ndline dashboard template with export from Grafana 3

steventux:
Grafana 3 needs a dataSource property specified for each dashboard,
this means dashboards from Grafana 1.9 fail to receive data unless
they are configured to use the Graphite source. This change is a
complete export of a working dashboard cloned in Grafana 3 so there
may be several format differences.


surminus:
:+1: well I guess it looks good

steventux:
Forgot to add it can be tested by importing into a Grafana 3 install

PR #4658, opened 2016-07-18
Update Grafana dashboards to 3.x format

steventux:
Update existing Grafana dashboards to use 3.x Graphite data source.
These are exports from Grafana 3, with the added Graphite data source so the diffs might be a bit confusing.


jackscotti:
I'm not familiar with Grafana: are tests not required for these changes?

rjw1:
Looks fine to me.

PR #4660, opened 2016-07-19
Add dfid-transition worker

rgarner:
[dfid-transition](http://github.com/alphagov/dfid-transition) is a repo containing nothing but rake tasks and sidekiq workers. This PR is for the setup of those workers across environments.

The procfile worker for sidekiq is currently enabled only in integration.

This PR did have a Jenkins job for `load:attachments` but we can just as easily SSH into a backend and run those manually (since we need the ability to use ad-hoc environmental params like `SPARQL_ENDPOINT`.

## Outstanding issue

We're only interested in getting the sidekiq workers running and addressable via `service stop|start`. It looks like an `app` type of `'procfile'` works best here, but we need a port, which I've ended up setting to a rather unsatisfactory 9999 and combining with `enable_nginx_vhost => false`. Dean suggested we might need to use a port registry of which I am blissfully unaware (even if we don't use the port)

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -22,6 +22,7 @@ deployable_applications: &deployable_applications
   - courts-api
   - courts-frontend
   - designprinciples
+  - dfid-transition
   - EFG
   - email-alert-api
   - email-alert-frontend
@@ -289,6 +290,10 @@ govuk::apps::frontend::nagios_memory_warning: 1200
 govuk::apps::frontend::nagios_memory_critical: 1400
 
 govuk::apps::govuk_cdn_logs_monitor::processed_data_dir: '/mnt/logs_cdn/data'
+
+govuk::apps::dfid_transition::redis_host: 'redis-1.backend'
+govuk::apps::dfid_transition::redis_port: '6379'
+govuk::apps::dfid_transition::enable_procfile_worker: false
 
 govuk::apps::publisher::redis_host: 'redis-1.backend'
 govuk::apps::publisher::redis_port: '6379'
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -36,6 +36,7 @@ govuk::node::s_development::apps:
   - 'content_tagger'
   - 'contentapi'
   - 'designprinciples'
+  - 'dfid_transition'
   - 'efg'
   - 'email_alert_api'
   - 'email_alert_frontend'
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -23,6 +23,7 @@ govuk::apps::publicapi::backdrop_host: 'www.preview.performance.service.gov.uk'
 govuk::apps::rummager::aws_s3_bucket_name: govuk-api-elasticsearch-snapshots-integration
 
 govuk::apps::smartanswers::expose_govspeak: true
+govuk::apps::dfid_transition::enable_procfile_worker: true
 govuk::apps::specialist_publisher::publish_pre_production_finders: true
 govuk::apps::specialist_publisher_rebuild::publish_pre_production_finders: true
 govuk::apps::specialist_publisher_rebuild_standalone::enabled: true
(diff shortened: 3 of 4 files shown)


surminus:
I'm going to quickly answer your questions then do a proper review (but a quick scan looks fine):

1. I think there is somewhere that you record ports used, I'll see if I can dig it out. 
2. The difference between the two hieradata structures between `common.yaml` and `development.yaml` is that `common.yaml` is basically listing the apps that will appear in the Jenkins "deploy app" job and so take the names of what's found in `govuk-app-deployment`, and `development.yaml` is listing [which manifests are included](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/node/s_development.pp#L39) which take the form of [the per app puppet class name](https://github.com/alphagov/govuk-puppet/tree/master/modules/govuk/manifests/apps). 
tl;dr they do different things and so it's inconsistent, but doesn't really matter
3. I think by it's very nature a trigger isn't manual. If you wish to run a manual task it's best to create it as a separate task or option in your Jenkins job.

rgarner:
Great, thanks Laura 👍 

On 2, Dean suggested we deploy this only to integration first. Is there a place I should be putting `dfid-transition` to have it deploy only to integration? There doesn't seem to be a natural home in `integration.yaml`.
On 3, would I be ok just omitting the `triggers` section? This is a job we'll want to run once to enqueue a lot of downloads/asset indexing from the wider web, so it genuinely shouldn't have a trigger apart from "we click 'Build'".

surminus:
To be honest, you have to explicitly do deploys so just having the job in the job list on Staging and Production isn't a problem unless there's a rogue developer who wants to deploy it :) 

If it's just a single rake task you wish to run once there is a specific "run rake task" Jenkins job, so perhaps you could just use that to build the stuff you need (if it's an operation that doesn't happen very often)?

rgarner:
Hi Laura, thanks and sorry for not getting back to you yesterday – train broke down and we were waiting in 40ºC + to be towed back to Biggleswade, where they'd laid on ambulances for people suffering from heatstroke. My journey home took 6 hours 😭

These rake tasks need to happen in a relatively ad-hoc fashion at the moment (likely at least several, probably dozens of times) but they talk to publishing-api, rummager and asset_manager. This rake task is consistent with our previous rake task which has to [reference the bearer tokens](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/templates/jobs/dfid_transition_import.yaml.erb#L32-L37) for two of those services. I think that might disqualify the Jenkins "run rake task" job?

Apologies, I'm not sure I understand with reference to deployment – if we want to have this job running only in integration, are you saying that we can just leave the reference in `common.yaml`? Or do we have to do something to this PR?

Thanks again!

tuzz:
For that last point, I think you could add `govuk::apps::dfid_transition::enable_procfile_worker: false` to `common.yml` and then add `govuk::apps::dfid_transition::enable_procfile_worker: true` to `integration.yml`.

tuzz:
I believe @surminus is referring to this rake task:
https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/

It lets you do an ad-hoc run on a rake task on integration. Can you access that?

rgarner:
@tuzz not without merging this first – dfid-transition isn't in that app list yet. But more critically, it doesn't look like that task allows ad-hoc environment variables. We're wanting to vary `SPARQL_ENDPOINT` according to the use case.

tuzz:
Ok, let's login and run the rake task manually for now, in that case.

How do we register an application's port? I've not done that before? Perhaps @surminus or @deanwilson could point us to some documentation? Thanks.

rgarner:
Ok, I've removed the Jenkins job. We'll SSH to run these. I've also (I think properly now) enabled the procfile worker only in integration.

rgarner:
BTW, I don't think the checks ever complete if they're run on a non-`origin` fork. If someone can give me push to this I'll re-open it so the checks will complete :)

PR #4661, opened 2016-07-19
Add latency monitoring for Sidekiq

rjbate:
Add an icinga check for a single (not per-host) metric on the monitoring box for the stats.gauges.govuk.app.rummager.workers.queues.default.latency metric.


deanwilson:
The code looks fine. Worth merging for integration testing. 👍

PR #4663, opened 2016-07-19
Update input raw argument for Gor

surminus:
When testing version 0.14.1 on Staging I discovered that it didn't like
using the interface name "localhost" and threw the following error in
the upstart logs:

```
2016/07/19 17:54:16 Can't find interfaces with addr: localhost. Provide
available IP for intercepting traffic
```

I don't think it hurts to explicitly use the loopback address and this
resolved the error when testing on the host itself.

### DIFF
--- a/modules/router/manifests/gor.pp
+++ b/modules/router/manifests/gor.pp
@@ -32,7 +32,7 @@ class router::gor (
 
   class { 'govuk_gor':
     args   => {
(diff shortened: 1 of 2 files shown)


deanwilson:
This might cause an issue with all our ipv6 traffic but we should be fine for now.

Merge away.

PR #4664, opened 2016-07-20
Set redis env variables for signon

mgrassotti:
Define environment constants for signon, in order to initialize the
govuk_sidekiq gem properly.


deanwilson:
Code looks fine. merge away.

PR #4665, opened 2016-07-20
Enable Gor in Bouncer Staging

surminus:
Upgrading Gor broke traffic replay for Bouncer previously, so we should
enable it in Staging replaying to Integration to see if we can spot
these problems before any upgrades in Production.


jennyd:
🎉  thanks!

deanwilson:
Merge away.

PR #4666, opened 2016-07-20
Rewrite the erb template slightly

deanwilson:
There were a couple of syntax issues, undefined variables and incorrect erb statement syntax,
that stopped this producing the correct values so I've fixed them and revised the spec tests.

This also removes the `profile` option as it doesn't seem to work, or be
mentioned in the mongo config file.


surminus:
:+1: merge when green

PR #4668, opened 2016-07-20
Add dfid transition worker

tuzz:
Reopening from @rgarner's forked branch so that Jenkins will build.

https://github.com/alphagov/govuk-puppet/pull/4660


tuzz:
Merging as @rgarner is the original creator of this pull request.

PR #4669, opened 2016-07-20
Enable anonymous grafana access

steventux:
We're not using authentication for access to Grafana in any serious
way so allow anonymous access. Grafana is only accessible internally
so this makes it usable to radiator screens.


rjw1:
:+1:

PR #4670, opened 2016-07-21
Check existence of lock file before reboot

djsd123:
What
Unattended reboots when invoked, interrupt backups. We don't want to reboot while
a backup is in progress.

How
While running backups we create a lock file under /var/lock.
The lock file only exists for the lifetime of the backup job.  While the lock file exists,
the machine doesn't get rebooted.

### DIFF
--- a/modules/govuk_unattended_reboot/manifests/init.pp
+++ b/modules/govuk_unattended_reboot/manifests/init.pp
@@ -13,10 +13,21 @@
 #   A hash containing a `username` and a `password` to access monitoring
 #   which is protected by basic authentication.
 #
+# [*lock_files*]
+#   An array of lock files owned by jobs that are in progress.
+#   Add full path of the lock file to hiera.
+#
+# === Variables
+#
+# [check_lockfiles*]
+#   Formatted array of lockfiles. Wrapped in doubles quotes and space separated.
+#
 class govuk_unattended_reboot (
   $enabled = false,
   $monitoring_basic_auth = {},
+  $lock_files = []
 ) {
+
 
   validate_bool($enabled)
   validate_hash($monitoring_basic_auth)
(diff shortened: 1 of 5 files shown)


rjw1:
The fix me should be a separate commit if not a pull request.
Th commit message doesnt follow our style guide. "Why" isnt a very good summary :)

modules/govuk_unattended_reboot/manifests/init.pp
```diff
@@ -67,6 +79,16 @@
...
   }
 
+  # This script will check for the existence of any lock files relating to backups in progress as not to interrupt them
+
+  file { "${check_scripts_directory}/03_backups":
+    ensure  => $file_ensure,
+    mode    => '0755',
+    owner   => 'root',
+    group   => 'root',
+    content => template("govuk_unattended_reboot${check_scripts_directory}/backups.erb"),
```
deanwilson: This path is incorrect.

modules/govuk_unattended_reboot/spec/classes/govuk_unattended_reboot_spec.rb
```diff
@@ -35,5 +36,23 @@
...
+      :enabled    => true,
+      :lock_files => [
+          '/var/lock/alvin',
+          '/var/lock/simon',
+          '/var/lock/theodore'
+      ]
+    }}
+
+    context 'It should contain a correctly formatted bash array' do
+      it { is_expected.to contain_file("#{check_scripts_directory}/03_backups").with_content(/("\/var\/lock\/alvin"\s"\/var\/lock\/simon"\s"\/var\/lock\/theodore")/) }
```
deanwilson: You can remove the first set of parens `(` as we're not using the capture.

deanwilson:
This code won't work, due to the way `setlock` handles the lock, and we should go over the details in person.

djsd123:
@deanwilson Hmmm, What if I added a remove lock file command to the tail end of the script?

djsd123:
I added a remove command to the tail end of the script as `setlock` does not remove the file it creates. Rather it reuses the same file again which would cause the server to never reboot.

deanwilson:
After a chat we've decided to change how this works slightly and make the unattended reboot module resposible for creating a lock directory and then blocking the reboot if any files are dropped in by "other" processes. This means we only have to add the lock in the script that wants to do the locking rather than including the information in two places.

PR #4671, opened 2016-07-21
Mongodb backup s3 script

djsd123:
While script was running it would log a warning "unary operator expected".

Compound commands should be wrapped in double braces or both variables should
be wrapped in quotes.


surminus:
:+1: fftm

PR #4676, opened 2016-07-21
Enable the dfid-transition nginx host

rgarner:
We've added the Sidekiq queue-monitoring Rack app for this short
shelf-life project, this lets us see it


rgarner:
Closing – will push direct to this and delete fork

PR #4677, opened 2016-07-21
Add redis location to dfid_transition_import

rgarner:
The rake `load:outputs` job now relies on the `AttachmentIndex`, which is
backed by redis. Without it, the hourly import – which is used to
semi-rapidly iterate and communicate changes in user experience to DFID
– will not work.


tuzz:
merge when green

PR #4681, opened 2016-07-22
Suggest upgrading Vagrant on hostname provisioning error

36degrees:
I was seeing the error in `change_host_name.rb` listed in the README when running Vagrant 1.8.1, and upgrading to 1.8.5 fixed it. Upgrading Vagrant seems like a better solution than editing random gem files, so let's suggest that as the first thing to try.


deanwilson:
Looks good.

PR #4683, opened 2016-07-22
Upgrade Gor to 0.14.1

surminus:
This was reverted in e454b9a0442e7e5079137de270954da86112dd9e, but now
after some testing I think it is ready to try again.

This will fully upgrade Gor for both the cache servers and bouncer, and
also update some options that the new version of Gor requires, along
with adding the '-http-original-host' setting which keeps the original
Host header in place (required for bouncer).

It also keeps Gor running in Staging on cache-1 and bouncer-1 which
replays traffic to Integration.

### DIFF
--- a/modules/govuk_bouncer/manifests/gor.pp
+++ b/modules/govuk_bouncer/manifests/gor.pp
@@ -27,13 +27,17 @@ class govuk_bouncer::gor (
   }
 
   class { 'govuk_gor':
-    args   => {
+    args    => {
       '-input-raw'          => ':80',
       '-output-http'        => $gor_targets,
-      '-output-http-method' => [
+      '-http-allow-method'  => [
         'GET', 'HEAD', 'OPTIONS',
       ],
+      '-http-original-host' => '',
     },
-    enable => $gor_enable,
+    envvars => {
+      'GODEBUG' => 'netdns=go',
+    },
+    enable  => $gor_enable,
   }
 }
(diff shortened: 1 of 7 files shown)


modules/govuk_bouncer/manifests/gor.pp
```diff
@@ -27,13 +27,17 @@
...
       '-output-http'        => $gor_targets,
-      '-output-http-method' => [
+      '-http-allow-method'  => [
         'GET', 'HEAD', 'OPTIONS',
       ],
+      '-http-original-host' => '',
     },
-    enable => $gor_enable,
+    envvars => {
+      'GODEBUG' => 'netdns=go',
```
tuzz: A quick google reveals that this is the default, I think? Would it make sense to simplify this manifest and remove this environment variable?

deanwilson: We were seeing failures when this was not explicit.

tuzz:
Looks sane. Is this something we should add explicit monitoring for?

rjw1:
Merge when ready.

deanwilson:
@tuzz we have https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_gor/manifests/init.pp#L74 as a basic check.

surminus:
I've removed the reference to removing `/usr/bin/gor` as it doesn't exist on a server with a new package. 

I will deploy this to Staging and make sure we're seeing the right stuff replayed, and then move to Production for the real test.

PR #4684, opened 2016-07-22
Add dfid-transition hostname

rgarner:
Add dfid-transition as a hostname so we can see
the Sidekiq queue monitor at
https://dfid-transition.integration.publishing.service.gov.uk/


tuzz:
merge when green

PR #4685, opened 2016-07-22
Add dfid transition to load balancer

rgarner:
Was missing and still unreachable. Restrict this to integration only while we're here.


tuzz:
merge on green

PR #4686, opened 2016-07-22
Rework request time per endpoint graph

steventux:
There was a bit of duplication in the request time graph on the publishing api dashboard, this
was the result of an attempt to aggregrate times by request id.
Instead just graph @fields.duration for every v2 endpoint.


deanwilson:
Merge when ready.

PR #4689, opened 2016-07-22
Extra disk for s3backups

djsd123:
What
It has been agreed that backups should be temporarily stored on separate partitions as
to not cause any issues for other services should the backup process somehow use
up all it's disk space.

How
Create separate mounts for backups.  Here I have added Logical Volume Manager
config for Mongodb instances performing s3backups.

### DIFF
--- a/hieradata/class/api_mongo.yaml
+++ b/hieradata/class/api_mongo.yaml
@@ -12,6 +12,9 @@ lv:
   data:
     pv: '/dev/sdc1'
     vg: 'mongodb'
+  s3backups:
+    pv: '/dev/sde1'
+    vg: 'mongo'
 
 mount:
   /var/lib/mongodb:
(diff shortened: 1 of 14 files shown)


surminus:
Do mounts needed to be added for other node classes? I'm tired so sorry if I'm missing stuff. 

I'm going to try to unblock this now.

djsd123:
No, You're correct. It probably makes sense to add these to a base mongo class if one exists.

deanwilson:
Looks fine to me but it's not part of the system i know about so I'm hesitant to say merge without another set of eyes. @surminus / @alexmuller can we get another +1 please?

alexmuller:
I think the disks still need to be created on the machines - for example `/dev/sde1` doesn't exist on the `mongo-n` machine which will cause Puppet to fail.

djsd123:
Context to Alex's comment.  I'll need to create a new PR on the provisioning repo to add disks for staging and prod.

djsd123:
So https://github.gds/gds/govuk-provisioning/pull/558 has now been merged. Before this gets deployed we'll need to add the disks in vCloud director.

deanwilson:
Disks have been added. Merging.

PR #4693, opened 2016-07-22
Bouncer Gor bugfix

surminus:
After upgrading to the new version of Gor, bouncer stopped replaying traffic. It seems that listening on the nginx port does not replay traffic, but to do so on the unicorn port works as expected. I have not deduced why this is yet.

In the screen shot below, Puppet is deployed at roughly 14:40 and the machines update themselves to the new version. Traffic significantly falls off (the remaining traffic is for healthchecks and automated tests). At about 15:50 I change the port it's listening on a single bouncer machine, and a small amount of traffic starts to re-appear. At 16:30 I enable this across all machines, leaving it enabled with Puppet disabled.

screen shot 2016-07-22 at 17 45 55

### DIFF
--- a/modules/govuk_bouncer/manifests/gor.pp
+++ b/modules/govuk_bouncer/manifests/gor.pp
@@ -28,7 +28,7 @@ class govuk_bouncer::gor (
 
   class { 'govuk_gor':
     args    => {
-      '-input-raw'          => ':80',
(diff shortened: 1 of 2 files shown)


deanwilson:
Merge as a temp fix but we should investigate this more.

PR #4697, opened 2016-07-25
Add brute-force option to pg_resync_slave

surminus:
This script fixes replication on a Postgres slave. It works very well,
other than it copies the data directory before restoring, which means we
need to have two full copies of the data directory on disk for the
restore. If the database is large, and we're tight on disk space, then
we can add a brute force option to delete the data directory and then do
the restore.


rjbate:
👍

PR #4699, opened 2016-07-27
Add `content-tagger` to backend

gpeng:
This commit adds `content-tagger` to the `hosts::production::backend::app_hostnames:` in
hierdata/common.yaml. Signon has started timing out when trying to update users on the Content Tagger app. This appears to be caused by a missing entry in `/etc/hosts`. DevOps-ing this fixed the issue.


tijmenb:
👍 sorry we missed this when setting up.

PR #4700, opened 2016-07-27
Removing cronjob after it was replaced in a previous PR

djsd123:
I renamed this cron job in another PR which has been merged.  Although there are no errors caused by having two of the same cronjob. It's best to remove.


rjbate:
MWG

PR #4703, opened 2016-07-27
Decrease amount of indices to keep

surminus:
We have a lot of data, too much for our current elasticsearch setup. To keep it ticking over while we reconfigure our elasticsearch stuff we can decrease the number of days that we keep our logs for, from 15 down to 13.

This also turns it to a parameter in a manifest from being hardcoded in the template because it's easier to find within the s_logs_elasticsearch manifest itself.


rjbate:
👍

PR #4704, opened 2016-07-28
Licensing: Remove automongobackup 

djsd123:
What
It's been decided that we will only implement s3 backups on the new licensing
Mongo vms running Mongodb 3.2.  We already have a PR to add separate mounts for s3backups.
This PR is to remove the `/dev/mapper/backup-mongodb` mount and the mongodb::backup class  from the Licensing mongo vms.

How
Remove the hiera config from this (puppet) repo that provisions the logical volumes/groups.
Remove the hiera config that provisions the disks in provisioning repo. (separate PR)
Remove mount resource from base class. Exclude mongodb::backup class.

It's agreed that the existing vms will be destroyed and new vms provisioned.

### DIFF
--- a/hieradata/class/licensing_mongo.yaml
+++ b/hieradata/class/licensing_mongo.yaml
@@ -12,9 +12,6 @@ hosts::production::licensify::app_hostnames:
   - 'licensing-web-forms'
 
 lv:
-  mongodb:
-    pv: '/dev/sdb1'
-    vg: 'backup'
   databases:
     pv: '/dev/sdc1'
     vg: 'mongodb'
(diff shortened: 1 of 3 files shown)


surminus:
Do we need to create a mount for S3 backups directory? Might be better to replace the current setup.

As with all disk related things, it might be easier to spin up all these machines afresh since they currently have no data?

surminus:
I'm just wondering how and/or if Puppet will handle removing/renaming VGs and PVs and whatnot. I think we could just add the new mount now and recreate these machines quickly to handle that. 

Basically, when it comes to storage stuff, I see problems.

deanwilson:
Code looks fine. I agree with @surminus though. It'll probably be easier to rebuild the hosts rather than fiddle with the LVM configs in hiera.

djsd123:
Yep, That's the intention.  Can I note that this PR is to remove automongo backup resources only.  I stated in the message above that these machines will be re provisioned. See https://github.com/alphagov/govuk-puppet/pull/4689

PR #4709, opened 2016-07-28
Add a new panel to the Publishing API Metrics dashboard

cbaines:


modules/grafana/templates/dashboards/publishing_api_overview.json.erb
```diff
@@ -829,6 +829,91 @@
...
+      "panels": [
+        {
+          "aliasColors": {},
+          "bars": false,
+          "datasource": "Graphite",
+          "editable": true,
+          "error": false,
+          "fill": 1,
+          "grid": {
+            "threshold1": null,
```
cbaines: I would love to get the variables here [1] in to this template, but after far too long trying to work out how puppet works, I've had to give up, but if someone has any advice, please let me know :)

1: https://github.com/alphagov/govuk-puppet/blob/86350f0a6e2ebc9e4be9c1df96e24739b63d5ddf/modules/govuk/manifests/node/s_redis_base.pp#L17-L18

deanwilson: Puppet templates are ruby ERB files under the hood. You'd probably want something like this

```<%
  redis_max_memory = (@memorysize_mb / 4 * 3).floor

  redis_mem_warn = $redis_max_memory * 0.8
  redis_mem_crit = $redis_max_memory * 0.9
%>

Literal string = <%= redis_mem_warn %>
```

Which is a duplication of the code that works out the values but is also much easier than trying to get the values from redis to a graphing host

cbaines: Does the `@memorysize_mb` then come from the machine with the Grafana templates on, and not the machine running redis?

deanwilson: Ah, yep. You're correct. We'd have to do an explicit puppetdb lookup or export a resource to link them together. Which is a lot more work.

cbaines: I believe the desired values are already exported through the icinga::check::graphite resources [1] as the warning and critical attributes, but... I'm still no closer to a solution, I just can't work out what "realize"'ing them in the s_redis_base class does, and how that helps getting closer to accessing the value of the attributes... puppet is making me sad :(

1: https://github.com/alphagov/govuk-puppet/blob/86350f0a6e2ebc9e4be9c1df96e24739b63d5ddf/modules/govuk/manifests/node/s_redis_base.pp#L20

PR #4710, opened 2016-07-28
Licensing s3 mount

djsd123:
Hiera config for s3backup mount on licensing mongo servers.


surminus:
:+1:

djsd123:
mikael allison [1:44 PM]  
@lauram: , @dwilson , @russ 
The licensing mongos initially provisioned nicely. 
Only one unexpected event so far was that the mongodb-databases mount was mounted on the disk intended for the s3backups invariably leaving it's intended disk available.  This doesn't appear to be a problem as far as I can tell.

rjw1:
merge when ready

PR #4712, opened 2016-07-28
Revert "Add brute-force option to pg_resync_slave"

surminus:
This reverts commit 4956a81ecc5a8437d8b5872c93eae96d56cbab55.

This did not work like I was expecting and I don't want it to be
deployed out. Needs more testing.


MatMoore:
👍

PR #4713, opened 2016-07-29
Enable dfid_transition in staging/production

tuzz:


hieradata/integration.yaml
```diff
@@ -23,8 +23,6 @@ govuk::apps::publicapi::backdrop_host: 'www.preview.performance.service.gov.uk'
 govuk::apps::rummager::aws_s3_bucket_name: govuk-api-elasticsearch-snapshots-integration
 
 govuk::apps::smartanswers::expose_govspeak: true
-govuk::apps::dfid_transition::enabled: true
-govuk::apps::dfid_transition::enable_procfile_worker: true
```
rgarner: Not sure I'd want to remove these yet; if there are any problems this is the least contentious available production-like environment in which to conduct tests. Can we leave this for a post-live PR?

tuzz: dfid_transition will still be enabled in integration. I've set these to `true` in common.yml which this environment inherits from.

rgarner: Ignore me. `common.yaml`.

rgarner:
LGTM :+1:

PR #4714, opened 2016-07-29
Mongodb s3 backups 1

djsd123:
What
Inconsistent errors while performing s3 backups due to trailing slashes not being present at the end of bucket URLs.
As a result of of the errors, the script never gets to a point where it can cleanup the tarball(s) it
creates causing the host to run low on disk space.
s3 backups are gated in the base backup class but are currently deployed and performed on every provisioned host with mongodb in integration.

How
Add trailing slash to bucket directories/URLs as manual tests show the script works with or without trailing slash.
Implement trap in bash script which calls a function that performs housekeeping regardless of the exit status.
Removed gated boolean value from common integration data. Applied the value in node specific files in hiera.


surminus:
:+1: LGTM

PR #4716, opened 2016-07-29
Resync Whitehall scheduled queue after data sync

surminus:
The scheduled queue goes out of sync after a full data sync, and this rake task needs to be run to resync it again. 

/cc @gpeng


gpeng:
:+1:

PR #4717, opened 2016-07-29
Licensing s3 mount 2

djsd123:
This change was left behind in a previous PR https://github.com/alphagov/govuk-puppet/pull/4689 and is required to mount /var/lib/s3backup to the newly provisioned disk.


surminus:
:+1: mwg

PR #4719, opened 2016-07-29
Licensing s3 mounts 3

djsd123:
We want the mapping for the s3 mount to be `/dev/mapper/mongo-s3backups`. My previous PR contained incorrect disk mappings. 
This has been corrected to provision `/dev/mapper/'volume_group'-'logical_volume'` as stated before.


surminus:
:+1: fftm

PR #4720, opened 2016-08-01
Remove Ben Hyland account

djsd123:
De-provisioning Ben's account as he has moved to Gaap.

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -52,7 +52,6 @@ govuk_jenkins::config::admins:
   - alextorrance
   - andrewgarner
   - anikahenke
-  - benhyland
   - bobwalker
   - brendanbutler
   - carlosvilhena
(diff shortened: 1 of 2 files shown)


deanwilson:
Can we add a `FIXME:` to remove the absent user once the deletes have gone through please. Other than that Merge away.

alexmuller:
Users are purged so we can just delete the file without ensuring it's absent: https://github.com/alphagov/govuk-puppet/blob/0db6b25636c74f970852000efd6d0f31bd3ebdaa/modules/users/manifests/init.pp#L38

deanwilson:
This will leave the SSH key in the home directory.

alexmuller:
:+1:

djsd123:
@deanwilson , @alexmuller , Think I should heed Aaron's words that we should have a leaver's ticket from "Finding Things".

PR #4721, opened 2016-08-01
Check existence of lock file before reboot

djsd123:
What
Unattended reboots when invoked, interrupt backups. We do not want to reboot the host while
a backup is in progress.

How
When a backup is initiated we create a file under /etc/unattended-reboot/no-reboot.
The file should only exists for the duration of the job. While the file exists,
the machine does not reboot.

This relates to https://github.com/alphagov/govuk-puppet/pull/4670

### DIFF
--- a/modules/govuk_unattended_reboot/manifests/init.pp
+++ b/modules/govuk_unattended_reboot/manifests/init.pp
@@ -13,6 +13,13 @@
 #   A hash containing a `username` and a `password` to access monitoring
 #   which is protected by basic authentication.
 #
+# === Variables
+#
+# [*lock_dir*]
+#   path => /etc/unattended-reboot/no-reboot/
+#   The directory in which a process places a file while in progress to
+#   prevent the host from rebooting.
+#
 class govuk_unattended_reboot (
   $enabled = false,
   $monitoring_basic_auth = {},
@@ -57,6 +64,24 @@ class govuk_unattended_reboot (
     group   => 'root',
     content => template("govuk_unattended_reboot${check_scripts_directory}/01_alerts.erb"),
     require => File['/usr/local/bin/check_icinga'],
+  }
+
+  $lock_dir = "${config_directory}/no-reboot"
+
+  file { $lock_dir:
+    ensure => directory,
+    mode   => '0777',
+    owner  => 'root',
+    group  => 'root',
+  }
+
+  file { '03_no_reboot':
+    ensure  => file,
+    path    => "${check_scripts_directory}/03_no_reboot",
+    mode    => '0755',
+    owner   => 'root',
+    group   => 'root',
+    content => template('govuk_unattended_reboot/03_no_reboot.erb'),
   }
 
   file { '/usr/local/bin/check_icinga':
--- a/modules/govuk_unattended_reboot/templates/03_no_reboot.erb
+++ b/modules/govuk_unattended_reboot/templates/03_no_reboot.erb
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -e
+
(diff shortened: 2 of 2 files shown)


modules/govuk_unattended_reboot/manifests/init.pp
```diff
@@ -59,6 +66,26 @@
...
+    ensure  => directory,
+    mode    => '0777',
+    owner   => 'root',
+    group   => 'root',
+    require => File[$config_directory],
+  }
+
+  file { 'no_reboot':
+    ensure  => file,
+    path    => "${check_scripts_directory}/no_reboot",
```
surminus: Should this be named with a numbered prefix to match [what we already do](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_unattended_reboot/manifests/init.pp#L50)?

modules/govuk_unattended_reboot/manifests/init.pp
```diff
@@ -59,6 +66,26 @@
...
+  }
+
+  file { '03_no_reboot':
+    ensure  => file,
+    path    => "${check_scripts_directory}/03_no_reboot",
+    mode    => '0755',
+    owner   => 'root',
+    group   => 'root',
+    content => template('govuk_unattended_reboot/no_reboot.erb'),
+    require => File[$check_scripts_directory],
```
deanwilson: This `require` can go too.

deanwilson:
Should the shell script (in the `erb` file) give a reason the reboot has not happened? or do we exit silently in the existing stuff? @alexmuller

modules/govuk_unattended_reboot/templates/03_no_reboot.erb
```diff
@@ -0,0 +1,22 @@
...
+set -e
+
+# Redirect stdout and stderr to syslog
+exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
+
+LOCK_FILES=$(ls -A <%= @lock_dir %>)
+
+  # If any files exist within the lock directory. The Server will not reboot
+
+  if [[ $LOCK_FILES ]]; then
```
surminus: Does this the right status code when there are files in the directory compared to when there are no files in the directory? 

```
lauramartin@xenial-test-1:~$ mkdir foo
lauramartin@xenial-test-1:~$ ls -A foo/
lauramartin@xenial-test-1:~$ echo $?
0
lauramartin@xenial-test-1:~$ touch foo/bar
lauramartin@xenial-test-1:~$ ls -A foo/
bar
lauramartin@xenial-test-1:~$ echo $?
0
```

djsd123:
Sorry Just had to check.  So for example:

`#!/bin/bash

set -e


LOCK_FILES=$(ls -A $1)


  if [[ $LOCK_FILES ]]; then

    echo "exit 1"
    exit 1

  else

    echo "exit 0"
    exit 0

  fi
`

djsd123:
Sorry had to make sure. So for example:

`#!/bin/bash`
`LOCK_FILES=$(ls -A $1)`
` if [[ $LOCK_FILES ]]; then`
`     echo "exit 1"`
`     exit 1`
`else`
`     echo "exit 0"`
`     exit 0`
`fi`

The I test it works:

`vagrant@ubuntu1:~$ mkdir foo`
`vagrant@ubuntu1:~$ mkdir bar`
`vagrant@ubuntu1:~$ touch foo/en_flique`
`vagrant@ubuntu1:~$ ./test.sh foo`
`exit 1`
`vagrant@ubuntu1:~$ ./test.sh bar/`
`exit 0`
`vagrant@ubuntu1:~$`

Hope I'm making some sense!?

PR #4723, opened 2016-08-01
Mongodb s3 backup 2

djsd123:
In my last PR where I added a function to tidy up on exit. I inadvertently trumped the Nagios passive check.
This change will call both functions as the script exits, so that the Nagios passive check
runs as well as the tidy up.


deanwilson:
Merge away.

PR #4724, opened 2016-08-02
Tidy old Ruby versions

alexmuller:
I don't understand why we need to use `defined()` for this. The `rbenv::version` type has an absent parameter which purges a package and removes a directory [1].

[1]: https://github.com/gds-operations/puppet-rbenv/blob/v1.2.0/manifests/version.pp#L72-L84

I think the intention was that new Ruby versions should be added to cleanup at the same time they are installed but in practice that doesn't happen (5014ad4, 6ff24fe).

Instead I think we can set the classes to `ensure => absent` when we want to remove them like we do for other stuff.


deanwilson:
Looks fine and makes sense. Merge away.

PR #4725, opened 2016-08-02
Mongodb s3 backup 3

djsd123:
What
This realtes to https://github.com/alphagov/govuk-puppet/pull/4721
The directory in which we touch a file to not only prevent the cronjobs from running asynchronously, but to also prevent the host from being rebooted while a cronjob is in progress has been modified.
Remove FIXME comment and relating cron resource.

How
Modified cron resource to touch file in the no-reboot directory.
Added comment explaining cron resource command.
Added clean up action for touched file to the s3 backup script for when the scripts exits.
Removed FIXME comment and relating cron resource.


surminus:
:+1: happy to merge this once assuming the unattended reboots thing has also been merged

PR #4726, opened 2016-08-02
Add nsca proc check

deanwilson:
We use this service so we should ensure it's running.

As per trello - https://trello.com/c/tZY8weAc


surminus:
:+1: mwg

PR #4728, opened 2016-08-03
Fix the links on the 2nd line health Grafana dashboard

cbaines:
These were broken with a Grafana update.


tijmenb:
👍 good to merge

PR #4729, opened 2016-08-03
Fix rummager grafana dashboards

steventux:
Grafana dashboards for Rummager are throwing javascript errors due to the upgrade from 1.9.x to 3.x
_something something y axis_

To fix I have added the expected properties locally, this made the dashboards visible, then updated the data sources and exported again. This _seems_ to make the dashboards appear properly again, see screenshots.

screenshot from 2016-08-03 14 11 22
screenshot from 2016-08-03 14 12 50

cc @tijmenb


tijmenb:
👍 MWG. Finding Things thanks you!

PR #4730, opened 2016-08-03
Exclude bouncer 4XXs from Edge health graph

boffbowsh:
We care less about Bouncer 4XXs, they're mostly noise against
origin 4XXs which are more likely to be real errors.

/cc @jennyd @tijmenb


tijmenb:
I approve 👍

alexmuller:
I understand the reasoning for this change but I'm still a little concerned about it. Not sure what the right thing to do is.

boffbowsh:
We could perhaps add another gauge/graph for bouncer-specific 4XXs but, I think it'd still just be noise.

PR #4731, opened 2016-08-03
Configure environment for Sidekiq monitoring

andrewgarner:
Ensures that the Redis config for all applications that use Sidekiq is stored in the environment.

The configuration is then made available to the Sidekiq monitoring application so that we can ensure the monitoring tools point to the same Redis configuration as the application.

Ticket: https://trello.com/c/Qa02qGyb/460-fix-sidekiq-monitoring-medium

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -247,8 +247,8 @@ govuk::apps::bouncer::db_hostname: "transition-postgresql-slave-1.backend"
 govuk::apps::bouncer::nagios_memory_warning: 1400
 govuk::apps::bouncer::nagios_memory_critical: 1500
 
-govuk::apps::collections_publisher::redis_host: 'redis-1.backend'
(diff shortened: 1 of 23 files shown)


hieradata/common.yaml
```diff
@@ -1210,6 +1250,9 @@ router::nginx::robotstxt: |
   User-agent: deepcrawl
   Disallow: /
 
+sidekiq::host: 'redis-1.backend'
```
boffbowsh: Can someone in @alphagov/team-gov-uk-infrastructure-admins check this is ok please?

andrewgarner: I've spoken to @alexmuller and agreed we should remove the namespace so as not to cause confusion that this could be related to a Sidekiq module. This follows the convention set nu `app_domain`, `deployable_applications` etc. See c179cbbdab20247d10b1663c78b00183119de9fc

PR #4739, opened 2016-08-05
Revert govuk puppet locks stuff

surminus:
Pushing this into Integration caused problems, so I'm rolling back because I think it needs better testing and I don't want it to block deployments of Puppet for a functionality that we can easily live without.


cbaines:
:+1:

surminus:
I think this is working fine so would be happy to push to Staging/Prod with this stuff reverted.

PR #4740, opened 2016-08-05
Add env vars for Signon to Transition

alexmuller:
These aren't going to be used yet, the file is still overwritten on deployment.


tijmenb:
👍

PR #4743, opened 2016-08-05
Ensure the dfid_transition_import job is removed

rgarner:
This only ever ran on integration and was for samples of data. Now we've
gone live we're part of Data_Sync and see production data every day.


boffbowsh:
This has lint errors

rgarner:
Now fixed, thanks @boffbowsh

PR #4745, opened 2016-08-05
Elasticsearch Repo Housekeeping

djsd123:
Removing FIXME comment to include housekeeping class when applying backup class.
Modified class params names.


cbaines:
Its not clear from the commit message why these changes are happening, but I'll merge as the tests pass.

PR #4747, opened 2016-08-08
Fix Travel Advice Publisher Sidekiq Monitoring

andrewgarner:
Adds missing location to the Nginx config.

Port is defined in https://github.com/alphagov/sidekiq-monitoring/blob/master/Procfile#L11


tijmenb:
👍 MWG

PR #4748, opened 2016-08-08
Grafana dashboard fixes

cbaines:


modules/grafana/templates/dashboards/vpn_diagnostics.json.erb
```diff
@@ -1,486 +0,0 @@
-{
```
boffbowsh: This will need removing from here https://github.com/alphagov/govuk-puppet/blob/9c036f2a6b8d69dc545f31432ba6a4b03efc7ce5/modules/grafana/manifests/dashboards.pp#L34

PR #4750, opened 2016-08-08
Update Transition PostgreSQL servers

surminus:
This is a piece of work which came out of the potential move to the Government PaaS where they're currently running PostgreSQL 9.5. We decided it would be a good reason to upgrade our Puppet module and version of PostgreSQL. On Ubuntu LTS 16.04 (Xenial), the default version of PostgreSQL is 9.5 so it made sense to also try and upgrade that, too. 

We _also_ took this opportunity to rename the Transition PostgreSQL servers to primary/standby rather than master/slave, which is consistent with our other PostgreSQL machines.


surminus:
Tests are failing after the upgrade to the puppetlabs postgresql module. I haven't been able to figure out why (the error is related to a call to `concat::fragment`), but it is possible to get the tests to pass if you explicitly pass in `target => 'foo'` [here](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_postgresql/manifests/db.pp#L123-L128)

PR #4651, opened 2016-07-15
[DO NOT MERGE] Install Ruby 2.3.1

gpeng:
This adds the latest available stable ruby.

PR #4755, opened 2016-08-10
Add ruby 2.3.1

h-lame:
Fix the Gemfile.lock and add ruby 2.3.1 (slightly more info in each commit).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -126,6 +126,3 @@ DEPENDENCIES
   rspec-puppet
   sshkey (= 1.7.0)
   webmock (~> 1.20.0)
-
-BUNDLED WITH
-   1.11.2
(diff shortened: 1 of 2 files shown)


andrewgarner:
This will also need a Ruby 2.3.1 package being created and added to our repository. @gpeng can probably explain better than I can.

h-lame:
Oh :( I hoped that rbenv would "just know", thanks for the heads up.

andrewgarner:
https://github.gds/pages/gds/opsmanual/infrastructure/howto/adding-a-ruby-version.html

h-lame:
Adding DNM tag while I build the ppa.

alexmuller:
The opsmanual instructions are for debuild but you may find it more simple to build it with fpm. Either will work.

alexmuller:
Also see #4651 :)

alexmuller:
I think this is ready to go out now:

```
fab $environment puppet_class:govuk::node::s_app_server do:'apt-cache policy rbenv-ruby-2.3.1'
```

You could use a newer version of Bundler if you wanted to but either way :+1:

elliotcm:
@h-lame @alexmuller This is causing problems in production for our EFG servers which seem to not have the package `rbenv-ruby-2.3.1` available to them.

PR #4756, opened 2016-08-10
Fix the memory used metrics to work on all environments

cbaines:
Suffixes (e.g. _integration) are used on staging and integration, so add a
wildcard to handle this.


danielroseman:
👍

PR #4759, opened 2016-08-11
Mongo s3 restore 5

djsd123:
The mongodb::s3backup::restore class is gated in mongodb::backup.
To enable it we need to override the class param to true.


alexmuller:
Is there any reason for choosing `mongo-3` specifically? Either way please can you put in the commit message that there is/isn't a reason so we can refer back to it?

djsd123:
Mob says This fine to merge.

PR #4763, opened 2016-08-11
Set environment variables for Stagecraft creds

alexmuller:
These creds are already defined in the deployment repo so we can make use of them with hiera lookups.


djsd123:
👍

PR #4765, opened 2016-08-12
Run licensify data sync daily instead of weekly.

elliotcm:
We're doing some work to improve these apps and
having more up to date data will help that process.


alexmuller:
Looks correct to me :+1:

PR #4766, opened 2016-08-12
Add CSV mime type to nginx

sihugh:
Local-links-manager generates and serves static CSVs.  In testing on Integration, we noticed that the CSVs were served with a `Content-Type` of `text/plain` where they should have been `text/csv`.

https://tools.ietf.org/html/rfc4180#page-4

### DIFF
--- a/modules/nginx/files/etc/nginx/mime.types
+++ b/modules/nginx/files/etc/nginx/mime.types
@@ -7,6 +7,7 @@ types {
 	application/x-javascript		js;
 	application/atom+xml			atom;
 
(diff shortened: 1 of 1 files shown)


sihugh:
Closing in favour of other approaches

sihugh:
Reopening as we like this again! 👍

alexmuller:
I think this change is good but somebody else should probably have a look too :+1:

sihugh:
I've checked that this type wasn't previously present (and then removed for some reason) by running `git log -p -S "text/csv"`.

PR #4768, opened 2016-08-12
Permit GovDelivery topic overwriting in staging and integration.

elliotcm:
Allows us to keep the integration/staging account
of GovDelivery in sync with production, making QA
easier.

See https://github.com/alphagov/email-alert-api/blob/master/lib/tasks/data_hygiene.rake#L23-L29
and https://trello.com/c/PTDavZWZ/1-sync-integration-and-staging-topic-ids


alexmuller:
This looks like it's the right way round to me :+1:

PR #4771, opened 2016-08-16
Update crawler dependencies

tijmenb:
This will make `govuk_seed_crawler` use the sitemap instead of contentapi. See https://github.com/alphagov/govuk_seed_crawler/pull/3

cc @alexmuller


alexmuller:
Looks correct to me :+1:

You could do an `ensure => absent` to clean up the old mirrorer gem but either way I think this is fine to merge :)

PR #4773, opened 2016-08-16
Don't phone for travel advice when GitHub is down

alexmuller:
If GitHub is unavailable the travel advice check will fail which will notify PagerDuty.

Instead we use the TextFinder Jenkins plugin to check for output that indicates GitHub is down, and use that to mark the build as unstable.

/cc @boffbowsh


boffbowsh:
I owe you things 😍

PR #4774, opened 2016-08-17
Add more PostgreSQL backup space

surminus:
Add another disk to increase the amount of backup space on postgresql_primary.


djsd123:
👍

PR #4776, opened 2016-08-17
Govuk Release app Postgresql backend

djsd123:
What
Govuk is moving away from MySql in favour of Postgres. This relates to https://github.com/alphagov/release/pull/110

How
Using the govuk_postgresql defined type, I have added the config parameters for the
release app's database configuration file.


djsd123:
Note:
Class params; postgresql_release and backend_ip_range need to be defined.

alexmuller:
I wonder if we should keep the MySQL database defined in Puppet until we're ready to destroy it (after the data has been migrated and we're running PostgreSQL in production).

boffbowsh:
We should definitely do that thing that Alex said

deanwilson:
Other than the horribly named `$mysql_release` this looks ok to my naive eyes. @alexmuller / @surminus mergeable?

Do we need two separate passwords or should we keep them insync between the two DB engines and have a release_db_password value instead of different ones for mysql and postgres?

surminus:
I think this looks mergeable. Having just one password might be tidier, but I'm not fussed either way.

alexmuller:
Going to close this for the time being because things have changed in the time since this was done.

PR #4780, opened 2016-08-17
Revert "Update Jack's key"

jackscotti:
I had to return the new laptop and I will get another one. For now, I need my old key back here. 😑 

Reverts alphagov/govuk-puppet#4778


Davidslv:
👍

PR #4784, opened 2016-08-18
Remove purged package reference

surminus:
For whatever reason this was making my local tests fail. It's been
deployed since it was added in 586a64ff9b70ac41c11837726bd4c60ee8c77f2f
so this should be safe to remove.


surminus:
:+1: reviewed by @deanwilson over my shoulder.

PR #4785, opened 2016-08-19
Show the reason Puppet is disabled in Icinga

alexmuller:
This commit changes the check of the Puppet agent to also check whether there's a lockfile on disk which is preventing Puppet from running.

Unfortunately this means that we need to make a fairly large change to `govuk_puppet` which allows all exit codes to let the script continue rather than just 0 or 2.


deanwilson:
This looks mergeable as is.

As an extra discussion - should this also look in the same places as the revised `govuk_puppet` to catch when it's disabled via one of the batch job lock files?

PR #4786, opened 2016-08-19
Upgrade Transition PostgreSQL to 9.5

surminus:
We want to move Transition to the PaaS, which is using PostgreSQL 9.5 by default. This would be a good opportunity to move to new machine names and classes using the primary/standby naming convention in use by our other machines.

It's also an opportunity to upgrade the puppetlabs-postgresql module.


modules/govuk_postgresql/manifests/server/primary.pp
```diff
@@ -42,12 +42,17 @@
...
-    'checkpoint_segments':
-      value => 8;
     'wal_keep_segments':
       value => 256;
   }
 
+  if $::postgresql::globals::version < '9.5' {
+    postgresql::server::config_entry {
+      'checkpoint_segments':
+        value => 8;
```
deanwilson: this can be a comma now as it's an extracted resource.

modules/govuk_postgresql/manifests/server/primary.pp
```diff
@@ -42,12 +42,17 @@
       value => 'hot_standby';
     'max_wal_senders':
       value => 3;
-    'checkpoint_segments':
-      value => 8;
     'wal_keep_segments':
       value => 256;
   }
 
+  if $::postgresql::globals::version < '9.5' {
```
deanwilson: We'll eventually have to convert this to a https://docs.puppet.com/puppet/latest/reference/function.html#versioncmp for when we move to puppet 4. I'm fine leaving it for later though.

Puppetfile
```diff
@@ -12,7 +12,7 @@ mod 'puppetlabs/concat'
 mod 'puppetlabs/gcc'
 mod 'puppetlabs/java_ks'
 mod 'puppetlabs/mysql'
-mod 'puppetlabs/postgresql', '4.1.0'
+mod 'puppetlabs/postgresql', '4.8.0'
```
deanwilson: should we remove the old versions tarball in this PR if we're adding the new one?

deanwilson:
Other than a few comments this looks good to merge.

PR #4787, opened 2016-08-19
Add PostgreSQL mirror

surminus:
This would allow us to upgrade to PostgreSQL 9.5 on Trusty.


modules/govuk/manifests/node/s_apt.pp
```diff
@@ -93,7 +93,12 @@
       location => 'http://repo.percona.com/apt',
       release  => 'trusty',
       key      => '1C4CBDCDCD2EFD2A';
+    'postgresql':
+      location => 'http://apt.postgresql.org/pub/repos/apt/',
+      release  => 'trusty',
```
alexmuller: Does the release need `-pgdg` as a suffix? https://wiki.postgresql.org/wiki/Apt

surminus:
@alexmuller great spot, thanks!

PR #4788, opened 2016-08-19
Revert "Added routing to make prep prod finders be handled by special…

tuzz:
We ran out of time for our deploy and the rest of the
day is booked up, so we need to roll these changes back.


carvil:
looks good to me 👍

PR #4791, opened 2016-08-22
Blacklist DFID research outputs from the mirrors

alexmuller:
URLs under this path have filenames which are too long, which causes errors in the crawler worker when writing to disk.

Blacklist this path temporarily until the publisher has been fixed to limit URLs to 255 characters.

@tuzz: Please would you add something to the story in your backlog to revert this change once the fix is complete?


deanwilson:
Merge away.

tuzz:
Added: https://trello.com/c/PVT5Fnpd/255-truncate-over-length-dfid-urls-medium

alexmuller:
Thanks!

PR #4792, opened 2016-08-23
Integration data sync complete: clear Sidekiq queues before requeuing…

jennyd:
… things

We need to requeue scheduled publishings after the data sync, so shouldn't
clear the queues after that.


gpeng:
:+1:

PR #4793, opened 2016-08-23
Disable workers in development

boffbowsh:
Workers should be run manually using `bowl` in development

/cc @alphagov/team-gov-uk-finding-things


MatMoore:
👍

PR #4794, opened 2016-08-23
MongoDB S3 Restore 6

djsd123:
Lack of attention on my part.  The auto loader could not resolve the dependency and therefore could not load the class 'Package' and require the needed resource.

For future reference. Puppet master would throw the following error:

----
Aug 23 13:26:14 puppetmaster-1 puppet-master[1542]: Invalid relationship: File[/usr/local/bin/mongodb-restore-s3] { require => Class[Mongodb::S3backup::Package] }, because Class[Mongodb::S3backup::Package] doesn't seem to be in the catalog
----
Aug 23 13:26:15 puppetmaster-1 puppet-master[1545]: (Scope(Class[main])) Could not look up qualified variable '::govuk_node_class';
----

### DIFF
--- a/modules/mongodb/manifests/s3backup/restore.pp
+++ b/modules/mongodb/manifests/s3backup/restore.pp
@@ -46,6 +46,7 @@ class mongodb::s3backup::restore(
 ){
 
   include ::backup::client
+  contain ::mongodb::s3backup::package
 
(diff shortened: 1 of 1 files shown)


deanwilson:
Merge when green.

PR #4795, opened 2016-08-23
Remove s3 Backup lvm config from Exception Handler

djsd123:
Exception Handler's data does not need to be backed up. The lvm config was added in error.

also 

Removed the gating to allow s3 backups for the exception handler.

relates to: https://github.gds/gds/govuk-provisioning/pull/564


deanwilson:
Merge away.

PR #4797, opened 2016-08-23
Add extra partition to integration postgres primary

deanwilson:
The bulk of this work was done in
https://github.com/alphagov/govuk-puppet/pull/4774 but integration had an
override file so the change needs to be present here too.


surminus:
I knew this would come back to bite (annoy) us. We should have just redeployed the machine so we had consistent disks and partitions instead adding random hieradata 

Anyway, :+1:

djsd123:
Just when I thought I knew my alphabet.

deanwilson:
We should get in to the habit of always rebuilding hosts when the disks change. It'd save a lot of hassle and ensure we're good with restores

PR #4798, opened 2016-08-24
Remove users no longer at Gov.uk

brenetic:
- Removed Ben Hyland's manifest.
- Remove Russell Garner's ssh key. Manifest file to be removed later.


modules/users/manifests/russellgarner.pp
```diff
@@ -1,6 +1,8 @@
 # Creates the russellgarner user
 class users::russellgarner {
+  #Once PR has been merged delete this file
   govuk_user { 'russellgarner':
+    ensure   => absent,
```
alexmuller: This `ensure => absent` won't be run because he's been removed from the lists above.

brenetic:
@alexmuller  Thanks dude. Have reverted the yaml file.

PR #4799, opened 2016-08-24
Always deploy Licensify using deployment master

alexmuller:
This mirrors the way all of our other apps are deployed.


elliotcm:
👍

PR #4801, opened 2016-08-24
Switch Licensing over in production

alexmuller:
This switches Licensing to use new app servers in production


elliotcm:
Good to go

PR #4802, opened 2016-08-24
Remove  s3 Backup lvm Config

djsd123:
Licensify data does not need to be backed up to s3.  This config was added in error.


deanwilson:
Merge when green.

PR #4803, opened 2016-08-24
MongoDB  s3 Restore 7

djsd123:
What
While verifying I noticed that s3 restore was missing values in hiera for staging and prod.

How
Added values for the $s3_bucket parameter for Staging and Production.
Modified restore class paramter $s3_bucket to use value from backup class.
Reworked gating so s3::backup + restore are provisioned together.
This is because s3 restores use hostname to determine where to find data to pull from s3.

When s3backups are not enabled puppet should remove the scripts and cron jobs to avoid confusion/errors.


surminus:
This looks fine to me. We'll need to add buckets and secrets to get this to work. :+1:

PR #4804, opened 2016-08-24
Run unattended-upgrades on the dev VM as part of the puppet run

h-lame:
When we run govuk_puppet on a dev VM we expect that it will update all
the managed software on the machine, however for package updates it
relies on unattended-upgrades to actually perform the updates.  These
are set to run at some small hour in the morning, but most dev VMs are
not powered on at this time, so the updates don't actually run.

We noticed this when postgres 9.3.14 came out which introduced a breaking
change to the `\connect` line in sql dumps.  The servers were upgraded
but most dev VMs weren't and so replicating data generated on the servers
running 9.3.14 would silently fail when run on the dev VMs running 9.3.12.

See https://github.gds/gds/development/pull/334 for some extra context on the change.

I fully expect that this isn't the correct way to invoke the run, or that some other approach would be better.  Please chip in with better ideas.

### DIFF
--- a/modules/govuk/manifests/node/s_development.pp
+++ b/modules/govuk/manifests/node/s_development.pp
@@ -151,4 +151,14 @@ class govuk::node::s_development (
     'vegeta':         ensure => installed; # vegeta is used by the router test suite
     'mawk-1.3.4':     ensure => installed; # Provides /opt/mawk required by pre-transition-stats
   }
+
+  exec { 'unattended_upgrades':
+    command => '/usr/bin/unattended-upgrades',
+    require => Notify['unattended_upgrades'],
+  }
+
+  notify { 'unattended_upgrades':
+    message  => 'Updating apt packages',
+  }
+
 }


modules/govuk/manifests/node/s_development.pp
```diff
@@ -151,4 +151,9 @@
     'vegeta':         ensure => installed; # vegeta is used by the router test suite
     'mawk-1.3.4':     ensure => installed; # Provides /opt/mawk required by pre-transition-stats
   }
+
+  exec { 'unattended_upgrades':
+    command     => "/usr/bin/unattended-upgrades",
```
brenetic: Linter doesn't like double quoted strings :(

surminus:
I think this is a decent approach as it'll always be caught in a puppet run. You might want to also add a `notify => 'Updating packages'` or something because otherwise it'll silently sit there for a while and people will start to get twitchy.

h-lame:
Forgot to pay attention to the linter.  Thanks for pointing it out.

jennyd:
I'm getting this error from the `notify`:

`Error: Parameter notify failed on Exec[unattended_upgrades]: No title provided and "Updating apt packages" is not a valid resource reference at /var/govuk/govuk-puppet/modules/govuk/manifests/node/s_development.pp:158`

jennyd:
I tried this instead:

```puppet
  exec { 'unattended_upgrades':
    command => '/usr/bin/unattended-upgrades',
    require => Notify['unattended_upgrades'],
  }

  notify { 'unattended_upgrades':
    message  => 'Updating apt packages',
  }
```

...and it looks like it's working (I presume it's now doing a lot of upgrades):

```
Notice: Compiled catalog for development.development in environment development in 30.70 seconds
Info: Applying configuration version '1472216804'
Notice: Updating apt packages
Notice: /Stage[main]/Govuk::Node::S_development/Notify[unattended_upgrades]/message: defined 'message' as 'Updating apt packages'
```

jennyd:
...but then I got a bunch of errors, and my postgres version is still the same as before (I had 9.3.13):

```
Notice: Updating apt packages
Notice: /Stage[main]/Govuk::Node::S_development/Notify[unattended_upgrades]/message: defined 'message' as 'Updating apt packages'
Notice: /Stage[main]/Govuk::Node::S_development/Exec[unattended_upgrades]/returns: executed successfully
Notice: /Stage[main]/Harden/File[/usr/bin/rlogin]/ensure: removed
Notice: /Stage[main]/Harden/File[/usr/bin/rcp]/ensure: removed
Error: Cannot create /etc/unattended-reboot/no-reboot; parent directory /etc/unattended-reboot does not exist
Error: /Stage[main]/Govuk_unattended_reboot/File[/etc/unattended-reboot/no-reboot]/ensure: change from absent to directory failed: Cannot create /etc/unattended-reboot/no-reboot; parent directory /etc/unattended-reboot does not exist
Error: Could not set 'file' on ensure: No such file or directory - /etc/unattended-reboot/check/03_no_reboot20160826-20903-ft13xs.lock at 85:/var/govuk/govuk-puppet/modules/govuk_unattended_reboot/manifests/init.pp
Error: Could not set 'file' on ensure: No such file or directory - /etc/unattended-reboot/check/03_no_reboot20160826-20903-ft13xs.lock at 85:/var/govuk/govuk-puppet/modules/govuk_unattended_reboot/manifests/init.pp
Wrapped exception:
No such file or directory - /etc/unattended-reboot/check/03_no_reboot20160826-20903-ft13xs.lock
Error: /Stage[main]/Govuk_unattended_reboot/File[03_no_reboot]/ensure: change from absent to file failed: Could not set 'file' on ensure: No such file or directory - /etc/unattended-reboot/check/03_no_reboot20160826-20903-ft13xs.lock at 85:/var/govuk/govuk-puppet/modules/govuk_unattended_reboot/manifests/init.pp
```

h-lame:
Thanks @jennyd, made that change.

h-lame:
#4815 seems to fix the other errors.  For the postgres not actually upgrading, I think this might be an order thing where unattended upgrades gets run early in the puppet process, but telling unattended upgrades to ask for a new version of postgres (or other packages) happens later in the process.  Running govuk_puppet a second time seems to achieve what we want which backs up that theory.  Is it possible to make this exec run as one of the last things that the puppet run does?

deanwilson:
As an alternative can we do https://github.com/alphagov/govuk-puppet/pull/4839 instead? We'd like to avoid a puppet run suddenly taking a lot longer with no visible output.

h-lame:
Data point of 1: I mostly run govuk_puppet on my VM as part of the update-all.sh script, so any output in there is lost halfway through, between the git and bundler parts.  I'm not sure I'd spot the output in #4839 to prompt me, I wonder if others would be the same.  Is there a way to run untattended-upgrades so that we get some output, so a long period of nothing is less likely?

As an aside: putting the unattended-upgrades run where you've put the warning message might be a good idea as it would guarantee that it happens at the end of the puppet run.

deanwilson:
I'm happy to add the command where the warning is. I was just a little worried about people suddenly thinking puppet runs take an extra few minutes when nothing changes. I'll update the other PR to suit.

h-lame:
@deanwilson that sounds like a good idea.  I'll close this PR in preference to yours.

PR #4805, opened 2016-08-24
Jenkins job to check Publishing API validity

kevindew:
This is a Jenkins job that should _hopefully_ fire off the run rake task build each day for the Publishing API.

I confess I worked out the syntax for the yaml file based off the other files and am unsure if it has any problems not having a `builders` attribute like all the other ones in the directory have.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/publishing_api_check_validity.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/publishing_api_check_validity.yaml.erb
@@ -1,0 +1,17 @@
+---
+- job:
+    name: Publishing_API_Check_Validity
+    display-name: Publishing_API_Check_Validity
+    project-type: freestyle
+    description: "This job checks the data within the Publishing API for any invalid data"
+    logrotate:
+      artifactNumToKeep: 10
+    publishers:
+        - trigger-parameterized-builds:
+          - project: run-rake-task
+            predefined-parameters: |
+              TARGET_APPLICATION=publishing-api
+              MACHINE_CLASS=backend
+              RAKE_TASK=duplicate_content_items:all
+    triggers:
+        - timed: 'H 5 * * *'
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -123,6 +123,7 @@ govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::job::network_config_deploy
   - govuk_jenkins::job::passive_checks
   - govuk_jenkins::job::performance_platform_smokey
+  - govuk_jenkins::job::publishing_api_check_validity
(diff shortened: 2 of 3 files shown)


rjw1:
Does the rake task alert people if it finds invalid data?

kevindew:
Yup - it pushes errors into Errbit via Airbrake.

modules/govuk_jenkins/templates/jobs/publishing_api_check_validity.yaml.erb
```diff
@@ -0,0 +1,16 @@
+---
+- job:
+    name: Publishing_API_Check_Validity
+    display-name: Publishing_API_Check_Validity
+    project-type: freestyle
+    description: "This job checks the data within the Publishing API for any invalid data"
+    logrotate:
+      artifactNumToKeep: 10
+    publishers:
```
surminus: I think this will need to be used as [`trigger-parameterized-builds`](http://docs.openstack.org/infra/jenkins-job-builder/publishers.html#publishers.trigger-parameterized-builds)

modules/govuk_jenkins/templates/jobs/publishing_api_check_validity.yaml.erb
```diff
@@ -0,0 +1,16 @@
...
+       - project: run-rake-task
+         predefined-parameters: |
+          TARGET_APPLICATION=publishing-api
+          MACHINE_CLASS=backend
+          RAKE_TASK=duplicate_content_items:all
+    triggers:
+       - timed: 'H 5 * * *'
```
surminus: Just to confirm, this runs once a day?

modules/govuk_jenkins/templates/jobs/publishing_api_check_validity.yaml.erb
kevindew: Doh - you're quite right. I'll sort that.

modules/govuk_jenkins/templates/jobs/publishing_api_check_validity.yaml.erb
kevindew: Yup - I was aiming for once each day at around 5am. 

Although I realise now that at that time integration is powered off - so I guess I should probably set it later unless it remains pending until integration is switched on (it will go out to prod eventually)

modules/govuk_jenkins/templates/jobs/publishing_api_check_validity.yaml.erb
kevindew: @surminus I've added `trigger-parametized-builds`in now, hopefully all good now.

cbaines:
:+1:

PR #4806, opened 2016-08-25
Collect sidekiq logs for Publishing API

kevindew:
We're not sending through the Publishing API to Kibana, which seems to be an oversight. I've set this up based on how it was set up for whitehall.


rjw1:
:+1: merge when ready

PR #4807, opened 2016-08-25
Make Nginx proxy with HTTP/1.1 for Licensing

alexmuller:
This commit adds a feature flag so that Nginx on the app servers can proxy to the Licensing app on localhost using HTTP/1.1.

Nginx docs:

> By default, version 1.0 is used. Version 1.1 is recommended for use with keepalive connections and NTLM authentication.

http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_http_version

Licensing has been upgraded to a new version of Scala/Play which is picky about which HTTP version it needs (because it uses chunked responses for file downloads).

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -253,6 +253,10 @@
 # Configure the amount of time the nginx proxy vhost will wait for the
 # backing app before it sends the client a 504. It defaults to 15 seconds.
 #
+# [*proxy_http_version_1_1_enabled*]
+#   Boolean, whether to enable HTTP/1.1 for proxying from the Nginx vhost
+#   to the app server.
+#
 define govuk::app (
   $app_type,
   $port = 0,
@@ -288,6 +292,7 @@ define govuk::app (
   $hasrestart = false,
   $depends_on_nfs = false,
   $read_timeout = 15,
+  $proxy_http_version_1_1_enabled = false,
 ) {
 
   if ! ($app_type in ['procfile', 'rack', 'bare']) {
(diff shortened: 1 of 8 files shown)


modules/govuk/manifests/app.pp
```diff
@@ -253,6 +253,10 @@
 # Configure the amount of time the nginx proxy vhost will wait for the
 # backing app before it sends the client a 504. It defaults to 15 seconds.
 #
+# [*proxy_http_version_1_1_enabled*]
+#   Boolean, whether to enable HTTP/1.1 for proxying from the Nginx vhost
+#   to the app server.
```
elliotcm: Could you add the default to this (in another PR if you like)

modules/govuk/manifests/app/config.pp
```diff
@@ -12,6 +12,10 @@
 # [*nagios_memory_critical*]
 #   Memory use, in MB, at which Nagios should generate a critical alert.
 #
+# [*proxy_http_version_1_1_enabled*]
+#   Boolean, whether to enable HTTP/1.1 for proxying from the Nginx vhost
+#   to the app server.
```
elliotcm: Ditto default

modules/nginx/manifests/config/vhost/proxy.pp
```diff
@@ -97,6 +101,7 @@
   $ensure = 'present',
   $alert_5xx_warning_rate = 0.05,
   $alert_5xx_critical_rate = 0.1,
+  $proxy_http_version_1_1_enabled = false,
```
elliotcm: Is this passed automatically to the template?

elliotcm:
Do we need to enable this on the feed app?

modules/govuk/manifests/app.pp
alexmuller: Sure, will do it in another pull request

modules/nginx/manifests/config/vhost/proxy.pp
alexmuller: Yep, it'll go through

alexmuller:
I'm not sure about the feed app but as it has a vhost and it's the same Scala app I don't think it will hurt.

elliotcm:
👍 when green

PR #4808, opened 2016-08-25
Tidy up removal of DFID transition app.

elliotcm:
Setting ENV vars doesn't work if you've removed
the app because the directory isn't there anymore.

Also removes the procfile worker properly.


rjw1:
merge when green :+1:

PR #4809, opened 2016-08-25
Avoid installing Ruby 2.3.1 on non-`trusty` machines.

elliotcm:
This package is not available for `precise`
machines and the only Ruby code we have running on
them is EFG which currently uses 2.1.4.


deanwilson:
Looks fine. Merge away.

PR #4810, opened 2016-08-25
Bugfix mysql backups

surminus:
This PR adds various bugfixes to the MySQL S3 backup scripts.

It ensures the script doesn't bomb out if there is no file for archiving, fixing incorrect logic for checking if a backup already exists. 

Detail within commit messages.

### DIFF
--- a/modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_base.erb
+++ b/modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_base.erb
@@ -19,7 +19,7 @@ function nagios_passive () {
 trap nagios_passive EXIT
 
 # Set a new backup date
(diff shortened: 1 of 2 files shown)


deanwilson:
Can we rename the `BACKUPEXISTS` variable please as discussed?

surminus:
The commits are a little confusing, but this actually happens in 97fed5ddc67af02bfcf942d872ec0568892c8ca5 when the `BACKUPEXIST` variable is removed completely. We looked at a2bff4011f065d2b8e29af631d633a161d2f0a5f and saw it there, but I split this commit to specifically mention the reasons for a double-pipe at the end of the command.

In summary, my commits were a little confusing, but I think it's OK to merge?

deanwilson:
It's better now than it was before you started work on it so merge away.

PR #4811, opened 2016-08-25
Make slugs for ET and EAT the same as V1

tuzz:


tuzz:
Tested on integration.

PR #4814, opened 2016-08-26
Remove dfid-transition app

tuzz:
Now that we've run `ensure => absent`, we can get rid of this completely.

https://trello.com/c/Ke4GpYqS/249-tear-down-dfid-transition-smediumish


surminus:
:+1: feel free to merge

PR #4815, opened 2016-08-26
Use file/directory_ensure for unattended reboots

alexmuller:
We defined these at the top of the file to match whether the class is enabled or not. Puppet will fail in cases where the `config_directory` doesn't exist (eg new development VMs).

/cc @h-lame @jennyd


jennyd:
👍  this fixes those errors for me

h-lame:
:+1: ditto

PR #4816, opened 2016-08-26
Serve styled errors for licensing

alexmuller:
This was done in October 2012 in a9d04ee2ba674cffcb86b317014c9aba494c4652 and I can't think of a reason we'd want it to exist in 2016.


deanwilson:
Merge away.

PR #4818, opened 2016-08-30
Update licensing log format to JSON

daibach:
Licensing logs in JSON format, but this isn't recognised by Kibana.

I don't know if this is the right way to change this (or if this even solves that issue) - I'd appreciate suggestions.


alexmuller:
Yep, this is right. Should it also be set to true for the other licensing apps as well?

daibach:
I've added the same log format line to `licensify_feed` and `licensify_admin` apps

deanwilson:
Looks fine. Merge when ready.

PR #4820, opened 2016-08-31
Remove backup for licensing mongo

leenagupte:
In commit 0e3cc049da5097d039881f7ff560cd92e4e825cb, we disabled the auto mongo db
backup for licensing, but were still trying to collect a backup that does not
exist.


alexmuller:
:+1:

Infrastructure need to discuss backup strategy for licensing.

PR #4821, opened 2016-08-31
Change unified_search to search

rubenarakelyan:
This commit changes the nginx config to point all external requests for `/api/search.json` to the rummager endpoint `/search.json` rather than `/unified_search.json`. This provides consistency between the internal and external APIs in terms of naming.

See https://github.com/alphagov/rummager/pull/694 for the equivalent changes to rummager.

Trello: https://trello.com/c/cj8UX2jX


tijmenb:
👍 when green

PR #4822, opened 2016-08-31
Enable rummager publishing listener in development

boffbowsh:
This enables the creation of the permissions and queues in RabbitMQ.
If we enable these, we should also enable the listener in order to
not let the queue back up with unprocessed messages.

/cc @tijmenb


tijmenb:
👍 MWG

PR #4825, opened 2016-08-31
Adds assets-origin (the asset root) to /etc/hosts

nickcolley:
When testing government-frontend and trying to run internal tests
the assets root does not resolve properly on the development vm

Service Manual team also encountered the same and used a work around https://github.com/alphagov/service-manual-frontend/pull/37/files#diff-04c6e90faac2675aa89e2176d2eec7d8R41

I'm not sure if adding this to apps does anything else than just add it to hosts, so not sure if this is the right place?


boffbowsh:
👍 merge when green

PR #4826, opened 2016-08-31
Only enable postgresql mirror for trusty

surminus:
We don't need the mirror on precise machines, and *if* we add Xenial machines in the future the version of PostgreSQL we need the mirror for is available by default, so only enable it for trusty machines.


deanwilson:
Merge when green

PR #4827, opened 2016-08-31
Mongodb s3 Restore 8

djsd123:
prefixing 's3cmd' sub-command with 'envdir'.

While verifying the restore script an error occurred stating that there was no '.s3cfg' file.
When invoking s3cmd it performs a lookup for aws auth tokens by checking for cmd-line arguments, env-variables and it's own config '.s3cfg'.
This occured while executing a sub-command that was not prefixed with the envdir command which is how we pass env-varibles.

Removed 'var/' sub-directory as this does not get created outside my testing environment.

This has been tested on a host running MongoDB in integration.

### DIFF
--- a/modules/mongodb/templates/mongodb-restore-s3.erb
+++ b/modules/mongodb/templates/mongodb-restore-s3.erb
@@ -14,7 +14,7 @@ cd $BACKUP_DIR
 TIME="$(date +%s)"
 
 # Download and decrypt backup
-  /usr/bin/envdir <%= @env_dir %>/env.d /usr/local/bin/s3cmd get --force `/usr/local/bin/s3cmd ls s3://${S3_BUCKET}/mongodump* | \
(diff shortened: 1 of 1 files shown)


surminus:
I think this change looks fine, but can we have some detail about why stuff is changing in line 10 and 33 respectively?

djsd123:
Added additional comment.

modules/mongodb/templates/mongodb-restore-s3.erb
```diff
@@ -14,7 +14,7 @@ cd $BACKUP_DIR
 TIME="$(date +%s)"
 
 # Download and decrypt backup
-  /usr/bin/envdir <%= @env_dir %>/env.d /usr/local/bin/s3cmd get --force `/usr/local/bin/s3cmd ls s3://${S3_BUCKET}/mongodump* | \
+  /usr/bin/envdir <%= @env_dir %>/env.d /usr/local/bin/s3cmd get --force `/usr/bin/envdir /etc/mongo_s3backup/env.d /usr/local/bin/s3cmd ls s3://${S3_BUCKET}/mongodump* | \
```
deanwilson: Is there a reason one of these uses `<%= @env_dir %>` and the other is the literal `/etc/mongo_s3backup/env.d`?

deanwilson:
Merge away.

PR #4828, opened 2016-09-01
Add Bevan Loon, new FTA dev on Core team

binaryberry:


boffbowsh:
Can Bevan submit this instead? :)

binaryberry:
ok! Will close this one and let @bevanloon open another - according to the test failures he needs a stronger SSH key too.

PR #4829, opened 2016-09-01
Add new setting for the licensify apps

deanwilson:
These need to know their own environment to pick the correct
bucket name


alexmuller:
Looks good to me :+1:

PR #4832, opened 2016-09-02
Add govdelivery syncing to staging data sync complete.

elliotcm:
This is supposed to be on integration and staging (BUT NOT PRODUCTION!!!!!1)


h-lame:
As far as I can tell this does what we need. :+1:

PR #4835, opened 2016-09-02
Tweak Puppet run

alexmuller:
In 73b49aa205b7226900ab90510cc2a6616edef865 I tried to get Icinga to
show the reason Puppet is disabled, but I had to remove the check
for exit code 2 on the Puppet run.

Puppet runs can have all kind of exit codes (especially when run with
`--detailed-exitcodes`) so we should probably let every exit code
through and catch problems as part of the script that checks the
run.

I've added an additional check that the run is recent enough.

### DIFF
--- a/modules/puppet/files/usr/lib/nagios/plugins/check_puppet_agent
+++ b/modules/puppet/files/usr/lib/nagios/plugins/check_puppet_agent
@@ -3,6 +3,8 @@
 require 'json'
 require 'yaml'
 
+WARNING = 1
+
 puppet_agent_summary_filename = "/var/lib/puppet/state/last_run_summary.yaml"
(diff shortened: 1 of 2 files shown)


modules/puppet/files/usr/lib/nagios/plugins/check_puppet_agent
```diff
@@ -13,6 +15,13 @@ if File.file?(puppet_agent_disabled_lockfile)
   exit 1
 end
 
+hours_since_last_run = (Time.now - File.mtime(puppet_agent_summary_filename)) / 3600
```
deanwilson: There is a small oddity here in that this file won't exist when the machine is built but before puppet runs. But as we deploy this test with puppet this'll be fine.

deanwilson:
Merge away.

PR #4840, opened 2016-09-06
Upgrade elasticsearch for logs on integration

h-lame:
We want to test out upgrading to 1.7.x as a precursor to trying this on
staging and eventually production.

### DIFF
--- a/hieradata/class/integration/logs_elasticsearch.yaml
+++ b/hieradata/class/integration/logs_elasticsearch.yaml
@@ -1,3 +1,3 @@
 ---
 
-govuk_elasticsearch::version: '1.5.2'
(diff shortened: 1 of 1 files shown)


h-lame:
Flagged as DNM because we need to disable puppet on the logs machines and roll this change out manually one by one to each server.

alexmuller:
Good to merge whenever you want 👍

alexmuller:
This should be a rolling upgrade: https://www.elastic.co/guide/en/elasticsearch/reference/1.7/setup-upgrade.html#rolling-upgrades

h-lame:
We've disabled puppet on the 3 machines, so we're ready to merge.

PR #4841, opened 2016-09-06
Update the govuk_puppet script.

deanwilson:
We'd like to be able to establish multiple locks at the same
time so cron jobs don't accidentally re-enable each other when
they release their own locks.

### DIFF
--- a/modules/puppet/templates/govuk_puppet
+++ b/modules/puppet/templates/govuk_puppet
@@ -6,13 +6,190 @@ if [ "$(id -u)" -ne "0" ]; then
   exec sudo "$BASH_SOURCE" "$@"
 fi
 
-START_TIME=`date +%s%3N`
+## variable defaults and setup
+app_name=$(basename "$0")
+
+puppet_lock=/var/lib/puppet/state/agent_disabled.lock
+lock_dir=/var/run/lock/puppet
+
+lock_name=default # TODO validate this
+lock_reason='' # default to no value to match current invocations
+
+start_time=`date +%s%3N`
+
+##############################################
+
+usage() {
+cat<<EOH
+
+Options:
+
+ -d [lock_directory]
+  The directory to store the locks in. Defaults to $lock_dir
+
+ -h
+  Shows this help information
+
+ -l
+  List locks. Show all the existing locks
+
+ -n [lock_name]
+  The name this lock should be taken out under. Defaults to 'default'
+  for compatibility with older invocations.
+
+ -r [lock_reason]
+  The reason puppet has been disabled. This is currently optional
+  to support the older invocation style.
+
+Examples:
+
+  # deprecated lock and unlock commands:
+  # to lock puppet the current way with a default lock and no reason
+  $app_name --disable
+
+  # and then to reenable puppet
+  $app_name --enable
+
+
+
+  # newer style with explicit lock names
+  # be explicit with the reason and the name of the lock
+  $app_name -n postgres_backups -r 'Running postgres restore' --disable
+
+  # show the lock exists
+  $app_name -l
+
+  # remove the named lock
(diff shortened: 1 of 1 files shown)


surminus:
I'm going to test this in Vagrant.

surminus:
```
vagrant@jumpbox-1:~$ govuk_puppet
govuk_puppet: lock directory /tmp/puppet-tests is not accessible
```

alexmuller:
This approach looks good to me! Some thoughts:

- We could probably remove `-d` as a command line option, I don't think it's useful except for testing.
- Could the filename be automatically generated? That means we wouldn't need a `-n` option.
- Then the option to disable could be `govuk_puppet -d {reason}` which makes it simpler to call.
- If somebody doesn't provide a `-d {reason}` the script can hard exit 1, I'm ok with forcing people to supply a reason.

deanwilson:
@alexmuller 

> We could probably remove -d as a command line option, I don't think it's useful except for testing.

I'd like to keep this for that very reason. When i test on my local dev environment i'd rather not hard code overridden paths and then forget to remove them on commit.

* Could the filename be automatically generated? That means we wouldn't need a -n option.

Doing this would substantially change the way we wanted this to work. One of the aims in this revision was to force things to declare what was asking for the lock both for tracking and to make it possible to consistently revoke the correct lock after. So I'm going to say "no" to this.

>Then the option to disable could be govuk_puppet -d {reason} which makes it simpler to call.
If somebody doesn't provide a -d {reason} the script can hard exit 1, I'm ok with forcing people to supply a reason.

After a discussion we'll update the script that call this in the old way in the puppet repo and exit hard if no reason is given.

surminus:
It errors uncleanly when using disable in the old way with no reason:
```
vagrant@jumpbox-1:~$ govuk_puppet --disable
govuk_puppet: No explicit lock name reason was given. Using defaults.
/usr/local/bin/govuk_puppet: line 154: lock_reason: unbound variable
```

surminus:
I don't seem to be able to get a lock or disable when running: 

`vagrant@jumpbox-1:~$ govuk_puppet -n foo -r bar` 

It still seems to run Puppet - is this the correct invocation? Not sure if my Vagrant setup is making life difficult, mind.

modules/puppet/templates/govuk_puppet
```diff
@@ -6,13 +6,176 @@ if [ "$(id -u)" -ne "0" ]; then
...
+
+locked=$(has_locks)
+if [ "$locked" == 1 ];then
+  list_locks
+  exit 1
+fi
 
 set +e
-RBENV_VERSION=1.9.3 /usr/local/bin/govuk_setenv default puppet agent --onetime --no-daemonize "$@"
+RBENV_VERSION=1.9.3 /usr/local/bin/govuk_setenv default puppet agent --onetime --no-daemonize "${puppet_args}"
```
surminus: The double quotes here `"${puppet_args}"` break the script (for some unknown reason).

cbaines: I believe the double quotes force the value to be interpreted as a single parameter, which has different semantics to `"$@"` as `$@`is an array like thing and `puppet_args` is just a string.

```
lucida:  /tmp/foo 
→ ARGS="-l -a"

lucida:  /tmp/foo 
→ ls $ARGS
total 152
drwxr-xr-x  2 gds  gds    4096 Oct 10 10:11 .
drwxrwxrwt 35 root root 147456 Oct 10 10:11 ..
-rw-r--r--  1 gds  gds       0 Oct 10 10:11 bar
-rw-r--r--  1 gds  gds       0 Oct 10 10:11 baz

lucida:  /tmp/foo 
→ ls "$ARGS"
ls: invalid option -- ' '
Try 'ls --help' for more information.
```

PR #4842, opened 2016-09-06
Revert "Upgrade elasticsearch for logs on integration"

h-lame:
Reverts alphagov/govuk-puppet#4840

We did the upgrade which went smoothly, but the redis river stoppped consuming data from redis and it's not clear how to fix it, so we're downgrading for now.  We may have to use a different redis river plugin, or, as rivers are deprecated change our approach entirely. 

Note that downgrading may mean we can no longer read the indexes but as it's integration we'll be able to delete all the data and start again.


alexmuller:
👍 😞 

This is a good indication of how creaking and old our logging infrastructure is and how badly it needs to be replaced.

PR #4843, opened 2016-09-06
Switch licensify-admin to a better healthcheck

alexmuller:
This healthcheck now exists and checks things we care about.

/cc @daibach


deanwilson:
Code looks fine. Merge when ready.

PR #4844, opened 2016-09-06
Bugfix for MySQL backups to S3

surminus:
This commit follows on from e00e8265b8cb0a35ff2e24b123a5428131717b20. This has a failure condition where when it moves the temporary filename to be named "base.xbcrypt", it actually just moves the temporary filename into a directory named "base.xbcrypt". This means that we haven't lost any data, but when we come to restore it would be called a file we don't expect (we call it s3://<bucket>/latest/base.xbcrypt so when we restore we can easily find the latest backup). This seems to happen because of the recursive option being set, and since this option does not seem to be needed for any of the incremental backup tasks, I think we can safely remove this from the base backup task and ensure the condition described above no longer occurs. This removes the `--recursive` flag from the `s3cmd mv` command.

**Note**: I've put a fix in place in Production and disabled Puppet on the machine where this occurs. It's currently breaking and causing files to pile up when they should be archived.


deanwilson:
Merge away.

PR #4845, opened 2016-09-06
Add contacts-admin envars

brenetic:
Contacts admin now uses the new world of passing secrts in via envars.

[Trello card](https://trello.com/c/YbcsenXK/473-contacts-admin-dependency-maintenance-3)

Related to: https://github.gds/gds/deployment/pull/1072

### DIFF
--- a/hieradata/class/backend.yaml
+++ b/hieradata/class/backend.yaml
@@ -1,6 +1,10 @@
 ---
 
 govuk::apps::contacts::vhost: contacts-admin
+govuk::apps::contacts::enable_admin_routes: 'enabled'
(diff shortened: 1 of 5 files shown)


hieradata/class/backend.yaml
```diff
@@ -1,6 +1,10 @@
 ---
 
 govuk::apps::contacts::vhost: contacts-admin
+govuk::apps::contacts::enable_admin_routes: 'enabled'
+govuk::apps::contacts::db_hostname: 'master.mysql'
```
rjw1: is this still the hostname we want to use?

brenetic: @rjw1  Have updated 👍

PR #4846, opened 2016-09-07
Allow multiple procfile processes

boffbowsh:
Adds support for a `process_count` parameter to
`govuk::procfile::worker`. This uses instanced upstart jobs to allow us
to run a configurable number of procfile-defined processes.

The existing procfile upstart job becomes a parent job, which manages a
set of child jobs. The original logic is moved into the child job
configuration. When the parent job is started, it starts the requested
number of instances of the child job. When it is stopped, it stops all
child jobs. This enables us to scale processes easily with hieradata,
and makes upstart tidy up all the child jobs whether they’re configured
or not.

This maintains the ability to stop the existing job that is running
directly under what is now the parent job. On stopping the parent job
(in this instance, on the puppet run that upgrades the job layout), it
finds any parent jobs that have a process attached and kills them.
These would have the format `foo-profile-worker start/running, process
1234` in `initctl list`.

### DIFF
--- a/modules/govuk/manifests/procfile/worker.pp
+++ b/modules/govuk/manifests/procfile/worker.pp
@@ -14,19 +14,25 @@
 #
 # [*process_type*]
 #   The type of process to spawn, defined in the application's Procfile.
-#   This variable is used by the govuk/procfile-worker.conf.erb template.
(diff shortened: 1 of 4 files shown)


tuzz:
Looks good based on my limited puppet knowledge.

PR #4847, opened 2016-09-07
Deploy Licensing Backups Everywhere

djsd123:
S3 backups are gated in the mongodb::backup class.  To enable s3 backups, we need to set the condition to true in the
node classifier.  This allows for deploying to specific nodes.


deanwilson:
Merge away.

PR #4849, opened 2016-09-07
Exclude the events table from the Publishing API database backup

cbaines:
This takes up a large proportion of the of the backup, and is used by very few people.

My testing on integration shows that this drops the size of the Publishing API backup by ~2GB (from 2.7GB to 723MB).

### DIFF
--- a/modules/govuk_postgresql/templates/etc/cron.daily/autopostgresqlbackup.erb
+++ b/modules/govuk_postgresql/templates/etc/cron.daily/autopostgresqlbackup.erb
@@ -344,6 +344,16 @@ dbdump () {
   touch $2
   chmod 600 $2
   for db in $1 ; do
+    <% if @environment == 'integration' %>
+    # The Publishing API events table takes up a lot of space in the
+    # backup, which is downloaded by lots of people in GOV.UK, and not
+    # needed by most of them, so exclude this from the backup on
+    # integration.
+    if [ "$db" = "publishing_api_production" ]; then
+      OPT="$OPT --exclude-table-data=events"
+    fi
+    <% end %>
+
     if [ -n "$SU_USERNAME" ]; then
       if [ "$db" = "$GLOBALS_OBJECTS" ]; then
         su $SU_USERNAME -c "pg_dumpall $PGHOST --globals-only" >> $2


modules/govuk_postgresql/templates/etc/cron.daily/autopostgresqlbackup.erb
```diff
@@ -344,6 +344,16 @@ dbdump () {
   touch $2
   chmod 600 $2
   for db in $1 ; do
+    <% if @environment != 'integration' %>
```
surminus: If I'm reading this right it looks like this will exclude the table in all environments except Integration - is this correct?

kevindew:
Personally I don't think the backups should differ in their contents from the production environment. It risks adding confusion and setting a bad precedent. 

I'd prefer it if we could resolve this table issue in a way that didn't require intervening with the backups.

modules/govuk_postgresql/templates/etc/cron.daily/autopostgresqlbackup.erb
cbaines: Your reading is, but the code is not. I've now fixed it such that it only includes this special case on integration.

cbaines:
> Personally I don't think the backups should differ in their contents from the production environment. It risks adding confusion and setting a bad precedent.
> 
> I'd prefer it if we could resolve this table issue in a way that didn't require intervening with the backups.

I agree.

However, I think that this is the best short term solution to reduce the impact that the Publishing API is having on the amount of data downloaded by the sync-postgresql script, as my tests on integration show that this will drop the size of the downloaded file from 3GB to 1.1GB.

I guess we could hold off on merging this until more people start complaining (I'm not even sure if anyone has complained yet) at the Publishing API team (and maybe we would end up sorting out this events situation in the mean time).

kevindew:
that sounds good to me 👍  lets see if we can get something moving on this in the next week.

cbaines:
> that sounds good to me :+1: 

I said several things, do you generally approve, or can you be more specific?

> lets see if we can get something moving on this in the next week.

If you mean moving in terms of merging, why wait until next week for merging this?

kevindew:
Sounds good to hold off until people start complaining.

It would be good to get alternative solutions moving forward in the next week so this can be avoided.

cbaines:
As this is being indefinitely put on hold (per the comments), I'm going to close this.

PR #4851, opened 2016-09-08
Add Go 1.7.1 and set as default

timblair:
Also remove old, unsupported versions of Go.

### DIFF
--- a/modules/golang/manifests/init.pp
+++ b/modules/golang/manifests/init.pp
@@ -7,11 +7,16 @@ class golang {
   include govuk_ppa
 
   class { 'goenv':
-    global_version => '1.7',
+    global_version => '1.7.1',
     require        => Class['govuk_ppa'],
   }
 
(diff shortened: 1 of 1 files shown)


modules/golang/manifests/init.pp
```diff
@@ -7,11 +7,16 @@
...
   }
 
-  goenv::version { ['1.3.3', '1.4.2', '1.4.3', '1.5.1', '1.5.3', '1.6.2', '1.6.3', '1.7']: }
+  goenv::version { ['1.6.3', '1.7', '1.7.1']: }
+
+  goenv::version { ['1.3.3', '1.4.2', '1.4.3', '1.5.1', '1.5.3', '1.6.2']:
+    ensure => absent,
+  }
+
   package { ['golang-gom', 'godep']:
```
gpeng: We don't need Gom anymore either do we?

timblair: No, but I thought I'd leave it there _just in case_ for now.

gpeng:
We've discussed Gom offline. Leave it on there for now.

PR #4852, opened 2016-09-08
Deploy s3 Backups Everywhere

djsd123:
S3 backups are gated in the mongodb::backup class. To enable s3 backups, we need to set the condition to true in the
node classifier. This allows for deploying to specific nodes.


surminus:
LGTM :+1:

alexmuller:
What's the reason for choosing machine number 2 in each case?

PR #4853, opened 2016-09-08
Support MongoDB 3 for unattended reboots

alexmuller:
MongoDB 3 introduces some additional data types as well as `ISODate()` like `NumberLong()`. Instead of stripping them all out by hand we should force MongoDB to give us valid JSON.

This allows MongoDB 3 machines to reboot unattended overnight.


deanwilson:
Merge away.

PR #4854, opened 2016-09-08
Reschedule Licensing realtime backup

surminus:
Do not merge yet so we can monitor how other machines perform with the every 15 minute backup schedule.

The dataset for Licensing mongo machines is large and isn't as vital as some of our other services. It causes high disk time when it runs, so paramaterise the cron times and specifically set them within hieradata.


djsd123:
Go for it 👍

PR #4855, opened 2016-09-09
Mongodb s3 Restore 9

djsd123:
What
Decided this script would be invoked interactively.  Verifying the script highlighted a number of niggles. 

1. gpg prompts for password
2. `mongorestore` warns when restoring to existing collection
3. `mongorestore` complains because it does not know what to do with the mongodump.tar.gz.gpg archive

How

- Added a conditional with help message to test whether the localhost is PRIMARY.
- Added `--drop` flag to `mongorestore` command to remove collections prior to restoring.
- Remove MongoDB archive file prior to restoring.
- Added trap function to tidy up regardless of exit status.
- Modified restore statement to point at `$BACKUP_DIR`

### DIFF
--- a/modules/mongodb/templates/mongodb-restore-s3.erb
+++ b/modules/mongodb/templates/mongodb-restore-s3.erb
@@ -7,35 +7,35 @@ exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
 
 S3_BUCKET=<%= @s3_bucket %>/<%= @fqdn %>
 BACKUP_DIR=<%= @backup_dir %>
+KEY_FINGERPRINT=<%= @private_gpg_key_fingerprint %>
+REPLSET=$(/usr/bin/mongo --quiet --eval "db.isMaster().setName")
+PRIMARY=$(/usr/bin/mongo --quiet --eval "db.isMaster().primary")
 
 cd $BACKUP_DIR
 
+function housekeeping {
+  # Purge /var/lib/s3backup
+  /bin/rm -rf $BACKUP_DIR/*
+}
+trap housekeeping EXIT
+
+housekeeping
 
 TIME="$(date +%s)"
-
-# Download and decrypt backup
-/usr/bin/envdir <%= @env_dir %>/env.d /usr/local/bin/s3cmd get --force `/usr/bin/envdir <%= @env_dir %>/env.d /usr/local/bin/s3cmd ls s3://${S3_BUCKET}/mongodump* | \
-  tail -1 | /usr/bin/awk '{print $4}'`| /usr/bin/awk '{print $4}' | /usr/bin/tr -d "'" | \
-  /usr/bin/xargs /usr/bin/gpg --yes --quiet --output mongodump.tar.gz --decrypt && /bin/tar xzf mongodump.tar.gz
+# Download and decrypt most recent backup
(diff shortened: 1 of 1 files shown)


djsd123:
We need to speak to @alexmuller regarding whether the restore script should prompt for a password during decryption??  As it stands now the script will not prompt because I have added `-r $KEY_FINGERPRINT`

surminus:
This only happens if you're restoring it to the same machine that you encrypted the backup on, which is fine. If you're decrypting it anywhere else then it will rightly prompt for a password.

modules/mongodb/templates/mongodb-restore-s3.erb
```diff
@@ -7,35 +7,37 @@ exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
...
 TIME="$(date +%s)"
 
 
-# Restore mongo database
-  /usr/bin/mongorestore var/ > /dev/null
+# Restore mongo database backup
+/usr/bin/mongorestore $BACKUP_DIR
 
-# Tidy up
-  /bin/rm -rf mongodump*
```
surminus: How come this gets removed? I can't see it detailed anywhere.

modules/mongodb/templates/mongodb-restore-s3.erb
```diff
@@ -7,35 +7,37 @@ exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
...
 
 TIME="$(($(date +%s)-TIME))"
-/usr/bin/printf "RESTORE COMPLETED IN: ${TIME}s\n"
-
-
```
surminus: Nitpick, but the empty lines makes me twitch.

surminus:
Just want to double-check on the priority stuff otherwise looks fine

surminus:
You are correct that the documentation states that to force a member to primary you need to [update the priority of all the nodes](https://docs.mongodb.com/manual/tutorial/force-member-to-be-primary/). 

My concern here is the process involved in forcing a node to be primary. We already manage the priority levels in Puppet, so we would also need to disable the puppet agent from running. I think it makes more sense to just provide the restore script on _all_ machines (whether or not they have the backup script). 

I do however appreciate that we said to only have all the scripts and envvars on a single machine, and so these would need to be added to each machine. I think this is more of a problem with process than anything, and how someone on 2ndline would go about restoring data from a backup safely.

djsd123:
I understand your concerns but I think you may have misunderstood  the way replica sets are being configured by puppet.  
Puppet configures a 4 member replica set so members [0 - 2] are set to the same priority and member [3] with a hard-coded priority of 0 (In an election it will never become primary).  This means the replica set will hold an election amongst members [0 - 2] and elect a member to become primary.  This is based on a number of factors like availability (pingMs), health and priority.  

When someone forces one of the members [0 - 2] to become primary, it does not prevent other members in that range from ever becoming primary again. It just means that at a point in time the member you have specified with the priority of 1 should become primary now. Which allows operations to be performed/started on it that otherwise could not be done on a secondary member.  It's entirely possible that an election will be held soon after setting priorities that will make a member other than the member you have just modified primary.  

In short, replica set member's roles are not static.  Unless the priority has been set to 0. 

Knowing this, if you still feel it should be changed and the restore script deployed everywhere?  Then I'll be happy to modify the process??

surminus:
I think I understand how puppet is configuring the priority of the nodes.

My question is: Do we want a 2ndline person in an emergency being told by the restore script to have to reconfigure priorities on the nodes, or to be told "I'm not currently the primary, log in and run it on this server instead"?

djsd123:
I have modified the script slightly.  I think this may be a better solution.

surminus:
:+1: I really like this solution. Means we don't repeat credentials etc and makes it easier for other people to use

PR #4857, opened 2016-09-12
Fix deprecation warning in vhost

tijmenb:
Running `govuk_puppet` on the VM currently gives this deprecation warning:

```
Warning: Variable access via 'proxy_http_version_1_1_enabled' is deprecated. Use '@proxy_http_version_1_1_enabled' instead. 

template[/var/govuk/govuk-puppet/modules/nginx/templates/proxy-vhost.conf]:43
```

This commit fixes that.


surminus:
:+1: merge away

PR #4858, opened 2016-09-12
Allow multiple procfile processes

boffbowsh:
Adds support for a `process_count` parameter to
`govuk::procfile::worker`. This uses instanced upstart jobs to allow us
to run a configurable number of procfile-defined processes.

The existing procfile upstart job becomes a parent job, which manages a
set of child jobs. The original logic is moved into the child job
configuration. When the parent job is started, it starts the requested
number of instances of the child job. When it is stopped, it stops all
child jobs. This enables us to scale processes easily with hieradata,
and makes upstart tidy up all the child jobs whether they’re configured
or not.

This maintains the ability to stop the existing job that is running
directly under what is now the parent job. On stopping the parent job
(in this instance, on the puppet run that upgrades the job layout), it
finds any parent jobs that have a process attached, and kills them.
These would have the format `foo-profile-worker start/running, process
1234` in `initctl list`.

I've added some Icinga alerts for this. Rather than alerting on the parent
job, it now alerts on the count of the child jobs similar to the unicorn process
worker check.

I've not been able to test the Icinga stuff locally, hoping to deploy the branch
to integration to test it.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -305,6 +305,7 @@ govuk::apps::publisher::redis_port: "%{hiera('sidekiq_port')}"
 
 govuk::apps::whitehall::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::whitehall::redis_port: "%{hiera('sidekiq_port')}"
+govuk::apps::whitehall::procfile_worker_process_count: 2
 
 govuk::apps::govuk_crawler_worker::airbrake_endpoint: "https://errbit.%{hiera('app_domain')}/notifier_api/v2/notices"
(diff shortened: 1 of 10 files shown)


boffbowsh:
I've tested this on Integration and bumped Whitehall to have two processes to counteract us reducing its thread count to 1 recently. Alerting worked correctly.

surminus:
:+1:

alexmuller:
What are the downsides to the multiple workers, increased memory and CPU?

(I don't object to this, just curious)

boffbowsh:
Additional CPU _cores_, and yes additional RAM. Previously when using threaded concurrency the worker would only use a single core. Now as it's two processes it'll use two cores. But probably not double the amount of CPU % because the CPU wasn't the bottleneck.

PR #4859, opened 2016-09-12
Remove Specialist Publisher from post-data sync run

tuzz:
We're not going to be writing any more migrations for
the old specialist publisher so this can go.


surminus:
:+1: fftm

PR #4864, opened 2016-09-13
Turn on slow query logging for Whitehall MySQL slaves

steventux:
We've seen some load spikes on whitehall-mysql-slave-1 and some
subsequent slow whitehall-frontend responses, 5xx spikes so turn on slow query log as part of
ongoing analysis.

### DIFF
--- a/modules/govuk/manifests/node/s_whitehall_mysql_slave.pp
+++ b/modules/govuk/manifests/node/s_whitehall_mysql_slave.pp
@@ -7,6 +7,9 @@ class govuk::node::s_whitehall_mysql_slave inherits govuk::node::s_base {
     tmp_table_size        => '256M',
     max_heap_table_size   => '256M',
     innodb_file_per_table => true,
+    # FIXME
+    # Slow query log is turned ON for a short period of analysis.
+    slow_query_log        => true,
   }
   include govuk_mysql::server::slave
 


boffbowsh:
What's the default threshold of a slow query?

alexmuller:
We [should have this in staging already](https://github.com/alphagov/govuk-puppet/blob/b76660fe1a2bd74dc182af1b80427e31a0ed9868/hieradata/staging.yaml#L79), do we need it in production?

steventux:
@boffbowsh: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_mysql/manifests/server.pp#L61  `1`

boffbowsh:
@alexmuller it's likely to be just a temporary measure to help debug the Whitehall issues.

👍  from me but will wait for Alex's approval

alexmuller:
Can you add a `FIXME` so we remember to switch it off again when we're done?

Happy with this as long as we think there are going to be differences between staging and production (`/var/log/mysql/mysql-slow.log` exists in staging).

steventux:
I'll have a look on staging and see what's been logged before thinking about putting this into production then, it should be very similar given traffic replay, thanks @alexmuller

steventux:
OK I'm gonna merge this, we experienced a spike in 502s and 504s on production a few minutes ago which traffic replay did not generate in staging.
Also (this is a reckon so please correct if untrue) replication is triggered by writes to master, and replication will invalidate the query cache which will influence the performance of the db.

PR #4865, opened 2016-09-13
EFG Training application

surminus:
This adds the configuration required to create the EFG Training application. It also removes and renames a parameter from the original EFG application, as well as adding some class docs.

/cc @alexmuller regarding SSL stuffs


alexmuller:
I think there will be some changes to the deployment repo at the same time:

- Renaming in line with the parameter change here
- Adding a new password for the training database in staging and production

alexmuller:
The vhost for `efg-training` should probably be hyphenated in `hieradata/common.yaml`, `hosts::development::apps` in `hieradata/development.yaml`, and the `govuk::app` and `govuk::app::envvar` in `modules/govuk/manifests/apps/efg_training.pp`.

Also the port for the training app will have to be different to the port for the normal app because they're running on the same box.

surminus:
Have updated it, however, in [`development.yaml`](https://github.com/alphagov/govuk-puppet/blob/master/hieradata/development.yaml#L17-L97) I kept it as underscore as per the convention from the rest of the apps listed.

djsd123:
The issue when running both training and efg on the same host is that you get conflicts. Seems `EFG_HOST` needs the value of `training` for running this app in training mode. See `training_mode.rb`.

alexmuller:
I think this is ok because `govuk::apps::efg_training::vhost_name` matches the regex `ENV["EFG_HOST"] =~ /training/` in each environment. The Puppet error that's happening is because there are 2 resources with the same name on the same machine.

alexmuller:
This is [an example of how we have multiple of the same environment variables on the same machine](https://github.com/alphagov/govuk-puppet/blob/563f9cd09a6e6d9bd49550e7bc44da6b2162490b/modules/govuk/manifests/apps/publishing_api.pp#L105-L124).

djsd123:
Made a slight change per the example from the publishing API.

surminus:
I think this looks good now, but I originally wrote it so probably need review from @alexmuller or @deanwilson

alexmuller:
Getting some errors from this:

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid relationship: File[/etc/govuk/efg-training/env.d/EFG_HOST] { notify => Govuk::App::Service[efg-training] }, because Govuk::App::Service[efg-training] doesn't seem to be in the catalog
```

(I think we might have an underscore/hyphen problem - the vhost should have a hyphen but Puppet resources should have an underscore.)

PR #4867, opened 2016-09-14
Revert monitoring value back to 8 minutes.

deanwilson:
We changed the value in ad9dd9 to cause less alerts
to be raised for a self correcting problem. This change didn't
prevent issues so we're reverting it back to the older, more tested,
value.


tijmenb:
:th

PR #4869, opened 2016-09-14
Link Deploys of EFG and EFG_TRAIN

djsd123:
Part of the EFG road map.
What
The stipulation on the deployment of the newly hosted EFG app is that EFG_Training must be deployed every time EFG (prod) is deployed.  Doing this within
the current 'Deploy_App' Jenkins job would prove quite complicated so we opted for a separate deploy job in Jenkins.

How
Created a new job template specifically for EFG/EFG_Training.
Statically set choice parameters as opposed to a generated list from a rake task.
Defined new post build task 'Trigger-Parameterized-Build'.  This will trigger the same Jenkins job it's defined in with the `TARGET_APPLICATION`parameter set to 'EFG_TRAINING'.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -22,7 +22,7 @@ deployable_applications: &deployable_applications
   - courts-api
   - courts-frontend
   - designprinciples
-  - EFG
+  - efg-training
   - email-alert-api
   - email-alert-frontend
   - email-alert-monitor
--- a/modules/govuk_jenkins/templates/jobs/deploy_efg.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_efg.yaml.erb
@@ -1,0 +1,79 @@
+---
+- scm:
+    name: alphagov-deployment_Deploy_App
+    scm:
+        - git:
+            url: git@github.com:alphagov/govuk-app-deployment.git
+            branches:
+              - master
+            wipe-workspace: true
+
+- job:
+    name: Deploy_EFG
+    display-name: Deploy_EFG
+    project-type: freestyle
+    description: "<% if @environment != 'integration' %><a href=\"http://www.flickr.com/photos/fatty/9158066939/\">\r\n  <img src=\"https://farm3.staticflickr.com/2835/9158066939_374360ed56_n.jpg\">\r\n</a>\r\n<% end %><h2>You can monitor the application health using the <a href=\"https://grafana.<%= @app_domain %>/#/dashboard/file/application_http_error_codes.json\">4XX and 5XX status dashboard</a></h2>\r\n"
+    <%- if @auth_token -%>
+    auth-token: <%= @auth_token %>
+    <%- end -%>
+    properties:
+        - github:
+            url: https://github.com/alphagov/govuk-app-deployment
+    scm:
+      - alphagov-deployment_Deploy_App
+    builders:
+        - shell: |
+            #!/usr/bin/env bash
+            export DEPLOY_TO="<%= @environment -%>"
+            export DEPLOY_TASK="$DEPLOY_TASK"
+            export TAG="$TAG"
+            export ORGANISATION="<%= @environment -%>"
+            export CI_DEPLOY_JENKINS_API_KEY="<%= @ci_deploy_jenkins_api_key -%>"
+
+            if [ "$DEPLOY_FROM_GITHUB_ENTERPRISE" == "true" ]; then
+              export GIT_ORIGIN_PREFIX="git@github.gds:gds-production-backup"
+            fi
+
+            ./jenkins.sh
+        - trigger-parameterized-builds:
+            - project: Deploy_EFG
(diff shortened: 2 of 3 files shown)


modules/govuk_jenkins/templates/jobs/deploy_efg.yaml.erb
```diff
@@ -0,0 +1,80 @@
...
+            if [ "$DEPLOY_FROM_GITHUB_ENTERPRISE" == "true" ]; then
+              export GIT_ORIGIN_PREFIX="git@github.gds:gds-production-backup"
+            fi
+
+            ./jenkins.sh
+        - trigger-parameterized-builds:
+            - project: Deploy_EFG
+              predefined-parameters: |
+                TARGET_APPLICATION=EFG_TRAINING
+                TARGET_APPLICATION_GIT_REPO=$WORKSPACE/$TARGET_APPLICATION
```
surminus: This probably needs to the be same repo (i.e. not `$TARGET_APPLICATION`)

modules/govuk_jenkins/templates/jobs/deploy_efg.yaml.erb
```diff
@@ -0,0 +1,80 @@
...
+              condition: 'UNSTABLE_OR_BETTER'
```
surminus: I wonder if we should only trigger a deploy of efg-training after a `SUCCESS` status rather than `UNSTABLE_OR_BETTER`

modules/govuk_jenkins/templates/jobs/deploy_efg.yaml.erb
djsd123: Yep. Changed to `EFG`. Thanks

surminus:
I've added an additional commit: f4cd16688d96d7ac6582b1146e9f82b14e1eed9a 

I haven't squashed this so we can see the changes I've made more clearly.

modules/govuk_jenkins/manifests/job/deploy_efg.pp
```diff
@@ -0,0 +1,11 @@
+# == Class: govuk_jenkins::job::deploy_efg
+#
+# Create a file on disk that can be parsed by jenkins-job-builder
+#
+class govuk_jenkins::job::deploy_efg {
+  file { '/etc/jenkins_jobs/jobs/deploy_efg.yaml':
+    ensure  => present,
+    content => template('govuk_jenkins/jobs/deploy_efg.yaml.erb'),
+    notify  => Exec['jenkins_jobs_update'],
+  }
```
alexmuller: Would it be better to include this file resource as part of the `govuk_jenkins::job::deploy_app` class? Otherwise this class will have to be included in the list of Jenkins jobs.

There are lots of variables used in the template like `@ci_deploy_jenkins_api_key` which is undefined here.

hieradata/common.yaml
```diff
@@ -22,7 +22,7 @@ deployable_applications: &deployable_applications
   - courts-api
   - courts-frontend
   - designprinciples
-  - EFG
+  - efg-training
```
alexmuller: We should remove both apps from this list, right? Otherwise `efg-training` will show up as part of the "deploy app" job.

modules/govuk_jenkins/manifests/job/deploy_efg.pp
surminus: Ah, I thought I had removed that variable from the template. It's not required.

surminus:
The way that I thought it would work is that we have a separate Deploy_EFG job, which then triggers the Deploy_App job, which can also be used to deploy `efg-training` as a standalone deployment. This was to avoid Deploy_EFG triggering itself, which would then trigger itself, which would then trigger itself... and so on. 

I guess we should actually only include this job in Staging and Production. If people think there's a better way, maybe by just manually building in the efg-training deploy job into the builder shell itself, then go for it.

hieradata/common.yaml
djsd123: This is intended. EFG will have a separate deploy job which triggers the deploy app job (upon SUCCESS) with EFG_TRAINING as the `TARGET_APP`

modules/govuk_jenkins/templates/jobs/deploy_efg.yaml.erb
```diff
@@ -0,0 +1,75 @@
...
+    scm:
+      - alphagov-deployment_Deploy_EFG
+    builders:
+        - shell: |
+            #!/usr/bin/env bash
+            export DEPLOY_TO="<%= @environment -%>"
+            export DEPLOY_TASK="$DEPLOY_TASK"
+            export TAG="$TAG"
+            export ORGANISATION="<%= @environment -%>"
+            export CI_DEPLOY_JENKINS_API_KEY="<%= @ci_deploy_jenkins_api_key -%>"
```
surminus: I don't think is needed.

PR #4870, opened 2016-09-15
Add environment variables for Devise in EFG

alexmuller:
Configuration should be passed in as environment variables, not using files.


deanwilson:
Merge when green.

PR #4871, opened 2016-09-15
Use api-elasticsearch for calculators-frontend in integration

h-lame:
For: https://trello.com/c/vaecOj9d/493-upgrade-rails-on-licence-finder-2

This is step one towards us getting rid of the 0.90 cluster.

The only deployed app that actually uses elasticsearch directly on
calculators-frontend is licencefinder and it can create it's index from
nothing so we can easily point the servers at a new cluster and create
the data again.

Tariff is still listed as an app for these machines, and it uses
elasticsearch, but it's since been deployed to the PaaS so we can ignore it.

### DIFF
--- a/hieradata/class/integration/calculators_frontend.yaml
+++ b/hieradata/class/integration/calculators_frontend.yaml
@@ -1,0 +1,6 @@
+---
+
+govuk_elasticsearch::local_proxy::servers:
+  - 'api-elasticsearch-1.api'
+  - 'api-elasticsearch-2.api'
+  - 'api-elasticsearch-3.api'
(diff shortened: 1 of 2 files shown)


h-lame:
Flagged as DNM because we need to roll this out with some care:

1. disable puppet on calculators-frontend machines
2. disable puppet on frontend loadbalancer machines
3. take one of the calculators-frontend machines out of the loadbalancer config for licencefinder on both machines and restart nginx
4. apply puppet (using this branch) to the removed calculator-frontend machine to make it use the api-elasticsearch cluster (restart the app if neccessary)
5. run the rake task to recreate the search index on the new elasticsearch cluster
6. check that the old cluster and new cluster have the same documents
7. add the machine back to the loadbalancer config for licencefinder and restart nginx
8. apply puppet (using this branch) to the remaining calculator-frontend machines (restart the app if neccessary)

h-lame:
Removed DNM because I deployed this to integration and it seems to work (thanks to @alexmuller's intervention with the firewall rules).  Ready to review + merge.

PR #4873, opened 2016-09-16
Revert "Turn on slow query logging for Whitehall MySQL slaves"

steventux:
This reverts commit c740b3fbe407e3d7d0d673d07840067eeb990113.

Once we've captured enough information in the slow query log we should turn it off again.


steventux:
@gpeng @boffbowsh we've captured 15MB of slow log since it was turned on yesterday. Are you both happy if I switch off the slow log today?

PR #4874, opened 2016-09-16
Configure Manuals Publisher application

Rosa-Fox:
- Part of the process to rename specialist-publisher to manuals-publisher

### DIFF
--- a/modules/govuk/manifests/apps/manuals_publisher.pp
+++ b/modules/govuk/manifests/apps/manuals_publisher.pp
@@ -1,0 +1,130 @@
+# == Class: govuk::apps::manuals_publisher
+#
+# Publishing App for manuals.
+#
+# === Parameters
+#
+# [*port*]
+#   The port that publishing API is served on.
+#   Default: 3064
+#
+# [*asset_manager_bearer_token*]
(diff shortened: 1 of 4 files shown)


alexmuller:
This looks good to me. I think it's worth removing all references to `nagios_memory_` in this pull request if we're not going to set it - we can add it later if the app ends up needing it.

steventux:
@alexmuller / @deanwilson this PR has been amended, would one of you mind reviewing please.

deanwilson:
It looks fine to me but as this is my first app add i'd like @alexmuller to have the final say.

modules/govuk/manifests/apps/manuals_publisher.pp
```diff
@@ -0,0 +1,130 @@
...
+#
+# [*redis_port*]
+#   Redis port for sidekiq.
+#   Default: 6379
+#
+class govuk::apps::manuals_publisher(
+  $port = 3064,
+  $asset_manager_bearer_token = undef,
+  $email_alert_api_bearer_token = undef,
+  $enabled = false,
```
steventux: Should this be set to `true` initially?

PR #4876, opened 2016-09-19
Add publishing-api envvar for RabbitMQ password

alexmuller:
Rather than using a file which is copied in during deployment.

This is for https://github.com/alphagov/publishing-api/pull/513.


alexmuller:
Sorry, updated when I realised I didn't also need to change the deployment repo, so another review would be appreciated.

djsd123:
Re-reviewed. 👍

PR #4878, opened 2016-09-19
Add environment variables for EFG

alexmuller:
These will help configure the app.


alexmuller:
> Will these envvars be needed for the training application?

Yep, afraid so!

PR #4886, opened 2016-09-20
Revise mongo backups

deanwilson:
Soon as this is deployed we'll remove the dead code from the PR.

### DIFF
--- a/hieradata/node/licensing-mongo-2.licensify.integration.publishing.service.gov.uk.yaml
+++ b/hieradata/node/licensing-mongo-2.licensify.integration.publishing.service.gov.uk.yaml
@@ -1,3 +1,4 @@
 mongodb::backup::s3_backups: true
-mongodb::s3backup::cron::realtime_hour: '*/6'
-mongodb::s3backup::cron::realtime_minute: '0'
+mongodb::s3backup::cron::hour: '5,14,19'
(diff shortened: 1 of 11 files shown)


hieradata/node/licensing-mongo-2.licensify.integration.publishing.service.gov.uk.yaml
```diff
@@ -1,3 +1,4 @@
 mongodb::backup::s3_backups: true
-mongodb::s3backup::cron::realtime_hour: '*/6'
-mongodb::s3backup::cron::realtime_minute: '0'
+mongodb::s3backup::cron::hour: '5,14,19'
```
surminus: Interestingly when testing, this failed:
`Error: Parameter hour failed on Cron[mongodb-s3backup-incremental]: 5,14,19 is not a valid hour at /var/govuk/govuk-puppet/modules/mongodb/manifests/s3backup/cron.pp:33`

Looks like this will need to be [set as an array as described in the documentation "Note that all periodic attributes can be specified as an array of values"](https://docs.puppet.com/puppet/3.8/reference/types/cron.html#cron-description)

surminus:
I think this has been superseded by https://github.com/alphagov/govuk-puppet/pull/4927

PR #4888, opened 2016-09-21
Bugfix EFG training

alexmuller:


modules/govuk/manifests/apps/efg_training.pp
```diff
@@ -28,7 +28,7 @@
 ) {
   validate_string($vhost_name)
 
-  $app_name = 'efg_training'
+  $app_name = 'efg-training'
```
surminus: I made a similar backwards change here: https://github.com/alphagov/govuk-puppet/commit/02b24011f57a2161bc8a92075d50b0dd2ab1a015

I thought that was a fix for the `govuk::app::envvar` thing, so if so that'll break, otherwise I was just doing random things it seems

alexmuller: Your commit was correct compared to the previous behaviour! It made the `$app` parameter of the `govuk::app::envvar` resource be set to the title of a `govuk::app` resource which existed, rather than one that didn't. But I think both of those should be a hyphen rather than an underscore.

surminus: OK thanks!

PR #4889, opened 2016-09-21
Add MySQL S3 backups everywhere

surminus:
This adds all the configuration which sets up MySQL S3 backups in all environments, on both Whitehall MySQL backup and MySQL backup machines. It also adds the restore scripts the appropriate masters.

The majority of this work had already been completed but we have changed focus so we are no longer doing an automated restore from Production to Staging/Integration, but should still provide the tools to allow restoring from backups for the appropriate environments.

Ref: [Trello story](https://trello.com/c/AsknJrSt/102-script-to-restore-s3-mysql-backups)

### DIFF
--- a/modules/govuk/manifests/node/s_whitehall_mysql_master.pp
+++ b/modules/govuk/manifests/node/s_whitehall_mysql_master.pp
@@ -4,11 +4,27 @@
 #
 # === Parameters
 #
-# [*whitehall_fe_password*]
-#   Password for a MySQL account called `whitehall_fe`.
+#  [*whitehall_fe_password*]
+#    Password for a MySQL account called `whitehall_fe`.
+#
+#  [*aws_access_key_id*]
+#    AWS access key ID for the S3 bucket
+#
+#  [*aws_secret_access_key*]
+#    AWS secret key for the S3 bucket
+#
+#  [*s3_bucket_name*]
+#    Name of the S3 bucket
+#
+#  [*encryption_key*]
+#    Encryption key that decrypts the backup
 #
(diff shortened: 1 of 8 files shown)


modules/govuk/manifests/node/s_whitehall_mysql_master.pp
```diff
@@ -36,4 +52,13 @@
   }
 
   Govuk_mount['/var/lib/mysql'] -> Class['govuk_mysql::server']
+
+  if $s3_bucket_name {
```
surminus: @rjw1 mentioned that it might be good to have _all_ the required pieces defined in the conditional here: the define will already [get upset if we don't already have all the parts](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_mysql/manifests/xtrabackup/backup.pp#L37-L40), so the conditional is there just to ensure Puppet doesn't break if we have a bucket defined. I just wonder if it's worth having a long  line of `if $s3_bucket_name and $aws_access_key_id and $aws_secret_access_key and $encryption_key {` for this purpose?

djsd123:
Is this intended to pull data from integration or production on a restore as the bucket name now points to the integration bucket.

surminus:
I provided the answer to the question in PR summary:

"...we have changed focus so we are no longer doing an automated restore from Production to Staging/Integration, but should still provide the tools to allow restoring from backups for the appropriate environments."

So Integration pulls from the Integration bucket, Staging pulls from the Staging bucket and so on. We are not doing "environment syncs" with these scripts (yet).

deanwilson:
Thanks for the clarification. Please merge when ready.

PR #4890, opened 2016-09-21
Add firewall rule for licence-finder Elasticsearch

alexmuller:
licence-finder needs to use Elasticsearch to display licence search results. We want to move it from the old Elasticsearch to the new one so that we can switch the old one off.

/cc @h-lame


h-lame:
Is there any reason that we shouldn't merge this even if we aren't yet ready to merge the other PR for licence-finder ( #4871 ) that prompted this?

alexmuller:
Nope, just me forgetting about it. Sorry!

PR #4891, opened 2016-09-21
Add timeout command for MySQL S3 backups

surminus:
This issue was raised because there was an unusual situation where gof3r was unable to reach the S3 bucket, and while gof3r threw an error, because of the pipe structure the streaming backup continued to stream the backup without finishing because it had no valid endpoint (as far as I can gather).

It would be safer to explicitly set a timeout value for both the incremental and base backup pushes. If they don't complete in the given time it's likely that they have run into an error.


surminus:
I missed this being approved and it's happened again, booo! Merging now.

PR #4894, opened 2016-09-22
Cdn logrotate rotate

djsd123:
What
Logrotate never removes CDN's log files. It appears that logrotate compresses CDN's log files but due to what I think is a config error.  They never get removed.  The result is log files building up and causing alerts for disk space.

How
Modify logrotate to remove the logs after 30 days.  The previous attempt to amend this was done in error by setting the `rotate` file count to 720.
Please see https://linuxconfig.org/logrotate-8-manual-page for more info.

### DIFF
--- a/modules/govuk_cdnlogs/templates/etc/logrotate.cdn_logs_hourly.conf.erb
+++ b/modules/govuk_cdnlogs/templates/etc/logrotate.cdn_logs_hourly.conf.erb
@@ -4,7 +4,7 @@
 -%>
 
 {
-  rotate 720
+  maxage 30
   dateext
   dateformat -%Y%m%d-%s
   compress
--- a/modules/govuk_cdnlogs/spec/classes/govuk_cdnlogs_spec.rb
+++ b/modules/govuk_cdnlogs/spec/classes/govuk_cdnlogs_spec.rb
@@ -117,7 +117,7 @@ ruleset\(name="cdn-giraffe"\) \{
         it 'should rotate the elephant logs hourly' do
           is_expected.to contain_file('/etc/logrotate.cdn_logs_hourly.conf')
             .with_content(%r{^/tmp/logs/cdn-elephant\.log$})
-            .with_content(/rotate 720/)
+            .with_content(/maxage 30/)
         end
       end
     end


alexmuller:
I'm having trouble seeing which disk is running low on space, where are you seeing the alert?

render-2

carvil:
@alexmuller we saw an alert in the morning on `logs-cdn-1.management.production`. We then went into `/mnt/logs_cdn` and saw a large number of old rotated log files from last year still there.

Because those log files have a date associated with them (e.g. `cdn-govuk.log-20160922`) we thought that the existing configuration of logrotate would not work appropriately (i.e. we would never reach 30 rotations on a file with that name because the date changes every day). Is this a fair assumption?

Those old files from 2015 would have been deleted if we looked into the creation date of a file, rather than the number of rotations of a particular file.

We have seen this across other boxes for application logs too (not during this 2nd line shit, but in past shifts). Because we don't look at the creation date of files, when logrotate fails, we end up with old files hanging around not being properly cleaned up.

Does this make sense?

djsd123:
@alexmuller,
@carvil already cleared some files which some were last touched December 2015.  Which is probably why you didn't see the alert.

I think that rotating every 720 files is a bit excessive in any case?

alexmuller:
Ok, I think we should revisit the alerting thresholds for that disk because it's never gone below 400GB free:

render-8

alexmuller:
It looks like the alert might have been for the root partition?

```
[2016-09-22 08:56:14] SERVICE ALERT: logs-cdn-1.management.publishing.service.gov.uk;low available disk space on root;WARNING;HARD;4;DISK WARNING - free space: / 9033 MB (19% inode=87%):
[2016-09-22 09:21:14] SERVICE ALERT: logs-cdn-1.management.publishing.service.gov.uk;low available disk space on root;OK;HARD;4;DISK OK - free space: / 28220 MB (62% inode=88%):
```

I'm a bit confused by this graph:

render-9

modules/govuk_cdnlogs/templates/etc/logrotate.cdn_logs_hourly.conf.erb
```diff
@@ -4,7 +4,7 @@
 -%>
 
 {
-  rotate 720
+  maxage 30
```
surminus: This is an hourly log rotation, so this would rotate after 30 hours worth of rotations. I appreciate it didn't work as expected, but I think this change is incorrect.

PR #4896, opened 2016-09-23
Upgrade ruby version

deanwilson:


h-lame:
This PR should also correct the BUNDLED_WITH section of Gemfile.lock to match the version of bundler pinned in `modules/govuk_rbenv/manifests/init.pp`.

h-lame:
Get's a :+1: from me!

alexmuller:
I'm having some trouble running this in Vagrant:

```
Error: Could not run: cannot load such file -- gpgme_n
```

And I'm going to be nitpicky about the commit message, sorry! This is absolutely the kind of change where I can imagine myself looking back at it in a few weeks or months time and being frustrated at not being able to remember things.

Please can you put all these things in the same commit? They're all the same logical change and it'll be useful to see them together.

Then it would be useful to add the discussion we had in the office:

The first Puppet run will always use system Ruby, which for Trusty is 1.9.3 but for Xenial is 2.3.1. The only way to get around that is to set up rbenv during the preamble which I don't think anyone wants to do.

As soon as the first Puppet run is complete, all future ones will use rbenv's Ruby.

deanwilson:
We've not had time to reproduce this error so I'm closing this.

djsd123:
Reopening this as it is a prereq for https://trello.com/c/uD3LBdKG

SamLR:
@alexmuller can we merge this?

deanwilson:
There's be no progress on this so we're going to live without the resilience and reclose the pull request.

PR #4905, opened 2016-09-26
[do not merge] updated to replace manuals publisher deploy with specialist publisher

Seraphiana:
specialist publisher deployment has been using a combination of manuals publisher (previously
called specialist publisher) and specialist publisher rebuild. This deployment should work alongside [this pull request](https://github.com/alphagov/govuk-app-deployment/pull/49) 
to replace this with a single deployment of specialist publisher rebuild, living at the address previously
occupied by manuals publisher. It also removes the custom routing to enable the combined deployment previously utilised.

### DIFF
--- a/hieradata/class/backend.yaml
+++ b/hieradata/class/backend.yaml
@@ -33,7 +33,7 @@ govuk::apps::share_sale_publisher::mongodb_nodes:
   - 'mongo-2.backend'
   - 'mongo-3.backend'
 
-govuk::apps::specialist_publisher::mongodb_name: 'specialist_publisher_production'
+govuk::apps::specialist_publisher::mongodb_name: 'govuk_content_production'
 govuk::apps::specialist_publisher::mongodb_nodes:
   - 'mongo-1.backend'
   - 'mongo-2.backend'
--- a/hieradata/class/backend.yaml
+++ b/hieradata/class/backend.yaml
@@ -33,6 +33,12 @@ govuk::apps::share_sale_publisher::mongodb_nodes:
   - 'mongo-2.backend'
   - 'mongo-3.backend'
 
+govuk::apps::specialist_publisher::mongodb_name: 'specialist_publisher_production'
+govuk::apps::specialist_publisher::mongodb_nodes:
+  - 'mongo-1.backend'
+  - 'mongo-2.backend'
+  - 'mongo-3.backend'
+
 govuk::apps::specialist_publisher_rebuild::mongodb_name: 'govuk_content_production'
 govuk::apps::specialist_publisher_rebuild::mongodb_nodes:
   - 'mongo-1.backend'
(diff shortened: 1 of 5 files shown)


alexmuller:
There are a couple of things changing here and I think it'd be good to split them out into smaller pull requests so we're all sure what's going on.

I think the first thing to do is to get both the `specialist-publisher` and `specialist-publisher-rebuild` versions of the apps running the same code. Then we can worry about removing the routing weirdness later.

For that to happen I think the only thing that needs to change is adding some environment variables to the old app?

Seraphiana:
@alexmuller I'm not entirely sure what you mean. I've pulled out the changes to the routing, but not sure what environment variables should be updated? I assumed since SPR and SPv1 were sharing the same environment variables previously then it should be fine?

steventux:
Differences in ENV vars between specialist-publisher and specialist-publisher-rebuild (where both env vars are present):
`GOVUK_APP_NAME, GOVUK_APP_LOGROOT, PUBLISHING_API_BEARER_TOKEN, PORT`

hieradata/class/backend.yaml
```diff
@@ -33,6 +33,12 @@ govuk::apps::share_sale_publisher::mongodb_nodes:
   - 'mongo-2.backend'
   - 'mongo-3.backend'
 
+govuk::apps::specialist_publisher::mongodb_name: 'specialist_publisher_production'
```
steventux: Should be `govuk_content_production`

hieradata/development.yaml
```diff
@@ -163,6 +163,8 @@ govuk::apps::signon::enable_procfile_worker: false
 govuk::apps::smartanswers::expose_govspeak: true
 govuk::apps::specialist_publisher::enable_procfile_worker: false
 govuk::apps::specialist_publisher::publish_pre_production_finders: true
+govuk::apps::specialist_publisher::mongodb_nodes: ['localhost']
+govuk::apps::specialist_publisher::mongodb_name: 'specialist_publisher_development'
```
steventux: I think this should also be `govuk_content_development`

hieradata/class/backend.yaml
Seraphiana: link doesn't work ::::/

hieradata/development.yaml
Seraphiana: I'm a little concerned that the one below is apparently a production database, and is different to this one. Surely they should be using the same db_name?

hieradata/class/backend.yaml
steventux: You possibly don't have access to this repo, it shows that specialist publisher uses a common mongo db.

steventux:
Closing in favour of https://github.com/alphagov/govuk-puppet/pull/4923

PR #4906, opened 2016-09-27
Amend the mongodb_name this got lost in translation

steventux:
https://github.gds/gds/alphagov-deployment/blob/master/specialist-publisher/to_upload/mongoid.yml#L2
Manuals Publisher should use the same collection as Specialist Publisher v1 did.


steventux:
thanks @rjw1

PR #4912, opened 2016-09-28
Cdn logrotate rotate

djsd123:
What
Logrotate never removes CDN's log files. It appears that logrotate compresses CDN's log files but due to what I think is a config error. They never get removed. The result is log files building up and causing alerts for disk space.

How
Set cdn logrotate's hourly config to rotate cdn logs hourly.

### DIFF
--- a/modules/govuk_cdnlogs/templates/etc/logrotate.cdn_logs_hourly.conf.erb
+++ b/modules/govuk_cdnlogs/templates/etc/logrotate.cdn_logs_hourly.conf.erb
@@ -4,7 +4,8 @@
 -%>
 
 {
-  maxage 30
+  rotate 720
+  hourly
   dateext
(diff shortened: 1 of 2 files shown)


modules/govuk_cdnlogs/templates/etc/logrotate.cdn_logs_hourly.conf.erb
```diff
@@ -4,7 +4,7 @@
 -%>
 
 {
-  rotate 720
+  hourly
```
surminus: I think this should be added in addition to `rotate 720`. I think we still require the number to rotate on, but also need to specify the `hourly` parameter.

djsd123: Changes made.!

PR #4914, opened 2016-09-28
Schedule Auto PostgreSQL Backups

djsd123:
What
Auto postgreSQL backups run daily (every morning) via cron and fail.
The error output:
`pg_dump: [archiver (db)] query failed: ERROR:  cache lookup failed for index 506280`

It's not entirely clear but the backup runs while the envsync is in progress, which would cause
some indexes to not be available. This is further clarified by running the script and it exiting with no
errors after the envsync has completed.

How
Move the main backup script from cron daily to `/usr/local/bin`.
Create a cron job to run the script at `08:30`, `07:30` UTC time.
The Icinga alert we're adressing only occurs on Staging and Integration, so we set the backup to run at `6:30`, `5:30` UTC in production which is the time it would normally run with `cron.daily`.


modules/govuk_postgresql/manifests/backup.pp
```diff
@@ -1,11 +1,42 @@
...
+class govuk_postgresql::backup (
+  $auto_postgresql_backup_hour = '7',
+  $auto_postgresql_backup_minute = '30'
+) {
+
     $threshold_secs = 28 * 3600
     $service_desc = 'AutoPostgreSQL backup'
 
-    file {'/etc/cron.daily/autopostgresqlbackup':
+    file {'/etc/cron.d/autopostgresqlbackup':
```
surminus: This should be created in `/usr/local/bin`. The Cron::Crondotdee defined type will create the file in `/etc/cron.d`, you just have to pass in the command that needs to be run.

modules/govuk_postgresql/manifests/backup.pp
```diff
@@ -1,11 +1,42 @@
...
         mode    => '0755',
-        content => template('govuk_postgresql/etc/cron.daily/autopostgresqlbackup.erb'),
+        content => template('govuk_postgresql/usr/local/bin/autopostgresqlbackup.erb'),
+    }
+
+    cron::crondotdee { 'auto_postgresql_backup':
+        command => '/usr/local/bin/autopostgresqlbackup',
+        hour    => $auto_postgresql_backup_hour,
+        minute  => $auto_postgresql_backup_minute,
+        require => File['/etc/cron.d/autopostgresqlbackup'],
```
surminus: This should be sorted in the define itself.

modules/govuk_postgresql/manifests/backup.pp
```diff
@@ -20,7 +51,7 @@
     #    PREBACKUP="/etc/postgresql-backup-pre"
     file {'/etc/default/autopostgresqlbackup':
         source  => 'puppet:///modules/govuk_postgresql/etc/default/autopostgresqlbackup',
-        require => File['/etc/cron.daily/autopostgresqlbackup'],
+        require => File['/etc/cron.d/autopostgresqlbackup'],
```
surminus: Probably change this to `/usr/local/bin`?

modules/govuk_postgresql/manifests/backup.pp
```diff
@@ -1,11 +1,42 @@
-# FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
-class govuk_postgresql::backup {
+# Class: govuk_postgresql::backup
```
surminus: Good stuff on class docs

modules/govuk_postgresql/manifests/backup.pp
```diff
@@ -1,11 +1,42 @@
...
+#
+# [*threshold_secs*]
+#  The time period in which an alert should be raised if no passive check
+#  submissions have been received.
+#
+# [*service_desc*]
+#  A simple description of the check that is alerting.
+#
+class govuk_postgresql::backup (
+  $auto_postgresql_backup_hour = '7',
```
surminus: This should be an integer, not a string

modules/govuk_postgresql/manifests/backup.pp
```diff
@@ -1,11 +1,42 @@
...
+  $auto_postgresql_backup_minute = '30'
```
surminus: This should be an integer, not a string

modules/govuk_postgresql/manifests/backup.pp
djsd123: I would work either way but technically you're right.  changed! 👍

modules/govuk_postgresql/manifests/backup.pp
djsd123: changed 👍

modules/govuk_postgresql/manifests/backup.pp
djsd123: My bad.  Corrected

modules/govuk_postgresql/manifests/backup.pp
djsd123: Same As Above. SAA

modules/govuk_postgresql/manifests/backup.pp
djsd123: SAA

surminus:
Functionally I think this looks fine. My final comment would be whether we're happy to do PostgreSQL backups in Production at 07:30 in the morning, especially as traffic starts to come. I think the problem you're trying to solve only affects Integration and Staging, so I wouldn't want to cause a problem in Production solving that.

djsd123:
Can we please have:

* A comma on the last class param.
* Change the requires around so that the binary requires the config file.

modules/govuk_postgresql/manifests/backup.pp
```diff
@@ -2,6 +2,14 @@
 #
 # This class sets up local backups for PostgreSQL.
 #
+# === Parameters
+#
+# [*$auto_postgresql_backup_hour*]
+#  The hour(s) that auto_postgresql_backup(s) run(s).
+#
+# [*$auto_postgresql_backup_minute*]
```
deanwilson: Can we remove the `$` from this param name please?

modules/govuk_postgresql/manifests/backup.pp
```diff
@@ -2,6 +2,14 @@
 #
 # This class sets up local backups for PostgreSQL.
 #
+# === Parameters
+#
+# [*$auto_postgresql_backup_hour*]
```
deanwilson: Can we remove the `$` from this param name please?

PR #4916, opened 2016-09-29
Use api-elasticsearch for calculators-frontend in staging

h-lame:
As a follow up to #4871 we make staging calculators-frontend machines
use the new api-elasticsearch cluster.  We want to run this for a few
days to check it's ok before moving production to the new cluster.


h-lame:
Flagged as DNM for now as, like #4871, there's a complicated process for rolling this out to avoid outages.

h-lame:
Removing DNM because I've disabled puppet on the machines this would affect.  Merging so I can deploy.

PR #4919, opened 2016-09-29
Delete last trade tariff config

tijmenb:
https://governmentasaplatform.blog.gov.uk/2016/09/29/first-service-paas/ reminded me we can do this now.

https://trello.com/c/sjlfSlqT/232-clean-up-trade-tariff


alexmuller:
This will also need to be removed from `hieradata/class/calculators_frontend.yaml` because that's where this class gets included on the machine.

alexmuller:
And there's a bit of hieradata in `common.yaml` for this class and `govuk::apps::tariff::publishing_api_bearer_token` in the deployment repo.

tijmenb:
Hah great, I missed like 95% of the things.

alexmuller:
Sorry! You are [doing our work for us though](https://trello.com/c/sjlfSlqT/232-clean-up-trade-tariff) so thanks ;)

tijmenb:
Well then I guess I'm webops now!

Deployment: https://github.gds/gds/deployment/pull/1102
Development repo: https://github.gds/gds/development/pull/343

hieradata/common.yaml
```diff
@@ -1122,8 +1115,6 @@ monitoring::checks::smokey::features:
     feature: smartanswers
   check_static_mirrors:
     feature: mirror
-  check_tariff:
-    feature: tariff
```
alexmuller: I think this one should stay, so that Smokey is still in Icinga.

tijmenb: You'll need to explain that to me sometime. In the meantime, updated branch.

alexmuller: This is the subset of Smokey tasks that we create Icinga checks for - the less important Smokey tasks only run in Jenkins.

PR #4923, opened 2016-09-29
Update Specialist Publisher application env

steventux:
Specialist Publisher will use the codebase from Specialist Publisher Rebuild very shortly, so ensure that the same environment is created for both applications to ease deployment.

This means the same ENV variables should be made available and any config from alphagov-deployment should be added to the environment where relevant.

Secrets here https://github.gds/gds/deployment/pull/1103 (this has now been merged)

### DIFF
--- a/hieradata/class/backend.yaml
+++ b/hieradata/class/backend.yaml
@@ -43,6 +43,12 @@ govuk::apps::share_sale_publisher::mongodb_nodes:
   - 'mongo-2.backend'
   - 'mongo-3.backend'
 
+govuk::apps::specialist_publisher::mongodb_name: 'govuk_content_production'
+govuk::apps::specialist_publisher::mongodb_nodes:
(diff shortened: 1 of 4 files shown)


alexmuller:
I pushed an additional commit which I think makes the overall set of changes in this pull request much easier to review - feel free to either squash it in or discard it, but it shouldn't be merged with the message as it exists :)

steventux:
@alexmuller: This is the branch I test deployed to integration, it supplied the correct environment vars for specialist-publisher and used the same cookie secret. Would you mind reviewing when you have a moment.

alexmuller:
There are lots of other places where it's worth using the `$app_name` variable from the final commit.

Apart from that this looks good to me :+1:

Nitpick: the first 2 commits could be squashed together.

steventux:
@alexmuller thanks :+1: I've amended

PR #4924, opened 2016-09-30
Use api-elasticsearch on all calculators-frontend environments

h-lame:
This finishes the work started in #4871 and continued in #4916 to make
the calculators_frontend machines use the api-elasticsearch.api cluster
instead of the elasticsearch.backend cluster.


h-lame:
Flagged as DNM because, just like the precursors, there's a process beyond just deploying a puppet for rolling this out to avoid downtime.

h-lame:
Removed DNM, going to deploy puppet (will disable puppet on the machines we care about on production to follow rollout).

PR #4926, opened 2016-09-30
Start enabling future parser

alexmuller:


hieradata/class/jumpbox.yaml
```diff
@@ -2,3 +2,5 @@
 
 govuk_safe_to_reboot::can_reboot: 'careful'
 govuk_safe_to_reboot::reason: 'Warn connected users before rebooting'
+
+puppet::future_parser: true
```
djsd123: Why is this set on the jumpbox?

alexmuller: Because the jumpbox is a simple machine where the future parser works - we'll have to spend more time testing the more complex machines.

PR #4928, opened 2016-10-03
Remove Felicity Moon

brenetic:
Felicity has left GOV.UK


alexmuller:
I think this is the same as #4921.

brenetic:
ah yes! I did not spot that PR

PR #4929, opened 2016-10-03
Remove load balancer config for specialist document paths

steventux:
Cleanup from https://trello.com/c/m9ZWAIbQ/284-rename-the-specialist-publisher-v1-to-manuals-publisher-large 

Specialist Publisher is now running the v2 rebuild code for specialist documents,
Manuals Publisher manages manuals content.
We no longer need to direct traffic for specialist document routes to a different
application as we've repurposed Specialist Publisher, so remove this config.


alexmuller:
Duplicate of #4915 I think, but either one is good.

steventux:
:+1: I'll leave https://github.com/alphagov/govuk-puppet/pull/4915 open then.

PR #4931, opened 2016-10-04
Add PostgreSQL WAL-E restore script

surminus:
While writing documentation it occurred to me that we did not have a basic restore script for restores from the same environment. This commit adds the script. The script is essentially the same as the env_sync script which has been successfully tested, except that it does not contain any Production key passphrases or references to a different envdir directory than the one that should already be present on the machine that sends backups to S3. 

It also removes the env_sync details added to Integration that was used for testing restores from Production. We are currently not using this method so it should be removed until we decide to use it.

### DIFF
--- a/modules/govuk_postgresql/manifests/wal_e/backup.pp
+++ b/modules/govuk_postgresql/manifests/wal_e/backup.pp
@@ -194,4 +194,11 @@ define govuk_postgresql::wal_e::backup (
         host_name           => $::fqdn,
         notes_url           => monitoring_docs_url(postgresql-s3-backups),
       }
+
+    file { '/usr/local/bin/wal-e_restore':
+      ensure  => present,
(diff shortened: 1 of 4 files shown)


modules/govuk_postgresql/manifests/wal_e/backup.pp
```diff
@@ -194,4 +194,12 @@
         host_name           => $::fqdn,
         notes_url           => monitoring_docs_url(postgresql-s3-backups),
       }
+
+    file { '/usr/local/bin/wal-e_restore':
+      ensure  => present,
+      content => template('govuk_postgresql/usr/local/bin/wal-e_restore.erb')
```
deanwilson: Missing ,

deanwilson:
Other than the missing `,` this looks mergeable once the tests pass.

surminus:
Thanks @deanwilson, please hit the approve button!

PR #4935, opened 2016-10-04
Add future parser

surminus:
**We need a controlled merge of this for testing purposes.**

This enables to help test the upgrade to Puppet 4.

The `future_parser` config is only required on the puppetmaster, but it doesn't hurt to have it on the client config.

The `stringify_facts` config fixes any problems we have with how data types are evaluated in Puppet 4.

The `strict_variables` config is set to help us debug issues easier 
because Puppet 4 has a stricter checking for undefined variables.

### DIFF
--- a/modules/puppet/manifests/init.pp
+++ b/modules/puppet/manifests/init.pp
@@ -8,11 +8,16 @@
 #   Whether this environment should be controlled by a master
 #   or should run from a local catalog.
 #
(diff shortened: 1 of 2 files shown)


deanwilson:
I think this is worthwhile merging if we default it to false (as we do) and remove the hiera entry enabling it.

PR #4938, opened 2016-10-05
Do the housekeeping before notifying nagios

cbaines:
Notifying nagios can fail, and if it does, this causes the housekeeping
function not to run, causing other problems with the disk space.

Running the housekeeping function before notifying nagios will ensure
that this always runs, regardless of issues with nagios.


alexmuller:
Thanks!

PR #4939, opened 2016-10-05
Add Lyndon Garvey to Jenkins

sihugh:
Lyndon's MOJ and needs to be able to see Integration builds for
Specialist Publisher and related things.

Also moved Laura up a bit as A comes before E (even after C)

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -79,9 +79,10 @@ govuk_jenkins::config::admins:
   - kanemorgan
   - katiesmith
   - kevindew
+  - lauramartin
   - leelongmore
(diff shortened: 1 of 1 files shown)


alexmuller:
Do we have a date when Lyndon won't be working on specialist publisher anymore? I suspect we might forget to do a leaver's ticket.

sihugh:
No, but I am responsible for it :-).  I've added a [Trello card to our planning board](https://trello.com/c/sEXbz7x2/438-raise-leaver-s-ticket-for-lyndon-garvey) so we won't forget.

brenetic:
MWR

PR #4944, opened 2016-10-06
Set environment variables for govuk_delivery

davbo:
Changes to set environment variables for the configuration of
govuk_delivery. Using environment variables we can store the credentials
for govdelivery in hiera eYAML.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -337,6 +337,15 @@ govuk::apps::govuk_crawler_worker::root_urls:
   - 'https://assets.publishing.service.gov.uk/'
   - 'https://www.gov.uk/'
 
+govuk::apps::govuk_delivery::redis_host: "%{hiera('sidekiq_host')}"
+govuk::apps::govuk_delivery::redis_port: "%{hiera('sidekiq_port')}"
+govuk::apps::govuk_delivery::mongodb_hosts:
+  - 'mongo-1.backend'
+  - 'mongo-2.backend'
+  - 'mongo-3.backend'
+govuk::apps::govuk_delivery::mongodb_database: 'govuk_delivery'
+govuk::apps::govuk_delivery::list_title_format: '%s'
+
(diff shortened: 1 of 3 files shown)


hieradata/common.yaml
```diff
@@ -337,6 +337,15 @@ govuk::apps::govuk_crawler_worker::root_urls:
   - 'https://assets.publishing.service.gov.uk/'
   - 'https://www.gov.uk/'
 
+govuk::apps::govuk_delivery::redis_host: 'redis-1.backend'
```
boffbowsh: Should probably use `"%{hiera('sidekiq_host')}"` as elsewhere in this file. I know this is for configuring Celery rather than Sidekiq, but I can't think of a reason to separate them.

modules/govuk/manifests/apps/govuk_delivery.pp
```diff
@@ -1,22 +1,108 @@
...
 class govuk::apps::govuk_delivery(
-  $enabled = true,
+  $redis_host,
+  $redis_port,
+  $mongodb_hosts,
+  $mongodb_database,
+  $list_title_format,
+  $govdelivery_username = undef,
+  $govdelivery_password = undef,
+  $govdelivery_account_code = undef,
```
boffbowsh: Missing the puppetdoc for this one

davbo:
@boffbowsh addressed your comments, thanks for reviewing!

modules/govuk/manifests/apps/govuk_delivery.pp
```diff
@@ -1,22 +1,108 @@
...
+  $govdelivery_hostname = undef,
+  $govdelivery_signup_form = undef,
   $port = '3042',
-  $enable_procfile_worker = true,
+  $enable_procfile_worker = true
```
alexmuller: Sorry for the annoying nitpick comment, please can we keep the trailing comma in here to keep future diffs smaller? Apart from that this is good to go!

PR #4945, opened 2016-10-07
Add EFG rebuild app

surminus:
The new EFG application will be developed alongside and deployed on the same machine as the current application. It also shares the same EFG database, so a new user has been created which will be granted access to the current EFG database.


hieradata/production.yaml
```diff
@@ -41,6 +41,9 @@ environment_ip_prefix: '10.3'
 govuk::apps::efg::ssl_certtype: 'sflg'
 govuk::apps::efg::vhost_name: "www.sflg.gov.uk"
 
+govuk::apps::efg::ssl_certtype: 'sflg'
+govuk::apps::efg::vhost_name: "rebuild.sflg.gov.uk"
```
surminus: I was just doing what I was told [in the story](https://trello.com/c/xkAb7yDA/237-run-a-rebuild-copy-of-the-efg-app) 😉

modules/govuk/manifests/apps/efg_rebuild/db.pp
```diff
@@ -0,0 +1,30 @@
...
+  mysql::db { 'efg_production':
+    user     => 'efg_rebuild',
+    host     => '%',
+    password => $mysql_password,
+  }
+
+  mysql_database { 'efg_il0':
+    charset => 'utf8',
+    collate => 'utf8_general_ci',
+  }
```
surminus: True, probably just need to set grants instead.

surminus:
@alexmuller addressed comments. govuk::apps::efg_rebuild::db now creates a user and sets a grant rather than using the mysql::db resource again.

hieradata/production.yaml
alexmuller: Interesting, sorry! Either is fine :)

PR #4951, opened 2016-10-11
Add Machine, Processes and Bouncer dashboards

cbaines:
I built these on the production grafana instance, but adding them to puppet for
persistence, and so that they appear in all environments.

The Machine and Processes dashboards are generic dashboards to provide easier
access to the metrics in Graphite about machine and app process metrics
respectively.

I created the Bouncer dashboard while investigating a high number of errors in
production.

### DIFF
--- a/modules/grafana/files/dashboards/bouncer.json
+++ b/modules/grafana/files/dashboards/bouncer.json
@@ -1,0 +1,476 @@
+{
+  "id": 12,
+  "title": "Bouncer",
+  "originalTitle": "Bouncer",
+  "tags": [],
+  "style": "dark",
+  "timezone": "browser",
+  "editable": true,
+  "hideControls": false,
+  "sharedCrosshair": false,
+  "rows": [
+    {
+      "collapse": false,
+      "editable": true,
+      "height": "450px",
+      "panels": [
+        {
+          "aliasColors": {},
+          "bars": false,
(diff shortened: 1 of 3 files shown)


modules/grafana/files/dashboards/bouncer.json
```diff
@@ -0,0 +1,470 @@
...
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "span": 12,
+          "stack": true,
+          "steppedLine": false,
+          "targets": [
+            {
+              "refId": "A",
+              "target": "aliasByNode(bouncer-*_redirector.nginx.nginx_requests, 0)"
```
surminus: Will this generically export correctly to Staging and Integration? From what I can recall we put some graphs into templates because the target machine adds `_staging` on to the machine name, for example. I guess you could probably just add a wildcard `*` to be safe.

cbaines: I did test this, but I think I messed up exporting the dashboard somehow. I've exported it again, and updated the branch.

cbaines: But yes, a wildcard is needed there to make this work in staging and integration.

cbaines:
Great, thanks for your review @surminus :)

PR #4952, opened 2016-10-13
Configure the service uri for trade tariff

cbaines:
The content-store can register the route for trade tariff, which without
this configuration will set the backend to the incorrect url.


modules/govuk/manifests/apps/content_store.pp
```diff
@@ -89,5 +89,8 @@
     "${title}-ERRBIT_API_KEY":
       varname => 'ERRBIT_API_KEY',
       value   => $errbit_api_key;
+    "${title}-PLEK_SERVICE_TARIFF_URI":
+      varname => 'PLEK_SERVICE_TARIFF_URI',
+      value   => $trade_tariff_uri;
```
boffbowsh: I don't believe this is enough to bring the variable in from hieradata. The variable should be namespaced too.

hieradata/production.yaml
```diff
@@ -621,3 +621,5 @@ shell::shell_prompt_string: 'production'
 
 mongodb::s3backup::backup::s3_bucket: 'govuk-mongodb-backup-s3-production'
 mongodb::s3backup::backup::s3_bucket_daily: 'govuk-mongodb-backup-s3-daily-production'
+
+trade_tariff_uri: 'https://trade-tariff.cloudapps.digital/'
```
daibach: This isn't the correct URL for Trade Tariff production.  It should be https://tariff-frontend-production.cloudapps.digital/

modules/govuk/manifests/apps/content_store.pp
cbaines: I don't have much experience using Puppet, but I've had another guess which I hope is more correct... https://github.com/alphagov/govuk-puppet/pull/4952/commits/7a0854e19362ab05cce923fd46174da951bfd703#diff-317c81f58cd20afec981e1e6c339703fR82

hieradata/production.yaml
cbaines: Thanks, now updated.

hieradata/staging.yaml
```diff
@@ -564,3 +564,5 @@ users::pentest_machines:
 
 mongodb::s3backup::backup::s3_bucket: 'govuk-mongodb-backup-s3-staging'
 mongodb::s3backup::backup::s3_bucket_daily: 'govuk-mongodb-backup-s3-daily-staging'
+
+trade_tariff_uri: 'https://tariff-frontend-staging.cloudapps.digital/'
```
boffbowsh: I'm wondering if this should be namespaced more. Something like `plek_overrides::tariff`? This is probably more a question for someone on infrastructure

cbaines: I'm unsure as I don't have much Puppet experience. My only comment is that this bit of information may be used by other classes, and not just when configuring Plek, so it may be useful to not use that as the namespace (and perhaps use something like `service::` instead).

rjw1:
vagrant.yaml will need a value so that tests can pass.

hieradata/staging.yaml
surminus: I think namespacing in that way might imply there's a class called `plek_overrides` so perhaps it'd be clearer if it were called `plek_trade_tariff_uri`. I think it's fine either way though, personally.

hieradata/integration.yaml
```diff
@@ -335,3 +335,5 @@ users::usernames:
   - tomridd
   - tomrussell
   - tomwhite
+
+trade_tariff_uri: 'https://tariff-frontend-dev.cloudapps.digital/'
```
surminus: This could probably go in `common.yaml` and just have overrides in Staging and Production as it would be useful to set this default on dev or local VMs, too.

cbaines: Ok, I've now moved it to common.yaml

PR #4953, opened 2016-10-13
Prepare to remove sp rebuild apps

steventux:
Mark specialist-publisher-rebuild and standalone as absent.
Please shout if more puppet config can be removed in this PR.


steventux:
Thanks @alexmuller updated.

PR #4955, opened 2016-10-14
Set higher disk time threshold

surminus:
Set a higher disk time threshold for mongo machines that take regular backups so it's less likely we will be alerted on them. We would still be alerted on a sustained spike of disk usage of over 10 minutes.


deanwilson:
It feels like this should be something we should be able to set in a mongo class rather than on a per-node hiera file basis. We should chat about this in the team meeting and see if the setting has a better home.

alexmuller:
I moved this so it affects all the backend MongoDB machines because those are the only ones which are alerting in prod right now - does that make sense?

surminus:
That's _probably_ OK I reckon. It's only `mongo-2` doing the backups but I guess it doesn't hurt to have it across the stack.

PR #4958, opened 2016-10-17
Delay cron process to dump elastic search indexes

pmanrubia:
Delays the execution of the cron process `es_dump` from 3am to 4am

We are having intermittent failures in the [process that copies data 
from production to staging][1]. The process usually fails when importing
`/tmp/es_dump_production/government.zip` See executions 437, 478, 475

We have not been able to identify the cause of this issue, but as 
we have another process that runs daily at 03:00, we want to explore if
increasing the gap between them prevent the timeout from happening.

[1]: https://deploy.publishing.service.gov.uk/job/copy_data_to_staging/


surminus:
It might be worth just changing the time it runs in Staging only rather than everywhere. In the Staging hieradata, set `govuk_elasticsearch::dump::run_es_dump_hour: 4`

pmanrubia:
Thank you @surminus 

Superseded by https://github.com/alphagov/govuk-puppet/pull/4962

PR #4961, opened 2016-10-17
Ensure lock directory exists

surminus:
When 78c9abf57e65b4cb136c31e877f6fb84e99215ff was deployed, the govuk_puppet script failed to run on all machines. This is because the lock directory does not exist. Add logic in the script to ensure the directory exists.

I chose to add the logic within the script rather than Puppet to avoid having to manually update all machines that already have the newest version of the script.


deanwilson:
After a discussion we'll manage this resource in puppet.

PR #4963, opened 2016-10-17
[DO NOT MERGE] Add openconnect-ppa remote so we can use backported versions

deanwilson:
Part of the supporting work for https://trello.com/c/t8FsXwzE


alexmuller:
Instead of doing this we can copy the packages into our Launchpad PPA with the Launchpad interface (done here: https://launchpad.net/~gds/+archive/ubuntu/govuk) and then update the mirror of our own PPA. @afda16 and I started work on this yesterday and it's done in integration.

PR #4966, opened 2016-10-18
Update docs to use latest Vagrant

Davidslv:


development-vm/README.md
```diff
@@ -26,7 +26,7 @@ checking for other manufacturers.
   * You have [OSX GCC tools installed](https://github.com/kennethreitz/osx-gcc-installer) (or via XCode)
   * You have [VirtualBox](https://www.virtualbox.org/) installed
   * You have an account on [GitHub.com](https://github.com)
-  * You have [Vagrant 1.7.4](https://releases.hashicorp.com/vagrant/1.7.4/vagrant_1.7.4.dmg)
+  * You have [Vagrant 1.8.6](https://releases.hashicorp.com/vagrant/1.8.6/vagrant_1.8.6.dmg)
```
boffbowsh: Would be better to say "the latest version of Vagrant" I think and link to the main download page.

tijmenb:
Does this mean 1.8.6 works with vagrant-dns now?

boffbowsh:
I didn't know it wasn't working. It's been fine for me.

development-vm/README.md
Davidslv: What if the next latest version doesn't work for some reason?

boffbowsh: We'll worry about that if and when it happens I think

tijmenb: The reason it was "pinned" was that vagrant-dns didn't work (https://github.com/alphagov/govuk-puppet/commit/3ac94574a4c1fe2fd9535d6c70fd963ab235227b). If that's solved then recommending the latest release is indeed the way to go.

boffbowsh: I said link to the main download page, then people can download the latest.

Davidslv:
@tijmenb that might have been related with this: https://github.com/alphagov/govuk-puppet/tree/master/development-vm#errors-with-vagrant-dns-having-updated-vagrant

@boffbowsh I think I've addressed your comments

Thanks both!

PR #4969, opened 2016-10-19
Don't call people for a machine being down

rjw1:
Since we mostly have clusters of machines a machine being down shouldn't be enough to call someone out of hours.
The effects of machines being down should be what calls people.

### DIFF
--- a/modules/icinga/files/etc/icinga/conf.d/generic-host_nagios2.cfg
+++ b/modules/icinga/files/etc/icinga/conf.d/generic-host_nagios2.cfg
@@ -14,7 +14,7 @@ define host {
   notification_interval           0
   notification_period             24x7
   notification_options            d,r
-  contact_groups                  urgent-priority
+  contact_groups                  high-priority
   register                        0       ; DONT REGISTER THIS DEFINITION - ITS NOT A REAL HOST, JUST A TEMPLATE!
 }
 


deanwilson:
I think we'll need to add some extra work to ensure things like the DB masters still alert on host failure.

rjw1:
So a db master being down will cause other things to alert. hopefully. If nothing else we should test that somewhere :)

surminus:
I love this concept, but a part of me says we should still be told about snowflakes going offline. In most instances this will be the quickest way to find and resolve the issue.

deanwilson:
Add a per host hiera override for the snowflakes and enable it for everything else?

rjw1:
@surminus icinga would still show them offline so people triaging problems should notice it. Just they wouldn't be woken up for it. 

@deanwilson indeed. I did think about that but couldnt be bothered. the snowflakes could have it in their host config. 

I might get to this when im bored at devopsdays :)

samjsharpe:
I heartily approve of you doing this. 

It is the way it is because back in the distant past there were many more snowflakes (sorry) - the intention was always that alerts should be from things that users will notice.

If either a machine being down will not have a serious effect on the users, or a machine being down will cause a higher-order effect which causes an alert, then machines themselves should not scream as they die, nor should you mourn their passing.

surminus:
Rebased.

rjw1:
Closing while me and @dazahern redo this to allow overrides.

PR #4973, opened 2016-10-20
Move assets s3 sync

surminus:
The S3 attachment sync causes the machine to run extremely low on memory, which *may* have caused problems with the asset-master to connect to it. The asset-master virus scanning process copies each file it scans to each slave, and if it's unable to connect it will eventually time out, but it will also cause files to backup in processing if there are a lot of them.

This makes it less likely that we will see an impact of this during the daytime when there are more files being uploaded for processing.

### DIFF
--- a/modules/govuk/manifests/node/s_asset_base.pp
+++ b/modules/govuk/manifests/node/s_asset_base.pp
@@ -193,11 +193,6 @@ class govuk::node::s_asset_base (
         mode    => '0755',
       }
 
-      # FIXME: remove after this has been deployed
-      file { '/etc/cron.daily/push_attachments_to_s3':
-        ensure => absent,
-      }
-
       cron { 'push_attachments_to_s3':
         command => '/usr/bin/setlock -n /var/run/virus_scan/push-attachments.lock /usr/local/bin/push_attachments_to_s3.sh /mnt/uploads',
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/node/s_asset_base.pp
```diff
@@ -193,15 +193,10 @@
...
-      # FIXME: remove after this has been deployed
-      file { '/etc/cron.daily/push_attachments_to_s3':
-        ensure => absent,
-      }
-
       cron { 'push_attachments_to_s3':
         command => '/usr/bin/setlock -n /var/run/virus_scan/push-attachments.lock /usr/local/bin/push_attachments_to_s3.sh /mnt/uploads',
         user    => 'assets',
-        hour    => 6,
+        hour    => 21,
```
andrewgarner: Is the rational for this that it's less likely to wake someone on call than if it ran at say 0100?

surminus: There is a full sync from the asset master that happens in the early hours of the morning and I didn't want it to clash with that. The less other maintenance tasks it clashes with the better I reckon. I don't think it'll be very unlikely to ever ring us for oncall, at least, I'd be very upset if it did :)

PR #4975, opened 2016-10-20
Mysql xtrabackup restore

djsd123:
See Trello: https://trello.com/c/AsknJrSt

What 
Initially, to address comments made in Trello card.  Running the script in Integration produced different errors not relating to previous comments made in Trello.
Breakages in script:
- Streamed data from `gof3r` to `xbcrypt` failed due to wrong data chunk size.  reference: https://github.com/mdkent/percona-xtrabackup/blob/master/src/xbstream_read.c
- `xbcrypt` is incapable of handling multiple chunks of streamed data

Refactoring to make use of newly added features to `govuk_puppet`.

How
Streamed data from `gof3r` must first go through `xbstream -x` when piping to `xbcrypt`.  This is due to `xbcrypt` not handling multiple chunks of data that well.  
We changed the process to decrypt the data after it had been saved to disk. We then have `xbcrypt` iterate over each file.  
We now change  directory to perform operations. 

We modified the temporary directory structure for clarity.

Added flag `--parallel` to `innobackupex` to improve the performance of the process.

Govuk Puppet:
- Remove conditional that checks for the existence of a lock.  The `govuk_puppet` script now allows for multiple locks to exist.
- Rewrite (`govuk_puppet`) disable command to be more explicit.

### DIFF
--- a/modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
+++ b/modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
@@ -31,11 +31,19 @@ else
     rm -rf $dir/*
   done
 
-  # Pull the latest backups from S3
-  envdir <%= @env_sync_envdir %> gof3r get --endpoint s3-<%= @aws_region -%>.amazonaws.com -b <%= @s3_bucket_name %> -k latest/base.xbcrypt | xbcrypt --encrypt-algo=AES256 --decrypt --encrypt-key="<%= @encryption_key %>" | xbstream -x -C $BASE_DIR
+  # Pull, then decrypt the latest backups from S3
+  cd $BASE_DIR && envdir <%= @env_sync_envdir %> gof3r get --endpoint s3-<%= @aws_region -%>.amazonaws.com -b <%= @s3_bucket_name %> -k latest/base.xbcrypt | xbstream -x -C $BASE_DIR
 
-  envdir <%= @env_sync_envdir %> gof3r get --endpoint s3-<%= @aws_region -%>.amazonaws.com -b <%= @s3_bucket_name %> -k latest/incremental.xbcrypt | xbcrypt --encrypt-algo=AES256 --decrypt --encrypt-key="<%= @encryption_key %>" | xbstream -x -C $INCR_DIR
+  for i in `find . -iname "*\.xbcrypt"`; do
+    xbcrypt --encrypt-algo=AES256 --decrypt --encrypt-key="<%= @encryption_key %>" < $i > $(dirname $i)/$(basename $i .xbcrypt) && rm $i
+  done
 
+  cd $INCR_DIR && envdir <%= @env_sync_envdir %> gof3r get --endpoint s3-<%= @aws_region -%>.amazonaws.com -b <%= @s3_bucket_name %> -k latest/incremental.xbcrypt | xbstream -x -C $INCR_DIR
+
+  for i in `find . -iname "*\.xbcrypt"`; do
+    xbcrypt --encrypt-algo=AES256 --decrypt --encrypt-key="<%= @encryption_key %>" < $i > $(dirname $i)/$(basename $i .xbcrypt) && rm $i
+  done
+  
   # Decompress the backups
   for dir in "$BASE_DIR" "$INCR_DIR"; do
     innobackupex --decompress $dir
--- a/modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
+++ b/modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
@@ -5,9 +5,10 @@ exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
 
 set -e
 
-DATE=$(date +%s)
-BASE_DIR=/tmp/mysql/basedir/$DATE/
-INCR_DIR=/tmp/mysql/incremental/$DATE/
+DATE=$(date +%F)
+CURRENT_RESTORE=/tmp/mysql/$DATE
+BASE_DIR=$CURRENT_RESTORE/basedir
+INCR_DIR=$CURRENT_RESTORE/incremental
 
 echo "Starting MySQL environment sync"
 
@@ -21,8 +22,13 @@ else
   govuk_puppet --disable "MySQL environment sync (by $0)"
 
   # Create directories to restore into
-  for dir in "$BASE_DIR" "$INCR_DIR"; do
+  if [[ ! -d "CURRENT_RESTORE" && ! -d "$BASE_DIR" && ! -d "$INCR_DIR" ]]; then
+    for dir in "$BASE_DIR" "$INCR_DIR"; do
     mkdir -p $dir
+    done
+  else
+    for dir in "$BASE_DIR" "$INCR_DIR"; do
+    rm -rf $dir/*
   done
 
   # Pull the latest backups from S3
--- a/modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
+++ b/modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
@@ -5,7 +5,7 @@ exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
 
 set -e
 
-DATE=$(date +%F)
+DATE=$(/bin/date +%F)
 CURRENT_RESTORE=/tmp/mysql/$DATE
 BASE_DIR=$CURRENT_RESTORE/basedir
 INCR_DIR=$CURRENT_RESTORE/incremental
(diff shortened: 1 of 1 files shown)


modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
```diff
@@ -18,29 +19,42 @@ if test -f /var/lib/puppet/state/agent_disabled.lock; then
   exit 1
 else
 
-  govuk_puppet --disable "MySQL environment sync (by $0)"
+  /usr/local/bin/govuk_puppet --disable "MySQL environment sync (by $0)"
```
surminus: Oooh in theory this could now use @deanwilson's new method to apply locks. I think the method would be something like `govuk_puppet -r 'MySQL environment sync' -u '$0' --disable`, but double check the script.

surminus:
Good troubleshooting! I appreciate this isn't exactly the easiest restore script in the world. It all looks like it makes sense and I'm not really surprised by the xbcrypt thing. 

The one thing that I am interested in though is the `xtrabackup` traversing issue. Full paths are provided throughout, so what errors did you see regarding that? 

Otherwise great work getting this working! Did the different versions of `xtrabackup` end up making no difference?

djsd123:
Hey @surminus 
Yep, sorry.  The traversing issue was due to xbcrypt not working.  `Innobackupex` evals the type (not name) of the files it needs and produces a rather confusing error message when the file is not what it expects.  Modified commit message.

Unfortunately, the version disparity was not relevant to the errors.  Worth addressing though!

modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
djsd123: Cool, I'll get that in and test.

djsd123: Done

modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
```diff
@@ -5,42 +5,49 @@ exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
...
+  /usr/local/bin/govuk_puppet -n mysql_restore -r "MySQL environment sync (by $0)" --disable
 
   # Create directories to restore into
-  for dir in "$BASE_DIR" "$INCR_DIR"; do
-    mkdir -p $dir
+  if [[ ! -d "$CURRENT_RESTORE" && ! -d "$BASE_DIR" && ! -d "$INCR_DIR" ]]; then
+    for dir in "$BASE_DIR" "$INCR_DIR"; do
+      /bin/mkdir $dir
+    done
+  else
```
djsd123: Reverted. Without the loop, the code looks less linear.  I added a comment for clarity.  Hope it reads well?

` /bin/rm -rf {$BASE_DIR/*,$INC_DIR/*}`

alexmuller:
I think this would be easier to review if the commit messages followed [the styleguide](https://github.com/alphagov/styleguides/blob/master/git.md#commit-messages) to explain what was changing and why. For example for the first commit:

```
Change restore directories for MySQL Xtrabackup

https://trello.com/c/AsknJrSt

Tools from `percona-xtrabackup` toolset are bad at traversing the file system
and fail to find binary logs necessary for operations.
```

alexmuller:
For the changes I can see here I'd probably structure the commits a bit like this:

```
Use full paths for running commands

[Not sure about this, was it not working before?]
```

```
Change restore directories for MySQL Xtrabackup

Tools from `percona-xtrabackup` toolset are bad at traversing
the file system and fail to find binary logs necessary for operations.
```

```
Use parallel flag for innobackupex

Decompressing can take a long time when working on a large dataset.
```

```
Refactor MySQL Xtrabackup to make use of newly added features to `govuk_puppet`

Remove conditional that checks for the existence of a lock.
The `govuk_puppet` script now allows for multiple locks to exist.
```

```
Call xbcrypt multiple times in MySQL Xtrabackup

`xbcrypt` is incapable of handling multiple chunks of streamed data.

Streamed data from `gof3r` must first go through `xbstream -x` when piping to `xbcrypt`.  This is due to ` xbcrypt ` not handling multiple chunks of data that well.

We changed the process to decrypt the data after it had been saved to disk. We then have `xbcrypt` iterate over each file.
```

I'm also a bit confused by the change to the `mkdir -p`. Is there a reason to remove the `mkdir -p $dir` and put it inside a conditional that checks whether the directory exists? It seems like that change is basically a no-op. Would the same result come from running this?

```
for dir in "$BASE_DIR" "$INCR_DIR"; do
  rm -rf $dir/*
  mkdir -p $dir
done
```

djsd123:
Hey @alexmuller,
I'm confused by your confusion?? :-)
What you're doing is removing contents of whatever `$dir` is and then running a `mkdir -p` on the directory structure that presumably already exists??

Sorry if it's not clear. 
- I'm checking whether or not the directory structure exists?  
- If any of the directories are missing? We create it with the loop.
- Otherwise, we empty them.

surminus:
Alternatively couldn't we just do `mkdir -p $dir || rm -rf $dir/*`? 

Try to create the directory, and if it already exists, remove the contents?

djsd123:
I see your point but the exit status of `mkdir -p $dir` would be 0.

surminus:
Oh yeah, my bad.

surminus:
Ah sorry, just catching up. @alexmuller was suggesting a change that @deanwilson originally suggested and then was reverted for readability. I don't think it's worth blocking this change on a pretty trivial function.

The only other thing I think would be good to echo @alexmuller's comments on style would be to correctly title the commits because otherwise finding a specific commit would be more difficult (currently there are 4 commits called "MySQL Xtrabackup Restore").

modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
```diff
@@ -21,8 +22,13 @@ else
...
   # Create directories to restore into
-  for dir in "$BASE_DIR" "$INCR_DIR"; do
-    mkdir -p $dir
+  if [[ ! -d "CURRENT_RESTORE" && ! -d "$BASE_DIR" && ! -d "$INCR_DIR" ]]; then
+    for dir in "$BASE_DIR" "$INCR_DIR"; do
+    mkdir $dir
+    done
+  else
+    for dir in "$BASE_DIR" "$INCR_DIR"; do
+    rm -rf $dir/*
```
alexmuller: We think it would be more clear to replace all of the loops and conditionals above with just these 2 commands. One to delete the directories no matter whether they exist or not and one to create them empty.

```bash
rm -rf $BASE_DIR $INCR_DIR
mkdir -p $BASE_DIR $INCR_DIR
```

Ana pointed out that if `$BASE_DIR` or `$INCR_DIR` in set to an empty string this change would remove the root directory.

alexmuller:
The 3rd commit (5e185000e20333bdae77eda030366f4108996127) has a lot of changes to paths which I think makes it harder to see the `--parallel` change - I'm also not sure if those path changes are required or just nice to have.

modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
```diff
@@ -52,27 +60,26 @@ else
...
+  /bin/chown -R mysql:mysql /var/lib/mysql
 
   # Start the service
-  service mysql start
+  /usr/sbin/service mysql start
 
   # Clean up
   echo "Cleaning up"
-  rm -rf /var/lib/$TEMPDIR $BASE_DIR $INCR_DIR
+  /bin/rm -rf /var/lib/$TEMPDIR $BASE_DIR $INCR_DIR
```
deanwilson: You should set the full path in the first place `$TEMPDIR` is set.

modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
```diff
@@ -52,27 +60,26 @@ else
   # the directory is a mount point so we need to copy
   echo "Backing up current dir"
   TEMPDIR=mysql-$(date '+%s')
```
deanwilson: You should rename this to something other than `$TEMPDIR` as this is often already a value in the shell.

modules/govuk_mysql/templates/usr/local/bin/xtrabackup_s3_restore.erb
```diff
@@ -5,42 +5,50 @@ exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
...
+  # Create directories to restore into OR Purge {$BASE_DIR,$INCR_DIR} if they already exist
+  if [[ ! -d "$CURRENT_RESTORE" && ! -d "$BASE_DIR" && ! -d "$INCR_DIR" ]]; then
+    for dir in "$BASE_DIR" "$INCR_DIR"; do
+      /bin/mkdir $dir
+    done
+  else
+    for dir in "$BASE_DIR" "$INCR_DIR"; do
+      /bin/rm -rf $dir/*
+    done
+  fi
```
deanwilson: Change to 
```
rm -rf "$BASE_DIR" "$INCR_DIR"
mkdir -p "$BASE_DIR" "$INCR_DIR"
```

PR #4978, opened 2016-10-21
Adding Aaron so I can see Jenkins things

Shotclog:


boffbowsh:
This will also give you a UNIX user, but you haven't granted yourself access to SSH. This is ok but probably not what you wanted?I think the entry in the jenkins section is sufficient.

Shotclog:
Yes, I think I only want to see build details. I'll remove the second part.

PR #4980, opened 2016-10-21
govuk_puppet Fix

djsd123:
What
Whenever a lock is created the `govuk_puppet` script would exit with the status code `1` when bash option `-e` is passed. It's not an issue when invoking the script interactively. However, invoking the script from another script would cause the latter to exit immediately, preventing the script from going any further.

How
Call to `exit` once the lock has been created.


surminus:
also inb4 people complain about "GovUK" rather than "GOVUK" 😉

PR #4983, opened 2016-10-24
Reduce the nodes and shards used in the cluster

deanwilson:
They are numbers but the class uses strings to represent them so
I'm doing the same here.

We are reducing the size of the ES cluster, this ensures a machine holds all its own data.


alexmuller:
I think this is right:

> The discovery.zen.minimum_master_nodes allows to control the minimum number of master eligible nodes a node should "see" in order to operate within the cluster.

>  index.number_of_replicas: The number of replicas each shard has. 

- https://www.elastic.co/guide/en/elasticsearch/reference/0.90/modules-discovery-zen.html
- https://www.elastic.co/guide/en/elasticsearch/reference/0.90/indices-update-settings.html

PR #4985, opened 2016-10-24
Update puppet aptly module

surminus:
We had an issue with the version of the puppet-aptly module that caused GPG key errors when bringing up an apt machine. We need to upgrade to at least 0.6.0, which has a dependency of the puppetlabs-apt module of >= 2.0.0. This also has an updated dependency of puppetlabs-stdlib.

In version 2.0.0 of puppetlabs-apt, the unattended_upgrades class was broken out into it's own module, so this installs that as well, and updates the references in our code to use this as a standalone module.


deanwilson:
This looks really good, one small item that's related to this - `stdlib` will start raising deprecation warnings on `validate_` and `is_` functions in 4.13, which this PR moves us to. Can we pin to 4.12 instead please?

surminus:
Thanks @deanwilson! I've updated to pin to 4.12.0 for puppetlabs/stdlib. We can test later on for >= 4.13.0

PR #4989, opened 2016-10-26
[WIP] Measure unprocessed whitehall uploads

MatMoore:
Measure the number of unprocessed whitehall uploads and alert if it gets too high.

We looked into also tracking the number of infected/cleared files, but these subdirectories contained a lot of files and nested subdirectories and the collectd plugin seems to hang while recursing through them. Because of this, we decided to start with just the unprocessed files.

I think the icinga alert still needs testing on integration, and the thresholds need to be set based on what's normal.

with @davidbasalla and @andrewgarner

### DIFF
--- a/modules/collectd/spec/defines/collectd__plugin__file_count_spec.rb
+++ b/modules/collectd/spec/defines/collectd__plugin__file_count_spec.rb
@@ -9,9 +9,9 @@ describe "collectd::plugin::file_count", :type => :define do
 
   it {
     is_expected.to contain_collectd__plugin("file count: #{title}").with_content(<<-EOS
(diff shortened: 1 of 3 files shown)


MatMoore:
This needs more work - tried testing it on integration and got some puppet errors - or some errors reporting puppet errors:
```
search-2.api.integration.publishing.service.gov.uk;puppet last run errors;UNKNOWN;HARD;1;UNKNOWN: Could not parse puppet summary file
```

surminus:
We have something which alerts us regarding slow virus scanning which was added [here](https://github.com/alphagov/govuk-puppet/pull/5547), so I'm going to close this PR.

PR #4991, opened 2016-10-26
PostgreSQL Backup Logs

djsd123:
If the script must remove it's logs. It would be handy to keep them around for a while for debugging purposes.

Remove autopostgresql backup logs that are older than 2 weeks

### DIFF
--- a/modules/govuk_postgresql/templates/usr/local/bin/autopostgresqlbackup.erb
+++ b/modules/govuk_postgresql/templates/usr/local/bin/autopostgresqlbackup.erb
@@ -637,9 +637,8 @@ if [ -s "$LOGERR" ]
     STATUS=0
 fi
 
-# Clean up Logfile
-rm -f "$LOGFILE"
-rm -f "$LOGERR"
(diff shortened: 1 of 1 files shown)


modules/govuk_postgresql/templates/usr/local/bin/autopostgresqlbackup.erb
```diff
@@ -637,9 +637,8 @@ if [ -s "$LOGERR" ]
     STATUS=0
 fi
 
-# Clean up Logfile
-rm -f "$LOGFILE"
-rm -f "$LOGERR"
+# Clean up Logfiles older than two weeks
+/usr/bin/find /var/lib/autopostgresqlbackup -iname *.log -mtime +14 -type f -delete
```
surminus: If we're setting the $BACKDIR variable earlier in the file I think it'd better to set it here because if we change location of backups then they won't get removed due to the hardcoded location.

deanwilson:
Merge away

PR #4992, opened 2016-10-26
Remove access for Tom {Russell, Ridd, White}

kevindew:
They've all left


deanwilson:
Have they left or is this a 2ndline shift change?

PR #4994, opened 2016-10-26
Change monitoring from BST to UTC

alexmuller:


alexmuller:
Let's do that in 6 months time when we have to look at this code again ;)

PR #5003, opened 2016-10-27
[DO NOT MERGE] Remove deploy_govuk_content_schemas

cbaines:
govuk-content-schemas is now deployed through the Deploy App job in
Jenkins.

[DO NOT MERGE] as I think this should wait for the deployment changes for the govuk-content-schemas to be tested in production.


cbaines:
Deployment for govuk-content-schemas via the Deploy App Jenkins job happened successfully today, so this old deployment path can be removed.

PR #5007, opened 2016-10-28
Remove deleted Elasticsearch nodes from cluster

alexmuller:
This is used when monitoring the cluster and needs to reflect the machines that exist.


alexmuller:
We missed this in #4983.

PR #5008, opened 2016-10-28
Goodbye Jack Scotti

kevindew:
👋😢

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -68,7 +68,6 @@ govuk_jenkins::config::admins:
   - henryturner
   - ikennaokpala
   - isabelllong
-  - jackscotti
   - jamesrobinson
   - jennyduckett
   - kanemorgan
@@ -284,7 +283,6 @@ users::usernames:
   - henryturner
   - ikennaokpala
   - isabelllong
-  - jackscotti
   - jamesrobinson
   - jennyduckett
   - kanemorgan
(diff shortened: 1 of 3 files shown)


djsd123:
I thought the correct way to remove a user was to `ensure => absent`.

cbaines:
> I thought the correct way to remove a user was to ensure => absent.

Hmm, I can see how that might be necessary.

In this case as its just integration, I'd expect the effect to be the same, but maybe its worth for consistency always removing with `ensure => absent`?

kevindew:
@djsd123 sure is there an example of that?

cbaines:
> @djsd123 sure is there an example of that?

This is the most recent change: 7889704d10ad4bd25b425840159e4c6227ca5cf5

kevindew:
@cbaines Thanks, I'll get this updated.

kevindew:
Ok - I've updated this now 👍

cbaines:
I'm not sure, but I would guess that you need to keep `jackscotti` in hieradata for the `absent` thing to have an effect?

djsd123:
Sweet. I think the way you did first partially removes a user and leaves behind ssh config/stuff.  No harm......
FYI; declaring the resource name is all you need but makes no difference dude.  merge away 👍 
```
govuk_user { 'jackscotti':
  ensure   => absent,
}
```

djsd123:
Hi @kevindew,

@cbaines is correct.  To purge a user, you'll need to keep their name in hiera.  I'd just do it the way you were going to in the first place. Everything is unjustly coupled in this repo. Sorry to stick my oar in.

kevindew:
@djsd123 No problem - Does seem better to remove the files but a tad painful. I've reset this to how it was now.

PR #5012, opened 2016-10-31
Log a publishing api related header from NGinx

cbaines:
This HTTP header is useful for request tracing as it allows identifying
what requests have been caused by dependency resolution, and the value
is the content id of the item on which dependency resolution took place.


djsd123:
How does this value `"$http_govuk_publishing_api_dependency_resolution_source_content_id"` get set?

If it's empty what happens?  Will it be an empty string/header?

cbaines:
> How does this value "$http_govuk_publishing_api_dependency_resolution_source_content_id" get set?

I believe this is a way to access arbitrary headers: https://nginx.org/en/docs/http/ngx_http_core_module.html#var_http_

> If it's empty what happens? Will it be an empty string/header?

No idea, I'm hoping nothing bad :) I'm assuming it will be fine as some of the other logged values could have the same behaviour.

PR #5017, opened 2016-11-01
ENV vars for Content API

kevindew:
Related to https://github.gds/gds/deployment/pull/1134


modules/govuk/manifests/apps/contentapi.pp
```diff
@@ -55,12 +70,24 @@
   }
 
   govuk::app::envvar {
+    app => 'contentapi',
```
deanwilson: This resource needs to have a `varname` and a `value` set.

PR #5023, opened 2016-11-01
A simple theme for Jenkins

kevindew:
Aim of this work is to make it easier to differentiate deploy environments.

Brings the Jenkins header more in line with that of sign-on and other apps.

## Before:

screen shot 2016-11-01 at 17 40 17

## After:

screen shot 2016-11-01 at 17 35 28
screen shot 2016-11-01 at 17 36 23
screen shot 2016-11-01 at 17 38 09


issyl0:
😮

hieradata/production.yaml
```diff
@@ -66,6 +66,9 @@ govuk_bouncer::gor::ip_address: '195.225.216.149'
 govuk_jenkins::config::banner_colour_background: '#ff3300'
 govuk_jenkins::config::banner_colour_text: 'white'
 govuk_jenkins::config::banner_string: 'Carrenza PRODUCTION'
+govuk_jenkins::config::theme_colour: '#ff3300'
```
tijmenb: We could use `#df3034` like the publishing apps here:

https://github.com/alphagov/govuk_admin_template/blob/master/app/assets/stylesheets/govuk_admin_template/_theme.scss#L46-L48

Davidslv:
👍

PR #5026, opened 2016-11-02
Set up pact_broker in CI

deanwilson:


djsd123:
Something like this for your TODO

```
nginx::config::vhost::proxy { 'pact-broker':
  ensure                  => present,
  to                          => [$jenkins_agents],
  ssl_only                => true,
  extra_app_config => ['Host $host',
    'X-Real-IP $remote_addr',
    'X-Forwarded-For $proxy_add_x_forwarded_for',
    'Proxy ""',
    'X-Forwarded-Proto $scheme'],
}
```

surminus:
I think this is ready, perhaps a final check over from someone else? 

The `pact_broker` class will also need including. Not having this PR merged is conflicting with other work so I've removed it, so we may have to add it wherever depending on the order of merge.

hieradata/class/ci_agent.yaml
```diff
@@ -6,3 +6,10 @@ mongodb::server::replicaset_members:
 
 govuk_ci::agent::postgresql::mapit_role_password: 'mapit'
 postgresql::globals::version: '9.3'
+
+pact_broker::hosts:
+  - 'ci-agent-1.ci'
```
deanwilson: This should be deployed on the master.

djsd123:
Include in `master.pp`

deanwilson:
Please rebase

surminus:
I've rebased this. Should it run on _all_ agents, or just one? /cc @deanwilson @djsd123

deanwilson:
We're going to redo this as a complete application rather than one embedded in the puppet code (decided in the standup on 2016/12/13) so we're closing this direct port.

PR #5031, opened 2016-11-02
Tell dev vm publishing api where content schemas are

h-lame:
More notes in the commit, and we also fix Gemfile.lock.

### DIFF
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -142,6 +142,7 @@ govuk::apps::publishing_api::draft_content_store: 'http://draft-content-store.de
 govuk::apps::publishing_api::enable_procfile_worker: false
 govuk::apps::publishing_api::rabbitmq::configure_test_exchange: true
 govuk::apps::publishing_api::suppress_draft_store_502_error: '1'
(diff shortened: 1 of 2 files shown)


hieradata/development.yaml
```diff
@@ -142,6 +142,7 @@ govuk::apps::publishing_api::draft_content_store: 'http://draft-content-store.de
 govuk::apps::publishing_api::enable_procfile_worker: false
 govuk::apps::publishing_api::rabbitmq::configure_test_exchange: true
 govuk::apps::publishing_api::suppress_draft_store_502_error: '1'
+govuk::apps::publishing_api::govuk_content_schemas_path: '/var/govuk/govuk-content-schemas'
```
kevindew: Should that be `/var/govuk/govuk-content-schemas/dist` as per "Ensure GOVUK_CONTENT_SCHEMAS_PATH env variable is set and points to the dist directory of govuk-content-schemas"  or is our error wrong?

dougdroper: No, that error message is wrong, it now uses the GovukSchemas gem, which automatically looks in the dist folder

kevindew: Cool cool, don't mind me then 😃

PR #5032, opened 2016-11-02
Run relevant smokey tests in integration

tijmenb:
Smokey was configured long ago to exclude tests tagged with `@notintegration` when you run `rake test:integration`.

But currently we still run `rake test:production` on integration CI, which has a bunch of test failures
(https://deploy.integration.publishing.service.gov.uk/job/Smokey/11051).

This commit makes sure we actually run the subset of tests on integration. This will get us closer to getting smokey green there, which would be useful for development.

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -134,6 +134,8 @@ govuk_jenkins::job_builder::environment: 'integration'
 
 govuk_jenkins::job::deploy_puppet::commitish: 'integration'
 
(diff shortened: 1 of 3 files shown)


deanwilson:
Should there be a change to smokey to support `MYTASK` or does it already handle that env var?

tijmenb:
Yup, I should've mentioned that Smokey already supports `MYTASK` since 2012 https://github.com/alphagov/smokey/commit/bea42ad6ba816120824ad9f4b1f6647953a273be  :)

PR #5041, opened 2016-11-03
Set up MySQL  in CI

djsd123:
All CI agents should run MySQL server with identical configuration.

### DIFF
--- a/modules/govuk_ci/manifests/agent/mysql.pp
+++ b/modules/govuk_ci/manifests/agent/mysql.pp
@@ -1,0 +1,99 @@
+# == Class: govuk_ci::agent::mysql
+#
+# Installs and configures mysql-server
+#
+class govuk_ci::agent::mysql {
+  contain ::govuk_mysql
+
+  mysql::db {
+    'collections_publisher_test':
+      user     => 'collections_pub',
+      password => 'collections_publisher',
+      require  => Class['::mysql::server'];
+
+    'contacts_test':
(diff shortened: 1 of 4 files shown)


modules/govuk_mysql/manifests/server.pp
```diff
@@ -3,6 +3,13 @@
 # This is an abstraction for puppetlabs/mysql to enforce a set of sensible
 # defaults and setup associated monitoring.
 #
+# === Parameters
+#
+# [*innodb_flush_log_at_trx_commit*]
+#   Controls if the filesystem flushes after every transactiopn commit.
```
surminus: Typo in "transactiopn"

modules/govuk_ci/manifests/agent/mysql.pp
```diff
@@ -0,0 +1,100 @@
+# == Class: govuk_ci::agent::mysql
+#
+# Installs and configures mysql-server
+#
+class govuk_ci::agent::mysql {
+  contain ::govuk_mysql
```
surminus: This is a bare class. I think it should be `::govuk_mysql::server` instead. Then we wouldn't also need to include `::mysql::server`, because that's included in the aforementioned class.

surminus:
It feels like 1f4d830a79902a01ca5912086d730b69bc68c455 has accidentally snuck in here.

surminus:
I've rebased this removing the commit.

PR #5047, opened 2016-11-04
Upgrade to concat 2.0

deanwilson:
The new version is implemented as a puppet type so it's a lot simplier,
faster and easier to test. This version also supports Ubuntu 16 and should run cleanly
under puppet 4.0 tests


deanwilson:
This needs to be rebased and retested after https://github.com/alphagov/govuk-puppet/pull/5247 is merged.

PR #5049, opened 2016-11-04
Allow `draft-whitehall-frontend.$app_domain` to serve whitehall assets

boffbowsh:
In production, the GOV.UK router expects anything under /government to
be served by `whitehall-frontend` unless a more specific route exists.
This includes assets such as uploaded PDFs. This is fine in the live
stack for immigrated content as the Whitehall Rails app will serve the
asset through nginx once it’s performed its own authorisation checks.

For migrated content such as detailed guides, content is previewed
using the draft stack. In this case, the router sends all requests for
/government to `draft-whitehall-frontend`. Before this change, this
would be an invalid vhost in production. This change will create the
vhost and simply serve the files from disk. We make a whitehall-backend
machine serve it as it has access to the NFS mount already. This is a
temporary solution until Whitehall’s asset management system is rebuilt
or ported to use asset-manager.

We don’t need to provide any additional authorisation checking for
these requests as `draft-origin` (the user’s only path to this
endpoint) is protected by `authenticating-proxy`.

/cc @alexmuller 

Todo: 
- [x] Test in integration before merging

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -833,6 +833,7 @@ hosts::production::backend::app_hostnames:
   - 'contacts-admin'
   - 'contentapi'
   - 'content-tagger'
+  - 'draft-whitehall-frontend'
   - 'email-alert-api'
   - 'errbit'
   - 'event-store'
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -243,6 +243,7 @@ hosts::development::apps:
   - 'content-tagger'
   - 'designprinciples'
   - 'draft-content-store'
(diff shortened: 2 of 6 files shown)


alexmuller:
If we're setting `ssl_only` to be true can we remove that param and only add it in if we need it in future? Unfortunately we'd still need the `enable_ssl` variable if you want this to work in development properly. But removing the `ssl_only` param would reduce the complexity in the template a bit.

boffbowsh:
Good point. Updated.

boffbowsh:
@alphagov/team-gov-uk-infrastructure-admins this is ready for review again

PR #5052, opened 2016-11-07
Creating master class for CI

djsd123:
I ended up running away with this PR a little bit. I need to stop doing that.

It refactors the govuk_jenkins class by breaking out user and package management. I began to do this to allow us to specify a different version of Jenkins to install. Installing the later version of Jenkins requires the later Puppet module [which is in a PR here](https://github.com/alphagov/govuk-puppet/pull/5026). The master also requires pact_broker, which is also dependent on that PR.

This also adds a credentials class which adds credentials to allow the CI master to talk to some third party stuff. I have included examples of secrets in `vagrant_credentials.yaml` file.

I've also added the latest Jenkins release in our apt repo.

Note: this was @surminus that wrote the above 😄

### DIFF
--- a/modules/govuk_jenkins/manifests/init.pp
+++ b/modules/govuk_jenkins/manifests/init.pp
@@ -5,11 +5,11 @@
 #
 # === Parameters:
 #
-# [*apt_mirror_hostname*]
-#   Hostname to use for the APT mirror.
-#
 # [*github_enterprise_cert*]
 #   PEM certificate for GitHub Enterprise.
+#
+# [*github_enterprise_hostname*]
+#   The hostname of Github Enterprise
 #
 # [*config*]
 #   A hash of Jenkins config options to set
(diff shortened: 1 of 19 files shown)


surminus:
I've removed the stuff related to "pact_broker". We can do this separately.

modules/govuk_jenkins/manifests/init.pp
```diff
@@ -6,51 +6,66 @@
...
-    require => [
-                  Class['govuk_java::openjdk7::jdk'],
-                  Class['govuk_java::openjdk7::jre'],
-                ],
-    notify  => Class['jenkins::service'],
+  class { 'govuk_jenkins::package':
+    apt_mirror_hostname => $apt_mirror_hostname,
+    config              => $config,
+    plugins             => $plugins,
+    version             => $version,
```
alexmuller: Instead of settings this on `govuk_jenkins` and passing it through to `govuk_jenkins::package` we can set this in hiera as `govuk_jenkins::package::version` and avoid having the default in 2 places.

modules/govuk_jenkins/manifests/init.pp
```diff
@@ -6,51 +6,66 @@
...
+  include ::govuk_jenkins::config
+  include ::govuk_jenkins::job_builder
 
-  user { 'jenkins':
-    ensure     => present,
-    home       => $jenkins_home,
-    managehome => true,
-    shell      => '/bin/bash',
+  class { 'govuk_jenkins::user':
+    home_directory => $jenkins_home,
```
alexmuller: Same comment with having it possible to set params in multiple places.

modules/govuk_jenkins/manifests/user.pp
```diff
@@ -0,0 +1,41 @@
...
+#
+# === Parameters:
+#
+# [*home_directory*]
+#   The home directory of the user
+#
+# [*username*]
+#   The username you wish to use
+#
+class govuk_jenkins::user (
```
djsd123: The govuk_jenkins base class already manages this user.

surminus:
After comments I've tidied this up a bit to make further use of hieradata and slightly less repetition in code. I've left in calling the following parameters:
`public_key`
`private_key`
`config`
`plugins`

My reasoning is that it would make it cleaner to be able to just call the `govuk_jenkins` class from elsewhere and include these parameters if you wanted to differentiate different config or plugins. 

However I've torn that it might just be better in hieradata, so happy to change it to that. 

As a further note I've tested bringing this up on a Vagrant VM and it successfully installs and configures Jenkins 2, although we still need to look into auth.

alexmuller:
There are a couple of failing specs:

```
rspec ./modules/govuk_jenkins/spec/classes/jenkins_spec.rb:16 # govuk_jenkins should contain Class[govuk_jenkins::user]
rspec ./modules/govuk_jenkins/spec/classes/jenkins_spec.rb:15 # govuk_jenkins should contain Class[govuk_jenkins::config]
```

surminus:
Fixed specs, had to add something in `spec/fixtures/hieradata/common.yaml` (and also extended another test and removed some unneeded code)

modules/govuk_jenkins/manifests/user.pp
surminus: This change specifically moves the management of this user from the base class to it's own class, and then includes it.

surminus:
This PR is too large so I'm breaking it down into small PRs.

PR #5057, opened 2016-11-10
Jenkins agent firewall rules

djsd123:
Amending firewall to allow network access for specified services.

### DIFF
--- a/modules/govuk_ci/manifests/agent.pp
+++ b/modules/govuk_ci/manifests/agent.pp
@@ -11,4 +11,10 @@ class govuk_ci::agent {
   include ::govuk_ci::agent::postgresql
   include ::govuk_ci::agent::mysql
 
+  ufw::allow { 'allow-jenkins-slave-swarm-to-listen-on-ephemeral-ports':
+    port  => '32768:65535',
+    proto => 'udp',
+    ip    => 'any',
+  }
+
 }


modules/govuk_ci/manifests/agent.pp
```diff
@@ -11,4 +11,16 @@
...
+    port  => '32768:65535',
+    proto => 'udp',
+    ip    => 'any',
+  }
+
+  ufw::allow { 'allow-cdn-acceptance-tests-8080-8090':
+    port  => '8080:8090',
+    proto => 'tcp',
+    ip    => 'any',
+  }
```
alexmuller: I don't think we need this second rule

modules/govuk_ci/manifests/agent.pp
```diff
@@ -11,4 +11,16 @@
   include ::govuk_ci::agent::postgresql
   include ::govuk_ci::agent::mysql
 
+  ufw::allow { 'allow-jenkins-slave-swarm-to-listen-on-ephemeral-ports':
+    port  => '32768:65535',
+    proto => 'udp',
+    ip    => 'any',
+  }
```
alexmuller: Do we have the swarm plugin installed yet? Shouldn't we install the plugin and add a firewall rule for it at the same time? Seems like this would be better done as part of [this story](https://trello.com/c/sH8ZfvdZ/294-connect-a-jenkins-agent-to-the-new-jenkins-master) so that we can keep track of what's going on.

djsd123:
@alexmuller and I have discussed.
Due to the nature of the current Epic.  i.e. security patching jenkins and porting it over to govuk-puppet.  I think this approach is appropriate because all parameters have been defined in ci-puppet.  We are just moving stuff over rather than re-engineering which should be left for when we define the new CI pipeline.

The plugins have already been defined in https://github.com/alphagov/govuk-puppet/pull/5061

PR #5058, opened 2016-11-10
Jenkins firewall rules for CI Master

djsd123:
Allows network access for specified services on the jenkins master.


alexmuller:
Similar comment to #5057, is this the right time to be making this change?

PR #5060, opened 2016-11-10
Add API users to CI

djsd123:
The user accounts that make api calls to jenkins from running jobs.

### DIFF
--- a/modules/govuk_ci/manifests/master.pp
+++ b/modules/govuk_ci/manifests/master.pp
@@ -1,8 +1,10 @@
-# == Class: govuk_ci::master
+# Class govuk_ci::master
 #
-# Class to manage continuous deployment master
+# === Parameters
 #
-# === Parameters:
+# [*agent_user*]
+#   The user Jenkins agent's will use to make api calls.
+#   This user's invoked from jenkins jobs.
 #
 # [*github_client_id*]
 #   The Github client ID is used as the user to authenticate against Github.
@@ -17,6 +19,7 @@
 #   The password to authenticate against the Github Enterprise VPN
 #
 class govuk_ci::master (
+  $agent_user = 'agent',
   $github_client_id,
   $github_client_secret,
   $ghe_vpn_username,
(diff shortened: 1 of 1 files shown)


alexmuller:
I think before we add user accounts it would be useful to understand what they're doing and add it to the commit message briefly so we have a reference:

- agent: absolutely no idea, should this exist? I can't think why the username would need to be configurable. The commit message here implies it's to do with connecting to the master but I don't know if that's necessary? https://github.com/alphagov/ci-puppet/commit/343395d352bed84b770955add5680e65924df453
- pingdom: monitoring maybe? Not sure, I don't think we need this one: https://github.com/alphagov/ci-puppet/commit/89cd52c7092ad5728c1c9f324675f09ad3f46faf
- github_build_trigger: used by the Heroku GitHub trigger to make builds happen
- deploy_jenkins: used by [some deploy jobs](https://github.com/alphagov/govuk-app-deployment/blob/fe36155a42e51da5b0b0067076985317bba67ec9/router/config/deploy.rb#L22) to download artefacts

This class also probably needs to be included somewhere.

surminus:
RE: class inclusion, because we have to keep continuing to rebase commits against changes I'm going to split out https://github.com/alphagov/govuk-puppet/pull/5052 to make it much simpler so we can merge it and then put additions to the master on top of that.

djsd123:
@alexmuller 
Fair point.  I would describe it like: Accounts used to authenticate against the Jenkins API by other services.

Putting more detail at this point might distract from the fact that this is a port of functionality from one repo to another.  At this point, determining whether these accounts are required would be a guess at best. At least until we have the Jenkins instances up and running with jobs loaded.

It's possible the agent account is still valid and poorly named.  Probably used by some webhook or something.  We won't know until we have the service up and running.

djsd123:
I'll add the master class to the node class in another PR

djsd123:
We're going to forward fix this.  We'll add accounts as we need them.

PR #5061, opened 2016-11-11
Jenkins CI Plugins

djsd123:
The plugins being used in our current CI environment.
Plugins extend the features of Jenkins.  They're primarily used for job support.

### DIFF
--- a/hieradata/class/ci_master.yaml
+++ b/hieradata/class/ci_master.yaml
@@ -1,0 +1,81 @@
+---
+
+govuk_jenkins::plugins:
+    greenballs:
+        version: "1.15"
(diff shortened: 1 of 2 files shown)


hieradata/class/ci_master.yaml
```diff
@@ -0,0 +1,81 @@
+---
+
+govuk_ci::master::plugin_hash:
```
surminus: I still can't make up mind whether it's nicer to do like this or just call `govuk_jenkins::plugins` directly

djsd123: Wouldn't that cause conflicts?

PR #5064, opened 2016-11-11
Refactor govuk jenkins class

surminus:
This refactors the govuk_jenkins class and breaks it out to make it more defined. It's main purpose is to be able to be able to set the Jenkins version easier for different machines.

It also makes it clear to be able to call that class with different config, plugins and SSH key pairs depending on what you need for a machine:

```
class { 'govuk_jenkins':
  private_key => 'foo',
  public_key  => 'bar',
  config      => $my_config,
  plugins     => $my_plugins,
}
```

### DIFF
--- a/modules/govuk_jenkins/templates/config/config.xml.erb
+++ b/modules/govuk_jenkins/templates/config/config.xml.erb
@@ -6,7 +6,7 @@
     <string>OldData</string>
     <string>jenkins.security.RekeySecretAdminMonitor</string>
   </disabledAdministrativeMonitors>
-  <version>1.554.2</version>
+  <version><%= @version %></version>
   <numExecutors>4</numExecutors>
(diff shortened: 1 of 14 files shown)


modules/govuk_jenkins/templates/config/config.xml.erb
```diff
@@ -6,7 +6,7 @@
     <string>OldData</string>
     <string>jenkins.security.RekeySecretAdminMonitor</string>
   </disabledAdministrativeMonitors>
-  <version>1.554.2</version>
+  <version><%= @version %> ></version>
```
djsd123: Extra `>`

surminus:
We tested this in Integration and it didn't affect the running of jenkins-1.management. I am tempted to merge this to unblock some of the other work we need to do.

PR #5066, opened 2016-11-11
Remove “contacts” as a frontend app

rubenarakelyan:
“contacts” (aka “contacts-admin”) is now purely a backend app (as of https://github.com/alphagov/contacts-admin/pull/250) and so it is removed here.

Trello: https://trello.com/c/wD10DEXt/260-migrate-the-hmrc-contacts-page-to-a-finder

### DIFF
--- a/hieradata/class/frontend.yaml
+++ b/hieradata/class/frontend.yaml
@@ -2,7 +2,6 @@
 govuk::node::s_base::apps:
   - canary_frontend
   - collections
-  - contacts
   - contacts_frontend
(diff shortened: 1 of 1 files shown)


tijmenb:
I guess we'll have to update this too: https://github.com/alphagov/govuk-app-deployment/blob/master/contacts/config/deploy.rb#L16-L19

rubenarakelyan:
Good call, see https://github.com/alphagov/govuk-app-deployment/pull/68

PR #5068, opened 2016-11-11
Add auth for Github OAuth plugin and VPN

surminus:
This explicitly calls the relevant classes to set the auth secrets required to first authenticate with the Github Enterprise VPN, then to authenticate with Github Enterprise itself.

We pass in parameters and call the classes themselves because when passing in secrets in encrypted hiera we can only specify per environment, and not by class or node specific hieradate. Adding parameters to this class allows us to set different credentials to the default Integration credentials.


surminus:
Depends on https://github.com/alphagov/govuk-puppet/pull/5064

surminus:
This is failing because of the way I've implemented it. It's trickier as I've described above.

`Error: Duplicate declaration: Class[Govuk_jenkins::Config] is already declared;`

PR #5069, opened 2016-11-11
Remove “enable_admin_routes” for contacts

rubenarakelyan:
Now that contacts-admin is a purely admin app with no frontend code, and only deployed to backend servers, we no longer need the ability to disable the admin routes on frontend servers.

Removed from contacts-admin by https://github.com/alphagov/contacts-admin/pull/253.


rubenarakelyan:
Will be merged after https://github.com/alphagov/contacts-admin/pull/253 is deployed.

PR #5076, opened 2016-11-14
Remove manage_config condition

surminus:
This is set to "true" everywhere, so I don't think it's needed. It was originally added so we could test the changes in Integration only, but was never removed when it was set everywhere.

This also recently caused a condition where the changes weren't tested in Vagrant because the default was "false", and when a test was deployed to Integration there were unexpected results.


surminus:
D'oh, this makes it difficult to test in Vagrant due to the auth stuff getting loaded in, which means you can't login to test.

PR #5080, opened 2016-11-15
Add user andreagrandi

mobaig:
Andrea has just joined Finding Things.


modules/users/manifests/andreagrandi.pp
```diff
@@ -0,0 +1,8 @@
+# Creates the carlosvilhena user
```
carvil: Could you please change the comment to `andreagrandi`?

andreagrandi: Good spot!

mobaig:
@andreagrandi 's adding a passphrase to his key.

PR #5081, opened 2016-11-15
Add plugins which are dependencies of other plugins

djsd123:


fofr:
Why did we need to do this?

alexmuller:
I believe this was because the Jenkins plugin installation doesn't resolve dependencies for us, but it would have been good to have that information in the commit message.

PR #5085, opened 2016-11-15
Add environment variables for assisted digital survey to feedback

h-lame:
These will hold the key of the spreadsheet to write survey results to and
the email and private key credentials of the service account to use to
perform the write.  Each environment will have their own values for this
and it'll come from the encrypted puppet hieradata.

To provide config for https://github.com/alphagov/feedback/pull/179/ (which is in service of https://trello.com/c/kZCX6dGB/53-assisted-digital-support-on-done-page-survey)


h-lame:
I added another commit to allow writing multiline values into envvars in a form that envdir supports. We probably need another approval before merging. /cc @surminus

PR #5088, opened 2016-11-16
Allow an alternative proxy ssl port to be specified

steventux:
https://trello.com/c/hCl0Jify/483-make-draft-content-store-accessible-large-3
[Secure connections to the draft content store require an alternative port to 443](https://github.com/alphagov/govuk-puppet/blob/0ec68d0244d498472e156fe3c1e096b326b25808/modules/govuk/manifests/node/s_api_lb.pp#L50).
This PR adds an additional variable to `nginx::config::vhost::proxy` which defaults to '443' but can be used to override the ssl port in the upstream proxy configuration.

### DIFF
--- a/modules/nginx/manifests/config/vhost/proxy.pp
+++ b/modules/nginx/manifests/config/vhost/proxy.pp
@@ -11,6 +11,10 @@
 #   Boolean whether the upstream servers are running on HTTPS/443.
 #   Default: false
 #
+# [*to_ssl_port*]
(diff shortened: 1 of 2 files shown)


alexmuller:
After further investigation I think our use of port 8443 for the draft content store should be entirely transparent to the apps, so I don't think we need this pull request.

steventux:
Closing as this isn't necessary, it's handled elsewhere.

PR #5092, opened 2016-11-16
Add memcached meta and entity store urls for contentapi

steventux:
These are moving out of alphagov-deployment config files, so they need to be turned into env vars.


tijmenb:
Interesting! I did https://github.com/alphagov/govuk_content_api/pull/272.

PR #5101, opened 2016-11-18
Add moar plugins for pipeline

surminus:
We installed the Pipeline plugin and all it's dependencies. I have also ordered the list.


surminus:
Thanks for adding them in alphabetical order @afda16! 🎉

PR #5104, opened 2016-11-18
Rework attachment syncs

surminus:
When the work was completed to add in the DR asset slave in the attachment syncs it caused a condition that can have a significant impact on the user.

If there is any sort of networking issue to the DR node, then the processing of each attachment can increase from ~10 seconds to over 2 minutes (due to the 2 minute timeout that is set). There should be no dependency on rsyncing the files across to both asset slaves before moving onto processing the next attachment on the asset master. The most important aspect of this process is to
scan the file, and move it on disk on the asset master so that the user is then able to access the file.

This commit looks to separate the processing and syncing processes. Instead of including the rsync within the loop, put the file that still needs to be synced into file. We create a list of files into one file
for each "processing-attachments" loop.

A separate process then goes through all the filelist files and attempts to sync each file onto both slaves (and an S3 bucket if set). The timeout has been reduced to 20 seconds because I don't think we will often have a file which should take longer than 20 seconds to copy. When each file
in a filelist has been copied, the filelist is removed and it moves onto the next one.

There is also a catch-all task that runs every morning to ensure the asset master and slaves are all in sync.

### DIFF
--- a/modules/govuk/templates/node/s_asset_base/copy-attachments-to-slaves.sh.erb
+++ b/modules/govuk/templates/node/s_asset_base/copy-attachments-to-slaves.sh.erb
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+set -eu
+
+FILELIST_DIR=$1
+
+# The default Icinga passive alert assumes that the script failed
+NAGIOS_CODE=2
+NAGIOS_MESSAGE="CRITICAL: Copy attachments to asset slaves failed"
+
+# Triggered whenever this script exits, successful or otherwise. The values
+# of CODE/MESSAGE will be taken from that point in time.
+function nagios_passive () {
+  printf "<%= @ipaddress %>\tCopy attachments to asset slaves\t${NAGIOS_CODE}\t${NAGIOS_MESSAGE}\n" | /usr/sbin/send_nsca -H alert.cluster >/dev/null
+}
+
+trap nagios_passive EXIT
+
+usage() {
+  echo "USAGE: $0 <filelist_dir>"
+  echo
+  echo "Copies files in all filelists within <filelist_dir> to each asset slave"
+  echo
+  exit 1
+}
+
+if [ ! "${FILELIST_DIR}" ]; then
+  usage
+fi
+
+ASSET_SLAVE_NODES=$(/usr/local/bin/govuk_node_list -c asset_slave)
+
+for FILELIST in $(ls $FILELIST_DIR); do
+  for FILENAME in $(cat $FILELIST); do
+    for NODE in $ASSET_SLAVE_NODES; do
+      if /usr/bin/timeout 20 rsync -e "ssh -q" --quiet --timeout=10 --relative "${FILENAME}" $NODE:$FILENAME; then
+        logger -t copy_attachments_to_slaves "File ${FILENAME} copied to ${NODE}"
+      else
+        logger -t copy_attachments_to_slaves "File ${FILENAME} failed to copy to ${NODE}"
+      fi
+    done
+    <% if @s3_bucket && @process_uploaded_attachments_to_s3 %>
+      if envdir /etc/govuk/aws/env.d /usr/local/bin/s3cmd --server-side-encryption put "$FILENAME" "s3://<%= @s3_bucket -%>${FILENAME}"; then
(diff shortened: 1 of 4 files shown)


modules/govuk/templates/node/s_asset_base/process-uploaded-attachments.sh.erb
```diff
@@ -5,25 +5,31 @@ set -e
...
-if [ $# -lt 3 ]; then
+if [ $# -lt 4 ]; then
  usage
 fi
 
-ASSET_SLAVE_NODES=$(/usr/local/bin/govuk_node_list -c asset_slave)
+# Creates a unique filename like "filelist-mnt-uploads-whitehall-clean-1479475223"
+FILE_LIST="${TMP_DIR}/filelist$(echo ${CLEAN_DIR} | sed 's/\//-/g')-$(date +%s)"
+
+if [[ ! -d ${TMP_DIR} ]]; then
```
deanwilson: Does this need to be a conditional if we're going to create it everytime it doesn't exist?

boffbowsh:
I think there's a race condition here. If the virus scanner is still scanning and writing to the file list when the rsync kicks in, then the rsync will miss files added to the list after the rsync script starts. Perhaps the file list could actually be a directory which contains links to the files to be rsynced?

surminus:
That is a very good point. When you say link, do you mean like a symlink? I'm just trying to think of the best way to read in the files. Open to more efficient alternatives. 

Alternatively I could just make shove the list into a `mktemp` file and copy it across when the loop has run.

surminus:
I've changed it so it writes to a temporary file, and then moves that to the directory that the rsync process reads from when the loop has finished.

modules/govuk/templates/node/s_asset_base/copy-attachments-to-slaves.sh.erb
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+set -e
+
+FILELIST_DIR=$1
+
+set -u
```
deanwilson: This should be moved up so we don't continue if the directory isn't given.

modules/govuk/templates/node/s_asset_base/process-uploaded-attachments.sh.erb
```diff
@@ -5,25 +5,28 @@ set -e
...
   exit 1
 }
 
-if [ $# -lt 3 ]; then
+if [ $# -lt 4 ]; then
  usage
 fi
 
-ASSET_SLAVE_NODES=$(/usr/local/bin/govuk_node_list -c asset_slave)
+mkdir -p ${TMP_DIR}
```
deanwilson: This should be in quotes rather than `{`

modules/govuk/templates/node/s_asset_base/process-uploaded-attachments.sh.erb
```diff
@@ -34,26 +37,15 @@ while IFS= read -r -d '' FILE
...
-        fi
-      <% end %>
+      echo "${CLEAN_DIR}/${FILE_PATH}" >> ${TMP_FILE}
       rm "$FILE"
     else
       rsync --remove-source-files --relative "$FILE_PATH" "$INFECTED_DIR"
     fi
 done < <(find "$INCOMING_DIR" -type f -print0)
 
+mv ${TMP_FILE} ${TMP_DIR}/
```
deanwilson: Quotes not curlies.

deanwilson:
@boffbowsh we're happy to merge this, any thoughts?

modules/govuk/templates/node/s_asset_base/process-uploaded-attachments.sh.erb
```diff
@@ -1,29 +1,30 @@
...
 
-if [ $# -lt 3 ]; then
+if [ $# -lt 4 ]; then
  usage
 fi
 
-ASSET_SLAVE_NODES=$(/usr/local/bin/govuk_node_list -c asset_slave)
+mkdir -p $TMP_DIR
+
+TMP_FILE=$(mktemp)
```
boffbowsh: A more descriptive name for the tempfile might be good for debugging purposes:
```
$ mktemp attachment_sync_list_XXXX
attachment_sync_list_qxMm
```

surminus:
Addressed comments!

PR #5105, opened 2016-11-18
Add common functions library for Jenkinsfile

afda16:
New multibranch pipeline jobs can share functions provided by this
library in the Jenkinsfile. For instance:

```
def jenkinsLib = load "/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy"
jenkinsLib.myfunction()
```


surminus:
I don't know groovy but this looks OK. Probably best have a review from someone else, too.

PR #5108, opened 2016-11-21
Add integration_deploy job to CI master

surminus:
This creates the job on the CI master that kicks off an application deploy on Jenkins deploy instance in Integration. This job is invoked by the Pipeline build which is created per application.


deanwilson:
Merge away.

PR #5109, opened 2016-11-21
Enable https_redirect for some loadbalancers

alexmuller:
In Kibana:

```
@source_host:backend-lb-* AND @fields.server_port:80 AND NOT @fields.application:json.event.access
@source_host:api-lb-* AND @fields.server_port:80 AND NOT @fields.application:json.event.access AND NOT @fields.application:rummager.publishing.service.gov.uk-json.event.access
```

There are no requests to the backend loadbalancer on port 80, so we can definitely redirect port 80 to 443 rather than responding on port 80.

Same thing for licensify on the API loadbalancer.


alexmuller:
Had to rebase, sorry

PR #5110, opened 2016-11-21
Add Puppet Jenkinsfile

afda16:
Test first Jenkinsfile


deanwilson:
The indentation will need to be cleaned up but that can wait until after it's been tested.

PR #5112, opened 2016-11-22
Upgrade ImageMagick

timblair:
See: https://lists.ubuntu.com/archives/ubuntu-security-announce/2016-November/003633.html

### DIFF
--- a/modules/imagemagick/manifests/init.pp
+++ b/modules/imagemagick/manifests/init.pp
@@ -4,7 +4,7 @@
 #
 class imagemagick {
   package { 'imagemagick':
-    ensure => '8:6.7.7.10-6ubuntu3.1',
+    ensure => '8:6.7.7.10-6ubuntu3.2',
   }
 
   file { '/etc/ImageMagick/policy.xml':


h-lame:
When running govuk_puppet on my dev VM I'm getting:

```
Error: Could not update: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold --force-yes install imagemagick=8:6.7.7.10-6ubuntu3.2' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Version '8:6.7.7.10-6ubuntu3.2' for 'imagemagick' was not found
Error: /Stage[main]/Imagemagick/Package[imagemagick]/ensure: change from 8:6.7.7.10-6ubuntu3.1 to 8:6.7.7.10-6ubuntu3.2 failed: Could not update: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold --force-yes install imagemagick=8:6.7.7.10-6ubuntu3.2' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Version '8:6.7.7.10-6ubuntu3.2' for 'imagemagick' was not found
```
Do we need to do something else to make sure the package is available?

tijmenb:
That's odd, I've just run it without trouble. The only thing I can think of is you have some local cache?

cbaines:
@h-lame Looks like you may need to run `apt update`, this should not be a problem on staging/production as I think unattended-upgrades will ensure that the local information apt has includes the new imagemagick version.

h-lame:
Thanks @cbaines, that did it.  Should `apt update` be part of the govuk_puppet run, even if only on the dev vm?  This might be similar to #4804 where changes that work properly on the deployed machines because they're on all the time and run unattended upgrades don't really work on the dev VM because it's not on and doesn't run things.

cbaines:
> Thanks @cbaines, that did it. Should apt update be part of the govuk_puppet run, even if only on the dev vm?

Maybe? I don't have enough context on the interactions between govuk-puppet and Debian/apt to suggest a good way of doing this.

PR #5114, opened 2016-11-22
Jenkins jobs for ci

afda16:
Add new define to create Pipeline jobs as exported resources from the application manifests. The jobs are collected in the govuk_ci class. The govuk_ci::master::jobs class should be removed if we follow this approach.

Change govuk_jenkins::config::manage_config default value to true.

### DIFF
--- a/modules/govuk_ci/manifests/job.pp
+++ b/modules/govuk_ci/manifests/job.pp
@@ -1,0 +1,33 @@
+# == Define: govuk_ci::job
+#
+# Create a Jenkins multibranch Pipeline job for GOV.UK CI
+#
+# === Parameters
+#
+# [*app*]
+# Application name (Default: title)
+#
+# [*repository*]
+# Repository name, it needs to match the Github repository name (Default: title)
+#
+define govuk_ci::job (
+  $app = $title,
+  $repository = $title,
+) {
+
+  $job_config_directory = '/var/lib/jenkins/jobs'
+  $application_directory = "${job_config_directory}/${app}"
+
+  file { [ $job_config_directory, $application_directory ]:
+    ensure => directory,
+    owner  => 'jenkins',
+    group  => 'jenkins',
+  }
+
+  file { "${application_directory}/config.xml":
+    ensure  => present,
+    content => template('govuk_ci/application_build_job.xml.erb'),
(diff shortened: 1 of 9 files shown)


modules/govuk_ci/manifests/job.pp
```diff
@@ -0,0 +1,40 @@
+# == Define: govuk_ci::job
+#
+# Define to manage a job for GOV.UK CI
```
deanwilson: This doc string could use a rewording.

modules/govuk_ci/manifests/job.pp
```diff
@@ -0,0 +1,40 @@
...
+    ensure => directory,
+    owner  => 'jenkins',
+    group  => 'jenkins',
+  }
+
+  file { $application_directory:
+    ensure  => directory,
+    owner   => 'jenkins',
+    group   => 'jenkins',
+    require => File['/var/lib/jenkins/jobs'],
```
deanwilson: This require isn't needed. Autorequire will do it for you.

modules/govuk_ci/manifests/job.pp
```diff
@@ -0,0 +1,40 @@
...
+#
+define govuk_ci::job (
+  $app = $title,
+  $repository = $title,
+) {
+
+  $job_config_directory = '/var/lib/jenkins/jobs'
+  $application_directory = "${job_config_directory}/${app}"
+
+  file { '/var/lib/jenkins/jobs':
```
deanwilson: Might be worth turning this in to an array of this and the following directory as the properties are all the same.

modules/govuk_ci/manifests/job.pp
```diff
@@ -0,0 +1,40 @@
...
+    ensure  => directory,
+    owner   => 'jenkins',
+    group   => 'jenkins',
+    require => File['/var/lib/jenkins/jobs'],
+  }
+
+  file { "${application_directory}/config.xml":
+    ensure  => present,
+    content => template('govuk_ci/application_build_job.xml.erb'),
+    require => File[$job_config_directory],
```
deanwilson: This should require `${application_directory}` not `$job_config_directory`

PR #5116, opened 2016-11-22
Export ASSET_MANAGER_BEARER_TOKEN

steventux:
This was passed to the Asset Manager class but not exported as an env var.
/cc @tijmenb


tijmenb:
Related: https://github.com/alphagov/govuk_content_api/pull/275

PR #5128, opened 2016-11-23
Jenkins swarm client

djsd123:
This commit adds the functionality to connect a Jenkins agent to a Jenkins master with the swarm client.

The swarm client will run on Jenkins agents as an upstart service.

The discovery address is hard coded as a temporary measure until we have a fact to resolve.

At the moment if we wanted to use another version of the swarm client.  We would have to reflect the change in two places with class parameters `$swarm_client_dest` and `$swarm_client_url`.

There is one manual step where we have to retrieve the api_token for the user we're authenticating against the Jenkins master with.
To automate this would mean opening up a potential security hole.
See: [jenkins.security.ApiTokenProperty.showTokenToAdmins](https://wiki.jenkins-ci.org/display/JENKINS/Features+controlled+by+system+properties)

Requires: https://github.com/alphagov/govuk-puppet/pull/5131, https://github.com/alphagov/packager/pull/121

### DIFF
--- a/modules/govuk_ci/manifests/agent/swarm.pp
+++ b/modules/govuk_ci/manifests/agent/swarm.pp
@@ -1,0 +1,94 @@
+# Class govuk_ci::agent::swarm
+#
+# Joins a jenkins agent machine to a jenkins master/cluster by downloading and invoking the swarm client.
+#
+# note: The jenkins master requires the swarm plugin before an agent can be assigned with this class.
+#       This class requires an associated API user to be created on the jenkins master. After the user
+#       has been created, you'll need to retrieve the user's API token. The API token is used in place of a
+#       a password to authenticate a jenkins user with the jenkins master. An API token can be
+#       found in the jenkins UI under the user's profile 'http://<JENKINS HOST>:<PORT>/user/<USERNAME>/configure'
+#
+# === Parameters
+#
+# [*swarm_user*]
+#   The user that invokes the swarm client
+#
+# [*agent_labels*]
+#   The agent labels advertise what resources a particular jenkins agent has available.
+#   This is useful if you wanted to isolate a job/task to a jenkins agent to
+#   make use of any unique resouces/dependencies a particular jenkins agent may offer.
+#
+# [*discovery_addr*]
+#   The broadcast IP address of an interface.  The swarm client uses udp to discovery to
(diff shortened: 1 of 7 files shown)


surminus:
I wonder if getting the swarm client jar would be a good candidate for https://github.com/alphagov/packager? It should be relatively trivial to create in fpm-cookery and makes more sense to me than doing the curl fetch etc in the Puppet code.

surminus:
Because I was bored and waiting for a task to finish running I had a look at creating a package of this: 
https://github.com/alphagov/packager/pull/121 

This means we could add it to our repo and have a little more control over what we're putting onto our systems. 

I did look at doing the upstart file but I can see why you have done it like you have, due to the username and token etc having to be passed in. I guess we could create a default upstart conf which pulls in some arguments from a `/etc/defaults/jenkins-agent` file or something... but I'm quickly making this much more complicated than it probably needs to be! 

Either way, having a package would _probably_ but easier to manage and you could just use a package resource to install it.

djsd123:
@surminus Sure will give it a try.

djsd123:
@surminus Just had a look and I agree.  The only problem is the upstart script.  While I have no problem with the file being static.  The agent labels won't be across the cluster.

Also, how do we hide the API token?  Would we `source` a file and pull it in from an ENV variable or  like `-password $(source /etc/blah.token)`????

We can discuss when you're in.

djsd123:
Sorry, just saw your comment.  I think hosting the jar would be a good idea as opposed to fetching from upstream.

modules/govuk_ci/manifests/agent/swarm.pp
```diff
@@ -0,0 +1,90 @@
...
+  $swarm_user           = 'jenkins',
+  $agent_labels         = [],
+  $discovery_addr       = '10.1.6.255',
+  $agent_user_api_token = undef,     # Corresponding user: 'jenkins_agent'
+  $swarm_client_package = 'jenkins-agent',
+  $swarm_client_dest    = "/usr/local/bin/${$swarm_client_package}",
+  $apt_mirror_hostname  = undef,
+) {
+
+  include ::curl
```
surminus: Is curl still required?

surminus:
This all looks OK. The only thing that I think is missing is something that ensures the service is running.

PR #5131, opened 2016-11-23
Jenkins agent_user

djsd123:
This user is required for Jenkins agents to authenticate with Jenkins master

relates to https://github.com/alphagov/govuk-puppet/pull/5128


alexmuller:
Does this need to be a class parameter - is there a reason that we'd ever want to change it on specific machines?

djsd123:
Done

PR #5136, opened 2016-11-24
Add env vars to whitehall

steventux:

Moving Whitehall to newer deployment pipeline, so adds database url config for admin and frontend versions of this application plus redis, panopticon, need API and some secret key base values.

Depends on https://github.gds/gds/deployment/pull/1161

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -314,6 +314,15 @@ govuk::apps::publisher::redis_port: "%{hiera('sidekiq_port')}"
 govuk::apps::short_url_manager::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::short_url_manager::redis_port: "%{hiera('sidekiq_port')}"
 
+govuk::apps::whitehall::admin_db_hostname: whitehall-master.mysql
+govuk::apps::whitehall::admin_key_space_limit: '262144'
+govuk::apps::whitehall::admin_db_name: whitehall_production
+govuk::apps::whitehall::admin_db_password: "%{hiera('govuk::apps::whitehall::db::mysql_whitehall_admin')}"
+govuk::apps::whitehall::admin_db_username: whitehall
+govuk::apps::whitehall::db_hostname: whitehall-slave.mysql
+govuk::apps::whitehall::db_name: whitehall_production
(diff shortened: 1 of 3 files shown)


hieradata/common.yaml
```diff
@@ -314,6 +314,15 @@ govuk::apps::publisher::redis_port: "%{hiera('sidekiq_port')}"
...
 
+govuk::apps::whitehall::admin_db_hostname: whitehall-master.mysql
+govuk::apps::whitehall::admin_key_space_limit: '262144'
+govuk::apps::whitehall::admin_db_name: whitehall_production
+govuk::apps::whitehall::admin_db_password: "%{hiera('govuk::apps::whitehall::db::mysql_whitehall_admin')}"
+govuk::apps::whitehall::admin_db_username: whitehall
+govuk::apps::whitehall::db_hostname: whitehall-slave.mysql
+govuk::apps::whitehall::db_name: whitehall_production
+govuk::apps::whitehall::db_password: "%{hiera('govuk::apps::whitehall::db::mysql_whitehall')}"
+govuk::apps::whitehall::db_username: whitehall_fe
```
cbaines: Is this right? The value in alphagov-deployment is just "whitehall".

steventux: I think it's right looking at https://github.gds/gds/alphagov-deployment/blob/master/whitehall/to_upload/database.yml#L5 vs. https://github.gds/gds/alphagov-deployment/blob/master/whitehall/admin_to_upload/config/database.yml.erb#L5

The former is the username for the frontend configuration of the app, the latter is the backend one.

cbaines: Ok, thanks :)

modules/govuk/manifests/apps/whitehall.pp
```diff
@@ -82,6 +185,16 @@
       }
       ",
     }
+
+    if $::govuk_node_class != 'development' {
+      govuk::app::envvar::database_url { $app_name:
+        type     => 'mysql',
```
h-lame: I think this needs to be mysql2 to match the name of the database adapter.  I'm getting:

```
*** [err :: whitehall-backend-1.backend.integration.publishing.service.gov.uk] Specified 'mysql' for database adapter, but the gem is not loaded. Add `gem 'mysql'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord).
```

When deploying and running the migrations.

PR #5143, opened 2016-11-25
New CI agent machines

alexmuller:


alexmuller:
Yep, I think we should add Postgres 9.5 in when we decide what to do about it.

PR #5144, opened 2016-11-25
Ci enable csrf protection

djsd123:
Enabling XSRF (Cross Site Remote Forgery) protection  …
See: https://trello.com/c/cFc17M7P
This option is enabled by default on more recent versions of Jenkins but only for new installs [not upgrades](https://wiki.jenkins-ci.org/display/JENKINS/CSRF+Protection)
This change affects the way we make (headless) API modifications to our Jenkins instance.  To authenticate we now need to include an `XSRF Token Crumb Header`, as well as a username and apiToken/password.
To retrieve a `CSRF Token Crumb Header` you'll have to use xpath.
``
curl -XGET 'http://'"${API_USERNAME}"':'"${API_TOKEN}"'@localhost:8080/crumbIssuer/api/xml?xpath=concat(//crumbRequestField,":",//crumb)'
```
This config has been omitted on our current deploy instance as this could potentially impact the service it provides.

### CODE REVIEW

**Comment by surminus** [comment]
Is anything else being added in the config file other than the below?
```
<crumbIssuer class="hudson.security.csrf.DefaultCrumbIssuer">
    <excludeClientIPFromCrumb>false</excludeClientIPFromCrumb>
</crumbIssuer>
```

That could quite easily be an `if` statement in the template, which could then be triggered with a `versioncmp` in the manifest. This would mean we avoid having to maintain two different template files. We'd also have to remember to create a new template everytime we upgrade Jenkins (though going on past behaviour this won't be a problem...)

If there are more differences then it does make sense having a newer file, but if it's only a smaller segment I'd suggest it's better to keep the config in the same bit.

PR #5149, opened 2016-11-25
Add Jenkins agent role

surminus:
This role is required to give the agent user appropriate permissions to connect to the CI master and become a fully fledged agent.

### DIFF
--- a/modules/govuk_jenkins/manifests/config.pp
+++ b/modules/govuk_jenkins/manifests/config.pp
@@ -59,6 +59,14 @@
 # [*version*]
 #   Specify the version of Jenkins
 #
+# [*create_agent_role*]
+#   If enabled, this creates a role and assigns an "agent" user to that role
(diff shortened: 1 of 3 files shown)


modules/govuk_jenkins/manifests/config.pp
```diff
@@ -77,6 +85,8 @@
   $admins = [],
   $manage_config = true,
   $version = $govuk_jenkins::version,
+  $create_agent_role = false,
+  $jenkins_agent_user = $govuk_ci::agent::swarm::swarm_user,
```
djsd123: Actually, I think this should be:
`$jenkins_agent_user = 'jenkins_agent',` ???

surminus: I agree with this. I was thinking about it and thought it felt backward to call that class from the main govuk_jenkins class.

PR #5151, opened 2016-11-28
Modifying User: jenkins home directory

djsd123:
This change is to keep things consistent.  The assumption that a jenkins user's home is set to a particular directory.

### DIFF
--- a/modules/govuk_ci/manifests/agent/swarm.pp
+++ b/modules/govuk_ci/manifests/agent/swarm.pp
@@ -1,3 +1,4 @@
+# FIXME Remove File resource '/home/jenkins' once merged
 # Class govuk_ci::agent::swarm
 #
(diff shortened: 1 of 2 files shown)


modules/govuk_ci/manifests/agent/swarm.pp
```diff
@@ -56,10 +61,18 @@
...
+    group  => $swarm_user,
+    mode   => '0775',
+  }
+
   user { $swarm_user :
-    comment    => 'I run the jenkins swarm client',
-    managehome => true,
-    shell      => '/bin/sh',
+    comment => 'I run the jenkins swarm client',
+    home    => '/var/lib/jenkins',
```
surminus: This could be set to `home => $swarm_user_home`

PR #5158, opened 2016-11-28
Add non deployable pipeline jobs and support for Github Enterprise

afda16:
Add support for the 'git' branch source plugin in our govuk_ci::job
template, so projects using Github Enterprise can be managed in the
same way. It requires credentials to connect with Github Enterprise
matching the ID supplied in the template.

Add Jenkins Pipeline jobs that are not part of the list of
deployable applications, but still can be managed as a standard
pipeline job

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -564,17 +564,24 @@ govuk_cdnlogs::service_port_map:
 govuk_ci::master::pipeline_jobs:
   <<: *deployable_applications
   async_experiments: {}
+  backdrop-transactions-explorer-collector: {}
   cdn-acceptance-tests: {}
   content-register: {}
+  deployment:
+    source: 'git'
   deprecated_columns: {}
   event-store: {}
   finder-api: {}
   finding-things-migration-checker: {}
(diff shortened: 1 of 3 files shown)


deanwilson:
Looks fine to me but i'll let someone with context give the final +1.

hieradata/common.yaml
```diff
@@ -541,7 +541,50 @@ govuk_cdnlogs::service_port_map:
...
 
-govuk_ci::master::pipeline_jobs: *deployable_applications
+govuk_ci::master::pipeline_jobs:
+  <<: *deployable_applications
+  async_experiments: {}
+  cdn-acceptance-tests: {}
+  content-register: {}
+  deployment:
+    source: 'git'
+    repository: 'git@github.gds:gds/deployment.git'
```
surminus: Could we do something clever with the repository name? It's almost always going to be `git@github.gds:gds/<repo name>.git`, so unless it's different we could probably make an assumption in the template?

afda16: yes I'll update that

PR #5163, opened 2016-11-28
Jenkins job to archive Publishing API events to S3

dougdroper:
This sets up a Jenkins job to run every Sunday at ~5am to run a Publishing API
rake task which archives events to S3.

Depends on https://github.com/alphagov/publishing-api/pull/614 running first


modules/govuk_jenkins/manifests/job/publishing_api_archive_events.pp
```diff
@@ -0,0 +1,14 @@
+# == Class: govuk_jenkins::job::publishing_api_check_validity
+#
+# Create a jenkins job to check the validity of data within the Publishing API
+#
+#
+# === Parameters:
+#
+class gevuk_jenkins::job::publishing_api_archive_events {
```
kevindew: govuk ;)

PR #5164, opened 2016-11-28
Ensure permissions are set

surminus:
Apparently the managehome parameter doesn't manage permissions, so it creates the user home but the user is unable to do anything inside the home directory.


deanwilson:
Merge away, it's worth testing this in integration

PR #5165, opened 2016-11-28
Add licensify and router-data deployment jobs to CI

afda16:
Licensify and router-data trigger specific jobs to deploy
the application on Integration. This commit adds the extra
jobs but we might want to investigate if this can be done
with integration-app-deploy instead.


deanwilson:
i think the code looks fine, I don't know when we should merge it. I'll defer to the others.

PR #5170, opened 2016-11-29
Add PATH on Jenkinsfile

surminus:
When trying to run builds on the new Jenkins agents using the swarm client, we were receiving the following error:

```
jenkins.sh: bundle: not found
```

We need to explicitly set the path so it is able to run bundler. This could be passed in a different way in the future but for now set it on the Jenkinsfile.

This didn't occur on the Jenkins master because we believe environmental variables are set by Jenkins itself.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,6 +12,9 @@ node {
 
     stage("Build") {
 
+      // Explicitly set the path for the Jenkins swarm service on the agent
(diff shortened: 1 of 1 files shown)


alexmuller:
I think we can set this in the global environment variables to avoid having to do it in every app.

surminus:
https://github.com/alphagov/govuk-puppet/pull/5172

PR #5175, opened 2016-11-29
Fix adapter name for whitehall database configs

steventux:
Provide the correct adapter name to the database config class. Whitehall uses the `mysql2` gem.


steventux:
closing in favour of https://github.com/alphagov/govuk-puppet/pull/5173

PR #5176, opened 2016-11-29
Add option to pass in different env vars

surminus:
If we need to move away from the default set environment variables which are set elsewhere, we need to be able to break them out and specify them by class as we are unable to differentiate by specific node class where
we set our secrets in hieradata.

This allows us to set a different PATH for the CI cluster.


alexmuller:
I think this will work but it's a little bit confusing:

- `govuk_ci::master` has a parameter set in hieradata which is passed through to `govuk_jenkins` and then to `govuk_jenkins::config`
- We also set hieradata on `govuk_jenkins::config` directly for the deployment Jenkins

So I think this works because class parameters that are passed in have higher precendence than hieradata.

I wonder if it would be more clear to add another parameter to `govuk::node::s_jenkins` and pass that through in the same way that `govuk_ci::master` passes things through (for consistency), and then change the existing hiera keys.

surminus:
Sounds like a sensible idea.

surminus:
@alexmuller I've added a new commit which makes this more explicit.

PR #5178, opened 2016-11-29
Scheduled rake tasks should be an array of hashes

steventux:

These tasks are parsed as JSON by an application initializer,
they need to be serialised as JSON in the puppet app module,
which means that they need to be passed to the module as an
array of hashes.


steventux:
Closing in favour of https://github.com/alphagov/govuk-puppet/pull/5183

PR #5180, opened 2016-11-29
Use api-elasticsearch for need-api in all envs

alexmuller:
This is working in integration so let's do it everywhere.

/cc @ikennaokpala


ikennaokpala:
@alexmuller thanks for raising this PR.

PR #5186, opened 2016-11-29
Internally trigger deploy job

surminus:
CI should just trigger the job on Integration internally. The public address won't resolve due to NAT.


alexmuller:
Still running? :disappointed:

modules/govuk_jenkins/templates/jobs/integration_app_deploy.yaml.erb
```diff
@@ -9,7 +9,7 @@
           JSON="{\"parameter\": [{\"name\": \"TARGET_APPLICATION\", \"value\": \"$TARGET_APPLICATION\"}, {\"name\": \"TAG\", \"value\": \"$TAG\"}, {\"name\": \"DEPLOY_TASK\", \"value\": \"$DEPLOY_TASK\"}], \"\": \"\"}"
 
           # Deploy to integration environment
-          echo curl -f -XPOST https://<%= @jenkins_integration_api_user %>:<%= @jenkins_integration_api_password %>@deploy.integration.publishing.service.gov.uk/job/Deploy_App/build --data-urlencode json="$JSON"
+          echo curl -f -XPOST https://<%= @jenkins_integration_api_user %>:<%= @jenkins_integration_api_password %>@jenkins-1.management/job/Deploy_App/build --data-urlencode json="$JSON"
```
alexmuller: Unrelated question: why is there an `echo` here?

alexmuller:
I think we solved this a different way in #5193.

PR #5187, opened 2016-11-29
Add auth token for Puppet

surminus:
We need this to trigger the job on deploy Jenkins


modules/govuk_jenkins/manifests/job/integration_deploy.pp
```diff
@@ -17,6 +17,7 @@
   $applications = undef,
   $jenkins_integration_api_user = undef,
   $jenkins_integration_api_password = undef,
+  $puppet_auth_token = undef,
```
deanwilson: Should this param have a docstring added?

surminus: Yes it should :) Added.

PR #5188, opened 2016-11-30
Add a Grafana dashboard for NGinx logs

cbaines:
On both production and staging if you would like to test.

nginx_logs

### DIFF
--- a/modules/grafana/files/dashboards/nginx_logs.json
+++ b/modules/grafana/files/dashboards/nginx_logs.json
@@ -1,0 +1,211 @@
+{
+  "id": 15,
+  "title": "NGinx Logs",
+  "originalTitle": "NGinx Logs",
+  "tags": [],
+  "style": "dark",
+  "timezone": "browser",
+  "editable": true,
+  "hideControls": false,
+  "sharedCrosshair": false,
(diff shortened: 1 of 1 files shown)


alexmuller:
I think the list of machines in this JSON file will get out of date quite quickly as people forget to add and remove them. What about templating this file and creating the list of machines dynamically using the output of `govuk_node_list` and maybe the Puppet `generate()` function?

cbaines:
> I think the list of machines in this JSON file will get out of date quite quickly as people forget to add and remove them. What about templating this file and creating the list of machines dynamically using the output of govuk_node_list and maybe the Puppet generate() function?

Shouldn't be necessary to template it, as Grafana will regenerate the list from the data in Graphite when the dashboard is opened in the users browser. I'd prefer to just remove all the entries if you don't want them in the repository?

alexmuller:
If they're generated by Grafana then not having them in this repo would be great, thanks! The same for domains and status codes if those are also generated.

cbaines:
> If they're generated by Grafana then not having them in this repo would be great, thanks! The same for domains and status codes if those are also generated.

Ok, now updated.

PR #5189, opened 2016-11-30
Pass environment to smokey

cbaines:
This will hopefully have an effect when combined with https://github.com/alphagov/smokey/pull/211


modules/monitoring/templates/smokey-loop.conf
```diff
@@ -6,4 +6,4 @@ description "Smokey loop outputs JSON which is consumed by monitoring"
 start on runlevel [2345] and stopped $UPSTART_JOB
 respawn
 
-exec /opt/smokey/tests_json_output.sh /tmp/smokey.json
```
cbaines: Have tested this in integration, the @environment comes out as "production", which is probably right, but not helpful, as I want to have the GOV.UK environment here. It also doesn't have a ending newline, but that can be fixed by removing the - before the closing % sign.

Does anyone know what I can put here to get the GOV.UK environment as a string?

cbaines: Had a look myself for similar things, and it seems like its all explicit in hiera for each class that needs to know, so I've done the same now for smokey  	9683414.

cbaines:
Now tested on integration, and it worked. This will also change how smokey is run on staging and production also.

cbaines:
The alert has now disapeared! Long may it last.

https://alert.integration.publishing.service.gov.uk/extinfo.cgi?host=monitoring-1.management.integration.publishing.service.gov.uk&type=2&service=Run+efg+normal+priority+tests

PR #5191, opened 2016-11-30
Configure scheduled signon rake tasks via jenkins

steventux:

Signon performs several rake tasks at regular intervals in order to
clean up certain user account specific data and fetch organisational
data. Previously this was done via the 'whenever' gem and the appropriate
cron configuration was passed from alphagov-deployment, this is going
away so we'll configure these tasks via jenkins per environment.

I'm not sure whether multiple jobs can be configured like this, given the structure of the jobs yaml file it's implied they are just an array of jobs.

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -113,6 +113,12 @@ govuk_jenkins::job_builder::environment: 'integration'
 
 govuk_jenkins::job::deploy_puppet::commitish: 'integration'
 
+govuk_jenkins::job::signon_cron_rake_tasks::configure_jobs: true
+govuk_jenkins::job::signon_cron_rake_tasks::rake_oauth_access_grants_delete_expired_frequency: '30 11 * * 2'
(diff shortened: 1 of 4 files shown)


alexmuller:
I think the multiple YAML is good. Instead of running an SSH command, could we get these jobs to trigger a downstream job [similar to how we do the data sync](https://github.com/alphagov/govuk-puppet/blob/8bebde9ed75595eb14a4dae6b4e0e39ed0f00847/modules/govuk_jenkins/templates/jobs/data_sync_complete_integration.yaml.erb#L42-L46)? That means we have one place which shows all the rake task history.

steventux:
That's a nice idea @alexmuller, do downstream like this need to be configured within `publishers` config or can they be left in the `builders` conf? ie is `publishers` downstream configuration?

steventux:
Thanks @alexmuller, I've updated, I'll deploy this branch to integration as I'm not 100% on the downstream config.

PR #5192, opened 2016-11-30
CI Next Build Number Plugin

djsd123:
This plugin is required to preserve the history sequence of builds

### DIFF
--- a/hieradata/class/ci_master.yaml
+++ b/hieradata/class/ci_master.yaml
@@ -103,6 +103,8 @@ govuk_jenkins::plugins:
       version: "1.1.1"
     monitoring:
       version: "1.62.0"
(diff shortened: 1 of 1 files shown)


hieradata/class/ci_master.yaml
```diff
@@ -190,3 +190,7 @@ govuk_jenkins::plugins:
       version: "2.11"
     ws-cleanup:
       version: "0.32"
+    next-build-number:
+      version: "1.4"
+    job-dsl:
+      version: "1.53"
```
alexmuller: Do we need the `job-dsl` plugin? It's listed as an optional dependency and is only needed to change the build number from within a job DSL script which I don't think we do anywhere.

PR #5195, opened 2016-11-30
Enable CSRF protection

djsd123:
Tightening security by enabling CRSF protection
This will only apply to Jenkins 2 instances

See: https://trello.com/c/cFc17M7P


surminus:
Ah, so this branch uses the old version of the Jenkinsfile which means builds are different to how we want. I'm going to rebase it and push.

PR #5204, opened 2016-12-01
Ci exclude release related branches from automated builds

afda16:
The new Multibranch pipeline jobs trigger a build for changes pushed
to branches. We want to exclude builds from branches used for
deployments, such as release, staging, etc.


alexmuller:
I think this is ok but I wonder if it's a regular expression? https://jenkins.io/doc/book/pipeline-as-code/ seems to say so. In which case it might need pipes between the words.

afda16:
I think it's not very clear, the include branches section says "Space-separated list of branch name patterns to consider" and the exclude section says "Branch name patterns to ignore"

I tried with a few variations of regular expressions and it didn't work

PR #4909, opened 2016-09-27
[DO NOT MERGE] Add Slack properties and publisher details

carvil:
This commit adds the slack configuration to the "Copy Data To Integration" job.

Before this can be merged, we need to:
- [ ] Add the Slack Access Token to hieradata in gds/deployment
- [ ] Test this against a local Jenkins.

PR #5206, opened 2016-12-01
Add Slack properties and publisher to jobs

carvil:
This commit adds the slack configuration to the following jobs:
- Copy Data To Integration;
- Smokey;
- Deploy Puppet.

This will alert us on some of the statuses on those jobs (more details on which alerts can be seen in the individual commits).

Requires:
- [x] https://github.gds/gds/deployment/pull/1172 to be merged;
- [x] https://github.gds/gds/deployment/pull/1177 to be merged;
- [x] testing in integration.

Original PR is [here](https://github.com/alphagov/govuk-puppet/pull/4909) for reference.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/copy_data_to_integration.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/copy_data_to_integration.yaml.erb
@@ -20,6 +20,16 @@
         - inject:
             properties-content: |
               PARALLEL_JOBS=2
+        - slack:
+            notify-start: false
+            notify-success: true
+            notify-aborted: true
+            notify-notbuilt: true
+            notify-unstable: false
+            notify-failure: true
+            notify-backtonormal: false
+            notify-repeatedfailure: false
+            include-test-summary: false
     scm:
       - env-sync-and-backup_Copy_Data_to_Integration
     logrotate:
@@ -60,6 +70,11 @@
             HOSTNAME=deploy.integration.publishing.service.gov.uk
             API_KEY=<%= @ci_alphagov_api_key %>
             AUTH_TOKEN=<%= @auth_token %>
+      - slack:
+          team-domain: <%= @slack_team_domain %>
+          auth-token: <%= @environment_variables['SECOND_LINE_SLACK_AUTH_TOKEN']%>
+          build-server-url: <%= @slack_build_server_url %>
+          room: <%= @slack_room %>
     wrappers:
         - ansicolor:
             colormap: xterm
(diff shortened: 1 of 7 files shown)


modules/govuk_jenkins/templates/jobs/copy_data_to_integration.yaml.erb
```diff
@@ -60,6 +70,11 @@
             HOSTNAME=deploy.integration.publishing.service.gov.uk
             API_KEY=<%= @ci_alphagov_api_key %>
             AUTH_TOKEN=<%= @auth_token %>
+      - slack:
+          team-domain: <%= @slack_team_domain %>
+          auth-token: <%= @environment_variables['SECOND_LINE_SLACK_AUTH_TOKEN']%>
```
surminus: Does this work? I thought this would need to be declared in the manifest that created the job. Still, I know there are some peculiarities between where variables are defined and used in templates so wouldn't be surprised if this does work.

If we're getting the secret by calling the hash into the job definition itself, does it also need to be set as a global env var?

carvil: @surminus we have also removed the environment variable from integration and staging in order to reduce the number of slack notifications in the channel. If you would like to check this in integration we will need to re-add the same environment variable in https://github.gds/gds/deployment/commit/fe0805b337bbcf9e7d6817e3e4497047e8286ba1 to the integration file.

surminus: Ah OK, we're setting it as a global env var so that it can be accessed by all the different jobs using it without having to repeat it everywhere - I see! That makes sense. If this works then I'm fine with this.

carvil: Yup, that's exactly it. It worked well in integration, and then we removed the environment variable from integration and staging so it doesn't get too spammy.

surminus:
If this is only enabled in Production, what's the behaviour of the Slack notification plugin if it's set in the job but the authentication token isn't present? The same template for Deploy Puppet is used in all three environments, without the token will it just silently fail?

carvil:
@surminus Good point. We tested this in integration again and we do need those environment variables to be in place (otherwise, `jenkins-jobs` doesn't produce the new configuration). However, the token can be invalid - the slack plugin handles that for us and doesn't impact the jobs. This means we can still prevent unwanted notifications in integration and staging.

There is a PR to add those variables again to the missing environments: https://github.gds/gds/deployment/pull/1177

With that branch added during the puppet deploy, the output of `jenkins-jobs` was successful and the jobs complete as they normally do. The post build action happens and the slack plugin fails silently.

The PR mentioned above needs to be merged before this PR is merged.

Is there anything else you think we should test?

PR #5210, opened 2016-12-02
CI golang plugin

djsd123:
I think this plugin will help simplify steps to build apps written in go.

See: https://wiki.jenkins-ci.org/display/JENKINS/Go+Plugin

### DIFF
--- a/hieradata/class/ci_master.yaml
+++ b/hieradata/class/ci_master.yaml
@@ -71,6 +71,8 @@ govuk_jenkins::plugins:
       version: "0.24"
     github:
       version: "1.22.4"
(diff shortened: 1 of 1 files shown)


alexmuller:
I'm a little confused by what this plugin does - is it just installing Go versions and setting up the path? Because our `golang` class installs versions and we have `/usr/lib/goenv/shims` which can go in the `$PATH` if it isn't already.

djsd123:
After evaluating this plugin, I feel that while useful in standard Jenkins management scenarios it isn't as useful to our needs as it doesn't offer pipeline support yet.

PR #5213, opened 2016-12-02
Enable integration-app-deploy in CI

afda16:
Enable the action in the integration-app-deploy CI job to trigger
the deploy build on Integration.


surminus:
Just while I remember, we probably also need to pass an auth_token

PR #5214, opened 2016-12-02
Enable signon jenkins rake tasks

steventux:

Missed in https://github.com/alphagov/govuk-puppet/pull/5191

These are only needed in integration and production.


steventux:
I've DNM'd this because it's possible that Puppet would be deployed before the application deploy config was updated/deployed and the tasks would be scheduled twice

PR #5218, opened 2016-12-02
Template config file for Jenkins 2

alexmuller:
Jenkins 2 is super picky about its config file. Jenkins 2 will decide to modify its config file when it starts so that it's formatted the way it wants it to be. Puppet will then revert that change and cause Jenkins to restart.

Here are some of the highlights of this ridiculousness:

- In Jenkins 2, things are sorted in some arbitrary way which is different to Jenkins 1. Don't try and sort things the old way. I could get on board with this except that it's less alphabetical than the sort order we were using in Jenkins 1.
- Jenkins 1 will add an exclamation mark before the XML for user IDs. Jenkins 2 will remove that exclamation mark.
- If Jenkins 2 decides that it needs to rewrite your perfectly fine config file, it will write out a POSIX-incompatible file because it doesn't have trailing newline character.

I am sick of XML and I am sick of using Jenkins and I feel awful for littering an XML template with conditionals but what are you going to do ¯\\\_(ツ)\_/¯.

### DIFF
--- a/modules/govuk_jenkins/manifests/config.pp
+++ b/modules/govuk_jenkins/manifests/config.pp
@@ -197,8 +197,13 @@ class govuk_jenkins::config (
       content => template('govuk_jenkins/config/jenkins.model.JenkinsLocationConfiguration.xml.erb'),
     }
 
+    # FIXME: Remove once all Jenkinses are upgraded to at least 2.0.0
     if versioncmp($version, '2.0.0') == 1 {
       $csrf_version = true
+      $jenkins_2 = true
+      $markup_formatter_version = '1.5'
+    } else {
+      $markup_formatter_version = '1.3'
     }
 
     file { '/var/lib/jenkins/config.xml':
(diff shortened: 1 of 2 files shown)


surminus:
Good work figuring out why it kept restarting. @deanwilson prevously made the argument about maintaining two separate `config.xml` files in Puppet for each version of Jenkins which I argued against - in light of this, I'm wondering whether that _is_ the right way to go instead of having conditionals all over the place? 

My original argument against that was that I didn't like the idea of having to maintain two different files.

Otherwise I think this looks fine. The re-ordering of the permissions based (or not, apparently) on alphabetical order is particularly, er, exciting? 😐

alexmuller:
Yeah, we could do 2 separate files - I'm not sure which approach I prefer. I'm hoping these conditionals won't be around for long because we'll upgrade the other Jenkins to be Jenkins 2 relatively soon. Let's go with this for now and see how we do.

PR #5219, opened 2016-12-05
Add check for travel advice smokey test

carvil:
This will run the following feature:
https://github.com/alphagov/smokey/blob/master/features/travel_advice.feature

cc/ @steventux


steventux:
Merge when green :star2:

PR #5215, opened 2016-12-02
Add reusable scripts for CI Jenkins

brenetic:
This adds some common use scripts across Ruby apps

PR #5223, opened 2016-12-05
Update Jenkinsfile

brenetic:
There are now reusable scripts available.

- [x] Merge after [this PR](https://github.com/alphagov/govuk-puppet/pull/5215) has been merged and Puppet deployed.


surminus:
Are we going to document the common/default functions anywhere? that'd be super useful (in fact, I'm going to use them now)

PR #5224, opened 2016-12-05
Remove reverse proxy for drug-analysis-prints

sihugh:
This relates to https://govuk.zendesk.com/agent/tickets/1453127

We're removing the reverse proxy configuration for `/drug-analysis-prints/` as MHRA have a new site which they want to manage mappings for.

The PDFs under `/home/groups/public/documents/sentineldocuments/` appear to all 404 now so we're removing this now.  I've requested that MHRA add mappings for these as there are 1436 of these without mappings.

### DIFF
--- a/modules/bouncer/templates/www.mhra.gov.uk_nginx.conf.erb
+++ b/modules/bouncer/templates/www.mhra.gov.uk_nginx.conf.erb
@@ -21,12 +21,6 @@ server {
       '/home/resources/',
       '/home/websites/',
 
-      # Drug Analysis Prints content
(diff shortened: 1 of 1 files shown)


sihugh:
We can't merge these yet as they've got to fix mappings at https://transition.publishing.service.gov.uk/sites/mhra/mappings?path_contains=%2Fhome%2Fgroups%2Fpublic%2Fdocuments%2Fsentineldocuments%2F&utf8=%E2%9C%93 first

sihugh:
They've put mappings in, so this is good to go!

PR #5228, opened 2016-12-05
Upgrade puppet-curl module

alexmuller:
This allows it to be installed alongside other Puppet modules (eg the Docker Compose one)

Depends on https://github.com/alphagov/puppet-curl/pull/10

/cc @carlosmartinez


alexmuller:
Ready to be reviewed.

PR #5237, opened 2016-12-06
Add user accounts for GOV.UK penetration test

alexmuller:
These accounts can be removed at the end of the week.


alexmuller:
SSH keys aren't strong enough for our spec tests.

timblair:
:| OK, let's get new keys from them when they get here.

PR #5245, opened 2016-12-08
Add Jenkins user to docker group in CI

afda16:
The Jenkins user needs to be able to build and manage containers
on the docker agents, so it needs to belong to the docker group.

### DIFF
--- a/modules/govuk_ci/manifests/agent/docker.pp
+++ b/modules/govuk_ci/manifests/agent/docker.pp
@@ -2,6 +2,12 @@
 #
 # Installs and configures Docker and Docker Compose
 class govuk_ci::agent::docker {
-  include ::docker
+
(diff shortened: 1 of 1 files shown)


modules/govuk_ci/manifests/agent/docker.pp
```diff
@@ -2,6 +2,12 @@
 #
 # Installs and configures Docker and Docker Compose
 class govuk_ci::agent::docker {
-  include ::docker
```
alexmuller: It has the same meaning as the `class { '::docker': ... }` definition below but that also provides options (ie `docker_users`)

carlosmartinez: gotcha. well let's merge this mother!

PR #5249, opened 2016-12-12
Add temporal CI Hiera key

afda16:
Add original govuk_jenkins::github_enterprise_cert_path key until
all the Hiera lookups are updated on all the environments to use
govuk_jenkins::github_enterprise_cert::certificate_path


afda16:
We think the property that is doing the Hiera lookup is not needed in the Deploy Jenkins instances, so we don't need this change any more

PR #5250, opened 2016-12-12
Remove unused ecryptfs and motd modules

alexmuller:
These were used for the old encrypted licensify MongoDB machines, but they're no longer needed.


alexmuller:
Who is will?

PR #5254, opened 2016-12-12
Fix default content schema branch in Jenkins config

suzannehamilton:
The content schema checkout step takes a git branch, with the default value of 'deployed-to-production'. In the old shell script, the default was inserted using shell parameter expansion. In the new Groovy script, I think the default should be inserted using a default method argument.

The implementation before this commit gave the error "Unimplemented pathspec magic" because the ":-" characters were being passed to git rather than being processed by the build script.

Note that this changes the signature of the base Jenkinsfile, but I think we might be the first people to use the `contentSchemaDependency` step - I've just added it to a branch of collections-publisher, and I couldn't find any references to it in other repos.


tijmenb:
👍

PR #5257, opened 2016-12-12
Add a way to blacklist packages

surminus:
We currently blacklist packages for unattended_upgrades, however when we bring up a machine Puppet will install packages, and some may install packages we do not want installed. This change makes it impossible to install the specified packages whatsoever.

An example of this is systemd, which affects the way the system behaves which we may not want to do yet. A newer version of postgresql installs systemd despite it not being a dependency which caused some problems.

### DIFF
--- a/modules/govuk_apt/manifests/package_blacklist.pp
+++ b/modules/govuk_apt/manifests/package_blacklist.pp
@@ -1,0 +1,20 @@
+# == Class: Govuk_apt::Package_blacklist
+#
+# Blacklists specific packages from being able to be installed on the system.
+# This will mean that even if a user tries to install a package use apt, the
+# package specified will simply appear as unavailable for install.
+#
+# === Parameters:
+#
+#  [*packages*]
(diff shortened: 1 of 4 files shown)


modules/govuk_apt/manifests/package_blacklist.pp
```diff
@@ -0,0 +1,20 @@
+# == Class: Govuk_apt::Package_blacklist
+#
+# Blacklists specific packages from being able to be installed on the system.
+# This will mean that even if a user tries to install a package use apt, the
```
deanwilson: "This means, if a user tries to install a package using apt, the"

modules/govuk_apt/manifests/package_blacklist.pp
```diff
@@ -0,0 +1,20 @@
...
+# package specified will simply appear as unavailable for install.
```
deanwilson: "install" installation.

PR #5260, opened 2016-12-13
Add a way to blacklist packages

surminus:
We currently blacklist packages for unattended_upgrades. However, when we bring up a new machine, Puppet will install packages from fresh, and some packages may install other packages we do not want installed. This change makes it impossible to install the specified packages whatsoever.

An example of this is systemd, which affects the way the system behaves which we may not want to do yet. A newer version of postgresql installs systemd which caused some problems. We want to block this from happening.

### DIFF
--- a/modules/govuk_apt/manifests/package_blacklist.pp
+++ b/modules/govuk_apt/manifests/package_blacklist.pp
@@ -1,0 +1,13 @@
+# == Class: Govuk_apt::Package_blacklist
+#
+# Blacklists specific packages from being able to be installed on the system.
+# This means, if a user tries to install a package using apt, the package
+# specified will appear as unavailable for installation.
+#
+class govuk_apt::package_blacklist {
+  apt::pin { 'blacklist_systemd':
+    packages => 'systemd',
+    priority => '-1',
+    origin   => '""',
+  }
+}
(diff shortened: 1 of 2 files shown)


modules/govuk_apt/manifests/package_blacklist.pp
```diff
@@ -0,0 +1,22 @@
...
+#
+#  [*packages*]
+#    An array of packages that you wish to blacklist.
+#
+class govuk_apt::package_blacklist (
+  $packages = [],
+){
+  validate_array($packages)
+
+  apt::pin { "blacklist_${packages}":
```
afda16: If we add a second package to the govuk_apt::package_blacklist::packages property and run Puppet again, this class generates a second file in addition to the first one, for instance '/etc/apt/preferences.d/blacklist_package1.pref' containing 'Package: package1', and a second  '/etc/apt/preferences.d/blacklist_package1package2.pref' containing 'Package: package1 package2', I don't know if this is what is expected

surminus: Oh, dammit, I forgot about weird puppet loops. 😞

surminus:
I have simplified my original approach as since we're only blacklisting a single package.

surminus:
Thanks. We should probably remember this change when come round to doing Xenial!

PR #5265, opened 2016-12-13
Add libmysqlclient-dev package to CI agents

alexmuller:
CI agents which run MySQL tests need to have this package installed.


suzannehamilton:
I haven't done enough with Puppet to review this properly, but it looks like this will fix the MySQL issue in Jenkins 2. Thanks!

PR #5266, opened 2016-12-13
Add ssh slaves to CI

surminus:
See commits for detail. This is a PR to replace the current Jenkins swarm plugin with the SSH slave plugin instead as we are seeing stability issues with the swarm protocol.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -635,6 +635,26 @@ govuk_ci::master::pipeline_jobs:
   ubuntu_unused_kernels:
     repo_owner: 'gds-operations'
 
+govuk_ci::master::ci_agents:
+  ci-agent-1:
+    agent_hostname: 'ci-agent-1.ci'
+    description: 'Build agent with docker'
+    labels: 'mongodb-2.4 docker'
+  ci-agent-2:
+    agent_hostname: 'ci-agent-2.ci'
+    labels: 'mongodb-2.4'
+  ci-agent-3:
+    agent_hostname: 'ci-agent-3.ci'
+    labels: 'mongodb-2.4'
+  ci-agent-4:
+    agent_hostname: 'ci-agent-4.ci'
+    labels: 'mongodb-3.2'
+  ci-agent-5:
+    agent_hostname: 'ci-agent-5.ci'
+    labels: 'mongodb-3.2'
+govuk_ci::master::credentials_id: 'jenkins-ssh-slave'
+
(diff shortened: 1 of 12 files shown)


modules/govuk_ci/manifests/master.pp
```diff
@@ -57,4 +59,9 @@
     mode => '0555',
   }
 
+  create_resources('::govuk_jenkins::ssh_slave', $ci_agents, {
+    'credentials_id' => $credentials_id,
+    'labels'         => 'mongo-3.2',
```
deanwilson: Is this to ensure that all CI hosts have this version of mongo? Should this be in the hiera with the other labels?

modules/govuk_jenkins/manifests/ssh_slave.pp
```diff
@@ -0,0 +1,60 @@
...
+#
+# [*jenkins_user*]
+#   The user which the master connects to the agent over SSH as.
+#
+define govuk_jenkins::ssh_slave (
+  $agent_hostname,
+  $credentials_id,
+  $labels = undef,
+  $description = 'Generic build agent',
+  $executors = '4',
```
surminus: I did it to annoy you ;) I'll change them. There's no reason other than absent mindedness.

modules/govuk_ci/manifests/master.pp
surminus: This is a default label. I put it in here to try and cut down the default hiera, but it probably makes more sense to be verbose in the hiera. WIll update.

hieradata/common.yaml
```diff
@@ -635,6 +635,26 @@ govuk_ci::master::pipeline_jobs:
   ubuntu_unused_kernels:
     repo_owner: 'gds-operations'
 
+govuk_ci::master::ci_agents:
+  ci-agent-1:
+    agent_hostname: 'ci-agent-1.ci'
+    description: 'Build agent with docker'
+    labels: 'mongodb-2.4 docker'
```
djsd123: Would you please do me a solid and amend the labels so they match the labels in our Jenkinsfiles?

i.e. `mongo-3.2` becomes `mongodb-3.2`

surminus: @djsd123 sorry yeah, I spotted that.

surminus:
I've deployed this to Integration (and stopped puppet on the master and agents). I believe it works as intended, and after some testing I've fixed some bugs.

PR #5268, opened 2016-12-14
Bugfix quotation in copy_attachments_to_slaves

surminus:
In 897ff22ac8a0858d1dac6189fdc27a3c4047fdf1 I mislaid a quotation which caused a problem with the rsync.


surminus:
Thanks, good shout.

PR #5287, opened 2016-12-19
Add per-project configuration for branches which are excluded by Jenkins

suzannehamilton:
By default, Jenkins does not build branches which represent releases, e.g. 'release' or 'deployed-to-production'. This makes sense for most projects, but it is sometimes useful to be able to trigger builds of release branches, such as for projects which are used as downstream tests of other projects like the content schemas.

This is the solution that Ana and I discussed - it avoids having to create a completely separate Puppet config for the schema tests. In a separate commit, I've also configured the first of the downstream builds to override the defaults so that deployed-to-production is a valid Jenkins branch build.

Marked as DO NOT MERGE until I've updated collections-publisher to handle 'deployed-to-production' builds (alphagov/collections-publisher#231).

@afda16 and @tijmenb - would you mind taking a look at this?

### DIFF
--- a/modules/govuk_ci/templates/application_build_job.xml.erb
+++ b/modules/govuk_ci/templates/application_build_job.xml.erb
@@ -38,7 +38,7 @@
 <%- end -%>
           <repository><%= @repository %></repository>
           <includes>*</includes>
-          <excludes>release deployed-to-* integration staging production</excludes>
+          <excludes><%= @branches_to_exclude.join(' ') =></excludes>
           <buildOriginBranch>true</buildOriginBranch>
           <buildOriginBranchWithPR>true</buildOriginBranchWithPR>
           <buildOriginPRMerge>false</buildOriginPRMerge>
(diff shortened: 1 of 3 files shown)


modules/govuk_ci/templates/application_build_job.xml.erb
```diff
@@ -38,7 +38,7 @@
 <%- end -%>
           <repository><%= @repository %></repository>
           <includes>*</includes>
-          <excludes>release deployed-to-* integration staging production</excludes>
+          <excludes><%= @branches_to_exclude.join(' ') =></excludes>
```
afda16: I think the closing block is misspelled, it needs to be '%>' instead of '=>'

suzannehamilton: Yes, I should have spotted that! Thanks. :)

alexmuller:
I think this is the right solution, but it will make Jenkins do a build whenever collections-publisher is deployed to production 👍

suzannehamilton:
@alexmuller: That's right. alphagov/collections-publisher#231 will handle that case: Jenkins will check a new build parameter and only run a full build of deployed-to-production if it's been triggered by the schema tests rather than a production deploy.

suzannehamilton:
alphagov/collections-publisher#231 is now on master, so I think this is ready to go.

PR #5289, opened 2016-12-19
Correctly docs.publishing.service.gov.uk to S3

tijmenb:
In https://github.com/alphagov/govuk-puppet/pull/5269 I introduced a nginx config to proxy `docs.publishing.service.gov.uk` to an S3 bucket. The configuration was wrong, which caused nginx to crash loop on the load balancers. Sad! The crash was fixed in https://github.com/alphagov/govuk-puppet/pull/5279.

This PR fixes the following:

- By default, the nginx configs puts a HTTP basic auth password on the proxy. We like the domain to be publicly accessible, so we disable this.
- To make S3 respond with the correct content, we set the `Hostname` HTTP header to the name of the bucket. If this isn't set, Amazon will try to find a bucket called `docs.publishing.service.gov.uk` and return 404.
- Add `docs` to the hostnames. This will make nginx respond on the new `docs.x` domain.

This time, I have employed a novel development technique called "testing your stuff locally" to make sure that this config actually works.

I tested it by doing the following. First, start a load balancer VM on locally with vagrant:

```
~/govuk/govuk-puppet $ vagrant up backend-lb-1.backend
```

Next, point the test domain to the running VM:

```
# /etc/hosts
10.1.3.101 docs.dev.gov.uk
```

After this, visiting the docs will do the correct thing:

http://docs.dev.gov.uk/

### DIFF
--- a/modules/govuk/manifests/node/s_backend_lb.pp
+++ b/modules/govuk/manifests/node/s_backend_lb.pp
@@ -107,7 +107,12 @@ class govuk::node::s_backend_lb (
     to => "https://admin.${perfplat_public_app_domain}/",
   }
 
+  # Allthough there are different staging, integration and production buckets
+  # created by Terraform on S3, we do not intend to use them. To avoid confusion
+  # we proxy all domains to the production bucket.
   nginx::config::vhost::proxy { "docs.${app_domain}" :
-    to => ['govuk-developer-documentation-production.s3-website-eu-west-1.amazonaws.com'],
+    to        => ['govuk-developer-documentation-production.s3-website-eu-west-1.amazonaws.com'],
+    protected => false,
+    http_host => 'govuk-developer-documentation-production.s3-website-eu-west-1.amazonaws.com',
   }
 }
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/node/s_backend_lb.pp
```diff
@@ -108,6 +108,8 @@
   }
 
   nginx::config::vhost::proxy { "docs.${app_domain}" :
-    to => ['govuk-developer-documentation-production.s3-website-eu-west-1.amazonaws.com'],
+    to        => ['govuk-developer-documentation-production.s3-website-eu-west-1.amazonaws.com'],
+    protected => false,
+    http_host => 'govuk-developer-documentation-production.s3-website-eu-west-1.amazonaws.com',
```
deanwilson: Should this be different per environment?

tijmenb: I've added a comment above. I'm open to making it per-environment if that's less surprising though!

deanwilson: Thanks for the clarifying comment. Without having any real insight in to the potential usage of the buckets I don't know if it's worth having per environment setups to enable future testing of changes. I'm happy to follow your lead.

PR #5292, opened 2016-12-19
Bump puppetlabs-rabbitmq module version.

deanwilson:
We're trying to run our code under `strict_variables` in prep for
a move to Puppet 4. The `puppet-staging` module, which we only use as a
dep to `puppetlabs-rabbitmq` needs to be upgraded for tests to pass.


afda16:
Should we remove the old puppetlabs-rabbitmq packages from vendor/puppet/cache?

deanwilson:
Yes we should. Nice catch

alexmuller:
There's also a `nanliu-staging-1.0.3.tar.gz`, sorry!

PR #5294, opened 2016-12-19
Add monitoring for jenkins agents

surminus:
We've had occasions where Jenkins agents have lost their connection to the Jenkins master, even when the processes and server seems to be OK. This hopefully allows us to monitor the agent status from the perspective of the master.

### DIFF
--- a/modules/icinga/files/usr/lib/nagios/plugins/check_jenkins_agent
+++ b/modules/icinga/files/usr/lib/nagios/plugins/check_jenkins_agent
@@ -1,0 +1,24 @@
+#!/bin/bash
+#
+# Desc : Plugin to check status of Jenkins agents. Should be run
+# on the master.
+
+PROGNAME=$(basename "$0")
+
+if [ -z "$1" ]
+then
+  echo -e "Use : $PROGNAME <Jenkins agent hostname> -- Ex : $PROGNAME ci-agent-1 \n "
+  exit 3
+fi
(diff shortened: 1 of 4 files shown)


modules/icinga/files/usr/lib/nagios/plugins/check_jenkins_agent
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+#
+# Desc : Plugin to check status of Jenkins agents. Should be run
+# on the master.
+
+PROGNAME=$(basename "$0")
+
+if [ "$1" = "" ]
```
deanwilson: Should this be `-z` instead?

modules/icinga/files/usr/lib/nagios/plugins/check_jenkins_agent
```diff
@@ -0,0 +1,24 @@
...
+then
+  echo -e "Use : $PROGNAME <Jenkins agent hostname> -- Ex : $PROGNAME ci-agent-1 \n "
+  exit 3
+fi
+
+if /usr/bin/curl -sk 'https://localhost:443/computer/api/json' \
```
surminus: Let's pair on this as I'm not quite sure how best to achieve this :)

alexmuller:
I think this approach is cool, I don't have a better one!

PR #5296, opened 2016-12-20
Add ImageMagick to CI agents

suzannehamilton:
ImageMagick is used by some projects' integration tests, e.g. whitehall, so it needs to be installed on Jenkins as part of the testing tools.


suzannehamilton:
@surminus Thanks! I've added a doc comment based on my understanding of the class. Do you know if there's anything else to add?

PR #5297, opened 2016-12-20
Remove old service manual indexing

tijmenb:
The service manual is now completely managed in service-manual-publisher. This means we can get rid of the special way it used to index the content.

This commit removes the `service_manual_rebuild_search_index` jenkins task.

https://trello.com/c/w9RIW7qz/539-remove-content-from-the-old-manual-from-search-index


deanwilson:
Puppet code looks fine.

PR #5302, opened 2016-12-21
Add new mount point for Jenkins data

surminus:
This assigns a dedicated disk to /var/lib/jenkins for Jenkins builds. Original disk size will be 128GB.


surminus:
I've added the disks and checked they're /dev/sdb!

PR #5303, opened 2016-12-21
Add user "Sam Cook"

SamLR:


modules/users/manifests/samcook.pp
```diff
@@ -0,0 +1,8 @@
+# Creates the deanwilson user
```
surminus: Incorrect description

PR #5308, opened 2016-12-21
Validate DNS YAML

SamLR:
Add new repo (govuk-dns-config) to CI. This will run validation checks against the YAML DNS file.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -570,7 +570,7 @@ govuk_ci::master::pipeline_jobs:
   backdrop-transactions-explorer-collector: {}
   cdn-acceptance-tests: {}
   deployment:
-    source: 'git'
+    source: 'github-enterprise'
   deprecated_columns: {}
   event-store: {}
   gapy: {}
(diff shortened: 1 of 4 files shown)


hieradata/common.yaml
```diff
@@ -599,6 +599,7 @@ govuk_ci::master::pipeline_jobs:
   govuk_navigation_helpers: {}
   govuk-provisioning:
     source: 'git'
+  govuk-puppet: {}
```
surminus: This was added manually because this list is also used by the "Deploy App" Jenkins job, and we don't want govuk-puppet to be in that job. It will just fail, but we need to be wary of where this list is used elsewhere.

hieradata/common.yaml
```diff
@@ -583,6 +583,8 @@ govuk_ci::master::pipeline_jobs:
   govspeak: {}
   govuk-app-deployment: {}
   govuk-diff-pages: {}
+  govuk-dns-config:
```
surminus: As other comment, this will now also appear in the Deploy App job, and I'm not sure if we want it there. It might be worth writing in some code to remove it from this job somehow? I'm not sure how much it matters.

SamLR:
@surminus I've removed the commit that added the 'govuk-puppet' to deploy-app and instead extended the commit explaining what is happening in the manifest. I've used the same pattern to add govuk-dns-config as well (rather than including it in the yaml).

PR #5310, opened 2016-12-22
Jenkins s3 plugin

djsd123:
Allows for storing, publishing and pulling artefacts backed by s3.

reference: https://wiki.jenkins-ci.org/display/JENKINS/Jackson2+API+Plugin


surminus:
As discussed we should check if we want this plugin to go in before we start work on artefacts and deployments.

PR #5311, opened 2016-12-22
Remove additional Stage in Jenkins library method

whoojemaflip:
It seems like a convention is emerging that the Jenkinsfile in an
individual project is responsible for defining the pipeline stages.

Some projects that use the `deployIntegration()` library method are showing multiple 'deploy to
integration' stages, as this defined a stage internally in addition to the stage that is defined in the project's own Jenkinsfile.

Not a big deal if people don't want this change to go through, but it
will improve the pipeline view for a few projects.

[Example](https://ci.integration.publishing.service.gov.uk/job/local-links-manager/job/master/)

screen shot 2016-12-22 at 12 27 28

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -163,10 +163,7 @@ def pushTag(String repository, String branch, String tag) {
 def deployIntegration(String repository, String branch, String tag, String deployTask) {
 
   if (branch == 'master') {
-    // Deploy on Integration
-    stage('Deploy on Integration') {
-      build job: 'integration-app-deploy', parameters: [string(name: 'TARGET_APPLICATION', value: repository), string(name: 'TAG', value: tag), string(name: 'DEPLOY_TASK', value: deployTask)]
-    }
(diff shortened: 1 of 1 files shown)


surminus:
This change might leave some projects without an explicit stage (i.e. the ones that were originally written knowing that the Staging convention was built into the groovy function). 

I don't mind this change but we shouldn't merge before all Jenkinsfiles are consistent. Personally I'm happy for the Staging to be built into the function itself.

surminus:
Having checked I think it _would_ be more consistent to keep the functions in the groovy scripts but the defined stages in the Jenkinsfile so I would be happy for this to be merged, however we just need to ensure that projects move over to this format before merging this.

surminus:
The following would need to be updated:

https://github.com/alphagov/authenticating-proxy/blob/master/Jenkinsfile#L23
https://github.com/alphagov/router-api/blob/eeb92fe0e87207e4a26e04fc352b7de62f5c3bc7/Jenkinsfile#L22
https://github.com/alphagov/frontend/blob/ec1225b2f0c6cf8bf5912643e101f29b48e6b6e1/Jenkinsfile#L48 (maybe?)
https://github.com/alphagov/publishing-api/blob/25db0c0010ab7ce49e20cc594763c7a5514b1a27/Jenkinsfile#L39
https://github.com/alphagov/content-store/blob/a9cdbbfd1c772f6c2460296deda9d1e0d5ff8c3c/Jenkinsfile#L33

surminus:
I was feeling charitable so made PRs for everyone. 🎁

PR #5312, opened 2016-12-22
[WIP] Xenial puppet fixes

deanwilson:


deanwilson:
Most of these changes are in https://github.com/alphagov/govuk-puppet/pull/5389 - the two that are not, and cause this PR to fail tests, require a lot more work than is in this request.

PR #5313, opened 2016-12-22
Strict vars test fix

deanwilson:


surminus:
Sadly node specs are now failing because they assign themselves a node class:

```
govuk::node::s_performance_mongo should compile
     Failure/Error: expect { subject.call }.not_to raise_error
       expected no Exception, got #<Puppet::Error: Cannot reassign variable govuk_node_class at line 2 on node performance-mongo-1.example.com
```

deanwilson:
This should be redone as separate requests.

PR #5316, opened 2016-12-28
Safely restart jenkins service

surminus:
When we manage files on disk, we have to restart the Jenkins service if we make a change. Doing a service refresh does not take into account if any builds may be happening and will cause them to fail.

This commit adds a class which pulls the jenkins-cli.jar and issues a safe restart. Use this class to notify rather than the service itself.

### DIFF
--- a/modules/govuk_jenkins/manifests/safe_restart.pp
+++ b/modules/govuk_jenkins/manifests/safe_restart.pp
@@ -1,0 +1,33 @@
+# == Class: Govuk_jenkins::Safe_restart
+#
+# This class will issue a "safe restart" of Jenkins, which means it will only
+# restart once all jobs have finished running.
+#
+# Note: this requires having a user enabled that has the appropriate public SSH
+# key assigned to it. This should be added manually via the UI in advance
+# otherwise the command will fail.
+#
+# === Parameters:
+#
+# [*jenkins_home*]
+#   The home directory of the Jenkins install, and where to put the Jenkins CLI
+#   jar.
+#
+class govuk_jenkins::safe_restart (
+  $jenkins_home = '/var/lib/jenkins',
+) {
+  require ::govuk_jenkins
+
+  curl::fetch { 'Jenkins CLI tool':
+    source      => 'http://localhost:8080/nlpJars/jenkins-cli.jar',
+    destination => "${jenkins_home}/jenkins-cli.jar",
+    timeout     => 0,
+    verbose     => false,
+  }
+
+  exec { 'Restart Jenkins':
+    command     => "/usr/bin/java -jar ${jenkins_home}/jenkins-cli.jar -s http://localhost:8080/ -i ${jenkins_home}/.ssh/id_rsa safe-restart",
+    require     => Curl::Fetch['Jenkins CLI tool'],
+    refreshonly => true,
+  }
+}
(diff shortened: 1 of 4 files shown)


modules/govuk_jenkins/manifests/safe_restart.pp
```diff
@@ -0,0 +1,38 @@
...
+# [*jenkins_user*]
+#   Which user to run the jar as.
+#
+class govuk_jenkins::safe_restart (
+  $jenkins_home = $::govuk_jenkins::jenkins_homedir,
+  $jenkins_user = $::govuk_jenkins::jenkins_user,
+) {
+  require ::govuk_jenkins
+
+  exec { 'Jenkins CLI tool':
```
rjw1: We could use [puppet-curl](https://github.com/alphagov/puppet-curl) here

alexmuller:
Wow this is weird/cool.

modules/govuk_jenkins/manifests/safe_restart.pp
```diff
@@ -0,0 +1,38 @@
...
+    command => "/usr/bin/wget http://localhost:8080/nlpJars/jenkins-cli.jar -O ${jenkins_home}/jenkins-cli.jar",
+    user    => $jenkins_user,
+    group   => $jenkins_user,
+    unless  => "/bin/ls ${jenkins_home}/jenkins-cli.jar 2>/dev/null",
+  }
+
+  exec { 'Restart Jenkins':
+    command => "/usr/bin/java -jar ${jenkins_home}/jenkins-cli.jar -s http://localhost:8080/ -i ${jenkins_home}/.ssh/id_rsa",
```
surminus: I just realised this command is wrong

modules/govuk_jenkins/manifests/safe_restart.pp
surminus: I think that would be nice, but it doesn't give the ability to specify the user/group that the file gets set as. I'm wondering if it needs to be run as Jenkins user at all, actually.

surminus:
I've changed this a little to use puppet-curl, and to just instigate the task as root.

alexmuller:
Does the `exec` resource need a `refreshonly => true` set? And then `notify => Exec['Restart Jenkins']` or something like that? Otherwise I'm a little worried the command might execute every Puppet run.

surminus:
The `notify` in the File resource should only call when the file changes. I call the `Class` instead of an `Exec` to ensure that the jenkins-cli.jar exists. We could order this in a different way, perhaps with a require on the exec itself?

alexmuller:
But I guess the `exec` will still execute every time because that's how it behaves, right? Unless it has an `onlyif`, `unless` or `refreshonly` or something like this.

djsd123:
@surminus For reference https://docs.puppet.com/puppet/3.8/type.html#exec-attribute-refreshonly

surminus:
So does `notify => Service['jenkins'],` behave differently to `notify => Class['Govuk_jenkins::Safe_restart'],`? 

That's what I don't understand, as I thought the notify would only call the class if the content changed on the File resource.

surminus:
I've added `refreshonly => true` anyway as it doesn't hurt.

surminus:
Thanks for the patience everyone (and for @deanwilson explaining), I misunderstood and got thrown by how the `Service` resource worked.

surminus:
@deanwilson updated so that it hardcodes the home directory.

PR #5320, opened 2016-12-29
Add ruby 2.4.0 as an available rbenv ruby

rjw1:


rjw1:
needs https://github.com/alphagov/packager/pull/127 to be merged and the resulting built package to put in our apt repo

PR #5331, opened 2017-01-05
[DO NOT MERGE] Prevent Varnish from removing an example A/B testing cookie

suzannehamilton:
Previously, all cookies were stripped from requests and responses, to prevent applications (or their dependencies) adding cookies that users had not been informed about, with the exception of the licence finder.

In order to A/B test changes to the navigation, we will need to add a cookie to allow us to ensure a user always sees the same version of the page.

This commit allows an example A/B testing cookie named 'abtest-example' to pass through Varnish. This will be used to develop A/B testing on Integration. Once that is done, it will be replaced with the name of the actual A/B testing cookie, or a configured whitelist of cookies.

The cookie filtering uses a slightly convoluted regex substitution process, because (AFAIK) this is the only option available in VCL. I've based it on the examples in the VCL wiki: https://www.varnish-cache.org/trac/wiki/VCLExampleRemovingSomeCookies

@alexmuller - you might want to take a look at this.

https://trello.com/c/QyDbSiR9/289-spike-build-a-whitelist-of-cookie-names-so-that-we-can-selectively-allow-cookies-to-be-returned-to-each-client

### DIFF
--- a/modules/varnish/templates/default.vcl.erb
+++ b/modules/varnish/templates/default.vcl.erb
@@ -79,9 +79,38 @@ sub vcl_recv {
   # Strip cookies from inbound requests. Corresponding rule in vcl_fetch{}
   # For simplicity and security most applications should not use cookies.
   # With the exception of:
-  #   - Licensing
+  #   - Licensing application
+  #   - AB Testing cookies in any application
   if (req.url !~ "^/apply-for-a-licence") {
-    unset req.http.Cookie;
+    if (req.http.Cookie) {
+      # Strip out any cookies which are not whitelisted using in-place regex
+      # substitution, since it is the only process available in VCL.
+
+      # Prepend a semicolon to make the first cookie consistent with subsequent
+      # cookies, which makes it easier to search for whitelisted cookies
+      set req.http.Cookie = ";" + req.http.Cookie;
+
+      # Remove any whitespace, so that whitespace can be added as a maker in
+      # later steps
+      set req.http.Cookie = regsuball(req.http.Cookie, "\s", "");
+
+      # Prepend a space to any whitelisted cookies (which are matched case-
+      # insensitively) as a marker to use in later steps
+      # TODO: Read from a cookie whitelist in a config file, rather than hard-coding allowed cookies
+      set req.http.Cookie = regsuball(req.http.Cookie, "(?i);abtest-example=", "; abtest-example=");
+
+      # Delete all cookies which do not have the space marker, i.e. those which
+      # are not whitelisted
+      set req.http.Cookie = regsuball(req.http.Cookie, ";[^ ]+", ";");
+
+      # Clean up the extraneous semicolons and whitespace which were added
+      # earlier or are left-over from deleting non-whitelisted cookies
+      set req.http.Cookie = regsuball(req.http.Cookie, "^;|\s|;$", "");
+
+      if (req.http.Cookie == "") {
+        unset req.http.Cookie;
+      }
+    }
   }
   <% end %>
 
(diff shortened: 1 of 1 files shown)


modules/varnish/templates/default.vcl.erb
```diff
@@ -79,9 +79,21 @@ sub vcl_recv {
...
   # With the exception of:
-  #   - Licensing
+  #   - Licensing application
+  #   - AB Testing cookies in any application
   if (req.url !~ "^/apply-for-a-licence") {
-    unset req.http.Cookie;
+    if (req.http.Cookie) {
+      set req.http.Cookie = ";" + req.http.Cookie;
+      set req.http.Cookie = regsuball(req.http.Cookie, "\s", "");
+      # TODO: Read from a cookie whitelist in a config file, rather than hard-coding allowed cookies
```
suzannehamilton: I've left this TODO in here because I'd like to test the simple, hard-coded version on Integration before adding a whitelist. If that's not appropriate, I can take it out and add it to a checklist on Trello.

suzannehamilton:
One other thought - does anyone have any ideas about how to add regression tests for this, e.g. to Smokey? If there's an example of an app which tries to add a `Set-Cookie` header which gets stripped out by Varnish, we could test that this happens even under the new rules.

I wrote a lot of examples in Ruby when I was testing the regex replace rules, but I'm not sure how to turn some of those into regression tests.

alexmuller:
I'm taking a look at this now - one thing I'm a little unsure of is that it looks like Rails sets multiple `Set-Cookie` headers and I'm not sure how Varnish deals with that.

alexmuller:
It looks like Varnish can only deal with the first header of any given type, so we can only support one type of header per request - this is fine, but we should document it somewhere.

alexmuller:
For the `Set-Cookie` behaviour I've verified that this works as expected 👍 

I'm probably missing something really obvious here but would this simpler code not behave the same way?

```vcl
# Strip cookies (except A/B testing or licence cookies) from outbound
# requests. Corresponding rule in vcl_recv{}
if (req.url !~ "^/apply-for-a-licence") {
  if (beresp.http.Set-Cookie) {
    if (beresp.http.Set-Cookie !~ "(abtest-example|some-other-cookie)") {
      unset beresp.http.Set-Cookie;
    }
  }
}
```

alexmuller:
I think the `Cookie` behaviour is correct! This is what I think it does:

- Prepend semicolon
- Remove all whitespace
- I don't really understand what the `(?i)` "optional i" is for, but add a space
- Remove every semicolon which doesn't have a space after it
- Remove all semicolons or whitespace from the beginning and end?

suzannehamilton:
> one thing I'm a little unsure of is that it looks like Rails sets multiple `Set-Cookie` headers and I'm not sure how Varnish deals with that.

> It looks like Varnish can only deal with the first header of any given type, so we can only support one type of header per request

Ah, this was a hole in my understanding - I didn't realise that there were multiple `Set-Cookie` headers. It sounds like [this is the standard way to set multiple cookies](https://tools.ietf.org/html/rfc6265#section-3):

> Origin servers SHOULD NOT fold multiple Set-Cookie header fields into a single header field.

Given all of that, I'm now a bit worried that my `beresp` code isn't doing the right thing. If there is a mix of matching and non-matching cookies, I wonder whether these lines will either strip all of them out or leave some blank `Set-Cookie` headers:

```
if (beresp.http.Set-Cookie == "") {
  unset beresp.http.Set-Cookie;
}
```

Similarly, how would your suggested change handle multiple headers?

```
if (beresp.http.Set-Cookie !~ "(abtest-example|some-other-cookie)") {
  unset beresp.http.Set-Cookie;
}
```

If there's a mix of cookies, does it leave them _all_ in?

Do you know if there is some way for me to test this before it's merged to master?

suzannehamilton:
> I think the `Cookie` behaviour is correct! This is what I think it does:
> Prepend semicolon
Remove all whitespace
I don't really understand what the `(?i)` "optional i" is for, but add a space
Remove every semicolon which doesn't have a space after it
Remove all semicolons or whitespace from the beginning and end?

The `(?i)` makes the matching case-insensitive, since `ABTest-Example` and `abtest-example` should be equivalent.

I've added some extra comments to explain each step. Do you think it helps?

suzannehamilton:
I've marked this as DO NOT MERGE, since we might have to change our approach and configure cookies and headers in the Fastly config instead, following something like the A/B approach documented here: https://www.fastly.com/blog/best-practices-for-using-the-vary-header

suzannehamilton:
I'm closing this PR because we now think that updating the CDN Varnish config (alphagov/govuk-cdn-config#13) is the right approach.

PR #5337, opened 2017-01-09
Add statsd and locksmithctl repos

surminus:
These are two packages which originally we had on our PPA, but for our upgrades to Xenial we are packaging using fpm and hosting on our local mirror.


SamLR:
Looks fine to me

PR #5338, opened 2017-01-09
Add instance_name envvar to short_url_manager

h-lame:
This is used when sending emails to help indicate to the recipients which
deployed environment the email came from, and thus if they should act on
it swiftly or not.  We follow the pattern established by signon to do this.

For: https://trello.com/c/SAcBmzf3/85-include-indication-of-which-environment-sent-the-email-in-short-url-manager

Also see: https://github.com/alphagov/short-url-manager/pull/70 which uses this env var.


h-lame:
Merging for now.  I think @tijmenb might be interested in the discussion about `DEPLOY_ENV` or whatever we call it when we get round to it.

PR #5340, opened 2017-01-09
Add nightly job to validate published DNS records

SamLR:
We want to make sure that the published DNS records match those saved
in the govuk-dns-config repository. To do this every night we will run
the validate_dns rake task which will match the YAML file data to
what is currently available in production.

This commit adds a validate_published_dns task to Jenkins and
configures it to run every night between midnight and 0259.


surminus:
Looks good! Can we ensure there's a mechanism to alert us (by email?) if there's a problem?

SamLR:
@surminus I've added email alerting based on `check_cdn_ip_ranges` so we should get told if the job fails.

PR #5346, opened 2017-01-11
Add andrew hilton (@mcgoooo)

mcgoooo:
Hi there, 

working with @nickcolley on template consolidation.

Cheers


nickcolley:
Looks like your SSH key is not strong enough by the looks from the failure.

PR #5350, opened 2017-01-12
Add note to update packages

chao-xian:
When running `govuk_puppet` to update the development-vm, sometimes there are errors when installing system packages where these packages can't be found. It may simply require getting the latest package list.


surminus:
Nope, we didn't. You could maybe put an `echo` in https://github.com/alphagov/govuk-puppet/blob/master/tools/puppet-apply-dev for people who forget to read docs

chao-xian:
@surminus I've added an echo but this is like the second time in my life I've ever bashed so hope it's ok.

PR #5352, opened 2017-01-12
Add cleanup of puppet reports

tomnatt:
Remove logs after 1 day
Fixes undisciplined logging from 95f86d28a76c7b1b6a7ec8e5b2dc9ef89912e5c4


alexmuller:
The tests are failing but it's only lint:

```
modules/puppet/manifests/master.pp - WARNING: double quoted string containing no variables on line 86
```

(Puppet interpolates variables in double quoted strings, but not single quoted ones)

PR #5353, opened 2017-01-13
Override Jenkins branches built for schema test repos

suzannehamilton:
Ensure that 'deployed-to-production' is included in the Jenkins branch builds for the 'whitehall' and 'content-tagger' repos. This is not normally required, but it is necessary for these repos because the
'deployed-to-production' branch is used to test changes pushed to the govuk-content-schemas repo.

https://trello.com/c/S4DZbsPa/285-jenkins-2-migration


suzannehamilton:
Thanks! Once some other projects are ready, I'll be creating another PR or two along the same lines. I'll try to explain it more clearly next time!

PR #5356, opened 2017-01-16
Dynamically lookup trade tariff hostname

alexmuller:
It turns out that nginx looks up the hostname of a proxy_pass only when it starts up, which means that if nginx hasn't restarted for a while and Amazon changes the location of an ELB, we'll send requests to the wrong place.

We've seen this before in ef64dda2105e7a055248d9dd9a7151315fe8acfe.

### DIFF
--- a/modules/router/templates/assets_origin.conf.erb
+++ b/modules/router/templates/assets_origin.conf.erb
@@ -48,8 +48,9 @@ server {
 
   <%- if @trade_tariff_host -%>
   location /tariff/ {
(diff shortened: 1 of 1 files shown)


dcarley:
Good sleuthing of a peculiar problem 😸

mtekel:
Given the experience, I'd probably double check if you need to do this for other (all?) redirects as well. The cloud is very dynamic and many times DNS is the discovery mechanism, so you have to keep polling all the time...

PR #5362, opened 2017-01-17
Purge mirrorer rabbitmq purge daily

surminus:
We run govuk_crawler_worker to test the functionality, and it adds messages to rabbitmq but these are never processed because we do not have mirrors in Integration.

This adds a function that purges the queue everyday so avoid get alerts related to rabbitmq memory because the queue has too many messages.

### DIFF
--- a/modules/govuk/manifests/node/s_mirrorer.pp
+++ b/modules/govuk/manifests/node/s_mirrorer.pp
@@ -1,6 +1,36 @@
-# Node defintion for mirrorer-?.management boxen
-class govuk::node::s_mirrorer inherits govuk::node::s_base {
+# == Class: Govuk::Node::S_mirrorer
+#
+#  Node definition for mirrorer boxes
+#
+# === Parameters:
+#
+# [*daily_queue_purge*]
+#   This purges the rabbitmq queue of all messages. Use with caution!
+#
+class govuk::node::s_mirrorer (
+  $daily_queue_purge = false,
+) inherits govuk::node::s_base {
   include govuk_crawler
   include govuk_rabbitmq
   include nginx
+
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/node/s_mirrorer.pp
```diff
@@ -1,6 +1,34 @@
...
   include nginx
+
+  if $daily_queue_purge {
+    # Create a file that only root can read so not to expose the rabbitmq root password.
+    file { '/root/purge_govuk_crawler_queue.sh':
+      ensure  => present,
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0700',
+      content => "/usr/local/bin/rabbitmqadmin --username=root --password=${::govuk_rabbitmq::root_password} purge queue name=govuk_crawler_queue",
```
deanwilson: Should this require the `govuk_rabbitmq` class to ensure the class has been included in the catalog at this point?

surminus: Probably!

djsd123:
This does not stop the alert from occurring.  It just makes the alert go away for a short period of time.  If the queue is not being processed, then we should probably consider not having an alert for "high watermark" in this environment.  I favoured increasing the threshold as the rabbitmq runtime has established a baseline of 1.5GB 'resident memory' running on a VM with 4GB of RAM, i.e. 40%.
Alerting on false positives desensitises staff to that alert.  That said, I'm not fussed so I'll close the other PR https://github.com/alphagov/govuk-puppet/pull/5371.

surminus:
@djsd123 I think you're right. If the alert pops up again even once then we should raise the threshold.

PR #5363, opened 2017-01-18
Adjust time of auto postgresql backup

djsd123:
There has been some improvement since https://github.com/alphagov/govuk-puppet/pull/4914 was merged, in that backups are not failing every day.

Bumping time by 30mins later to avoid race with env_sync

screen shot 2017-01-18 at 09 41 23

### DIFF
--- a/modules/govuk_postgresql/manifests/backup.pp
+++ b/modules/govuk_postgresql/manifests/backup.pp
@@ -20,8 +20,8 @@
 #  A simple description of the check that is alerting.
 #
 class govuk_postgresql::backup (
-  $auto_postgresql_backup_hour = 7,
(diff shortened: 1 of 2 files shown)


modules/govuk_postgresql/manifests/backup.pp
```diff
@@ -20,8 +20,8 @@
 #  A simple description of the check that is alerting.
 #
 class govuk_postgresql::backup (
-  $auto_postgresql_backup_hour = 7,
-  $auto_postgresql_backup_minute = 30,
+  $auto_postgresql_backup_hour = 8,
+  $auto_postgresql_backup_minute = 00,
```
surminus: I do wonder if having a double `0` will cause problems with cron

PR #5369, opened 2017-01-19
[DISCUSS] Configure Jenkins not to merge 'master' in schema tests

suzannehamilton:
Many Jenkins builds call the `mergeMasterBranch()` helper near the start of the build. This alerts us (and GitHub) to branches which cannot be cleanly merged with master. It uses the `--no-commit` flag to just check whether a merge would be valid, without actually performing the merge.

This works fine in the mainstream use-case, which is to test dev branches which have yet to be merged. However, it doesn't do the right thing for content schema tests, where we are running builds on the `deployed-to-production` branch. Unlike a dev branch, this branch is _behind_ `master`. So the merge being tested in this case is a fast-forward merge. Git doesn't apply the `--no-commit` behaviour for fast-forward merges, so it actually performs the merge! This means that we end up building the `master` branch rather than `deployed-to-production`.

The consequence of this is that the content schema tests were not strict enough: they were unintentionally running against the `master` version of the downstream repos, not `deployed-to-production`, which is the branch that we really care about breaking.

This is my proposed fix, which prevents the fast-forward merge if the build is a schema test. It makes an assumption that all schema tests set a build parameter or environment variable called `IS_SCHEMA_TEST`, which is a bit nasty: if that assumption is incorrect, then the schema tests will silently use `master` rather than `deployed-to-production`.

An alternative would be to put this logic into the individual projects' Jenkinsfiles. This would reduce the coupling to the environment variable, but would mean modifying a lot of other projects. And we could forget to put that check into a new project, and again the build would silently use the wrong branch.

Is this solution good enough? Can anyone suggest a better one? @tijmenb @surminus 

Trello: https://trello.com/c/S4DZbsPa/285-jenkins-2-migration

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -16,10 +16,18 @@ def cleanupGit() {
  * Try to merge master into the current branch, and abort if it doesn't exit
  * cleanly (ie there are conflicts). This will be a noop if the current branch
  * is master.
+ *
+ * Only merges master if this is a regular development branch. The merge is
+ * skipped if this is a content schema test, because schema tests should be run
+ * against the exact branch specified, rather than fast-forwarded to master.
  */
 def mergeMasterBranch() {
-  echo 'Attempting merge of master branch'
-  sh('git merge --no-commit origin/master || git merge --abort')
+  if (isContentSchemaTest()) {
+    echo "This is a schema test, so building the '${env.BRANCH_NAME}' without merging master"
+  } else {
+    echo 'Attempting merge of master branch'
+    sh('git merge --no-commit origin/master || git merge --abort')
+  }
 }
 
 /**
@@ -52,6 +60,15 @@ def initializeParameters(Map<String, String> defaultBuildParams) {
 }
 
 /**
+ * Check if current build is a content schema test. This relies on the build
+ * setting an environment variable 'IS_SCHEMA_TEST' to 'true', which could be
+ * done by setting a build parameter.
+ */
+def isContentSchemaTest() {
+  return env.IS_SCHEMA_TEST == "true"
+}
+
+/**
  * Check whether the Jenkins build should be run for the current branch, either
  * because it is a regular branch build or because it is being run to test the
  * content schema.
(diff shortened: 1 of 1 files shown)


MatMoore:
Would passing `--no-ff` help in any way?

The environment variable seems like it might be a bit confusing but I'd be fine with that if we can't think of a simpler solution ¯\_(ツ)_/¯

suzannehamilton:
Yeah, I thought about using `--no-ff`: #5366. 🙂 

The problem is that `'git merge --no-commit --no-ff` will still pull all the changes from `master`, it just won't commit them. So you still end up with the wrong version of the schemas sitting in the tmp directory.

There's another solution I've been thinking about which would avoid checking schema environment variables:

- Get the current SHA with `git rev-parse name-of-current-branch`
- Get the merge base for the current branch and `master` using `git merge-base name-of-current-branch master`
- Compare them
  - If they're the same, it means that the merge would be a fast-forward or a no-op (if the current branch _is_ `master`). Do _not_ perform the merge - run the rest of the build on the current branch without merging `master`.
  - If they're different, it means that this is a regular dev branch. Perform the `git merge --no-commit origin/master || git merge --abort` step as before.

This would affect every build of a merged branch or a deployed branch. I think this would be a feature rather than a bug - if you kick off a CI build of a branch that's strictly behind `master`, surely that means that you're trying to build that specific branch without a fast-forward merge of `master`?

MatMoore:
👍 I like that solution and agree with your conclusion.

P.S. this PR made me keenly aware of how much I don't know about git 😱

cbaines:
> There's another solution I've been thinking about which would avoid checking schema environment variables:
> 
>     Get the current SHA with git rev-parse name-of-current-branch
>     Get the SHA of the merge base for the current branch and master using git merge-base name-of-current-branch master
>     Compare them
>         If they're the same, it means that the merge would be a fast-forward or a no-op (if the current branch is master). Do not perform the merge - run the rest of the build on the current branch without merging master.
>         If they're different, it means that this is a regular dev branch. Perform the git merge --no-commit origin/master || git merge --abort step as before.
> 
> This would affect every build of a merged branch or a deployed branch. I think this would be a feature rather than a bug - if you kick off a CI build of a branch that's strictly behind master, surely that means that you're trying to build that specific branch without a fast-forward merge of master?

If the desired behaviour is to identify if the current commit/branch is contained in master, it might be simpler to just check that, and not use the merge base, e.g. you could do something like the following, and the exit code will indicate if this is the case or not:

```
git rev-list HEAD | grep $(git rev-parse master)
```

suzannehamilton:
@MatMoore Thanks for the feedback. I'll raise a separate PR once I've tried it out in a real build.

@cbaines Yes, that's what we're trying to work out. I think your approach is more intuitive, so I'll give it a go. Thanks!

suzannehamilton:
@MatMoore I've opened #5396 for feedback on the other approach. I'll close this PR without merging.

PR #5371, opened 2017-01-19
Adjust high watermark threshold

djsd123:
The threshold appears to be set too low for the rabbitmq runtime.  This causes the Nagios checks to raise alerts in Icinga (false positives) for integration.

What the threshold should be set to, is debatable. We are currently using the default value of 40% of resident memory for the mirrorer vms.
See: https://www.rabbitmq.com/memory.html

Alternatively, we could increase the memory of the VM to same as in production.

Rabbitmq consistently occupies currently set threshold for the high watermark

Affects: https://github.gds/gds/opsmanual/pull/868

### DIFF
--- a/hieradata/class/mirrorer.yaml
+++ b/hieradata/class/mirrorer.yaml
@@ -19,3 +19,6 @@ mount:
     disk: '/dev/mapper/crawler-worker'
     govuk_lvm: 'worker'
     mountoptions: 'defaults'
+
+rabbitmq::config_variables:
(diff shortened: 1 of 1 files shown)


surminus:
I looked into this and discovered it's because messages in the queue are never processed (I assume this is mirrorer in Integration?) I created a PR to purge them every morning in Integration: https://github.com/alphagov/govuk-puppet/pull/5362

Try running the command in that PR on the machine and see if it helps memory consumption (but only if it's mirrorer in Integration!)

PR #5373, opened 2017-01-19
Rename signonotron2 to signon

boffbowsh:
We can now delete the duplicate jobs in CI and Deployment


alexmuller:
👍 

Could you also remove `vhost_aliases` from `modules/govuk/manifests/apps/signon.pp`?

boffbowsh:
Done

PR #5374, opened 2017-01-19
Adding stephenharker user.

stephenharker:


surminus:
"SSH key for stephenharker is only 2048 bits and must be stronger" 🏋️

PR #5380, opened 2017-01-20
Add blue ocean plugin

surminus:
We should add the plugin incase people want to try it. Enabling currently does mean that you are unable to configure jobs in the UI, so would need to use classic mode, but it specifically plays nicely with pipeline jobs, which are the majority of jobs used on CI.

https://jenkins.io/projects/blueocean/
https://wiki.jenkins-ci.org/display/JENKINS/Blue+Ocean+Plugin


surminus:
I get the feeling people aren't keen

surminus:
Thanks @djsd123 :)

PR #5388, opened 2017-01-24
Add further dependencies for Blue Ocean

surminus:
Adds additional dependencies for Blue Ocean that are required from 11c43492825776db9496aeed5e1e53583e665560.


surminus:
Enabling Blue Ocean has caused problems elsewhere and doesn't really give us any decent functionality, so going to remove and close this PR.

PR #5394, opened 2017-01-24
Add govuk beat

afda16:
- Add pcfens-filebeat to Puppetfile
- Add new govuk_beat Puppet module to configure Filebeat for Logstash:
  - Override the 'govuk_beat::filebeat_outputs' parameter in Hiera with the Logstash information for the environment
  - Add 'filebeat::prospectors' to Hiera to configure the files that we want to send to Logstash
  - Include the 'govuk_beat' class to the environment or node configuration

### DIFF
--- a/modules/filebeat/manifests/config.pp
+++ b/modules/filebeat/manifests/config.pp
@@ -1,0 +1,28 @@
+#
+class filebeat::config {
+
+  $filebeat_config = delete_undef_values({
+    'beat_name' => $::fqdn,
+    'output'    => $filebeat::outputs,
+  })
+
+  file {'/etc/filebeat/filebeat.yml':
+    ensure  => file,
+    content => template('filebeat/filebeat.yml.erb'),
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0644',
+    notify  => Service['filebeat'],
+    require => File['/etc/filebeat/conf.d'],
+  }
+
+  file {'/etc/filebeat/conf.d':
+    ensure  => directory,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0755',
+    recurse => $filebeat::purge_conf_dir,
+    purge   => $filebeat::purge_conf_dir,
+  }
+
+}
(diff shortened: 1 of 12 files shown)


alexmuller:
This looks good to me! I think in order to get it working we need to add a prospector and include this class on a box. What about doing `include govuk_beat` inside `govuk::node::s_jumpbox` and then adding something like this to hieradata?

```
filebeat::prospectors:
  syslog:
    paths:
      - '/var/log/syslog'
```

surminus:
Look forward to this!

afda16:
@deanwilson questioned the amount of dependencies that the Filebeat module is pulling:

puppetlabs/stdlib (>=4.6.0 <5.0.0) -> Already included in our Puppetfile
puppetlabs/apt (>=2.0.0 <3.0.0) -> This is used if we don't set manage_repo to false, it's already included in our Puppetfile
puppetlabs/powershell (>= 1.0.1 < 2.0.0), lwf/remote_file (>= 0.2.0 < 2.0.0) -> These are used to install the package on Windows platforms. 'powershell' is supported by Puppetlabs, 'remote_file' is an approved module

Another simpler option is https://forge.puppet.com/ferrydg/filebeat, if you think it's better to use something with less dependencies I can have a look at that

surminus:
Yeah, it does seem a bit weird to pull the powershell module 😄 

The author of the alternative module also has one for [packetbeat](https://forge.puppet.com/ferrydg/packetbeat) which might be useful. 

Personally I wouldn't be opposed to writing our own simple set of classes if configuration is very minimal and default, but I know people have opinions on that!

afda16:
I'm having a look at https://forge.puppet.com/ferrydg/filebeat

afda16:
I updated the implementation:

- Add ferrydg-filebeat to Puppetfile: it needs ispavailability/file_concat
- Add new govuk_beat Puppet module to configure Filebeat for Logstash:
  - Override the 'govuk_beat::filebeat_outputs' parameter in Hiera with the Logstash information for the environment
  - Add 'govuk_beat::filebeat_prospectors' to Hiera to configure the files that we want to send to Logstash
  - Include the 'govuk_beat' class to the environment or node configuration

surminus:
This looks good, can we add default values in hiera for: 
  - `/var/log/*.log`
  - `/var/log/*/*.log`

modules/filebeat/manifests/config.pp
```diff
@@ -0,0 +1,28 @@
+#
+class filebeat::config {
+
+  $filebeat_config = delete_undef_values({
+    'beat_name'         => $::fqdn,
```
surminus: Lint might complain about the whitespace before the `=>`

afda16: Updated

SamLR:
LGTM

PR #5399, opened 2017-01-26
Add a feature flag to content-performance-manager

leenagupte:
Content Performance Manager has been deployed to staging and
integration, but it is not ready to be deployed to production.

The app has already been added to puppet though, and was throwing errors
in production.

Added a feature flag to fix the error, as per the docs:
https://github.com/alphagov/govuk-puppet/blob/master/docs/adding-a-new-app.md#feature-flags

### DIFF
--- a/modules/govuk/manifests/apps/content_performance_manager.pp
+++ b/modules/govuk/manifests/apps/content_performance_manager.pp
@@ -8,6 +8,10 @@
 #   The port that it is served on.
 #   Default: 3206
 #
+# [*enabled*]
+#   Whether the application should be enabled. Can be specified for each
+#   environment using deployment hieradata.
+#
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/apps/content_performance_manager.pp
```diff
@@ -40,6 +40,7 @@
 #
 class govuk::apps::content_performance_manager(
   $port = '3206',
+  $enabled = false,
```
danielroseman: The new parameter needs to be added to the documentation comments above.

leenagupte: Thanks @danielroseman. I've added the documentation and fixed up the change.

PR #5403, opened 2017-01-26
Remove smokey for performance platform

tijmenb:
This job has  been failing for almost 18 months, 4 times a day. It has never run successfully:

https://deploy.publishing.service.gov.uk/job/Smokey_Performance_Platform/1/

The long failure without somebody caring or noticing is evidence enough for me that the tests wouldn't add value. This PR is easily reverted if it turns out we need to support the tests again.

Slightly related: https://trello.com/c/WIunrfiT

After merging this:

- [ ] Remove the tests manually from the UI
- [ ] Remove the signon username/password from `gds/deployment`


alexmuller:
Correct decision 👍

PR #5405, opened 2017-01-26
[Don't merge] Clear up smoke test configuration

tijmenb:
While trying to upgrade smokey to capybara-webkit (https://trello.com/c/WIunrfiT) I was confused a lot. One of the things that bit me were that I didn't quite understand the role of all the Jenkins jobs.

This renames the puppet configuration and the jenkins jobs to be more descriptive.

After this is reviewed, I'll add a PR that adds the new hieradata in gds/deployment. After that this can be merged.

Best reviewed per-commit.


alexmuller:
I'd rather not change the `name` value because as you noted it will change the URL and break links, but more than happy for the `display-name` to change. But I don't feel that strongly about it :)

tijmenb:
@alexmuller Yeah, I went back and forth on that. I think breaking of the links is probably worth it in the long run, as the URLs will be more descriptive too. The downstream jobs are also more self-documenting (https://github.com/alphagov/govuk-puppet/pull/5405/commits/99f463cd64ec9618fe43310880377d7339f0fa8a). I've searched and haven't found any links to it, if that's worth anything.

PR #5409, opened 2017-01-27
Defining node labels for Jenkins agents

djsd123:
This is useful when investigation requires isolation. Relates to: https://trello.com/c/XazKRXGm


surminus:
Adding the labels as an array won't currently work in the way this was implemented, and would require a little refactoring in the [manifest](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/manifests/ssh_slave.pp) and [template](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/templates/ssh_slave_config.xml.erb).

djsd123:
changed to space separated string

PR #5413, opened 2017-01-30
Correctly parse graphite minimum thresholds

SamLR:
Graphite can be configured to treat a number as a minimum by
prepending the value with a '@', these values should still be
displayed as a line in charts. This changes the selector regex to
allow strings to start with a '@' whilst still correctly setting
the value of the constantLine.

### DIFF
--- a/modules/icinga/manifests/check/graphite.pp
+++ b/modules/icinga/manifests/check/graphite.pp
@@ -88,12 +88,14 @@ define icinga::check::graphite(
   $graph_height = 600
 
   # Only display lines when threshold is an integer or float.
+  # The '@' symbol indicates to icinga that the threshold is a minimum so we
+  # allow that and capture the numbers for display in graphite.
   $warn_line = $warning ? {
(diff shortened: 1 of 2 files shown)


modules/icinga/manifests/check/graphite.pp
```diff
@@ -88,12 +88,14 @@
   $graph_height = 600
 
   # Only display lines when threshold is an integer or float.
+  # The '@' symbol indicates to icinga that the threshold is a minimum so we
+  # allow that and capture the numbers for display in graphite.
   $warn_line = $warning ? {
-    /^[0-9.]+$/ => "&target=alias(dashed(constantLine(${warning})),%22warning%22)",
+    /^@?([0-9.]+)$/ =>  "&target=alias(dashed(constantLine(${1})),%22warning%22)",
```
rjw1: extra white space !

SamLR: Fixed

PR #5421, opened 2017-01-31
Remove Tariff databases from CI

alexmuller:
We don't run the tests for these apps anymore.


alexmuller:
Rebased!

alexmuller:
Rebased so it actually works this time.

PR #5425, opened 2017-02-01
Add PP data sync job

surminus:
Every couple of months we're asked to set up a data sync for Performance Platform. The way this happens is by editing and submitting PRs on different repos and/or by editing the Jenkins job that syncs *all* the data on GOVUK to Integration.

This is not an ideal way of doing things. Instead, create a separate job that can be run at anytime to sync data and give power back to the developers who want to do the work.

This is just a slimmed down version of the Copy_Data_to_Integration job, with only a single option for what to sync.

/cc @leelongmore


surminus:
Just realised I forgot to add the job to hiera. It'll only be present in Production.

PR #5426, opened 2017-02-01
[WIP] Update name and ssh key

binaryberry:
I have a new laptop therefore a new ssh key, and I got married so my name has
changed. I'm taking this opportunity to update my name and email. Happy to keep
my maiden name and email if there's a chance the new data will cause issues.

### DIFF
--- a/modules/users/manifests/tatianasoukiassian.pp
+++ b/modules/users/manifests/tatianasoukiassian.pp
@@ -1,8 +1,8 @@
 # Creates the tatianasoukiassian user
 class users::tatianasoukiassian {
   govuk_user { 'tatianasoukiassian':
-    fullname => 'Tatiana Soukiassian',
-    email    => 'tatiana.soukiassian@digital.cabinet-office.gov.uk',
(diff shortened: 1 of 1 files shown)


modules/users/manifests/tatianasoukiassian.pp
```diff
@@ -1,8 +1,8 @@
 # Creates the tatianasoukiassian user
 class users::tatianasoukiassian {
```
surminus: This also needs to be updated.

binaryberry: ok! My LDAP is still tatianasoukiassian. Might that cause an issue?

cbaines:
I've added WIP to the title as I believe this PR will be dormant for a little while, and this will get the Seal off my back.

deanwilson:
Can we bring this one back to life or close it off as it's been around for a while. The tests fail as the name isn't consistently changed everywhere, and if we do change it we might have a little discovery as to what breaks. Which might be worth doing incase someone else gets married. Hopefully it's just changing the local machines ssh config to use the new name.

surminus:
I'm going to close this PR, please feel free to raise a new one if you wish to update your name.

PR #5430, opened 2017-02-02
Add env vars for search-admin

rubenarakelyan:
search-admin is moving to the new deployment pipeline.

Trello: https://trello.com/c/tjUur3ec/462-move-search-admin-to-new-deployment-pipeline

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -462,6 +462,11 @@ govuk::apps::rummager::rabbitmq::enable_publishing_listener: true
 govuk::apps::rummager::redis_host: 'api-redis-1.api'
 govuk::apps::rummager::redis_port: '6379'
 
+govuk::apps::search_admin::db_name: 'search_admin_production'
+govuk::apps::search_admin::db_hostname: 'master.mysql'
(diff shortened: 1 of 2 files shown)


hieradata/common.yaml
```diff
@@ -458,6 +458,11 @@ govuk::apps::rummager::rabbitmq::enable_publishing_listener: true
 govuk::apps::rummager::redis_host: 'api-redis-1.api'
 govuk::apps::rummager::redis_port: '6379'
 
+govuk::apps::search_admin::db_name: 'search_admin_production'
+govuk::apps::search_admin::db_hostname: 'master.mysql'
+govuk::apps::search_admin::db_password: "%{hiera('govuk::apps::search_admin::db::password')}"
```
MatMoore: I think this line isn't necessary because the hiera call has the same name as the hiera value you're setting. So I'd expect this to be either shadowed entirely by the encrypted hieradata, or referring to itself?

rubenarakelyan: Yes good catch. This is actually coming from the encrypted hieradata with the same name so we don't need to have it here as well.

rubenarakelyan: The names are different (`db_password` vs `db::password`) :)

PR #5433, opened 2017-02-02
Explicitly set the `value` of an govuk_envvar to undef

deanwilson:
We currently default to undef so we might as well be explicit.
This also allows tests to pass under `strict`


surminus:
I think this is fine, but where do we default to undef already?

deanwilson:
I think you're right in that it'll end up being an empty string on puppet 3 and a failure on puppet 4 if nothing is provided.

PR #5441, opened 2017-02-06
[Don't merge] Revert removal of homepage no-cache rules

mobaig:
Revert "Revert no-cache rule for front-end homepage"
This reverts commit 9213f54c855b753931239a3f0fa87b237bb65db8.

We're running the Javascript detection experiment again. This reinstates
no caching rules for the tracking images.

Merge required prior to https://github.com/alphagov/frontend/pull/1115.

### DIFF
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -115,6 +115,7 @@ govuk::apps::email_alert_api::enable_procfile_worker: false
 govuk::apps::email_alert_service::rabbitmq_hosts: ['localhost']
 govuk::apps::event_store::enabled: true
 govuk::apps::event_store::mongodb_servers: ['localhost']
+govuk::apps::frontend::enable_homepage_nocache_location: false
 govuk::apps::hmrc_manuals_api::allow_unknown_hmrc_manual_slugs: true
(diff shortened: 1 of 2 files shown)


mobaig:
@alexmuller does this look ok to you? The code I'm reverting is a bit old - will this change have the intended effect of ensuring the tracking images aren't cached?

alexmuller:
What's the reason for disabling this in development? Can we remove all of that logic?

tijmenb:
I'm probably not understanding this, but what the reason for not caching this? All calls to the cache will be reported in Fastly anyway.

mobaig:
@tijmenb I believe this refers to the expires header set on responses to requests for the images - if you're still within the expiry time the browser won't request the image at all.

mobaig:
@alexmuller regarding the dev-specific config - not sure. I would assume it could be removed without issue. 

Original commit message https://github.com/alphagov/govuk-puppet/commit/e10bc998cc9229308c321f00 doesn't clear things up much. I also have no idea what `govuk_platform` is.

mobaig:
Closing for now, see: https://trello.com/c/ee87DdP0/350-determine-how-many-people-have-don-t-have-javascript#comment-58a2dfb0b53567cacc648184

PR #5445, opened 2017-02-07
Link inode alerts to the correct documentation

SamLR:
Low inode availability is directly linkable in the opsmanual so we
should link to it directly.


deanwilson:
Relates to docs added in internal/gds/opsmanual/pull/880

PR #5446, opened 2017-02-07
Set secret key base and errbit api key environment variables for info frontend

cbaines:


cbaines:
This has built successfully on CI, but GitHub said no when CI tried to set the status (rate limit exceeded).

PR #5447, opened 2017-02-07
Add frontend to draft-frontend

davidbasalla:
As part of migrating frontend to use the content store and the draft content store we need to deploy Frontend to draft-frontend as well.

https://trello.com/c/TOdNOEDQ/627-migrate-help-page-use-draft-stack-for-preview-2-3-3

### DIFF
--- a/modules/govuk/manifests/apps/frontend.pp
+++ b/modules/govuk/manifests/apps/frontend.pp
@@ -4,6 +4,10 @@
 #
 # === Parameters
 #
+# [*vhost*]
+#   Virtual host used by the application.
+#   Default: 'frontend'
+#
 # [*port*]
 #   The port that the app is served on.
 #   Default: 3005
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/apps/frontend.pp
```diff
@@ -33,14 +38,15 @@
   govuk::app { 'frontend':
     app_type               => 'rack',
     port                   => $port,
-    vhost_aliases          => ['private-frontend'],
+    vhost_aliases          => ['private-frontend', 'draft-frontend'],
```
surminus: is this alias required if the default vhost is going to be set on the draft-frontend machines?

brenetic: We thought that it was required for routing of requests for draft-frontend? If that's not the case we can remove it.

brenetic:
@surminus Thanks for the comments. Have made those changes and rebased. Tested this on Integration and it seems to do what we need

PR #5458, opened 2017-02-08
Remove hstore extension from email-alert-api's db definition

jennyd:
We've finished migrating the app to use json columns instead of hstore to
store the main matching criteria for email subscription lists.

This change isn't enough to actually disable the extension in the database (we
don't support `ensure => absent` in our puppet module for this) so we'll devops
it in each environment (`DROP EXTENSION hstore;`).

We've done similar things before: 3b8fc618febac824edf24aae719d5c2709ce8434

Part of https://trello.com/c/a2KUeoSD/69-finish-or-revert-email-alert-api-migration


jennyd:
I've dropped this extension in all environments now.

PR #5461, opened 2017-02-09
Monitor api_mongo disk time over 7 minute windows

boffbowsh:
api-mongo-2 (currently, may change?) does large backup jobs every 5
minutes, which show as large spikes on disk time and frequently trigger
our alerting. Taking the 50th percentile over a 7 minute window should
smooth the spikes out enough to not trigger the alert, but still warn
us if there’s a more consistent problem.

Before:
graphite publishing service gov

After:
graphite publishing service gov-1

### DIFF
--- a/hieradata/class/api_mongo.yaml
+++ b/hieradata/class/api_mongo.yaml
@@ -37,3 +37,5 @@ mongodb::server::replicaset_members:
   'api-mongo-4':
     hidden: true
     priority: 0
+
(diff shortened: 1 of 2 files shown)


SamLR:
Sounds good to me although I think we should probably investigate why the IO time seems to have been creeping up over the last few months (see below)
screen shot 2017-02-09 at 11 06 32

boffbowsh:
Good idea. Changed

PR #5467, opened 2017-02-09
Remove unicode character from description

surminus:
Jenkins Job builder doesn't like the unicode:

`UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 295: ordinal not in range(128)`

/cc @36degrees


36degrees:
Aah, sorry about that! Thanks for fixing 👍

PR #5470, opened 2017-02-09
Replicate mirrors to S3

SamLR:
We want to serve mirrors from S3 rather than from VMs.

This commit adds S3 as a target for the cron job that copies to our
mirrors. To achieve this the AWS secret & access keys are set by
`govuk_setenv` and the copy itself done using the AWS command line
tool.

DO NOT DEPLOY WITHOUT https://github.com/alphagov/govuk-terraform-provisioning/pull/94

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -41,6 +41,9 @@ govuk::deploy::setup::ssh_keys:
 govuk_cdnlogs::monitoring_enabled: false
 
 govuk_crawler::seed_enable: true
+govuk_crawler::sync_enable: true
+govuk_crawler::targets:
+  - 's3://govuk-mirror-integration'
 
 govuk_elasticsearch::dump::run_es_dump_hour: '9'
 
--- a/modules/govuk_crawler/manifests/config.pp
+++ b/modules/govuk_crawler/manifests/config.pp
@@ -1,0 +1,40 @@
+# == Class: govuk_crawler::config
+#
+# Class for configuring the AWS keys used to upload crawler results to S3.
+#
+# [*aws_access_key*] access key for AWS
+# [*aws_secret_key*] secret for chosen AWS key
+#
+# Will be set using environment variables, see:
+# http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-environment
+#
+
+class govuk_crawler::config (
+  $aws_access_key = undef,
+  $aws_secret_key = undef,
+) {
+  $env_name = 's3_sync_mirror'
+
+  file { ["/etc/govuk/${env_name}", "/etc/govuk/${env_name}/env.d"]:
+    ensure  => 'directory',
+    purge   => true,
+    recurse => true,
+    force   => true,
+    mode    => '0755',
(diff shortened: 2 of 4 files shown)


modules/govuk_crawler/manifests/init.pp
```diff
@@ -79,6 +79,8 @@
   $sync_enable = false,
   $targets = [],
 ) {
+  include govuk_crawler::config
```
afda16: Should this require the awscli package?

modules/govuk_crawler/manifests/config.pp
```diff
@@ -0,0 +1,31 @@
...
+  $env_name = 's3_sync_mirror'
+
+  file { ["/etc/govuk/${env_name}", "/etc/govuk/${env_name}/env.d"]:
+    ensure  => present,
+    purge   => true,
+    recurse => true,
+    force   => true,
+  }
+
+  file { "/etc/govuk/${env_name}/env.d/AWS_SECRET_ACCESS_KEY":
```
afda16: We are storing secrets in these files, maybe it needs to enforce the file mode and ownership

modules/govuk_crawler/manifests/config.pp
```diff
@@ -0,0 +1,36 @@
...
+#
+
+class govuk_crawler::config (
+  $aws_access_key = undef,
+  $aws_secret_key = undef,
+) {
+  $env_name = 's3_sync_mirror'
+
+  file { ["/etc/govuk/${env_name}", "/etc/govuk/${env_name}/env.d"]:
+    ensure  => present,
```
deanwilson: Should these be `ensure => 'directory'` ?

modules/govuk_crawler/manifests/config.pp
```diff
@@ -0,0 +1,36 @@
...
+    purge   => true,
+    recurse => true,
+    force   => true,
```
deanwilson: You'll need a `mode` here as the `File` underneath will set a default that might get picked up.

modules/govuk_crawler/manifests/config.pp
```diff
@@ -0,0 +1,37 @@
+#
```
surminus: Format the class docs to be consistent with other classes (i.e. add a `== Class: Govuk_crawler::Config` heading)

hieradata/integration.yaml
```diff
@@ -41,6 +41,8 @@ govuk::deploy::setup::ssh_keys:
 govuk_cdnlogs::monitoring_enabled: false
 
 govuk_crawler::seed_enable: true
+govuk_crawler::targets:
+  - 's3://govuk-mirror-integration'
```
surminus: We currently only [enable syncs in Production](https://github.com/alphagov/govuk-puppet/blob/master/hieradata/production.yaml#L165). I think it's worth testing this in Integration just to check the sync and bucket serving the site works, but we won't be able to test the full fastly failover, and it can be removed later.

SamLR: Done : )

PR #5471, opened 2017-02-09
Allow cross-origin requests for search & content-store APIs

tijmenb:
This allows cross origin requests for two of our public apis, rummager and content-store.

The reason I'd like to allow this is that this will enable us to experiment with API preview tools like Swagger. Preview tools like Swagger UI (http://petstore.swagger.io/ for example) are served from a different domain.

Since the APIs only respond to GET requests I think there is no security risk. But to be sure, I'd like the opinion of @alexmuller, @timblair, @boffbowsh or @tomnatt.

Precedent in this PR: https://github.com/alphagov/calendars/pull/114


tijmenb:
Great! Thanks. We'll keep an eye on that discussion.

PR #5473, opened 2017-02-10
Add Aptly rbenv-ruby repo for Xenial

afda16:
We need to build rbenv-ruby packages for Xenial and upload them to
Aptly. We are going to have the same package version built twice,
one for each distribution. There is an email thread that explains some
of the options to implement this:

https://groups.google.com/forum/#!topic/aptly-discuss/QhgkRlR577w

This commit adds a new repo for the Xenial distribution. We could still
publish the repo under the 'rbenv-ruby' prefix if we want to have
something like 'rbenv-ruby trusty' and 'rbenv-ruby xenial' in our sources
or use a different prefix if we want to keep package names.


surminus:
I personally think it'd be better to keep the single repo but with a different distribution as this would involve adding less Puppet code, as I think you can just add the packages to the repo in aptly?

PR #5476, opened 2017-02-10
Add Plek service uris for Draft Frontend

brenetic:
In the draft stack draft-frontend needs to make api calls to both Mapit and Local Links Manager neither of which are available with `draft` prepended to their uri.

We now set them directly based on the app_domain with and so Plek will return the necessary uri.


brenetic:
@tijmenb  I think this will stop you receiving those errbit emails?

PR #5478, opened 2017-02-13
Set correct database password for search-admin

rubenarakelyan:
This commit fixes the setting of the database password for search-admin in puppet, since https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/search_admin/db.pp#L8 is looking for `govuk::apps::search_admin::db::mysql_search_admin`.

Trello: https://trello.com/c/tjUur3ec/462-move-search-admin-to-new-deployment-pipeline

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -466,7 +466,7 @@ govuk::apps::rummager::redis_port: '6379'
 
 govuk::apps::search_admin::db_name: 'search_admin_production'
 govuk::apps::search_admin::db_hostname: 'master.mysql'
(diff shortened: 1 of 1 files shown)


leenagupte:
@rubenarakelyan Have you added a similar PR to the deployment repo?

rubenarakelyan:
I looked and turns out the encrypted hieradata is already using `mysql_search_admin`. But because this was getting set incorrectly the password was empty in the database URL.

leenagupte:
@rubenarakelyan Excellent!

PR #5436, opened 2017-02-03
Remove bits of EFG

alexmuller:

PR #5479, opened 2017-02-13
Remove bits of EFG (again)

alexmuller:
This contains all the commits from #5436 except for the VPN gateway, which can't be removed until the machines have been destroyed.


deanwilson:
Let's try this again.

surminus:
What's the status of EFG? When this is deployed what happens to EFG?

alexmuller:
> What's the status of EFG?

It's going away! @timblair is working with the British Business Bank to get rid of it and I'm just hanging around because I like deleting code :)

> When this is deployed what happens to EFG?

Nothing, it keeps running for a little bit longer on our infrastructure but at some point the new people will change DNS away from us and then we can delete _everything_

PR #5480, opened 2017-02-14
govuk offsite assets

djsd123:
Configures duplicity to backup assets to S3

https://trello.com/c/eGZsCDvm


surminus:
This looks good! See my comment in https://github.com/alphagov/govuk-terraform-provisioning/pull/96 - we could use the bucket and creds already available - I believe duplicity should correctly namespace inside the bucket.

PR #5481, opened 2017-02-14
Rename the module

deanwilson:
We currently use a local puppet module to manage puppetdb and in the future we'd like to both
remove this custom code and gain the benefits of using the upstream that supports newer versions of PuppetDB (that we would also eventually like to use.) For this to happen we need to namespace our own puppetdb code to avoid conflicts and incorrect loading of classes.

This is an early first step that will enable to Puppetfile the upstream and eventually start testing it.


surminus:
Could you explain _why_ "we will need to use the upstream puppetdb module from puppetlabs in the future" for those of us being deliberately naive?

deanwilson:
I've updated the description

PR #5482, opened 2017-02-14
[WIP] Remove apt mirror hostname

deanwilson:
We currently have an apt-mirror configured in our base hieraconfig and then alias that to a class specific value for every puppet class that uses it. This value is the same throughout our infrastructure so instead we'll create a static host entry and use that in all the classes.

This removes the need for all the aliases and class params.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -106,8 +106,6 @@ deployable_applications: &deployable_applications
   whitehall:
     branches_to_exclude: ['release', 'deployed-to-integration', 'deployed-to-staging']
 
-apt_mirror_hostname: 'apt.production.alphagov.co.uk'
-
 apt::apt_update_frequency: 'daily'
 apt::purge_preferences_d: true
 apt::purge_sources_list: true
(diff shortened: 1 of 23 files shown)


alexmuller:
I think this value won't work across environments, for example connecting to `10.3.0.75` from integration won't work.

surminus:
I'm wondering that if we're going through and  setting the hostname to `apt_mirror.cluster` we could just go through and hardcode this as `apt.publishing.service.gov.uk` which would achieve the same result by removing all the parameters. 

We would lose the benefit of being able to set it to point elsewhere for a different environment, but this is tricky since we want to ensure the apt repo is behind fastly.

deanwilson:
I think we should keep this as something that can be different per environment both to enable us to iterate on the current packaging pipelines and for future flexibility.

surminus:
I agree, I just can't think of a good way to do this, but I do want to remove all the `apt_mirror_hostname` everywhere. If only we could add an alias somewhere...

PR #5483, opened 2017-02-14
Enable filebeat

afda16:


hieradata/integration.yaml
```diff
@@ -31,6 +31,9 @@ govuk::apps::whitehall::basic_auth_credentials: "%{hiera('http_username')}:%{hie
 govuk::apps::whitehall::highlight_words_to_avoid: true
 govuk::apps::whitehall::enable_tagging_to_new_taxonomy: "yes"
 
+govuk_beat::hosts:
```
afda16: I'm not sure why we shouldn't just override govuk_beat::hosts in the credentials repo, does the logstash_host variable already exist?

hieradata/integration.yaml
```diff
@@ -31,6 +31,9 @@ govuk::apps::whitehall::basic_auth_credentials: "%{hiera('http_username')}:%{hie
...
+  - "%{hiera('logstash_host')}:5566"
```
afda16: I'm not sure why we shouldn't just override govuk_beat::hosts in the credentials repo, does the logstash_host variable already exist?

SamLR: @afda16 overriding it in the credentials repo is fine. This was me being bad at puppet

surminus:
Might be good to add `/var/log/syslog`, too?

afda16:
Added syslog

PR #5489, opened 2017-02-14
Ensure backup offsite jobs use newer gpg key

surminus:
When the GPG key was rotated in 7d4208b3b3 I only updated it in `production.yaml` and not `common.yaml`. This was misleading as the hiera that creates the job uses the key defined in the hiera, which means that offsite backups have been using the older key, although assets have been using the newer one.

This is problematic because we rotated the passphrase and secret key in our secret hiera, so if we wanted to unencrypt our non-assets offsite backups we'd need to delve back into the history of that repo to find the right key.

In the future we should refactor the way we include the offsite backup job hiera. The reason we have the data in both sets of hiera is because the defined type has required parameters, so the production ones should be the real ones, and the ones in common shouldn't matter.


surminus:
Realised I probably don't need to include graphite or cdn-logs, will amend.

PR #5491, opened 2017-02-15
ci-agent-6 on Xenial

djsd123:
We require an agent running on Xenial as Trusty is nearing EOS.
We've decided to defer provisioning elasticsearch, due to the complexity involved with jumping multiple major versions.

https://trello.com/c/VWChaAZu

### DIFF
--- a/modules/govuk_elasticsearch/manifests/init.pp
+++ b/modules/govuk_elasticsearch/manifests/init.pp
@@ -11,6 +11,9 @@
 #   The version of elasticsearch to install.  This must specify an exact
 #   version (eg 1.4.2)
 #
+# [*enabled*]
+#   Boolean.  Allows disabling of the this class at the node level.
+#
 # [*manage_repo*]
 #   Whether to configure an apt source for our mirror of the elasticsearch
 #   repo. Disable to install elasticsearch from a separately configured repo.
@@ -26,6 +29,7 @@
 #
 class govuk_elasticsearch (
   $version,
+  $enabled = true,
   $cluster_hosts = ['localhost'],
   $cluster_name = 'elasticsearch',
   $heap_size = '512m',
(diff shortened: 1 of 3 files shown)


surminus:
Trusty is only EOL in 2019, but we're upgrading to Xenial to be proactive. 

You'll also need to add the agent in here: https://github.com/alphagov/govuk-puppet/blob/master/hieradata/common.yaml#L743-L759

surminus:
I wonder if we need a label to say this is a Xenial test VM, and also something to ensure that jobs that require Elasticsearch don't run on it. 

Also, as agents have multiple packages that get installed, what are the version of MySQL and PostgreSQL that get installed? Might be worth having a label for them, too? or perhaps that should be inclusive of the 'xenial' label?

djsd123:
@surminus Addressed in https://github.com/alphagov/govuk-puppet/pull/5498

modules/govuk_elasticsearch/manifests/init.pp
```diff
@@ -41,107 +45,112 @@
   $backup_enabled = false
 ) {
 
-  validate_re($version, '^\d+\.\d+\.\d+$', 'govuk_elasticsearch::version must be in the form x.y.z')
-  include augeas
-  validate_bool($manage_repo)
+  if $enabled {
```
afda16: It looks like the 'enable' parameter is basically invalidating the whole class, isn't it? This is the same as not including the class in first place. I think it's better to configure a flag in govuk_ci::agent directly to not include this class if we don't want to

djsd123: Yep, pretty much.  With the absence of a "roles & profiles" design pattern we have to gate classes in this way.  Sure, I'll move te gate up.

PR #5494, opened 2017-02-15
Update conditional to install filebeat on Vagrant and Integration

afda16:
Use 'domain' fact instead of 'environment'

### DIFF
--- a/modules/govuk/manifests/node/s_base.pp
+++ b/modules/govuk/manifests/node/s_base.pp
@@ -15,7 +15,7 @@ class govuk::node::s_base (
   include backup::client
   include base
   # Remove conditional to install filebeat on other environments
-  if $::environment =~ /vagrant|integration/ {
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/node/s_base.pp
```diff
@@ -15,7 +15,7 @@
   include backup::client
   include base
   # Remove conditional to install filebeat on other environments
-  if $::environment =~ /vagrant|integration/ {
+  if $::domain =~ /^(dev|.*\.integration\.publishing\.service)\.gov\.uk/ {
```
surminus: I think this needs to be:
`^(.*\.dev|.*\.integration\.publishing\.service)\.gov\.uk`

Otherwise it won't match a dev domain such as "ci.dev.gov.uk" because the bit before "dev" isn't being matched.

afda16: :) I just saw that, updated, cheers

PR #5495, opened 2017-02-15
Reload configuration rather than restart

surminus:
In an attempt to stop interrupting jobs a class was created that issued a "safe-restart" from the jenkins-cli. Unfortunately this does not appear to work with pipeline jobs, which is problematic on our CI environment.

This changes the command to use "reload-configuration" which reloads all configuration that may have been made to disk, and should *hopefully* not interrupt any running jobs when it does it.


afda16:
I think we need to change the name of the file modules/govuk_jenkins/manifests/safe_restart.pp to match the new class name to be auto included?

surminus:
Thanks @afda16, updated.

PR #5498, opened 2017-02-16
CI Elasticsearch Label

djsd123:
We're now introducing an agent which does not offer elasticsearch. The implications to this are that jobs will have to explicitly define a label if it requires elasticsearch.  Otherwise, the job may be executed on an agent without it's requirements.


deanwilson:
Should we put a version number on the label liek we do with mongo for when we start testing ES5?

PR #5500, opened 2017-02-16
Use SSH credentials to access Github from CI

SamLR:
During busy parts of the day we have been hitting the Github rate
limit of 5,000 requests/hour. This causes an exception in Jenkins
and stops jobs building.

This commit changes the Multibranch Pipeline configuration to use
separate SSH credentials for checkout as to scanning. This should
mean that we are less likely to hit the rate limit.

**This assumes that the govuk-ci-ssh-key is enabled on github for the ci-user**


SamLR:
A test project using this can be seen here: https://ci.integration.publishing.service.gov.uk/job/aa-test/

PR #5501, opened 2017-02-16
Add taxonomy gem to CI

MatMoore:
See https://github.com/alphagov/govuk_taxonomy_helpers/pull/2

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -717,6 +717,7 @@ govuk_ci::master::pipeline_jobs:
   govuk_schemas_gem: {}
   govuk_security_audit: {}
   govuk_sidekiq: {}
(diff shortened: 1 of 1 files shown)


hieradata/common.yaml
```diff
@@ -712,6 +712,7 @@ govuk_ci::master::pipeline_jobs:
   govuk_security_audit: {}
   govuk_sidekiq: {}
   govuk_template: {}
+  govuk_taxonomy_helpers: {}
```
suzannehamilton: Should be before `govuk_template` to keep the list in order, but otherwise this is correct. 👍

MatMoore: :man_facepalming:

PR #5506, opened 2017-02-17
Include all dependencies for gems in jenkins builds

MatMoore:
I added this step in https://github.com/alphagov/govuk-puppet/pull/5502

I thought it was still useful to filter out development dependencies, but this
actually filters out things like govuk_lint which we need for CI.


tijmenb:
Could we rename the methods to something like `bundleForDevelopment` and `bundleForProduction`?

MatMoore:
Wouldn't that get confusing as well? I would read those as deployment environments, but that would be integration in both cases.

I'm thinking of this as more to do with the difference between apps and gems than environments.

- Apps should work with the specific set of dependencies that will be deployed
- Gems should work with anything that meets the gemspec 

In addition to the `--without development` flag, the app step also passes the `--deployment` flag, which ensures the gem lockfile is used to make builds repeatable. I don't think this works with gems either.

PR #5509, opened 2017-02-20
ci-agent-6 on Xenial

djsd123:
We require an agent running on Xenial.
We've decided to defer provisioning elasticsearch, due to the complexity involved with jumping multiple major versions.

https://trello.com/c/VWChaAZu

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -769,6 +769,10 @@ govuk_ci::master::ci_agents:
   ci-agent-5:
     agent_hostname: 'ci-agent-5.ci'
     labels: 'mongodb-3.2 ci-agent-5 elasticsearch'
+  ci-agent-6:
+    agent_hostname: 'ci-agent-6.ci'
+    labels: 'ci-agent-6 xenial'
+
 govuk_ci::master::credentials_id: 'jenkins-ssh-slave'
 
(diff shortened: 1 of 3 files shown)


surminus:
I'm still wondering how we defer jobs running on this machine (or if indeed we want to). With the 'xenial' label we can ensure jobs with that label always get built on it, but if a job has no labels then it may end up on the xenial machine, which may be problematic.

djsd123:
@surminus The way to achieve this is to modify the node config.  i.e. instead of "Use this node as much as possible". Change to "Only build jobs with label expressions matching this node".

surminus:
Ah OK, cool! Can we add that into the `govuk_ci::master::ci_agents` hash config?

djsd123:
@surminus It does not look like it is supported by the module.  I cannot see where `Availability` is defined.  Worth someone having a 2nd look just in case I missed it??

hieradata/common.yaml
```diff
@@ -769,6 +769,11 @@ govuk_ci::master::ci_agents:
   ci-agent-5:
     agent_hostname: 'ci-agent-5.ci'
     labels: 'mongodb-3.2 ci-agent-5 elasticsearch'
+  ci-agent-6:
+    agent_hostname: 'ci-agent-6.ci'
+    labels: 'mongodb-3.2 ci-agent-6 xenial'
```
surminus: if we're now making this exclusive we probably do not want mongodb-3.2 as a label

PR #5515, opened 2017-02-21
Replace publishing-e2e-docker with publishing-e2e-tests

kevindew:
This has been added so that we can trigger jenkins builds for https://github.com/alphagov/publishing-e2e-tests

It will eventually replace https://ci.integration.publishing.service.gov.uk/job/publishing-platform-docker/ which does not seem to be listed in this file. 

Note this is also a repo that needs docker in CI, not sure if I have to set anything up extra for that?

/cc @deanwilson @afda16

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -739,7 +739,7 @@ govuk_ci::master::pipeline_jobs:
   pp-manual:
     source: 'github-enterprise'
   pre-transition-stats: {}
-  publishing-e2e-docker: {}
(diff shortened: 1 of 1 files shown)


deanwilson:
You'll need to add the label it requires for Docker in the `Jenkinsfile`. Hopefully the existing one has the code you can crib from but if not we can have a look with @afda16

kevindew:
@deanwilson cheers - I think that's already sorted in the Jenkinsfile

PR #5517, opened 2017-02-22
Add memcached to frontend machines

tijmenb:
This adds memcached to the frontend machines. I've copied the config from the other machines with memcached (whitehall, mapit, backend nodes).

We'll use this mainly to cache HTTP responses, first from slimmer (https://github.com/alphagov/slimmer/pull/184) and hopefully later from gds-api-adapters.

cc @boffbowsh 

https://trello.com/c/D9HmkJwI

### DIFF
--- a/modules/govuk/manifests/node/s_calculators_frontend.pp
+++ b/modules/govuk/manifests/node/s_calculators_frontend.pp
@@ -21,4 +21,9 @@ class govuk::node::s_calculators_frontend inherits govuk::node::s_base {
 
   Govuk_mount['/data/vhost'] -> Class['govuk::apps::calculators']
 
+  include collectd::plugin::memcached
+  class { 'memcached':
+    max_memory => '12%',
+    listen_ip  => '127.0.0.1',
+  }
 }
--- a/modules/govuk/manifests/node/s_frontend.pp
+++ b/modules/govuk/manifests/node/s_frontend.pp
@@ -20,4 +20,10 @@ class govuk::node::s_frontend inherits govuk::node::s_base {
     host_name => $::fqdn,
     notes_url => monitoring_docs_url(nginx-high-conn-writing-upstream-indicator-check),
   }
+
+  include collectd::plugin::memcached
+  class { 'memcached':
+    max_memory => '12%',
+    listen_ip  => '127.0.0.1',
+  }
 }


boffbowsh:
How did you arrive at 12%? It looks about right based on frontend machines in production right now (and the hope that this will reduce memory bloat anyway) but it'd be good to have the number documented

tijmenb:
As noted, the config was copied from the existing configs. All the nodes are set at 12%. I'll try to find out why.

tijmenb:
The history of the 12%:

- It was first set in 2013 for the backend and development machines: https://github.com/alphagov/govuk-puppet/commit/055eaf8e8cfa46303395e09437aa3c4bb7f0084a. Unfortunately, it doesn't specify why. The original PR also doesn't say: https://github.gds/gds/puppet/pull/1167.
- Then Whitehall backend (Sep 9, 2014): https://github.com/alphagov/govuk-puppet/commit/e5d0a418877e4be9dad27df0dd46814041cba5c5
- Then Mapit (Jul 23, 2015): https://github.com/alphagov/govuk-puppet/commit/1cfca604304cb2104863d97341c33aa771f449ee
- Then Whitehall FE (Aug 18, 2015): https://github.com/alphagov/govuk-puppet/commit/f1b11ed27b8db794e29a139a89449f7dfde0de17

boffbowsh:
#wearethe12percent. Fine by me but probably needs a review from an infrastructure person first

tijmenb:
After some googling, I found that the docs in the puppet module that was originally used (saz/memcached) uses 12% as an example: https://github.com/saz/puppet-memcached#use-12-of-available-memory. Perhaps the source?

stephenharker:
Can this caching be done with Redis instead of memcached?

boffbowsh:
Technically yes, but Redis would have to be configured to do LRU based eviction, which it isn't by default. This would also mean we couldn't use the same Redis instances as we currently use, as they can't be configured for LRU eviction.

tijmenb:
Had a quick chat with @stephenharker about this. One important consideration is if we can still run applications on Heroku/PaaS without a configured memcached.

I'm closing this so it can thought about some more. It doesn't block https://github.com/alphagov/slimmer/pull/183, because apps could be configured to use `ActiveSupport::Cache::Store` and it would still be an improvement.

PR #5518, opened 2017-02-22
Use SSH key and wipe workspace

surminus:
Using the SSH key helps with git rate limits, and wipe the workspace to help ongoing maintenance.

As a note, the `scm.branches` must be specified as otherwise it checks out master.

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -13,6 +13,19 @@ def cleanupGit() {
 }
 
 /**
+ * Checkout repo using SSH key
+ */
+def checkoutFromGitHubWithSSH(String repository, String org = 'alphagov', String url = 'github.com') {
+  checkout([$class: 'GitSCM',
+    branches: scm.branches,
+    userRemoteConfigs: [[
+      credentialsId: 'govuk-ci-ssh-key',
+      url: "git@${url}:${org}/${repository}.git"
+    ]]
+  ])
+}
+
+/**
  * Try to merge master into the current branch
  *
  * This will abort if it doesn't exit cleanly (ie there are conflicts), and


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -13,6 +13,19 @@ def cleanupGit() {
...
 
 /**
+ * Checkout repo using SSH key
+ */
+def checkoutSSH(String repository, String org = 'alphagov', String url = 'github.com') {
+  checkout([$class: 'GitSCM',
+    branches: scm.branches,
+    userRemoteConfigs: [[
+      credentialsId: 'govuk-ci-ssh-key',
+      url: "git@${url}:${org}/${repository}.git"
```
cbaines: Because of `${org}/${repository}`, this will probably only work for GitHub instances, so maybe checkoutFromGitHubWithSSH would be a more descriptive name?

surminus: I explicitly added the `${url}` and `${org}` arguments so we could also pass in arguments for both `github.com:gds-operations/foo.git` and `github.gds:gds/foo.git`

cbaines: Yep, and that makes sense to me. What I meant is that the way the url is build looks more specific than the name might suggest, e.g. you might run in to difficulties if you tried to use this to checkout something not on a GitHub instance.

surminus: Oh, I see what you mean. I'll update the method name.

surminus: Updated!

cbaines: Great :+1:

PR #5519, opened 2017-02-22
Gem publishing function

alecgibson:
This change adds a common function for use in a `Jenkinsfile` to publish a gem, which used to be performed with the [Gem Publisher gem](https://github.com/alphagov/gem_publisher).

Getting this gem to work with Jenkins 2 was non-trivial, so this behaviour was migrated to Jenkins 2 on [`govuk_navigation_helpers`](https://github.com/alphagov/govuk_navigation_helpers/pull/24). Since then, this behaviour has also been moved successfully to `gds-api-adapters` and [`slimmer`](https://github.com/alphagov/slimmer/pull/186).

This functionality is now being commonised into Puppet.


cbaines:
Thanks for doing this Alex, this solves the problems I was encountering, and gave up trying to fix with the gds-api-adapters build :)

PR #5521, opened 2017-02-22
Add environment variable to content-performance-manager

leenagupte:
Add an environment variable to content-performance-manager that holds the
default value for the batch size of requests sent to Google Analytics.

Related to: https://github.com/alphagov/content-performance-manager/pull/76

### DIFF
--- a/modules/govuk/manifests/apps/content_performance_manager.pp
+++ b/modules/govuk/manifests/apps/content_performance_manager.pp
@@ -23,6 +23,9 @@
 #   Google authentication email
 #   Default: undef
 #
+# [*ga_pageview_batch_size*]
+#   Default batch size of requests to Google API
+#   Default: 3000
 #
 # [*db_hostname*]
 #   The hostname of the database server to use in the DATABASE_URL.
@@ -45,6 +48,7 @@ class govuk::apps::content_performance_manager(
   $google_analytics_govuk_view_id = undef,
   $google_private_key = undef,
   $google_client_email = undef,
+  $ga_pageview_batch_size = '3000',
   $db_hostname = undef,
   $db_username = 'content_performance_manager',
   $db_password = undef,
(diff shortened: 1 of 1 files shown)


deanwilson:
You could set the value on https://github.com/alphagov/govuk-puppet/pull/5521/files#diff-9ad5ee54919f692061c45771f662f350R51 and save an extra line in hieradata if this is the same between environments at the moment.

cbaines:
I'd normally set a default in the code (e.g. here https://github.com/alphagov/content-performance-manager/pull/76/files#diff-31e3b35bb88d5abfca18ceb7a5ff5c4eR4 ), as that makes the setup quite a bit simpler. Is there a specific reason why you wanted to only set this through puppet?

leenagupte:
@deanwilson @cbaines Both are good points. We were talking about having a configurable value, and mentioned using a environment variable, but you are right, setting a default in the code is easier.

deanwilson:
If this is going in to the code should this PR be closed?

leenagupte:
@deanwilson I'll discuss it with the team and update appropriately. In the mean time I'll add a "Do not merge" to the title.

leenagupte:
@deanwilson You're right, the batch size will be the same in all environments, so I've taken your advice and removed the line from hieradata.

leenagupte:
@cbaines Thanks for the advice. I've added a default value in the code as well as assigning an environment variable.

modules/govuk/manifests/apps/content_performance_manager.pp
```diff
@@ -23,6 +23,9 @@
 #   Google authentication email
 #   Default: undef
 #
+# [*ga_pageview_batch_size*]
+#   Default batch size of requests to Google API
+#   Default: undef
```
surminus: The default doesn't seem to be `undef`

leenagupte: Thanks! I'll fix it.

leenagupte: Fixed!

leenagupte:
As the value for the batch size is the same in allow environments, we have decided to just set this value in the application code.

PR #5524, opened 2017-02-23
[WIP] Add templating of Google credentials file

surminus:
Google Cloud does things differently in that you must have a credentials file to be able to authenticate with their API. This causes a problem as while we can provide a file to use for authentication, we want to ensure that when people run the Jenkins job they are passing in some form of authentication, because otherwise they could potentially cause serious impact in deployment of DNS.

This looks to template this credentials file, in which we can store most of the secrets of the file in encrypted hiera, but also prompt for a couple of parts of the file in the Jenkins job itself, and pass these in as environment variables. These secrets can be stored in our password store to use.

### DIFF
--- a/modules/govuk_jenkins/manifests/job/deploy_dns.pp
+++ b/modules/govuk_jenkins/manifests/job/deploy_dns.pp
@@ -11,16 +11,47 @@
 #   ID of the route53 DNS zone to upload to
 #
 # [*dyn_customer_name*]
-#    Customer account to use for Dyn (not to be confused with the Dyn user name)
+#   Customer account to use for Dyn (not to be confused with the Dyn user name)
+#
+# [*gce_zone_id*]
+#   The zone ID to deploy to
+#
+# [*gce_project_id*]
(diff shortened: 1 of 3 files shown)


modules/govuk_jenkins/manifests/job/deploy_dns.pp
```diff
@@ -11,16 +11,47 @@
...
 class govuk_jenkins::job::deploy_dns (
     $dyn_zone_id,
     $route53_zone_id,
     $dyn_customer_name,
+    $gce_zone_id,
+    $gce_project_id,
+    $gce_private_key,
+    $gce_client_name,
+    $gce_credentials = '/var/lib/jenkins/gce_credentials.json',
+    $gce_region= 'europe-west1-d',
```
deanwilson: Tiny comment - but a space before the `=`?

deanwilson:
We've discussed this PR and we're going to change the approach to use a group and limit the job to only admin users.

PR #5526, opened 2017-02-23
Restart carbon-aggregator service every 5 minutes in Staging

afda16:
Carbon-aggregator connections in Staging seem to be filled up from clients on
the licensify VPN, and the errors in the log file are filling the disk.

This is probably caused by network problems that are being investigated
with Carrenza, for now we can configure this cron job to restart the daemon
every 5 minutes.

### DIFF
--- a/modules/govuk/manifests/node/s_graphite.pp
+++ b/modules/govuk/manifests/node/s_graphite.pp
@@ -89,4 +89,13 @@ class govuk::node::s_graphite (
 
   include collectd::server
   include grafana
+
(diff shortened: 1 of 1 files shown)


surminus:
I think it might be better to just disable puppet on graphite-1 and add a cron job manually. We'll have an alert in Icinga to remind us Puppet is disabled and it won't require a Puppet deploy to add.

deanwilson:
The connections are staying in an established state so this is a good way to handle this while then underlying issue is investigated.

afda16:
I'll follow @surminus advice, easier to remember to remove it later

afda16:
Cron job added manually

PR #5530, opened 2017-02-24
[DO NOT MERGE] Jenkins Roles

djsd123:
We've introduced some deploy jobs that we'd prefer to be restricted to specified, authenticated users.  At present all interactive users hold the `admin` role.

This PR specifies a `dev` role with the same access as the previous admin role.  Except that the `devs` is restricted from building jobs that name matches the `Deploy_DNS` pattern and it's variants.

Relates to: https://github.com/alphagov/govuk-puppet/tree/gce-jenkins

See: https://trello.com/c/88BhiP5e

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -55,6 +55,13 @@ govuk_jenkins::config::theme_text_colour: 'black'
 govuk_jenkins::config::theme_environment_name: 'Integration'
 
 govuk_jenkins::config::admins:
+  - anafernandez
+  - deanwilson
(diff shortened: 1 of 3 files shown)


surminus:
We should be careful when we role this out because if there's any problems it'll cut off access to the devs. 

I'm very tempted to say we should just allow devs to be able to run the deploy DNS job. Most people who have access to Jenkins can be destructive if they want to be. For now, let's get the job out and the credentials file created and worry about if we want to auth later. I'm a little nervous at having this change deployed.

PR #5533, opened 2017-02-24
Change to using s3cmd rather than awscli

SamLR:
The `aws s3 sync` command does not complete. This tries using the alternate `s3cmd sync` which will hopefully succeed.

This is awaiting the successful completion of a manual run on `mirrorer-1.integration`.

### DIFF
--- a/modules/govuk_crawler/manifests/init.pp
+++ b/modules/govuk_crawler/manifests/init.pp
@@ -129,8 +129,12 @@ class govuk_crawler(
         provider => system_gem,
   }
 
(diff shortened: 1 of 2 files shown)


modules/govuk_crawler/manifests/init.pp
```diff
@@ -130,7 +130,7 @@
   }
 
   # Needed to copy to AWS S3
-  package { 'awscli':
+  package { 's3cmd':
```
surminus: can we ensure the awscli is set to `absent` also?

PR #5540, opened 2017-02-27
Reconfigure offsite backups

surminus:
This PR does two things:

 - reconfigures the offsite backups class so that we do not need the configuration in two places in hiera
 - removes the configuration in hiera from common to leave it only in Production

More detail in commits.


surminus:
Nervous about deploying this so going to wait.

surminus:
I've tested this in vagrant several times and am reasonably happy it'll work. Will do a careful rollout.

PR #5546, opened 2017-02-28
Revert "Update bucket to destination"

surminus:
This reverts commit 52f48dad5f8f31aab809b9acd2c79ebaea98dbc5.

I'm not convinced this is the fix, so would prefer to revert this for now. I want to test having a bucket without versioning or lifecycling enabled before updating the destination in this way.


surminus:
The `bucket` parameter is actually deprecated: https://github.com/alphagov/puppet-duplicity/blob/master/manifests/job.pp#L99

PR #5549, opened 2017-03-01
Ensure that MySQL is stopped before Puppet runs

surminus:
Unfortunately because the upstream module manages the MySQL service there's no good way of disabling it to start on boot. If it does try to start on boot, it will likely fail because the `/var/lib/mysql` dir won't have been initilised, however it will still place some ibdata stuff inside of it.

We do not want anything shadow mounted, so using the on-boot cron job, stop the mysql service (if it's already stopped then the exit code will be 1 thus the break), delete the directory and run puppet. Puppet will then create the directory, mount in tmpfs and do everything else.


surminus:
This isn't required because the changes in /etc/fstab will ensure the directory is mounted before the MySQL service even tries to start on boot.

PR #5559, opened 2017-03-02
Add documentation to Icinga virus scan alert

afda16:


modules/govuk/manifests/node/s_asset_master.pp
```diff
@@ -98,6 +98,7 @@
     critical  => 2400,
     desc      => 'Waiting time to pass virus scan',
     host_name => $::fqdn,
+    notes_url => monitoring_docs_url(virus-scanning),
```
SamLR: I think this needs a second argument, otherwise `monitoring_docs_url` will assume this is a section in the nagios page. The docs for the function are a bit vague https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/lib/puppet/parser/functions/monitoring_docs_url.rb#L20

afda16:
Updated, cheers!

PR #5560, opened 2017-03-02
Duplicate Terraform Declaration

djsd123:
Accidentally added a duplicate package declaration.  Removing package resource and specifying dependency.

See: https://trello.com/c/88BhiP5e


surminus:
This feels a little weird to me. Could we create a new class (`govuk_terraform`) and include that?

PR #5561, opened 2017-03-02
Use latest bundler for all rbenv rubies

mobaig:
Keeping our bundler versions up to date and consistent across rubies
gives us access to bug fixes, speed improvements, and new features.

This change also specifically addresses the Gemfile.lock fluctuations
that occur around the inclusion/removal of the RUBY VERSION stamp. This
behaviour was introduced in bundler 1.12. Any developer using 1.12 and
above (because they manually updated their bundler version on the VM,
let's say) will generate lock files that include the version stamp. A
developer running < 1.12 will generate lock files without it. This leads
to additional noise and confusion in pull requests when developers
running different versions end up reversing each other's changes.

### DIFF
--- a/modules/govuk_rbenv/manifests/all.pp
+++ b/modules/govuk_rbenv/manifests/all.pp
@@ -21,46 +21,46 @@ class govuk_rbenv::all (
   }
 
   rbenv::version { '2.1.2':
-    bundler_version => '1.6.5',
+    bundler_version => '1.14.5',
   }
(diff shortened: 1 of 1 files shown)


h-lame:
This seems great :clap:  

Probably want another reviewer or two for any potential gotchas first (this caution has nothing to do with me wanting #5562 to be merged first so you have to deal with the conflict. 😉)

PR #5567, opened 2017-03-03
Removes access for Douglas Roper

dougdroper:
👋


surminus:
Need to remove yourself from [here](https://github.com/alphagov/govuk-puppet/blob/master/hieradata/integration.yaml) too!

PR #5572, opened 2017-03-06
Clean up Integration data sync complete job

jennyd:
See individual commits for more details.


surminus:
Just also need to remove [the job from hiera](https://github.com/alphagov/govuk-puppet/blob/master/hieradata/class/integration/jenkins.yaml#L25), otherwise looks fine.

jennyd:
@surminus ah yes, thanks, forgot about that one :) I've removed that now.

PR #5573, opened 2017-03-06
Enable sync to S3 in staging

SamLR:
In order to test using S3 for mirroring we want to use staging for failover tests. To do this we need to provision the staging S3 mirror buckets with the mirrors. This enables the sync cron job in staging.

It also enables the crawler-seed job in order to bring Staging more in line with Production

### DIFF
--- a/hieradata/staging.yaml
+++ b/hieradata/staging.yaml
@@ -29,6 +29,7 @@ govuk::deploy::setup::ssh_keys:
 
 govuk_cdnlogs::monitoring_enabled: false
 
+govuk_crawler::seed_enable: true
 govuk_crawler::sync_enable: true
 govuk_crawler::targets:
   - 's3://govuk-mirror-staging'
--- a/hieradata/staging.yaml
+++ b/hieradata/staging.yaml
@@ -29,6 +29,10 @@ govuk::deploy::setup::ssh_keys:
 
 govuk_cdnlogs::monitoring_enabled: false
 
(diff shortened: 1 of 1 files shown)


surminus:
I agree we should have parity with Production, but we should try to find out why it's disabled incase it has some unintended side effects. I suspect @rjw1 knows

SamLR:
Discussion of this from infrastructure channel on Slack (edited for clarity):
```
<sam_cook>  any ideas as to why the seed-crawler job is disabled in staging?
<bob>       because we dont want it running there. we let it run in integration
<sam_cook>  will it cause problems if it does run there? We're going to enable
            the sync job in staging (mostly to prep the S3 bucket for testing
            mirroring) and it seems sensible that the crawler is primed
<bob>       it shouldn't. It may increase load on production though.
            for testing purposes it will be fine
<sam_cook>  cool I'll make sure we have a ticket to disable/remove the staging &
            integration buckets as well as the seed job in staging
<lauram>    I don’t think we necessarily need to remove the bucket in
            Integration. I like the fact we’re testing the actual sync function
            somewhere.
<bob>       it is possibly worth revisiting the decision about staging as well.
<sam_cook>  fair
<bob>       the datasync is possibly the reason also in integration it may not
            actually run because the machines are off :)
<lauram>    I suspect the job starts after they’ve been switched off.
            The job definitely runs because the rabbitmq queue used to fill up
            and run out of memory
```

PR #5578, opened 2017-03-06
Doug left on 3rd March

sihugh:
Hey Duggee!


sihugh:
Already done in https://github.com/alphagov/govuk-puppet/pull/5567

PR #5582, opened 2017-03-07
Use new duplicity opts

surminus:
Update the duplicity module and add the new options to the backup configuration. Detail in commits.


surminus:
When testing the backup spawned hundreds of backup processes and caused asset-slave-1 to have load of >300. 

Back to the drawing board.

surminus:
I've readjusted this, and added a new commit which includes mirroring a PPA for the latest release of duplicity, which has additional functionality.

surminus:
I want to do a managed roll out of this change.

PR #5584, opened 2017-03-07
Update my SSH key

mgrassotti:
as I changed laptop


mgrassotti:
This is my vagrant machine key

PR #5587, opened 2017-03-08
Don't lint release branches

MatMoore:
We lint changes by taking the diff with master. This works for dev branches but
not release branches, which are either behind or in sync with master.

When the commits on the current branch are already in master, diffing it to
master makes no sense.  govuk-lint-ruby picks up lines that wouldn't have been
linted when the changes in it were first made.  This is happening for the
deployed-to-production branches of our apps, and leads to a lot of spurious
lint errors when changing the content schemas, because we trigger CI jobs using
production app code with the new schemas.

This is the same situation as described in https://github.com/alphagov/govuk-puppet/pull/5396

This problem currently blocks https://github.com/alphagov/govuk-content-schemas/pull/560

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -26,6 +26,18 @@ def checkoutFromGitHubWithSSH(String repository, String org = 'alphagov', String
 }
 
 /**
+  * Check if the git HEAD is ahead of master.
+  * This will be false for development branches and true for release branches,
+  * and master itself.
+  */
+def isCurrentCommitOnMaster() {
+  sh(
+    script: 'git rev-list origin/master | grep $(git rev-parse HEAD)',
+    returnStatus: true
+  ) == 0
+}
+
+/**
  * Try to merge master into the current branch
  *
  * This will abort if it doesn't exit cleanly (ie there are conflicts), and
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -26,6 +26,16 @@ def checkoutFromGitHubWithSSH(String repository, String org = 'alphagov', String
 }
 
 /**
+  * Check if the git HEAD is ahead of master.
+  * This will be false for development branches and true for release branches,
+  * and master itself.
+  */
+def isCurrentCommitOnMaster = sh(
```
suzannehamilton: This looks like a variable assignment rather than a function definition, which I think means it will be evaluated when this helper script is loaded, and therefore will be run _before_ `git clone`. That seems wrong, but I might have just misunderstood what's going on.

MatMoore: It's most likely me not understanding as I haven't tested this. Will check it to tomorrow.

MatMoore:
@suzannehamilton have replaced with function syntax (I think!)

modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -34,11 +46,6 @@ def checkoutFromGitHubWithSSH(String repository, String org = 'alphagov', String
  * branch.
  */
 def mergeMasterBranch() {
-  def isCurrentCommitOnMaster = sh(
-    script: 'git rev-list origin/master | grep $(git rev-parse HEAD)',
-    returnStatus: true
-  ) == 0
-
   if (isCurrentCommitOnMaster) {
```
suzannehamilton: I think you need the parentheses in zero-arg function calls: http://groovy-lang.org/semantics.html#_optional_parentheses

Ditto for line 137.

PR #5588, opened 2017-03-08
Remove daily purge

surminus:
We now sync data to a bucket for Integration so this is no longer required as the rabbitmq queue will be processed.

### DIFF
--- a/modules/govuk/manifests/node/s_mirrorer.pp
+++ b/modules/govuk/manifests/node/s_mirrorer.pp
@@ -7,30 +7,20 @@
 # [*daily_queue_purge*]
 #   This purges the rabbitmq queue of all messages. Use with caution!
 #
-class govuk::node::s_mirrorer (
-  $daily_queue_purge = false,
-) inherits govuk::node::s_base {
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/node/s_mirrorer.pp
```diff
@@ -7,30 +7,20 @@
...
-      ensure  => present,
-      owner   => 'root',
-      group   => 'root',
-      mode    => '0700',
-      content => "/usr/local/bin/rabbitmqadmin --username=root --password=${::govuk_rabbitmq::root_password} purge queue name=govuk_crawler_queue",
-      require => Class['govuk_rabbitmq'],
-    }
+  # Create a file that only root can read so not to expose the rabbitmq root password.
+  file { '/root/purge_govuk_crawler_queue.sh':
+    ensure  => 'absent',
```
afda16: I'm not sure if lint should complain with the extra space in the arrow alignment , checks are passing so it should be ok

surminus:
Thanks, just added a FIXME, too.

PR #5595, opened 2017-03-09
Remove an unneeded dependency

deanwilson:
The package itself was removed in
https://github.com/alphagov/govuk-puppet/commit/3bec96c6f502698093d1e1ff7d72d3fb6215897c


surminus:
wot sam sed 👆

PR #5598, opened 2017-03-10
Change pushTag to not try and update a branch

cbaines:
This is breaking publishGem, as the repositories containing gems do
not usually have release branches.


surminus:
Do we know if there is anything that uses a release tag and that might break by removing this?

cbaines:
> Do we know if there is anything that uses a release tag and that might break by removing this?

I haven't checked if anything is using this and expecting it to push to a `release` branch. I'll have a look shortly though. There are some options, including modifying this to fail silently (e.g. have pushTag only update the release branch if it already exists) which would fix the immediate issue I'm encountering.

cbaines:
> Do we know if there is anything that uses a release tag and that might break by removing this?

So looking at a couple of repositories, can't see anything about updating the release branch. I also don't particularly like the idea of updating them all before I can get gem publishing to work. So I'll close this in favour of https://github.com/alphagov/govuk-puppet/pull/5601 .

PR #5599, opened 2017-03-10
Improve publish gem tag management

cbaines:


cbaines:
This should make publishing gems work ( @whoojemaflip and I are having issues with govuk_content_models and plek respectively), but the build will still fail as it will fail when pushTag attempts to update the non-existent release branch. If https://github.com/alphagov/govuk-puppet/pull/5601 works, then that will prevent pushTag from updating attempting to update the release branch.

36degrees:
It might just be my Friday afternoon brain, but I really can't get my head around the first commit message - I can't seem to parse it as a proper sentence.

MatMoore:
I think we could lose the `return` statements inside the `taggedReleaseExists` and `versionAlreadyPublished` checks. Then if either ones fails it should be safe to rerun the job.

modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -311,6 +298,19 @@ def publishGem(String repository, String branch) {
     sh("gem build ${repository}.gemspec")
     sh("gem push ${repository}-${version}.gem")
   }
+
+  def taggedReleaseExists = sh(
+    script: "git ls-remote --exit-status --tags origin v${version}",
```
36degrees: Should this be `--exit-code` as per https://git-scm.com/docs/git-ls-remote.html? Doesn't work locally, at least on my Mac (Git 2.9.0)

cbaines:
> It might just be my Friday afternoon brain, but I really can't get my head around the first commit message - I can't seem to parse it as a proper sentence.

I agree that its not great, would "Push version tag last in publishGem" be any better?

36degrees:
Sorry, it's the commit description I'm struggling with more than the summary…

> This means that issues with the tagging, either it failing and stopping the gem being published, or where the tag exists, but the gem is not published from stopping the gem being published.

modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
cbaines: Yep, good spot :)

cbaines:
> Sorry, it's the commit description I'm struggling with more than the summary…

Ah, I've tried rewriting it, but I'm still struggling to make it clear. I'm happy to try again if you think it still needs some work.

cbaines:
> I think we could lose the return statements inside the taggedReleaseExists and versionAlreadyPublished checks. Then if either ones fails it should be safe to rerun the job.

Ah yeah, that does make more sense. I've done this in 81d2dc9 .

36degrees:
New commit message is 👌 – thanks for changing that!

PR #5602, opened 2017-03-10
Jenkins Deploy Upgrade [stage 1]

djsd123:
Upgrading deploy Jenkins to version 2.

Plugin versions had to be modified to be compatible with upgraded Jenkins core.

plugin=github-oauth must be omitted when provisioning a vagrant instance.  This is because the plugin will cause Jenkins to not load its configuration due to the lack of an auth backend.

see: https://trello.com/c/RZCVbiHd

### DIFF
--- a/hieradata/class/jenkins.yaml
+++ b/hieradata/class/jenkins.yaml
@@ -1,5 +1,4 @@
 ---
-
 govuk_jenkins::config:
   NAME:
     value: jenkins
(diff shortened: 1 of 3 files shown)


hieradata/class/jenkins.yaml
```diff
@@ -1,5 +1,7 @@
 ---
 
+govuk_jenkins::version: '2.19.2'
```
surminus: I wonder if it's worth getting ahead of the curve and trying an upgrade to the latest?

modules/govuk_jenkins/manifests/init.pp
```diff
@@ -36,7 +36,7 @@
   $plugins = {},
   $ssh_private_key = undef,
   $ssh_public_key = undef,
-  $version = '1.554.2',
+  $version = undef,
```
surminus: What's the behaviour if we don't specify a version? is it better to pin here rather than having another line in hiera?

hieradata/class/jenkins.yaml
djsd123: Plugin Hell when I tried version 2.7.x

PR #5504, opened 2017-02-17
CI Bundle Cleanup

djsd123:
Jenkins Bundle directories take up a lot of disk space. Currently a bundle directory gets created for every project/branch that gets built.
These directories then remain there afterwards.
Consequence of this is that the first build of the day may take slightly longer depending.

PR #5604, opened 2017-03-13
Add gem directory parameter to bundler methods in Jenkinslib

suzannehamilton:
Add an optional parameter to both `bundle install` methods to allow projects' Jenkinsfiles to set the subdirectory that gems are installed to.

This will let individual projects override the default directory, which is the job name. Most govuk Jenkins jobs are multibranch pipelines with job names like `project_name/branch_name`. This means that every single branch has its own gem directory, which causes gems to be installed from scratch for each branch. This can take several minutes.

Projects will now be able to specify the gem directory to be the repository name, so that builds of new branches don't have to install the existing gems. This will hopefully knock a few minutes off the build times of projects which use it.

https://trello.com/c/sGdEJaaW/15-gem-installs-take-a-long-time

This should also reduce the disk usage of the Jenkins nodes, but I don't know how significant it will be.


MatMoore:
Could we parse the job name and default it to `project_name` instead of `project_name/branch_name` for multi branch jobs?

suzannehamilton:
@MatMoore Do you think that's safe to do for every project? I was a bit concerened that changing every project at once could cause hard-to-debug problems like the bundler upgrade did (because most project owners won't know about this change).

But I might be worrying about nothing. If people think it's safe to change everything at once then we can do that. :)

tijmenb:
I think using a directory per-app is what we used to do. I think we should try it! We can always revert if it causes problems.

tijmenb:
With this change, we perhaps won't need https://github.com/alphagov/govuk-puppet/pull/5504 anymore?

rjw1:
Do we want to go a step further and just have one bundle directory for everything?

tijmenb:
The only thing I can think of why that wouldn't be possible if there are concurrency issues when two projects write to the same bundle dir at the same time. I don't know the internals of bundler. We could try it?

suzannehamilton:
When I was testing the job throttling, I saw build failures when two projects were compiling gems in the same directory at the same time. And searching around, I see that [other](https://0xacab.org/leap/platform/issues/8684) [people](https://github.com/bundler/bundler/issues/2586#issuecomment-22918449) have had problems with parallel installs on CI machines.

I've actually just added the Lockable Resource Plugin to Jenkins (#5603). We could use a shared gem directory and then use a lock to prevent jobs from running parallel `bundle install` steps on the same node? `bundle install` usually just takes a few seconds if it doesn't have to install new gems, so it wouldn't block other jobs for very long.

I've also updated the commit with the per-project gem directory, so you can see what that looks like.

suzannehamilton:
I had a chat to @tijmenb, and we've decided to try out @rjw1's suggestion of a single gem directory for all projects.

I've added a `lock` around `bundle install` to prevent projects from trying to install gems in parallel. I don't think it will impact build times by more than a few seconds, but the lock is inside the statsd timing so we can test that assumption. 🙂

PR #5605, opened 2017-03-13
Allow building deployed-to-production for frontend

cbaines:
This is used for govuk-content-schema tests.


cbaines:
Thanks @36degrees :)

PR #5606, opened 2017-03-13
Add a Plex service URI for Licensify for draft frontend

issyl0:
- In the same vein as PR 5476. Frontend talks to Licensify to get licence details.

### DIFF
--- a/modules/govuk/manifests/node/s_draft_frontend.pp
+++ b/modules/govuk/manifests/node/s_draft_frontend.pp
@@ -14,6 +14,7 @@ class govuk::node::s_draft_frontend() inherits govuk::node::s_base {
     'PLEK_HOSTNAME_PREFIX': value => 'draft-';
     'PLEK_SERVICE_ERRBIT_URI': value => "https://errbit.${app_domain}";
     'PLEK_SERVICE_IMMINENCE_URI': value => "https://imminence.${app_domain}";
+    'PLEK_SERVICE_LICENSIFY_URI': value => "https://licensify.${app_domain}";
     'PLEK_SERVICE_LOCAL_LINKS_MANAGER_URI': value => "https://local-links-manager.${app_domain}";
     'PLEK_SERVICE_MAPIT_URI': value  => "https://mapit.${app_domain}";
     'PLEK_SERVICE_SEARCH_URI': value => "https://search.${app_domain}";


modules/govuk/manifests/node/s_draft_frontend.pp
```diff
@@ -14,6 +14,7 @@
     'PLEK_HOSTNAME_PREFIX': value => 'draft-';
     'PLEK_SERVICE_ERRBIT_URI': value => "https://errbit.${app_domain}";
     'PLEK_SERVICE_IMMINENCE_URI': value => "https://imminence.${app_domain}";
+    'PLEK_SERVICE_LICENSIFY_URI': value => "https://licensify-admin.${app_domain}";
```
issyl0: Should this be `...LICENSIFY_ADMIN_URI`?

cbaines: The environment variable name is constructed by plek, so if you say `Plek.find('cats')`, it will check PLEK_SERVICE_CATS_URI [1]. Looking at frontend, it asks Plek to find 'licensify' so I think that is right for the environment variable name.

One think I looked in to was the `licensify-admin` bit. Given that this was not specified before, on the live stack, asking Plek for licensify would have probably been falling through to `licensify.publishing.service.gov.uk`. I checked govuk-puppet, and it looks like licensify and licensify-admin are both valid hosts for the same service, so I think that is ok also.

1: https://github.com/alphagov/plek/blob/master/lib/plek.rb#L168

davidbasalla:
Tested this on integration, it configures the api as follows (and returns 404s):
`<GdsApi::LicenceApplication:0x007faf403e3270 @options={:logger=>#<NullLogger::Logger:0x007faf4052b358>, :cache_size=>0, :cache_ttl=>1, :endpoint_url=>"https://licensify-admin.integration.publishing.service.gov.uk"}, @endpoint="https://licensify-admin.integration.publishing.service.gov.uk">`

but on frontend it's (and this works and returns 200s):

`<GdsApi::LicenceApplication:0x007f3186b7b4c0 @options={:logger=>#<NullLogger::Logger:0x007f3186de0530>, :cache_size=>0, :cache_ttl=>1, :endpoint_url=>"https://licensify.integration.publishing.service.gov.uk"}, @endpoint="https://licensify.integration.publishing.service.gov.uk">`

So that would suggest we should get rid of the '-admin' bit...?

modules/govuk/manifests/node/s_draft_frontend.pp
issyl0: OK, thanks.

PR #5607, opened 2017-03-13
Instrument expensive steps

tijmenb:
This adds instrumentation to most functions that take a while to complete. We can use this to diagnose slow builds.

Stats are emitted to graphite with the key:

```
stats.timers.ci.APP_NAME.KEY_NAME
```

https://trello.com/c/5EGPSkfa

cc @suzannehamilton

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -9,20 +9,24 @@
  */
 def cleanupGit() {
   echo 'Cleaning up git'
-  sh('git clean -fdx')
+  withStatsdTiming("cleanup_git") {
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -9,20 +9,24 @@
  */
 def cleanupGit() {
   echo 'Cleaning up git'
-  sh('git clean -fdx')
+  withStatsdTiming("cleanup_git", {
```
suzannehamilton: Just a syntax suggestion, but Groovy gives you a neater way to place the parentheses when the last argument is a closure:

```
withStatsdTiming("cleanup_git") {
  // Function body
}
```

It's also how we call the Jenkins functions like `node` and `stage`.

tijmenb:
@suzannehamilton updated!

PR #5610, opened 2017-03-14
Fix ssh-key usage in gem push

tijmenb:
This wraps the `git ls-remote` script in a block to use the correct SSH key. This prevents the script from trying to create a tag when it already exists.

https://trello.com/c/ONAhXHP8

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -336,10 +336,14 @@ def publishGem(String repository, String branch) {
     sh("gem push ${repository}-${version}.gem")
   }
 
-  def taggedReleaseExists = sh(
-    script: "git ls-remote --exit-code --tags origin v${version}",
-    returnStatus: true
-  ) == 0
+  def taggedReleaseExists = false
+
+  sshagent(['govuk-ci-ssh-key']) {
+    taggedReleaseExists = sh(
+      script: "git ls-remote --exit-code --tags origin v${version}",
+      returnStatus: true
+    ) == 0
+  }
 
   if (taggedReleaseExists) {
     echo "Version ${version} has already been tagged on Github. Skipping publication."


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -336,10 +336,12 @@ def publishGem(String repository, String branch) {
     sh("gem push ${repository}-${version}.gem")
   }
 
-  def taggedReleaseExists = sh(
-    script: "git ls-remote --exit-code --tags origin v${version}",
-    returnStatus: true
-  ) == 0
+  sshagent(['govuk-ci-ssh-key']) {
+    def taggedReleaseExists = sh(
```
suzannehamilton: `taggedReleaseExists` is now defined in a closure. Is it available outside the closure's scope on line 346? (My guess is no, but I don't know enough Groovy to say for sure.)

cbaines: I think you would now need to declare taggedReleaseExists before the closure passed to sshagent, but I wasn't able to verify this locally.

modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -336,10 +336,14 @@ def publishGem(String repository, String branch) {
     sh("gem push ${repository}-${version}.gem")
   }
 
-  def taggedReleaseExists = sh(
-    script: "git ls-remote --exit-code --tags origin v${version}",
-    returnStatus: true
-  ) == 0
+  taggedReleaseExists = false
```
tijmenb: This should work according to http://stackoverflow.com/a/185879.

cbaines: Great, :shipit:

PR #5611, opened 2017-03-15
Add single-step build for most projects

tijmenb:
This adds functionality to perform the build process with a minimal Jenkinsfile. In most cases the following will be sufficient:

```groovy
#!/usr/bin/env groovy
node {
  def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
  govuk.buildProject()
}
```

It's been tested with:

- content-tagger (Rails app with database)
- collections & finder-frontend (Rails apps without database)
- rummager (Sinatra app)
- email-alert-service (non-web app)
- govuk_ab_testing (a Ruby gem)

The script uses some simple heuristics to determine how to build/test/release the project.

It's intended to fix a number of specific problems and make it easier to iterate on the build process.

- @36degrees discovered that some downstream schema testing projects aren't correctly checking out the correct branch, which causes them to tests against the `deployed-to-production` branch of the schemas, not against the branch they should be testing against
(https://trello.com/c/iZMj1jz9).
- The [throttling configuration][throttle] in some apps is wrong, causing slow builds.
- @mcgoooo discovered that the [contacts-admin tests][contacts] weren't being run because `RAILS_ENV` wasn't set.
- `checkout scm` often runs into trouble with rate limiting, but it's hard to change all the Jenkinsfiles (https://trello.com/c/amd0k0xe)

### SASS linting

The only thing we can't infer from the code is whether or not the project is supposed to run SASS linting. If you aren't ready to run this linter yet, you can disable it explicitly:

```groovy
#!/usr/bin/env groovy
node {
  def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
  govuk.buildProject(false)
}
```

### Trade-offs

- We have to clone the schemas for all projects. This is acceptable because most projects use the schemas, but it will slow down non-schema tests with a few seconds.
- It's easy to break CI for all projects by messing up this script. If that happens often we should be looking to add groovy unit tests or do some kind of integration test.
- Some projects will have to be updated to respond correctly to `bundle exec rake (default)`, which should run the tests. This is acceptable because it improves consistency across projects.
- It's not possible to specify the linting directories so [some projects](man) will have less linting. 

[throttle]: https://github.com/alphagov/content-tagger/blob/5fefdace31953b2582780f1932dc5198463a0f9c/Jenkinsfile#L15-L22
[contacts]: alphagov/contacts-admin#265
[man]: https://github.com/alphagov/manuals-publisher/blob/3cd802d065ad21e063f28311dd522c54455c5c78/Jenkinsfile#L70

https://trello.com/c/iDWjE1Vz

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -451,7 +451,10 @@ def buildProject(sassLint = true) {
     }
 
     stage("Run tests") {
-      runTests()
+      // Prevent a project's tests from running in parallel on the same node
+      lock("$repoName-$NODE_NAME-test") {
+        runTests()
+      }
     }
 
     if (hasAssets()) {
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -463,17 +463,19 @@ def buildProject(sassLint = true) {
       }
     }
 
-    if (isGem()) {
-      stage("Publish Gem to Rubygems") {
-        publishGem(repoName, env.BRANCH_NAME)
-      }
-    } else {
-      stage("Push release tag") {
-        pushTag(repoName, env.BRANCH_NAME, 'release_' + env.BUILD_NUMBER)
-      }
+    if (env.BRANCH_NAME == "master") {
+      if (isGem()) {
+        stage("Publish Gem to Rubygems") {
+          publishGem(repoName, env.BRANCH_NAME)
+        }
+      } else {
+        stage("Push release tag") {
+          pushTag(repoName, env.BRANCH_NAME, 'release_' + env.BUILD_NUMBER)
+        }
 
-      stage("Deploy to integration") {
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -367,4 +367,156 @@ def withStatsdTiming(key, fn) {
...
+
+    stage("Run tests") {
+      // Prevent a project's tests from running in parallel on the same node
+      lock("$repoName-$NODE_NAME-test") {
+        runTests()
+      }
+    }
+
+    if (hasAssets()) {
+      stage("Precompile assets") {
```
suzannehamilton: We can skip this step if it's a schema test. Whitehall already does it, but I think it applies everywhere.

modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -367,4 +367,156 @@ def withStatsdTiming(key, fn) {
...
+        precompileAssets()
+      }
+    }
+
+    if (isGem()) {
```
suzannehamilton: We should check that this is the master branch before running any of this. It's a redundant check because `deployIntegration`, etc. also check the branch, but it stops "Deploy to integration" from appearing in branch build summaries.

PR #5619, opened 2017-03-15
Increase api-mongo-2 disk time window

gpeng:
api-mongo-2 runs 15 minutely backups that routinely alert for high disk time. The disk time window is set to 10 minutes in the base mongo class and then reduced to 7 (to try and fix this problem) in api_mongo.

This commit resets to 10 minutes just for api-mongo-2.

### DIFF
--- a/hieradata/class/api_mongo.yaml
+++ b/hieradata/class/api_mongo.yaml
@@ -38,4 +38,4 @@ mongodb::server::replicaset_members:
     hidden: true
     priority: 0
 
-icinga::client::checks::disk_time_window_minutes: 7
+icinga::client::checks::disk_time_window_minutes: 10


surminus:
If this happens on Staging too then you'd also have to add another node class called `hieradata/node/api-mongo-2.api.staging.publishing.service.gov.uk.yaml`. 

Upon reflection it might just be easier to raise the default in the api_mongo class to bring it into line with the mongo nodes.

gpeng:
I can't see this happening on staging but I've removed the override from `api_mongo.yml` so all machines should have the default 10 minutes. If this is ok I'll remove the first commit before merging.

surminus:
Sorry @gpeng, this is slightly confusing. The `hieradata/class/api_mongo.yaml` file relates only to `api-mongo-1` etc machines, whereas `hieradata/class/mongo.yaml` relates to `mongo-1` etc, so it would need to be in both files.

djsd123:
@surminus 
Unless I have misunderstood. you want @gpeng to amend the "Time window" for the `mongo-x` replica set as well as the `api-mongo` replicaset?  I thought this only relates to the `api-mongo` cluster? 
In which case, What @gpeng has done so far, is all that is required??

@gpeng Do you see the same alerts for the `mongo-x` cluster too?

PR #5623, opened 2017-03-16
Set DISPLAY env variable for tests

tijmenb:
This will allow capybara-webkit to run without using `xvfb-run -a` first.

Will allow us to port https://github.com/alphagov/specialist-publisher/pull/1017 to the new jenkins.

https://trello.com/c/iDWjE1Vz


tijmenb:
"Know what I'm doing" is maybe an overstatement :) For posterity: the `99` value comes from [the original specialist-publisher script](https://github.com/alphagov/specialist-publisher/commit/a903ce76b2d1f15c023698525100236ee8b9d53e). Some [tutorials on the internet](http://elementalselenium.com/tips/38-headless) also use `99`.

PR #5624, opened 2017-03-16
Revert removal of homepage no-cache rules

carolinegreen:
This re-opens https://github.com/alphagov/govuk-puppet/pull/5441 and add in configuration used originally in puppet for the original experiment.

And is part of running the no-js experiment again https://gds.blog.gov.uk/2013/10/21/how-many-people-are-missing-out-on-javascript-enhancement/

### DIFF
--- a/modules/govuk/manifests/apps/frontend.pp
+++ b/modules/govuk/manifests/apps/frontend.pp
@@ -47,6 +47,11 @@ class govuk::apps::frontend(
     nagios_memory_warning  => $nagios_memory_warning,
     nagios_memory_critical => $nagios_memory_critical,
     vhost                  => $vhost,
+    nginx_extra_config     => '
+  location ^~ /frontend/homepage/no-cache/ {
+    expires epoch;
+  }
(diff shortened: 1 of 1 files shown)


carolinegreen:
Hi @alexmuller I know you've moved off of infrastructure but I have talked to @afda16 off of infrastructure and we're not sure if it's okay to merge this or not as the original discussion in https://github.com/alphagov/govuk-puppet/pull/5441 came to a halt.  We were wondering if you had any more thoughts on this.  @tijmenb I don't know if you have any more thoughts on this either, as you were involved in the original discussion too?

afda16:
I think `nginx_extra_config` needs to be passed to the govuk::app class

alexmuller:
I think this makes sense as long as the nginx config is valid 👍 

I think removing all of the development specific logic would be cool and would make it a lot more simple.

carolinegreen:
@alexmuller Thanks!  I paired with @afda16 to change that.

PR #5625, opened 2017-03-16
Report triggered schema tests status separately.

danielroseman:
For downstream builds triggered by govuk-content-schemas, send the
pending/success/failed notification to the triggering commit on the
schemas repo. This will show the status individually in the Github UI
and permit re-running of individual tests where necessary.

(This will also need a change to the schemas test itself, to pass the commit sha and not wait for the downstream results.)

https://trello.com/c/5psKmk85

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -393,13 +393,21 @@ def buildProject(sassLint = true) {
         [$class: 'StringParameterDefinition',
           name: 'SCHEMA_BRANCH',
           defaultValue: 'deployed-to-production',
-          description: 'The branch of govuk-content-schemas to test against']]
+          description: 'The branch of govuk-content-schemas to test against'],
+        [$class: 'StringParameterDefinition',
+          name: 'SCHEMA_COMMIT',
+          defaultValue: 'invalid',
+          description: 'The commit of govuk-content-schemas that triggered this build, if it is a schema test']],
     ],
   ])
 
   try {
     if (!isAllowedBranchBuild(env.BRANCH_NAME)) {
       return
+    }
+
+    if (params.IS_SCHEMA_TEST) {
+      setBuildStatus(repoName, params.SCHEMA_COMMIT, "Downstream ${repoName} job is building on Jenkins", 'PENDING')
     }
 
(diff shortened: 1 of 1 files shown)


danielroseman:
@tijmenb @suzannehamilton what do you think?

modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -393,7 +393,11 @@ def buildProject(sassLint = true) {
         [$class: 'StringParameterDefinition',
           name: 'SCHEMA_BRANCH',
```
suzannehamilton: I wonder if we could rid of this parameter and use `SCHEMA_COMMIT` everywhere instead?

modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -526,4 +540,25 @@ def hasDatabase() {
...
+ * Manually set build status in Github.
+ *
+ * Useful for downstream builds that want to report on the upstream PR.
+ *
+ * @param repoName Name of the repo being built
+ * @param commit SHA of the triggering commit on govuk-content-schemas
+ * @param message The message to report
+ * @param state The build state: one of PENDING, SUCCESS, FAILED
+ */
+def setBuildStatus(repoName, commit, message, state) {
```
suzannehamilton: Just wrapping my head around this - this reports the build status against the content schemas, so a build error will show up on the correct govuk-content-schemas PR?

danielroseman: Yes, that's the intention - see https://github.com/alphagov/govuk-content-schemas/pull/572 for an example.

suzannehamilton:
This looks great! 👏

tijmenb:
This looks like a great solution! Really happy with this.

modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
danielroseman: Potentially yes, but it'll be messy if we try to do that before all the dependent apps are moved over to use buildProject so probably continue with both params until then.

PR #5627, opened 2017-03-16
Suppress output of Google credentials

surminus:
This ensures that the credentials are not visible in the Jenkins job output.


SamLR:
Will you rotate the current GCE credentials in case they've been previously exposed?

surminus:
That's a good point, probably should do.

PR #5628, opened 2017-03-16
Don't upgrade Jenkins in Staging or Prod

surminus:
We're testing in Integration, but we don't want it in Staging or Production yet, so keep the version to the same as it is currently.


djsd123:
@surminus Ta

PR #5630, opened 2017-03-16
Jenkins2 Deploy Box [Stage 2]

djsd123:
New node class to provision a Jenkins 2 instance

See: https://trello.com/c/RZCVbiHd

Requires: https://github.gds/gds/govuk-provisioning/pull/622

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1117,6 +1117,10 @@ hosts::production::management::hosts:
     ip: '10.1.0.3'
     legacy_aliases:
       - "deploy.%{hiera('app_domain')}"
+  deploy-1:
+    ip: '10.1.0.4'
+    legacy_aliases:
+      - "deploy-1.%{hiera('app_domain')}"
   puppetmaster-1:
     ip: '10.1.0.5'
     legacy_aliases:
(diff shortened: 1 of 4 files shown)


afda16:
Jenkins build is failing, I think you need to add the new host to Hiera, in hosts::production::management::hosts:

surminus:
Can we revert any changes that were previously made regarding upgrading `jenkins-1`?

hieradata/common.yaml
```diff
@@ -1117,6 +1117,10 @@ hosts::production::management::hosts:
     ip: '10.1.0.3'
     legacy_aliases:
       - "deploy.%{hiera('app_domain')}"
+  deploy-1:
+    ip: '10.1.0.4'
+    legacy_aliases:
+      - "deploy.%{hiera('app_domain')}"
```
surminus: I think this will conflict, can we change it to something different during testing?

surminus:
I opened a PR to revert changes:

https://github.com/alphagov/govuk-puppet/pull/5644

This may affect what plugins you add, but we need to do this to get Puppet running on `jenkins-1` in Integration again (I have a new job I'd like to run on it).

PR #5637, opened 2017-03-16
Add delayed_job worker to cpm

chao-xian:
The Content Performance Manager app needs delayed_job to start as a worker. Adding this call to the delayed_job service worker that's already defined in our puppet manifests.


chao-xian:
Closing as we've decided to switch to `govuk_sidekiq` instead for better support.

PR #5653, opened 2017-03-22
Reduce content store TTL

gpeng:
We expect cacheing to be 15 minutes by default. This commit updates the default TTL for content store.


gpeng:
Closing... think I might have dreamt the 15 minutes. There are plans to improve our use of 'proper' HTTP cacheing so this will go away.

PR #5655, opened 2017-03-22
Add some instructions for CI

tijmenb:
This used to be in the opsmanual, but I think that it's better here because otherwise it might get out of date.

Needs to be reviewed together with https://github.gds/gds/opsmanual/pull/897.

https://trello.com/c/iDWjE1Vz


tijmenb:
Changed my mind, this is better in the jenkinslib file itself. Moved the `buildProject()` method to the top of the file to make it easy.

PR #5657, opened 2017-03-23
[DO NOT MERGE] Add a jenkins job to dump routes and content every day

thomasleese:
The routes dump artefact will be used in a separate job which checks for consistency between the content store and the router-api.


thomasleese:
This will be done in a better way.

PR #5662, opened 2017-03-24
Test gems with all the version of Ruby that we currently support

suzannehamilton:
Update the shared Jenkins build configuration to run tests of gem projects under different Ruby versions.

Specify ruby versions 2.2 and 2.3, since all GOVUK Rails apps are using one or the other.

cc @tijmenb 

https://trello.com/c/zFGCUY9X

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -37,8 +37,10 @@
  * ```
  *
  * @param sassLint Whether or not to run the SASS linter
+ * @param extraRubyVersions Optional Ruby versions to run the tests against in
+ * addition to the versions currently supported by all GOV.UK applications
  */
-def buildProject(sassLint = true) {
+def buildProject(sassLint = true, extraRubyVersions = []) {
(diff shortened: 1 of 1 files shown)


rjw1:
Since 2.4 is available and I assume people will move to it eventually should we also test under that.

suzannehamilton:
@rjw1 - Some gems (e.g. gds-api-adapters) have test failures with 2.4, so we can't add it yet. But it would be a shame to not test those gems which _can_ support 2.4.

I've updated this PR to add an optional parameter which contains additional Ruby versions to test against. What do you think?

cc @tijmenb

tijmenb:
Hmm, having it parameterised means that we can't make testing against versions mandatory centrally. I'd rather fix it in the gems. Are there a lot of errors?

tijmenb:
Hmm, having it parameterised means that we can't make testing against versions mandatory centrally. I'd rather fix it in the gems. Are there a lot of errors?

suzannehamilton:
As discussed in person, the parameter is just for additional versions so we can still make versions mandatory. This just gives us a way to try out newer versions on some gems first.

suzannehamilton:
Having said all of that, all the gems that currently use the shared Jenkins build are fine with Ruby 2.4, so I've added it to the list. 👍 

I've left the parameter in place to make it easier to upgrade to future versions.

PR #5668, opened 2017-03-27
Fix name of `govuk-client-url_arbiter` project in hieradata

suzannehamilton:
Update project name to match repo and gem name to fix Jenkins build.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -574,7 +574,7 @@ govuk_ci::master::pipeline_jobs:
   govuk-diff-pages: {}
   govuk_admin_template: {}
   govuk_ab_testing: {}
-  govuk_client_url-arbiter: {}
+  govuk-client-url_arbiter: {}
   govuk_content_models: {}
   govuk-content-schema-test-helpers: {}
   govuk-dummy_content_store: {}


hieradata/common.yaml
```diff
@@ -574,7 +574,7 @@ govuk_ci::master::pipeline_jobs:
   govuk-diff-pages: {}
   govuk_admin_template: {}
   govuk_ab_testing: {}
-  govuk_client_url-arbiter: {}
+  govuk-client-url_arbiter: {}
```
jennyd: The [url-arbiter app is dead](https://github.com/gds-attic/url-arbiter) and the client should be too - we should be able to delete this.

suzannehamilton: It's still being used by info-frontend. 🙁 I'm not sure what the arbiter was, but I think the gem is still being used to parse API responses.

I suspect it wouldn't be very hard to remove from info-frontend, but it's not completely trivial (at least, not for someone who hasn't used url-arbiter or info-frontend before) so I'm trying to update it because it seemed like an easier yak to shave. That might not be true, though, so I'll take another look at removing it tomorrow.

PR #5669, opened 2017-03-28
Remove the lyndongarvey user

cbaines:
They don't need access anymore.

### DIFF
--- a/modules/users/manifests/lyndongarvey.pp
+++ b/modules/users/manifests/lyndongarvey.pp
@@ -1,6 +1,7 @@
 # Create the lyndongarvey user
 class users::lyndongarvey {
   govuk_user { 'lyndongarvey':
+    ensure   => absent,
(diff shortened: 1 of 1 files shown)


cbaines:
> Can you remove him from https://github.com/alphagov/govuk-puppet/blob/fcdc7275168717154e72e97d5b47499c8c263d06/hieradata/integration.yaml#L99 and https://github.com/alphagov/govuk-puppet/blob/fcdc7275168717154e72e97d5b47499c8c263d06/hieradata/integration.yaml#L303 too please?

I don't have a complete understanding of this, but I think removing https://github.com/alphagov/govuk-puppet/blob/fcdc7275168717154e72e97d5b47499c8c263d06/hieradata/integration.yaml#L303 would defeat the point of setting `ensure => absent` as then puppet wouldn't go and remove the user?

PR #5677, opened 2017-03-29
Add CI job for govuk-developer-docs

tijmenb:
In light of the recent token issues, I'd prefer to run the tests for this on our own machines instead of CircleCI.


tijmenb:
@surminus this was added to the `govuk_ci::master::pipeline_jobs`, so I think it won't be added to the deploy list in jenkins?

surminus:
Oh, I totally didn't view the whole context of change there! You're right.

PR #5679, opened 2017-03-29
Add a job which checks for content consistency

thomasleese:
This adds a job which runs various rake tasks checking for consistency across the `router-api`, `content-store` and `publishing-api`.

[Trello card](https://trello.com/c/0uqM1YHL/899-create-alert-for-new-inconsistent-documents-3)

### DIFF
--- a/modules/govuk_jenkins/manifests/job/check_content_consistency.pp
+++ b/modules/govuk_jenkins/manifests/job/check_content_consistency.pp
@@ -1,0 +1,11 @@
+# == Class: govuk_jenkins::job::check_content_consistency
+#
+# Create a file on disk that can be parsed by jenkins-job-builder
+#
+class govuk_jenkins::job::check_content_consistency {
+  file { '/etc/jenkins_jobs/jobs/check_content_consistency.yaml':
+    ensure  => present,
+    content => template('govuk_jenkins/jobs/check_content_consistency.yaml.erb'),
+    notify  => Exec['jenkins_jobs_update'],
+  }
+}
--- a/modules/govuk_jenkins/templates/jobs/check_content_consistency.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/check_content_consistency.yaml.erb
@@ -1,0 +1,91 @@
+---
+- scm:
+    name: check-content-consistency
+    scm:
+      - git:
+          url: git@github.gds:gds/router-data.git
+          basedir: router-data
+          branches:
+            - master
+- job:
+    name: Check_Content_Consistency
+    display-name: Check_Content_Consistency
+    project-type: freestyle
+    description: |
+      This job checks for consistency between the router-api, content-store
+      and publishing-api.
+    scm:
+      - check-content-consistency
+    logrotate:
+      numToKeep: 10
+    triggers:
+        - timed: 'H 5 * * *'
+    builders:
+      - shell: |
+          #!/bin/bash
+          set -eu
+
+          cd "${WORKSPACE}"
+
+          ROUTER_BACKEND=$(govuk_node_list --single-node -c router_backend)
+          DRAFT_CACHE=$(govuk_node_list --single-node -c draft_cache)
+          CONTENT_STORE=$(govuk_node_list --single-node -c content_store)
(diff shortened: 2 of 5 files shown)


modules/govuk_jenkins/manifests/job/check_content_consistency.pp
```diff
@@ -0,0 +1,11 @@
+# == Class: govuk_jenkins::job::check_content_consistency
+#
+# Create a file on disk that can be parsed by jenkins-job-builder
+#
+class govuk_jenkins::job::check_content_consistency() {
```
surminus: the parenthesis isn't required if there are no parameters being set

modules/govuk_jenkins/templates/jobs/check_content_consistency.yaml.erb
```diff
@@ -0,0 +1,87 @@
...
+        - timed: 'H 5 * * *'
+    builders:
+      - shell: |
+          #!/bin/bash
+          set -eu
+
+          cd "${WORKSPACE}"
+
+          function run_on_router_backend() {
+            ssh deploy@router-backend-1.backend.<%= @app_domain -%> "$@"
```
surminus: if you're using `@app_domain` this will need to be added as a parameter to the manifest

modules/govuk_jenkins/templates/jobs/check_content_consistency.yaml.erb
```diff
@@ -0,0 +1,87 @@
...
+          function run_on_content_store() {
+            ssh deploy@content-store-1.backend.<%= @app_domain -%> "$@"
+          }
+
+          function run_on_draft_content_store() {
+            ssh deploy@draft-content-store-1.backend.<%= @app_domain -%> "$@"
+          }
+
+          function run_on_backend() {
+            ssh deploy@backend-1.backend.<%= @app_domain -%> "$@"
```
surminus: As a rule we tend to suggest that you use a randomly selected (available) machine from that machine class ie choose something from the results of `govuk_node_list -c backend` rather than always choosing `backend-1`, unless there a specific reason it needs to be tied to that machine. If we need to take down a machine for maintenance and/or remove it from Puppet this would mean that your job would still run without failing.

surminus:
What's the outcome if this job fails? Should an alert?

modules/govuk_jenkins/templates/jobs/check_content_consistency.yaml.erb
thomasleese: @surminus Do you know if I actually need `@app_domain` to SSH into the machine?

thomasleese:
@kevindew This is ready for re-review. It's working fine on integration! 🎉

kevindew:
great - have do you have a link to a successful run? (would just make reviewing it easier)

thomasleese:
I don't have a link to a successful run, because whenever puppet runs on Jenkins it gets cleared. I have an [example alert](https://errbit.integration.publishing.service.gov.uk/apps/537ddefe0da115bffa0002de/problems/5900591165786317c83e1700) though.

kevindew:
Cool thanks - I notice we have lots of: `"Handler (gone) does not match expected item handler (backend).", "Backend ID () does not match item rendering app (frontend)."` I was wondering if that was an issue since Backend ID is always empty in the output - but I'm guessing that's because gones and redirects don't have them? In that case should we just not check backend id? - I realise this is not really related to this PR just the wider feature.

modules/govuk_jenkins/templates/jobs/check_content_consistency.yaml.erb
```diff
@@ -0,0 +1,91 @@
+---
+- scm:
+    name: check-content-consistency
+    scm:
+      - git:
+          url: git@github.gds:gds/router-data.git
+          basedir: router-data
+          branches:
+            - master
```
kevindew: Do we use this git clone?

thomasleese: Yes, we use it to include routes in the CSV files that aren't in the router dump.

thomasleese:
@kevindew Yeah, it seems reasonable to not check the Backend ID if we have already got that error about inconsistent handlers. I'll raise a PR about it.

PR #5686, opened 2017-03-29
Remove robinwhittleton user

robinwhittleton:
Don’t think I’ll need this access before I leave (31st).


cbaines:
Could someone who knows about govuk-puppet decide if we want to use `ensure => absent` here?

PR #5687, opened 2017-03-29
Add tests for collectd plugins

SamLR:
This adds tests for the Fastly CDN, Python, Elasticsearch and Etcd
collectd plugins. These plugins all contain some amount of logic
(e.g. conditionals or loops).

### DIFF
--- a/modules/collectd/files/usr/lib/collectd/python/mongodb.py
+++ b/modules/collectd/files/usr/lib/collectd/python/mongodb.py
@@ -1,5 +1,6 @@
 #
 # Plugin to collectd statistics from MongoDB
+# Source https://github.com/sebest/collectd-mongodb
 #
 
 import collectd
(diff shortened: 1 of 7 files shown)


modules/collectd/templates/etc/collectd/conf.d/etcd.conf.erb
```diff
@@ -76,7 +76,7 @@ stats_self = {
   ],
 }
 -%>
-  <URL "<%= @client_endpoint -%>/v2/stats/self">
+  <URL "http://<%= @client_endpoint -%>/v2/stats/self">
```
surminus: is there a reason this changes to http? If so, can we add this in the commit message?

SamLR: I think this was an error, I'll change it back.

PR #5690, opened 2017-03-30
Remove Frontend app from backend machines

brenetic:
We no longer use private-frontend and therefore need to remove it from any backend machines that it was previously deployed to.

https://trello.com/c/JYT9roQm/709-turn-off-private-frontend

cc @emmabeynon


brenetic:
Tested this on Integration. Removed all Frontend related bits and pieces. Ready for review now.

PR #5697, opened 2017-03-31
Fix govuk-delivery script command in data sync complete job

jennyd:
Oops, forgot the govuk_setenv which this script actually needs :(


jennyd:
[Error from this morning](https://deploy.integration.publishing.service.gov.uk/job/Data_Sync_Complete/390/console)

PR #5702, opened 2017-04-03
Add the govuk-guix repository to CI

cbaines:


cbaines:
Great, thanks for the review Laura :)

PR #5703, opened 2017-04-03
Remove mikaelallison from puppet.

deanwilson:
He is no longer with us.


cbaines:
I was looking at another user removal request here https://github.com/alphagov/govuk-puppet/pull/5686 and I'm unsure if using `ensure => absent` is useful or not?

PR #5710, opened 2017-04-04
Run fake virus scan in development VM

rubenarakelyan:
This commit adds a cron job for the development VM to run a fake virus scan every minute. The fake scan just copies files from the `upload` folder to the `clean` folder to simulate a virus scan.

This allows users of the development VM to do things like publishing documents with attachments in whitehall without having to run the rake task manually each time.

### DIFF
--- a/modules/clamav/manifests/run_fake_virus_scan.pp
+++ b/modules/clamav/manifests/run_fake_virus_scan.pp
@@ -1,0 +1,14 @@
+# == Class: clamav::run_fake_virus_scan
+#
+# Copy uploaded files from the `uploads` folder to the `clean` folder.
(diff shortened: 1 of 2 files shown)


rubenarakelyan:
Yup, I've made it run as the `vagrant` user now as well as specifying the path to `govuk_setenv` and redirecting all output to `/dev/null` so that you don't get mail building up in `/var/mail`.

cbaines:
> Yup, I've made it run as the vagrant user now as well as specifying the path to govuk_setenv and redirecting all output to /dev/null so that you don't get mail building up in /var/mail.

Looks good :+1:

PR #5711, opened 2017-04-04
Rename link-checker to govuk-link-checker

kevindew:
Repo has been renamed: https://github.com/alphagov/govuk-link-checker

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -42,6 +42,7 @@ deployable_applications: &deployable_applications
   govuk_content_api: {}
   govuk-content-schemas: {}
   govuk_crawler_worker: {}
+  govuk-link-checker: {}
   govuk_need_api: {}
   hmrc-manuals-api: {}
   imminence: {}
@@ -50,7 +51,6 @@ deployable_applications: &deployable_applications
     repository: 'kibana-gds'
   licencefinder:
     repository: 'licence-finder'
-  link-checker: {}
   local-links-manager: {}
   manuals-frontend: {}
   manuals-publisher: {}


tijmenb:
Can you elaborate on the thinking behind the renaming?  `govuk-link-checker` looks inconsistent with other applications and I think it might not conform to the naming RFC: https://gov-uk.atlassian.net/wiki/pages/viewpage.action?pageId=115081296

kevindew:
Sure @tijmenb - we were asked by @timblair to prefix that and all other new repos with "govuk" for differentiation from other GDS teams. Which sounds like something that conflicts with the RFC.

so by that RFC we'd be `link-checker-api` if we plan to append with `govuk` that'd get us to `govuk-link-checker-api` but is naturally inconsistent with past repos and RFC. 🤔

I still want to call it `link-checking-service` but that helps no-one 😀

timblair:
@tijmenb: yes, I asked @kevindew to update this.

As the number of repos on alphagov continues to grow, it's getting more and more difficult to find GOV.UK's ones. Most (all?) other teams have a consistent prefix to identify their repos; it's not great that GOV.UK's are identifiable as "the ones that don't look like the rest."

I appreciate that there's some inconsistency with other repos (well, inconsistency seems to be the only consistent thing here...), but I'd like to start trying to be consistent going forward. We probably want to consider an update to the naming RFC.

kevindew:
@timblair would it be an idea to have different organisation namespaces? would make it easier to grok repos in an org

timblair:
Yes, that would certainly make it easier to group repos, but it's not something we're going to do right now.

tijmenb:
I agree with `govuk-` for everything except deployable applications. 

For consistency with existing applications, but also keep the app name the same between GitHub, hostname and puppet. Disparity between those is still causing problems with content-api/govuk_content_api/content_api.

I know that there's a lot of inconsistency, but there's currently no application prefixed with `govuk-`. Unless we have plans to rename everything, "govuk-link-checker" will be inconsistent with others for years to come. If I look at the list on https://docs.publishing.service.gov.uk/apps.html I think it'd be really weird to call it `govuk-x`.

If it helps, as part of our docs blitz we hope we'll be tagging all GOV.UK-owned repos consistently, so that search becomes easier (https://trello.com/c/CdGSaozQ).

timblair:
Your argument is convincing, and has convinced me. It still annoys me that our repos are identifiable as being the odd-ones-out, but I can learn to live with it.

Sorry, @kevindew, can you change your repo back again, please? 😢

PR #5714, opened 2017-04-05
Create whitehall asset directory structure in dev VM

rubenarakelyan:
This commit creates the whitehall asset directory structure in the development VM. This allows locally-uploaded assets to be served correctly.


rubenarakelyan:
Ana has also reviewed this from the infrastructure side

PR #5716, opened 2017-04-05
Add Link Checker API

steventux:
Adds the Link Checker API application to development and integration environments.

Depends on https://github.gds/gds/deployment/pull/1277

See also https://github.com/alphagov/govuk-app-deployment/pull/120

### DIFF
--- a/hieradata/class/backend.yaml
+++ b/hieradata/class/backend.yaml
@@ -62,6 +62,7 @@ govuk::node::s_base::apps:
   - hmrc_manuals_api
   - imminence
   - kibana
+  - link_checker_api
   - local_links_manager
   - manuals_publisher
   - maslow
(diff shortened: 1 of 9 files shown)


kevindew:
Hopefully I've added the missing bits that make this 🍏 - does it need entries for the load balancer?

Also I wonder if this belongs more on the api tier of servers than backend? Not really sure what criteria is for choosing between them.

steventux:
@kevindew I did wonder about this application belonging to the api tier, who can advise us on this?
https://github.com/alphagov/govuk-puppet/pull/5716/commits/b59ee3b218942d53a1c8fb45c2886ff1c4393c29 has loadbalanacer entry

kevindew:
Not really sure who we ask maybe @deanwilson ? failing that could just wander to infra and bug whoever looks up 😀 

ah cool I was wondering about definitions in the [hieradata](https://github.com/alphagov/govuk-puppet/blob/master/hieradata/integration.yaml#L174-L178) but looking more closely that seems to be just for special cases?

steventux:
@kevindew @boffbowsh I've compared available resources on backend and api class machines, backends look much more suited to running this application.

deanwilson:
I'm happy with @steventux take, I don't think we have ay hard rules on what goes where. I'll raise it in the meeting on Tuesday for future clarification.

PR #5731, opened 2017-04-06
Auto generate a self signed certificate

mgrassotti:
Auto generate a self signed certificate while provisioning the VM.


mgrassotti:
the `.sh` script works but it it is not automatically executed from the Vagrantfile...
Maybe something is wrong with the line ```config.vm.provision :shell, :path => "generate_self_signed_certificate.sh"```

Comments are welcome.

mgrassotti:
Ah thanks @cbaines. Well, now it should be done anyway...

rubenarakelyan:
I'm going to close this PR since we're going to go with the SSL certificates that are already in the VM.

PR #5732, opened 2017-04-06
Enable docker on all CI agents

deanwilson:


rjw1:
Does this even have to be feature flagged now? 
When are you going to not want docker on a ci-agent?

PR #5743, opened 2017-04-07
Add content training environment

afda16:
This commit adds all the configuration required to build a content training environment based on the development VM. The environment can be built using either Vagrant or Terraform in AWS. There are a number of provisioning scripts that are run as part of the build. A Jenkins job is also defined to plan/apply/destroy the VM as required, and this is scheduled to run once a week.


training-vm/Vagrantfile
```diff
@@ -0,0 +1,34 @@
...
+    aws.tags["Name"]= "training"
+    aws.block_device_mapping = [ 
+      {
+        'DeviceName' => "/dev/sda1",
+        'Ebs.VolumeSize' => 100,
+        'Ebs.DeleteOnTermination' => true
+      },
+    ]
+    aws.iam_instance_profile_name = ENV['AWS_IAM_INSTANCE_PROFILE_NAME']
+    aws.region = ENV['AWS_REGION']
```
SamLR: AMIs are locked to a region so either the AMI should be configurable (or, ideally, the latest imaged fetched) or the region should be locked to `eu-west-1`.

I've opened a support ticket to recreate default subnets on our default VPC so that (should) remove the need to manually set the `subnet_id`.

Otherwise this looks good to me.

PR #5744, opened 2017-04-07
Add env vars for Local Links Manager GA credentials

sihugh:
These are the credentials that are used to import Google Analytics data to
Local Links Manager.


sihugh:
These apply to the creds set up in https://github.gds/gds/deployment/pull/1278

sihugh:
Thanks @cbaines!

PR #5750, opened 2017-04-10
Add dashboards for most deployable apps.

suzannehamilton:
We have excluded the following deployable apps from this list:
    
* backdrop-read
* backdrop-write
* bouncer
* designprinciples
* email-alert-service
* errbit
* govuk-deliver
* govuk-cdn-logs-monitor
* govuk-content-schemas
* govuk_crawler_worker
* kibana
* performanceplatform-admin
* performanceplatform-big-screen-view
* router
* sidekiq-monitoring
* smokey
* spotlight
* stagecraft
    
We decided to focus on apps that speak HTTP and are part of GOV.UK
publishing.

https://trello.com/c/vJodL9Cc/36-generate-dashboard-from-a-template-in-puppet

Mobbed with @dwhenry and @MatMoore

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -755,8 +755,10 @@ grafana::dashboards::deployment_applications:
   contacts-frontend: {}
   content-performance-manager: {}
   content-store: {}
-  content-tagger: {}
-  email-alert-api: {}
+  content-tagger:
+    has_workers: true
+  email-alert-api:
+    has_workers: true
   email-alert-frontend: {}
   feedback: {}
   finder-frontend: {}
@@ -765,7 +767,8 @@ grafana::dashboards::deployment_applications:
   govuk_content_api: {}
   govuk_need_api: {}
   hmrc-manuals-api: {}
-  imminence: {}
+  imminence:
+    has_workers: true
   info-frontend: {}
   licencefinder: {}
   link-checker-api: {}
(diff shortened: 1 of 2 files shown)


jennyd:
govuk-delivery does very similar things to email-alert-api, so it would be helpful to have dashboards for both (even though we're trying to kill it) - is it more difficult because python?

MatMoore:
@jennyd I think it depends on whether we capture the metrics in the same way for email-alert-api. It would be good to avoid spending too much time on individual apps if we can avoid it but we haven't actually checked it yet (we'll make sure to do this before merging though). If it works we can definitely add it in.

MatMoore:
@boffbowsh what do you think about the list of apps in this PR? Are there any others we haven't included that we should have?

suzannehamilton:
@jennyd We (the mob) just had a look at creating a dashboard with our current setup for govuk_delivery. At this stage most of the graphs don't work. We think a dashboard makes sense but it would be a custom dashboard rather than autogenerated. This should be out of scope of this PR.

Once this PR is merged we will be getting feedback on the current dashboard and expect to make subsequent changes as a result of feedback.

jennyd:
@suzannehamilton that makes sense - I think lots of the standard stats are only set up for the `rack` app type, so depending on which stats you're using for the dashboards it's probably more work for anything non-ruby.

boffbowsh:
😍 for this PR. Just been taking a look on Integration - this is so good. 👏s all round!

PR #5758, opened 2017-04-11
Add building of Docker images in Jenkins

boffbowsh:
If a repository has a `Dockerfile`, build it. If tests pass (in the
traditional way, not using Docker yet), push the image to the public
Docker hub tagged with the branch name. If it’s the master branch, also
push an image tagged with the same name as the release tag, e.g.
`release_123`.

This isn't pre-judging the output of the infrastructure team's Q1 work, this is to allow us to easily run the end to end Publishing tests using externally hosted Docker images.

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -127,6 +127,12 @@ def buildProject(options = [:]) {
       }
     }
 
+    if (hasDockerfile()) {
+      stage("Build Docker image") {
+        buildDockerImage(repoName, env.BRANCH_NAME)
+      }
+    }
+
     if (hasLint()) {
       stage("Lint Ruby") {
         rubyLinter("app lib spec test")
@@ -182,6 +188,12 @@ def buildProject(options = [:]) {
       }
     }
 
+    if (hasDockerfile()) {
+      stage("Push Docker image") {
+        pushDockerImage(repoName, env.BRANCH_NAME)
+      }
+    }
+
     if (env.BRANCH_NAME == "master") {
       if (isGem()) {
         stage("Publish Gem to Rubygems") {
@@ -190,6 +202,12 @@ def buildProject(options = [:]) {
       } else {
         stage("Push release tag") {
           pushTag(repoName, env.BRANCH_NAME, 'release_' + env.BUILD_NUMBER)
+        }
+
+        if (hasDockerfile()) {
+          stage("Tag Docker image") {
+            pushDockerImage(repoName, env.BRANCH_NAME, "release_${env.BUILD_NUMBER}")
+          }
         }
 
         stage("Deploy to integration") {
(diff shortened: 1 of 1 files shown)


boffbowsh:
Example build with Groovy script edited with the "Replay" function: https://ci.integration.publishing.service.gov.uk/job/content-store/job/master/627/console

boffbowsh:
And the output of that build https://hub.docker.com/r/govuk/content-store/tags/

boffbowsh:
I should mention this isn't pre-judging the output of the infrastructure team's Q1 work, this is to allow us to easily run the end to end Publishing tests using externally hosted Docker images.

surminus:
I've been away so may have totally missed this, but if we're going to start building lots of images across lots of CI agent machines can we ensure we the relevant tooling in place for maintenance of the docker hosts (ie so we don't burn through disk space). It might help to build a "govuk" base image that uses common functionality rather than pulling from the "ruby" base image each time for each repo as well.

This comment may also just be totally superfluous :)

boffbowsh:
Yep, good point @surminus. Do we have any prior art for keeping them tidy?

I'm not sure about govuk images yet. We'd need to maintain images per ruby version, and we have varying binary dependencies per repo too (eg imagemagick, PDFkit) that don't need to be included everywhere. I think building from the main Ruby image is fine for this particular use case.

modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -192,6 +204,10 @@ def buildProject(options = [:]) {
           pushTag(repoName, env.BRANCH_NAME, 'release_' + env.BUILD_NUMBER)
         }
 
+        stage("Tag Docker image") {
```
boffbowsh: Thanks, done.

SamLR:
I added some notes about cleaning up docker to the [opsmanual](https://github.gds/pages/gds/opsmanual/2nd-line/nagios.html#low-available-disk-space-on-jenkins) which may be a reasonable starting point

surminus:
We haven't really written much tooling around docker yet because we were using it so minimally. 

May be worth considering having dockerhub doing the builds and pushes as it would be less dependence on having to maintain our infra, but that's probably a conversation for a different time.

boffbowsh:
How about this as a mitigation:

Overnight, every day:

1. Stop containers that have been running > 24 hours
2. Remove all exited containers
3. Remove (not force) images over 7 days old
4. Remove (not force) orphaned images

This should allow us to keep the cached layers that speed up the builds, while removing the stale application or gem dependency layers. I used a similar process at my old job and it worked quite well. We can tweak those thresholds as needed.

boffbowsh:
re: dockerhub builds - the lag between a github push and the build actually starting can be up to 15 minutes. I'd like to use this as a way of testing the publishing platform end to end tests. The alternative would be to build with CircleCI or similar I guess?

surminus:
Process for cleaning up sounds good :+1: 

Regarding dockerhub etc, I think that's for another time. I think this is fine as a way to get things built and tested for publishing e2e tests, certainly better than building everything each time.

surminus:
I've added dedicated mounts for docker to the CI machines so cleanup isn't entirely required anymore. On that basis this PR is ready to go.

deanwilson:
Looks good. Thanks @boffbowsh

PR #5761, opened 2017-04-11
Improve usage output for the govuk_puppet script

SamLR:


modules/puppet/templates/govuk_puppet
```diff
@@ -54,7 +70,7 @@ Examples:
 
   # newer style with explicit lock names
   # be explicit with the reason and the name of the lock
-  $app_name -n postgres_backups -r 'Running postgres restore' --disable
+  $app_name -n postgres_backups -r '<your name>: Running postgres restore' --disable
```
surminus: I think it would make more sense if the command was: 

`govuk_puppet -n lauramartin -r "doing some stuff" --disable` 

We added the name flag so that automated scripts could call themselves something and we would know it was an automated thing that locked puppet.

PR #5771, opened 2017-04-12
Add extra useful links to deployment dashboard

brenetic:
Icinga is also looked at often during deployments

### DIFF
--- a/modules/grafana/templates/dashboards/deployment_panels/_links.json.erb
+++ b/modules/grafana/templates/dashboards/deployment_panels/_links.json.erb
@@ -1,5 +1,10 @@
 {
-  "content": "<ul style=\"font-size: 20px;\">\n<li><a target=\"_blank\" href=\"https://errbit.<%= @app_domain %>/\">Errbit</a></li>\n</ul>\n",
+  "content": "<ul style=\"font-size: 20px;\">\n
+      <li><a target=\"_blank\" href=\"https://errbit.<%= @app_domain %>/\">Errbit</a></li>\n
+      <li><a target=\"_blank\" href=\"https://alert.<%= @app_domain %>/\">Icinga</a></li>\n
+      <li><a target=\"_blank\" href=\"https://docs.publishing.service.gov.uk/apps/<%= @docs_name %>.html\"><%= @app_name %> documentation</a></li>\n
+      <li><a target=\"_blank\" href=\"https://govuk.zendesk.com/agent/filters/135081425\">Dashboard feedback</a></li>\n
+    </ul>\n",
   "editable": true,
   "error": false,
   "id": 12,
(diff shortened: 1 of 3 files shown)


modules/grafana/templates/dashboards/deployment_panels/_links.json.erb
```diff
@@ -1,5 +1,10 @@
 {
-  "content": "<ul style=\"font-size: 20px;\">\n<li><a target=\"_blank\" href=\"https://errbit.<%= @app_domain %>/\">Errbit</a></li>\n</ul>\n",
+  "content": "<ul style=\"font-size: 20px;\">\n
+      <li><a target=\"_blank\" href=\"https://errbit.<%= @app_domain %>/\">Errbit</a></li>\n
+      <li><a target=\"_blank\" href=\"https://alert.<%= @app_domain %>/\">Icinga</a></li>\n
+      <li><a target=\"_blank\" href=\"https://docs.publishing.service.gov.uk/apps/<%= @app_name %>.html\"><%= @app_name %> documentation</a></li>\n
```
suzannehamilton: Do the dev docs use the same app name that the dashboards do? Or are they based on repo name, which doesn't always match?

modules/grafana/templates/dashboards/deployment_panels/_links.json.erb
```diff
@@ -1,5 +1,10 @@
...
+      <li><a target=\"_blank\" href=\"https://govuk.zendesk.com/agent/filters/135081425\">Feedback</a></li>\n
```
suzannehamilton: Maybe worth wording it as something like "Dashboard feedback" so that it's clear that it's not feedback on the application itself?

dwhenry: Did we want to add a github link or are we assuming they can get there themselves?

modules/grafana/templates/dashboards/deployment_panels/_links.json.erb
brenetic: That is a good point. The docs use the repo name. Therefore something like contacts-admin would not link correctly

modules/grafana/templates/dashboards/deployment_panels/_links.json.erb
brenetic: First link on the app page in the docs is the Github repo. We could save them an extra click . . . ¯\_(ツ)_/¯

brenetic:
@suzannehamilton @dwhenry Take 3

suzannehamilton:
LGTM.

Not sure about `docs_name` or `repository`. I wouldn't be very surprised if we had to use the repo name for something else, but `docs_name` makes the purpose clearer so it might be OK to stick with it for now and change it if we need to.

PR #5783, opened 2017-04-13
Update processor count panel

dwhenry:
Processor counts were not being displayed for all application, investigation revealed this was caused by limiting to servers with backend in the name. This change removes that limitation.

* move label to the right hand side
* add current processors net to label
* remove `backend` from graphite query
* Shorten name from full graphite name to server name


modules/grafana/templates/dashboards/deployment_panels/_processor_count.json.erb
```diff
@@ -51,7 +53,7 @@
         }
       ],
       "refId": "A",
-      "target": "*_backend*.processes-app-<%= @app_name %>.ps_count.processes",
+      "target": "aliasByNode(*.processes-app-whitehall.ps_count.processes,0)",
```
suzannehamilton: Should still be `@app_name`?

dwhenry: updated now.

PR #5786, opened 2017-04-13
Move error counts to table

dwhenry:
This reduces the number of single value panels on the dashboard

* Remove deploy count, this is better displayed with processor count and deploy
  lines
* move 5min errors and period errors into the error table
* allow highlighting when error exceed threshold, with threshold able to be customised
  for each app

https://trello.com/c/ODsAbilp/41-make-sure-puppet-triggers-a-restart-of-grafana-web-after-dashboards-are-changed-or-the-package-is-upgraded-downgraded

screenshot 2017-04-18 12 03 35


modules/grafana/templates/dashboards/deployment_panels/_error_counts_table.json.erb
```diff
@@ -23,36 +23,39 @@
...
         "rgba(237, 129, 40, 0.89)",
-        "rgba(50, 172, 45, 0.97)"
+        "rgba(245, 54, 54, 0.9)"
       ],
-      "decimals": 2,
+      "decimals": 0,
       "pattern": "/.*/",
-      "thresholds": [],
+      "thresholds": [
+        "<%= @error_threshold / 2 %>",
```
suzannehamilton: Does this do the right thing if it's not a whole number? (Possibly a silly question, but puppet and Grafana both seem to have a lot of restrictions so I thought I'd check.)

dwhenry: it should do as int / number = int of course @error_threshold is a string so need to cast it to int first. :( just testing it all locally now

modules/grafana/templates/dashboards/deployment_panels/_error_counts_table.json.erb
```diff
@@ -23,36 +23,39 @@
...
         "rgba(237, 129, 40, 0.89)",
-        "rgba(50, 172, 45, 0.97)"
+        "rgba(245, 54, 54, 0.9)"
       ],
-      "decimals": 2,
+      "decimals": 0,
       "pattern": "/.*/",
-      "thresholds": [],
+      "thresholds": [
+        "<%= @error_threshold.to_i / 2 %>",
```
MatMoore: I think this should be a separate `@warning_threshold` rather than fixing it to half the error threshold.

dwhenry: Done

PR #5788, opened 2017-04-13
Configure DNS for training.publishing

boffbowsh:
Configures a new Hosted Zone for training.publishing.service.gov.uk and
outputs the NS servers. These should be used to delegate to in the main
publishing.service.gov.uk zone.
Creates a`vm-1` A record pointing at the VM’s public IP, and CNAMEs `*`
to `vm-1`.

This PR will merge into @afda16's Terraform branch.


afda16:
We applied these changes in the terraform project and add-training-environment branch, feel free to delete

PR #5795, opened 2017-04-18
Increase disk space on mirrorer-1

surminus:
The size of the site we're pulling is now becoming sufficiently large that we need to increase the default size of the crawler mount (50GB). This commit adds a new disk to increase the size of this mount.


surminus:
Hmm, this may be redundant, there seem to be some large disks attached to the VMs but not mounted.

surminus:
The extra disks aren't being used so this change is safe to go ahead with.

PR #5798, opened 2017-04-18
Use events to display server deploys

dwhenry:
Event data exists for all environments which whisper (time stream)
data does not. As a result moving to use event data means deploys
are visible in all environments.

Events are documented: http://graphite.readthedocs.io/en/latest/events.html

https://trello.com/b/T9cPzxHP/gov-uk-search-doing

Event data is recorded slightly earlier than whisper data as you can see in the below picture, however as there is less than 5sec in it I feel we should change over.

screenshot 2017-04-18 12 19 01


suzannehamilton:
Tried this out manually on Integration and it looks good. 🚀

PR #5802, opened 2017-04-18
Rename run training environment script

rubenarakelyan:
This commit renames the “run training environment” script to make it clear it uses bowl rather than upstart. It also adds a comment to this effect in the file.


rubenarakelyan:
I've amended the comment to hopefully make it clearer what it's used for, which is development of the training environment.

cbaines:
> I've amended the comment to hopefully make it clearer what it's used for, which is development of the training environment.

Ok, thanks :)

PR #5803, opened 2017-04-18
Hide error/slow responses graphs if data is missing

MatMoore:
The graphs at the bottom of the deployment dashboard come from
elasticsearch. For a lot of dashboards, these are empty because the
events aren't logged in a consistent way.

For apps where the data is not available, hide these panels. Also hide
them for low usage apps where there won't be enough events to see
trends.

In the case of the error graph, the graph itself is only useful when
there are errors and you want to break it down by controller, so we
might want to change how we present this for apps that don't error all
the time. A series is only drawn on the graph if there is an event so
the graph is just blank if everything is fine, which is confusing.

https://trello.com/c/lVNtP9A4/47-hide-errors-by-controller-and-slow-responses-by-controller-graphs-on-dashboards-where-they-have-no-data

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -750,36 +750,85 @@ grafana::dashboards::machine_suffix_metrics: ''
 grafana::dashboards::deployment_applications:
   asset-manager: {}
   business-support-api: {}
-  calculators: {}
+  calculators:
+    # Status, duration, controller missing @fields. prefix
+    show_controller_errors: false
+    show_slow_requests: false
   calendars: {}
   collections: {}
-  collections-publisher: {}
-  contacts-frontend: {}
-  content-performance-manager: {}
+  collections-publisher:
+    # Low usage
+    show_controller_errors: false
+    show_slow_requests: false
+  contacts-frontend:
+    # Missing data in kibana
+    show_controller_errors: false
+    show_slow_requests: false
+  content-performance-manager:
+    # Missing duration, status, controller fields
+    show_controller_errors: false
+    show_slow_requests: false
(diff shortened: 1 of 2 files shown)


modules/grafana/manifests/dashboards/deployment_dashboard.pp
```diff
@@ -32,23 +32,39 @@
...
+  $show_controller_errors = true,
+  $show_slow_requests = true
 ) {
   if $has_workers {
     $worker_row = [['worker_failures', 'worker_successes']]
   } else {
     $worker_row = []
   }
 
+  if $show_controller_errors and $show_slow_requests {
```
suzannehamilton: Could we put these graphs into separate dashboard rows? It would make this logic simpler and I don't think it would change the appearance much.

hieradata/common.yaml
```diff
@@ -750,47 +750,105 @@ grafana::dashboards::machine_suffix_metrics: ''
 grafana::dashboards::deployment_applications:
   asset-manager: {}
   business-support-api: {}
-  calculators: {}
+  calculators:
+    # Status, duration, controller missing @fields. prefix
```
suzannehamilton: Mostly out of interest - is this a problem with the way the data is being reported? This app has a field called `controller` rather than the expected `@fields.controller`?

MatMoore: Yeah - I haven't looked into why yet though.

MatMoore:
Updated to separate the graphs onto different row, mainly to simplify the horrible puppet syntax when building up the array of partials.

PR #5805, opened 2017-04-18
Add HTTP basic auth to training environment

rubenarakelyan:
This commit adds HTTP basic auth to the training environment. This mimics the integration environment and stops search engines from indexing content or users accidentally visiting and mistaking it with production.

### DIFF
--- a/modules/govuk/manifests/app/nginx_vhost.pp
+++ b/modules/govuk/manifests/app/nginx_vhost.pp
@@ -94,11 +94,19 @@ define govuk::app::nginx_vhost (
     $nginx_extra_config_real = $nginx_extra_config
   }
 
+  # Force HTTP basic auth for the training environment, since it doesn't have
+  # the cache servers which do this in the integration environment
+  if $::govuk_node_class == 'training' {
+    $really_protected = true
+  } else {
+    $really_protected = $protected
+  }
+
   nginx::config::vhost::proxy { $vhost:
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/app/nginx_vhost.pp
```diff
@@ -94,11 +94,19 @@
     $nginx_extra_config_real = $nginx_extra_config
   }
 
+  # Force HTTP basic auth for the training environment, since it doesn't have
```
afda16: Instead of using if - else for a specific facter inside a generic class, I would override the parameter vhost_protected in Hiera, for instance:
hieradata/development.yaml:govuk::apps::whitehall::vhost_protected: false
inside hieradata/class/training.yaml with true

rubenarakelyan: That would be ideal, but it seems like only 1 or 2 apps have a per-app configuration for HTTP basic auth. The main integration auth is done on the cache servers (https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/node/s_cache.pp#L52 and https://github.com/alphagov/govuk-puppet/blob/master/hieradata/integration.yaml#L197) which we don't have in the training environment. Do you know if there's a better way of doing this?

rubenarakelyan:
Closing to look at a better solution for this

PR #5807, opened 2017-04-18
Docker cleanup cron task

kevindew:
This runs the commands listed in the opsmanual to clean up docker space
that is built up from dangling images.

A more thorough alternative to this could be to look at something like:
https://github.com/meltwater/docker-cleanup

### DIFF
--- a/modules/govuk_ci/manifests/agent/docker.pp
+++ b/modules/govuk_ci/manifests/agent/docker.pp
@@ -10,4 +10,15 @@ class govuk_ci::agent::docker {
 
   include ::docker::compose
 
+  cron::crondotdee { 'remove_docker_dangling_images' :
+    hour    => 10,
+    minute  => 0,
+    command => 'docker rmi $(docker images --filter "dangling=true" -q --no-trunc)',
+  }
+
+  cron::crondotdee { 'remove_docker_dangling_volumes' :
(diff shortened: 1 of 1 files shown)


modules/govuk_ci/manifests/agent/docker.pp
```diff
@@ -10,4 +10,13 @@
 
   include ::docker::compose
 
+  # Docker can hog a lot of space by not cleaning up after itself, this removes
+  # dangling images and volumes.
+  $remove_dangling_images = 'docker rmi $(docker images --filter "dangling=true" -q --no-trunc)'
+  $remove_dangling_volumes = 'docker volume rm $(docker volume ls -qf dangling=true)'
+  file { '/etc/cron.d/clean_dangling_docker_files':
```
SamLR: Rather than using a raw file could this be done with the crondotdee resource which will give you a more readable form? ([for example](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/node/s_mirrorer.pp#L20) & [source](https://github.com/alphagov/govuk-puppet/blob/master/modules/cron/manifests/crondotdee.pp))

kevindew:
Thanks @SamLR - I've made those changes, it looks a lot cleaner (less need for comments)

PR #5811, opened 2017-04-19
Add docker backend class.

deanwilson:
This will eventually hold the dockerised backend applications. It currently
does not include the docker class as we're testing `::govuk::node::s_base`
as an include rather than an inherits.


surminus:
:+1: still good

PR #5812, opened 2017-04-19
Docker changes

surminus:
This PR does two things:

* Adds docker by default on the CI agents
* Updates the govuk_docker class so that CI can make use of it

### DIFF
--- a/modules/govuk_docker/manifests/init.pp
+++ b/modules/govuk_docker/manifests/init.pp
@@ -2,9 +2,26 @@
 #
 # Install and run docker
 #
-class govuk_docker {
+# === Parameters:
+#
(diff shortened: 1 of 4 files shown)


modules/govuk_docker/manifests/init.pp
```diff
@@ -2,9 +2,26 @@
...
+#   Specify an array of users to add to the docker group
+#   Default is empty
+#
+# [*version*]
+#   Pin a version to use
+#   Default is latest
+#
+class govuk_docker (
+  $docker_users = [],
+  $version = 'latest',
```
deanwilson: I think we should pin this to `present`

PR #5814, opened 2017-04-19
Disable Rails 5 DB environment check

kevindew:
In ActiveRecord on Rails 5 before being able to drop databases you need
to run `rake db:environment:set` otherwise you can't run `db:drop` etc.

Unfortunately the `db:environment:set` command requires the DB to exist
which is a pain on CI. Setting this environment variable removes the
need.


tijmenb:
👍 this is totally fine I think, because we're setting `RAILS_ENV=test` here anyway.

kevindew:
@tijmenb nah unfortunately it still blows up with that - we had to get around it with [this](https://github.com/alphagov/link-checker-api/blob/master/Jenkinsfile#L6) on link-checker-api

PR #5818, opened 2017-04-20
First draft at managing apps as containers

surminus:
This commit adds a defined type to pull and run a specified container.

It's function is meant to be the beginning of a construction similar to the govuk::app class, but does the minimum neccessary to pull an image with the given tag, and run that image.

It allows adding environment variables using encrypted hieradata. These will be placed in an init upstart script for the time-being before we use proper secret management in docker.

We're using the [release](https://hub.docker.com/r/govuk/release/) app for testing our processes.

### DIFF
--- a/hieradata/vagrant.yaml
+++ b/hieradata/vagrant.yaml
@@ -31,8 +31,17 @@ govuk::apps::stagecraft::celerycam::enabled: true
 
 # An example configuration for env vars
 govuk_containers::apps::release::envvars:
-  DATABASE: 'password'
-  BLUE: 'green'
+  - "GOVUK_ENV=production"
+  - "NODE_ENV=production"
+  - "RACK_ENV=production"
+  - "RAILS_ENV=production"
+  - "ERRBIT_ENVIRONMENT_NAME=%{hiera('govuk::deploy::config::errbit_environment_name')}"
+  - "GOVUK_APP_DOMAIN=%{hiera('app_domain')}"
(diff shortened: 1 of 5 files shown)


surminus:
I've updated the image_tag to the latest release here: https://hub.docker.com/r/govuk/release/tags/

hieradata/vagrant.yaml
```diff
@@ -29,6 +29,20 @@ govuk::apps::stagecraft::worker::enabled: true
...
+  - "NODE_ENV=production"
+  - "RACK_ENV=production"
+  - "RAILS_ENV=production"
+  - "ERRBIT_ENVIRONMENT_NAME=%{hiera('govuk::deploy::config::errbit_environment_name')}"
+  - "GOVUK_APP_DOMAIN=%{hiera('app_domain')}"
+  - "GOVUK_ASSET_HOST=%{hiera('govuk::deploy::config::asset_root')}"
+  - "GOVUK_ASSET_ROOT=%{hiera('govuk::deploy::config::asset_root')}"
+  - "GOVUK_WEBSITE_ROOT=%{hiera('govuk::deploy::config::website_root')}"
+  - "DATABASE=password"
+  - "BLUE=green"
```
brenetic: 🔵 ❌ = 💚

surminus: 2+2=5

surminus:
Let's merge this and then think about how we manage global env vars better later.

PR #5819, opened 2017-04-20
Update Deploy_Training Jenkins job defaults

afda16:
Our current version of job-builder doesn't support parameterized schedules.
To be able to rebuild the AWS box automatically, we update the job parameters
defaults to do an apply and use a different instance name when the job runs.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_training.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_training.yaml.erb
@@ -47,7 +47,7 @@
                credential-id: <%= @credential_id_govuk_training_ssh_key %>
                variable: TF_VAR_training_ssh_key_path
     triggers:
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/templates/jobs/deploy_training.yaml.erb
```diff
@@ -47,25 +47,25 @@
...
             terraform ${ACTION} .
     parameters:
         - string:
             name: INSTANCE_NAME
-            default: govuk-training
-            description: Set VM name tag
+            description: Set VM name tag. If this parameter is empty, the builder uses govuk-training-YYYYMMDD
         - choice:
             name: ACTION
             description: Choose whether to plan (default), apply or destroy
```
rubenarakelyan: "Choose whether to plan, apply (default) or destroy"

PR #5826, opened 2017-04-20
Fix newStyleDockerTags

SamLR:
There was a problem where the timestamp wasn't being trimmed which meant
tags contains newlines causing builds to fail. This fixes the bug and
tidies some other bits of the code (e.g. adding guards and docs).


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -692,12 +692,26 @@ def buildDockerImage(imageName, tagName) {
   docker.build("govuk/${imageName}:${tagName}")
 }
 
+
+/*
+ * Push the image to the govuk docker hub and tag it. If asTag is set the image
```
deanwilson: "is set the image is give" needs rewording.

PR #5827, opened 2017-04-21
Enable CORS for elasticsearch on dev VM

MatMoore:
I'm trying out a tool that communicates with elasticsearch through the
browser, which needs CORS.

This shouldn't affect production.

### DIFF
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -199,6 +199,7 @@ govuk_lvm::no_op: true
 govuk_mount::no_op: true
 
 govuk_elasticsearch::version: '1.7.5'
+govuk_elasticsearch::cors_enabled: true
 
 govuk_logging::logstream::disabled: true
 
--- a/modules/govuk_elasticsearch/manifests/init.pp
+++ b/modules/govuk_elasticsearch/manifests/init.pp
@@ -24,6 +24,17 @@
 # [*backup_enabled*]
 #   Boolean. Whether backup class will be included.
 #
+# [*cors_enabled*]
+#   Boolean. Whether cross-origin-resource-sharing is enabled.  i.e. whether a
+#   browser on another origin can execute requests against Elasticsearch.
+#
+# [*cors_accept_headers*]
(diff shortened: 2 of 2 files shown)


hieradata/development.yaml
```diff
@@ -199,6 +199,8 @@ govuk_lvm::no_op: true
 govuk_mount::no_op: true
 
 govuk_elasticsearch::version: '1.7.5'
+govuk_elasticsearch::cors_enabled: true
+govuk_elasticsearch::cors_allow_headers: 'X-Requested-With, Content-Type, Content-Length, If-Modified-Since'
```
suzannehamilton: Why is `If-Modified-Since` in this list, but not in the list of defaults in the `govuk_elasticsearch` class?

MatMoore: I set the defaults to the elasticsearch defaults, but I think I'll change it so it's just a simple feature toggle in the hieradata.

modules/govuk_elasticsearch/manifests/init.pp
```diff
@@ -38,7 +48,10 @@
   $disable_gc_alerts = false,
   $manage_repo = true,
   $open_firewall_from_all = false,
-  $backup_enabled = false
+  $backup_enabled = false,
+  $cors_enabled = false,
+  $cors_allow_headers = 'X-Requested-With, Content-Type, Content-Length',
+  $cors_allow_origin = 'http://app.quepid.com'
```
suzannehamilton: Just checking my own understanding: are we adding the quepid config to the defaults so that we can start using quepid in production by just flipping the `cors_enabled` flag (or adding it to production hiera)?

MatMoore: Not exactly - we don't need to use production data for this tool, and I'm only using a free trial of Quepid at the moment - it's normally a paid service.

Thought I'd add in the host there because the default is to allow all hosts - if anyone does switches on CORS for something else they should probably specify the origin.

PR #5830, opened 2017-04-21
Configure bundler to use gemstash

SamLR:


modules/govuk/manifests/node/s_base.pp
```diff
@@ -109,4 +109,8 @@
   user { 'ubuntu':
     ensure => absent,
   }
+
+  if $::domain =~ /^.*\.(staging|integration)?\.publishing\.service\.gov\.uk/ {
```
surminus: Had a chat in meatspace about whether to have this enabled by default and disabled specifically in development.

modules/govuk/manifests/node/s_base.pp
```diff
@@ -109,4 +110,8 @@
   user { 'ubuntu':
     ensure => absent,
   }
+
+  govuk_bundler::config { 'deploy-bundler':
```
afda16: If the default of the user_home is related with the deploy user, I would put this resource next to the 'deploy' user, that I think it is govuk::deploy::setup, and make sure there is a require that ensures /home/deploy exists before creating this resource

modules/govuk_bundler/manifests/config.pp
```diff
@@ -0,0 +1,32 @@
...
+# [*user_home*]
+#   The home directory of the user to add the config to.
+#   Default: /home/deploy
+#
+# [*server*]
+#   The gemstash server to use
+#   Default: http://gemstash.cluster
+#
+define govuk_bundler::config(
+  $user_home = '/home/deploy',
```
afda16: If this is a generic define to be reused in different classes, I don't know if it's better to make the user_home mandatory, so we don't make any assumptions about a specific user

PR #5834, opened 2017-04-21
Add global environment variables

surminus:
This commit creates an "env_file" which is used by each container that is started using the govuk_containers::app defined type. It separates the management of app specific env vars and global env vars rather than trying to merge arrays of both into the same variable.

On testing they seem to correctly get loaded (using `docker inspect` to check):

  ```
  vagrant@docker-backend-1:~$ sudo docker inspect release |grep -A10 Env
            "Env": [
                "GOVUK_ENV=development",
                "NODE_ENV=development",
                "RACK_ENV=development",
                "RAILS_ENV=development",
                "ERRBIT_ENVIRONMENT_NAME=vagrant",
                "GOVUK_APP_DOMAIN=dev.gov.uk",
                "GOVUK_ASSET_HOST=http://static.dev.gov.uk",
                "GOVUK_ASSET_ROOT=http://static.dev.gov.uk",
                "GOVUK_WEBSITE_ROOT=http://www.dev.gov.uk",
                "FOO=BAR",
  ```

### DIFF
--- a/hieradata/vagrant.yaml
+++ b/hieradata/vagrant.yaml
@@ -29,20 +29,6 @@ govuk::apps::stagecraft::worker::enabled: true
 govuk::apps::stagecraft::beat::enabled: true
 govuk::apps::stagecraft::celerycam::enabled: true
 
-# An example configuration for env vars
-govuk_containers::apps::release::envvars:
-  - "GOVUK_ENV=production"
-  - "NODE_ENV=production"
-  - "RACK_ENV=production"
-  - "RAILS_ENV=production"
-  - "ERRBIT_ENVIRONMENT_NAME=%{hiera('govuk::deploy::config::errbit_environment_name')}"
-  - "GOVUK_APP_DOMAIN=%{hiera('app_domain')}"
-  - "GOVUK_ASSET_HOST=%{hiera('govuk::deploy::config::asset_root')}"
-  - "GOVUK_ASSET_ROOT=%{hiera('govuk::deploy::config::asset_root')}"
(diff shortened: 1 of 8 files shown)


hieradata/vagrant.yaml
```diff
@@ -29,19 +29,19 @@ govuk::apps::stagecraft::worker::enabled: true
...
-# An example configuration for env vars
 govuk_containers::apps::release::envvars:
-  - "GOVUK_ENV=production"
-  - "NODE_ENV=production"
-  - "RACK_ENV=production"
-  - "RAILS_ENV=production"
+  - "FOO=BAR"
+
+govuk_containers::app::config::global_envvars:
+  - "GOVUK_ENV=development"
```
afda16: I think govuk_env defaults to 'production' in the deploy::config class, and we only override with 'development' in:
`hieradata/development.yaml:govuk::deploy::config::govuk_env: 'development'`

modules/govuk_containers/manifests/app.pp
```diff
@@ -17,6 +17,10 @@
 # [*envvars*]
 #   An array of environment variables to start the container with.
 #
+# [*global_envs*]
```
afda16: minor update to global_env_file

afda16:
I think a bin file has gone in the commit:
modules/govuk_containers/manifests/app/.config.pp.swp

Where are now the ...::release::envvars configured?

surminus:
I've removed the bin file, thanks! Addressed comment.

The app env vars are configured in this PR: https://github.com/alphagov/govuk-puppet/pull/5837

PR #5835, opened 2017-04-21
Point /var/apps to correct app directory in development VM

rubenarakelyan:
This commit adds a `repo_name` variable to `govuk::app` and `govuk::app::package`, to be used where the app name is different from its repository name. Setting `repo_name` ensures that symlinks from `/var/apps` to `/var/govuk` in the development VM point to the correct directory.

### DIFF
--- a/modules/govuk/manifests/app/package.pp
+++ b/modules/govuk/manifests/app/package.pp
@@ -1,6 +1,7 @@
 # FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
 define govuk::app::package (
   $vhost_full,
+  $repo_name = $title,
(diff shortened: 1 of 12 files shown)


modules/govuk/manifests/app/package.pp
```diff
@@ -55,9 +56,13 @@
 
   # Otherwise, assume the apps are checked out straight into /var/govuk.
   } else {
+    if $repo_name == undef {
```
afda16: Tests are failing, I think because the new parameter is mandatory. I think it's better to do   $repo_name = $title in the define (), so it always takes a default value

rubenarakelyan: I wasn't sure if you could do that in the `define` - done!

PR #5836, opened 2017-04-21
Add jenkins jobs to run search benchmarks

MatMoore:
These tests run the rummager healthcheck, which benchmarks search
against a set of judgements. One job checks for the presence/absense of
particular documents in query results, and the other tests that the "did
you mean" returns suitable suggestions for misspelled queries.

These used to run on the old CI server, which we replaced. These take a
long time to run, so we're using the deploy jenkins and running them
nightly instead of incorporating it into the rummager CI pipeline.

We'd like to test this on integration before rolling it out to production.

Before rolling it out to production we'll create an alert page in the developer docs for these jobs (at the moment the notes url is is a dead link).

Paired with @brenetic 
Trello: https://trello.com/c/TXeHcOnY/53-revive-the-search-healthcheck

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/benchmark_search.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/benchmark_search.yaml.erb
@@ -1,0 +1,53 @@
+---
+- scm:
+    name: rummager_<%= @job_name %>
+    scm:
+        - git:
+            url: git@github.com:alphagov/rummager.git
+            branches:
+              - master
+
+- job:
+    name: <%= @job_name %>
+    display-name: <%= @job_name %>
+    project-type: freestyle
+    description: |
+      <p><%= @service_description %></p>
+      <p>More information:</p>
+      <ul>
+        <li><a href='https://github.com/alphagov/rummager/blob/master/docs/health-check.md'>Healthcheck docs on Github</a></li>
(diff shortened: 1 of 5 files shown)


modules/govuk_jenkins/templates/jobs/benchmark_search.yaml.erb
```diff
@@ -0,0 +1,46 @@
...
+      </ul>
+    scm:
+        - rummager
+    builders:
+        - shell: |
+            bundle install --path &quot;${HOME}/bundles/${JOB_NAME}&quot;
+            bundle exec bin/health_check --download
+            bundle exec bin/health_check --json=https://www-origin.<%= $app_domain %>/api/search.json --type=<%= @test_type %>
+    triggers:
+        - timed: '30 5 * * *'
```
suzannehamilton: Just to check: is it OK for both of these jobs to run at once?

MatMoore: That was unintentional. It's probably fine but I'll stagger them just in case.

suzannehamilton:
@MatMoore: I've added a third commit which adds the rate limit token. It uses the new health check parameter added in alphagov/rummager#773.

PR #5838, opened 2017-04-24
Add Haproxy profile for Docker machines

afda16:
- Add classes and defines to manage Haproxy in the docker backend
profile.

- Override upstream global options that expects rsyslog listening
on eth0, and add rsyslog snippet to capture Haproxy logs

### DIFF
--- a/modules/govuk_containers/manifests/balancermember.pp
+++ b/modules/govuk_containers/manifests/balancermember.pp
@@ -1,0 +1,42 @@
+# == Define: Govuk_containers::Balancermember
+#
+# A defined type which creates a Haproxy backend with a balancermember
+# for the container application
+#
+# === Parameters:
+#
+# [*port*]
+#   Container listening port.
+#
+# [*ipaddress*]
+#   Container listening ipaddress
+#
+# [*options*]
+#   Haproxy backend server options
+#   Default: 'check'
+#
+define govuk_containers::balancermember (
+  $port,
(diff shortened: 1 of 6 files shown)


surminus:
You need to squash my commit away

modules/govuk_containers/manifests/frontend/haproxy.pp
```diff
@@ -0,0 +1,64 @@
...
+        'httplog',
+        'accept-invalid-http-request',
+      ],
+      'use_backend'    => '%[req.hdr(host),lower,map(/etc/haproxy/domains-to-backends.map,bk_default)]',
+    },
+  }
+
+  haproxy::mapfile { 'domains-to-backends':
+    ensure   => 'present',
+    mappings => [
```
surminus: Could this be defined in hiera in the future? Just thinking about rapid adoption of all our apps as containers, and the less per app config we can add the better (IMO).

modules/govuk_containers/manifests/balancermember.pp
```diff
@@ -1,6 +1,6 @@
 # == Define: Govuk_containers::Balancermember
 #
-# A defined type which creates a Haproxy backend with alancermember 
+# A defined type which creates a Haproxy backend with balancermember
```
surminus: Squash this into the previous commit if possible.

surminus:
Couple of comments but otherwise looks good!

PR #5841, opened 2017-04-24
Add docker Haproxy class to rest of environments

afda16:
Update Hieradata to apply govuk_containers::frontend::haproxy to
the rest of environments


afda16:
Already in https://github.com/alphagov/govuk-puppet/pull/5842

PR #5854, opened 2017-04-28
Change image tag on containers

surminus:
The Puppet code that used the ::docker::image and ::docker::run classes did not fit into our current deployment model as it meant Puppet managed which images were running, whereas we need a job which deploys a newer image and then restarts the service.

the ::docker::run defined type is useful as it creates an init script, so we're going to have a generic "release_env" tag, and then use `docker image tag` to point this specific tag to a specific image.


afda16:
Looks good

PR #5856, opened 2017-04-28
Fix error caused by invalid chars in icinga desc

dwhenry:


modules/govuk_jenkins/manifests/job/search_test_spelling_suggestions.pp
```diff
@@ -11,8 +11,8 @@
 
   $test_type = 'suggestions'
   $job_name = 'search_test_spelling_suggestions'
-  $service_description = 'Check for spelling suggestions (see https://docs.google.com/spreadsheets/d/1JjSoy68vscNjrvQm8b9hHt0nbZgFxk8lrcTdqV08iHk)'
-  $job_url = "https://deploy.${app_domain}/job/search_test_spelling_suggestions/"
+  $service_description = 'Check for spelling suggestions'
+  $job_url = "https://deploy.${app_domain}/job/ /"
```
suzannehamilton: Should this line have been changed?

dwhenry: no.. no it should not..

reverted now

PR #5864, opened 2017-05-02
Enable Logspout for docker logging

SamLR:
Add a logspout dockerfile and use it for logging containers.

Merge with deploy [PR](https://github.gds/gds/deployment/pull/1304)


surminus:
This looks good but I think it might be doing that weird travis thing again...

PR #5873, opened 2017-05-03
Rate limit contact/govuk urls for POST requests only

h-lame:
For: https://trello.com/c/vNJvbs2g/119-simple-ip-rate-limiting-for-named-contact-requests-to-reduce-spam

We were rate limiting /contact/govuk/service-feedback and
/contact/govuk/problem_reports already, but for all requests.  The problem
is that /contact/govuk and /contact/govuk/foi are both restful and so we
want to allow the GET requests through, but limit the POST requests that
actually submit.  To do this we introduce a new map that toggles on both
the Rate-Limit-Token header and request method.  This map sets up a
variable that is blank for anything that isn't a POST request, and also
blank for a POST request that has a correct Rate-Limit-Token header.  All
other requests have the binary version of the remote address as the value.

This variable is then used in the "contact" `limit_req_zone` to decide which
requests to rate limit.  It'll let all blank values through, but rate limit
any others to 6 requests per minute (with a 5Mb cache).  Note that the limit
zone is unchanged, apart from the variable used as the key.

We then add all POST requests for the feedback app to the existing location
that uses this `limit_req_zone`.  We'll need to keep this in sync with any
new POST endpoints in the feedback app.

### DIFF
--- a/modules/router/templates/router_include.conf.erb
+++ b/modules/router/templates/router_include.conf.erb
@@ -51,7 +51,9 @@ location ~ ^/performance/ {
   proxy_pass http://varnish;
 }
 
-location ~ ^/contact/govuk/(service-feedback|problem_reports)$ {
+# This list of URLS should be kept in sync with the POST-able routes in the
+# feedback application
+location ~ ^/contact/govuk/(|service-feedback|problem_reports|foi|page_improvements|email-survey-signup)$ {
   limit_req zone=contact burst=4 nodelay;
   proxy_pass http://varnish;
 }
--- a/modules/router/templates/rate-limiting.conf.erb
+++ b/modules/router/templates/rate-limiting.conf.erb
@@ -7,9 +7,21 @@ map $http_Rate_Limit_Token $limit_req {
 }
 
 limit_req_zone $limit_req zone=rate:30m rate=10r/s;
-limit_req_zone $limit_req zone=contact:5m rate=6r/m;
 limit_req_zone $limit_req zone=performance:5m rate=5r/s;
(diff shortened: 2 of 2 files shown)


h-lame:
FWIW I've tested this locally in the dev VM by including this config (tweaked slightly because my local nginx isn't a cache node so has no varnish to `proxy_pass` onto) into the router.dev.gov.uk nginx config, and it seems to work! 

The [original change that introduced the map](https://github.com/alphagov/govuk-puppet/commit/35581af1f3d124fd3854e5381726c1da188d520a) expressed some concern that the map might add some overhead, so I'd be keen for someone having a good thing about that while reviewing because I've added another one here.

modules/router/templates/router_include.conf.erb
```diff
@@ -51,7 +51,9 @@ location ~ ^/performance/ {
   proxy_pass http://varnish;
 }
 
-location ~ ^/contact/govuk/(service-feedback|problem_reports)$ {
+# This list of URLS should be kept in sync with the POST-able routes in the
```
surminus: How are we going to ensure this is kept in sync? :)

h-lame: You mean this comment won't do it on it's own? ;)

Good question, I have no idea - is there a good place to document it?  I guess something in the README for the feedback app would be a good start.

surminus: I think that's the best we can do at the moment as nginx doesn't really support env vars in a nice way and that'd be the best thing IMO.

PR #5879, opened 2017-05-04
Bugfix removing interims

surminus:
These contractors were removed in 579a500570adc1880a441207998f62c77245d3e1, but not removed from all references in Puppet. Their user class was deleted, but still referenced in the, which caused Puppet to get upset.


h-lame:
Oops!  Sorry.

PR #5880, opened 2017-05-05
Tweak getting started instructions.

leenagupte:
Removed step to install GCC tools. Everything works without out this  tools and doesn't appear that later versions of OSX are compatible with them.


leenagupte:
@cbaines Thanks 😸

PR #5881, opened 2017-05-05
Add calendars as dependency to publisher

h-lame:
The default message for the fact-check email in publisher needs to know how to
calculate working days and it does so by reading one of the bank holidays json
files hosted by the calendars app.  This seems like a small dependency, but the
default message for the fact-check email is rendered on every visit to the
email edit form.  If calendars isn't running then we break publisher which
would probably be painful in dev.

For: https://trello.com/c/VSg9iWPO/156-factcheck-email-include-the-date-in-email

See also: 

* https://github.com/alphagov/publisher/pull/621 - introduces the dependency on calendars to publisher
* https://github.com/alphagov/gds-api-adapters/pull/702 - introduces the new api adapter for calendars

This probably shouldn't be merged until the above PRs are.

### DIFF
--- a/development-vm/Pinfile
+++ b/development-vm/Pinfile
@@ -57,7 +57,7 @@ process :'multipage-frontend' => [:static, :'content-store']
 process :need_api
 process :'performanceplatform-admin' => [:stagecraft]
 process :'policy-publisher' => [:'publishing-api', :rummager]
-process :publisher => [:'publishing-api', 'publisher-worker'] # for some requests also uses: mapit
+process :publisher => [:'publishing-api', 'publisher-worker', :calendars] # for some requests also uses: mapit
 process :'publishing-api' => [:'content-store', :'draft-content-store', :'router-api', :'draft-router-api', :'publishing-api-worker']
 process :'publishing-api-worker' => [:'content-store', :'router-api']
 process :release


tijmenb:
Just a random thought: might it be better to treat calendars as an external dependency and point publisher to production GOV.UK?

h-lame:
Oooh.  Interesting.  It would make [this line](https://github.com/alphagov/publisher/pull/621/files#diff-ee054dafd7eadee4c8baf208077209baR18) in publisher look a bit weird/inconsistent though right? It'd be a raw www.gov.uk URL instead of using `Plek`.  Maybe that's fine, but then it'd be harder to configure to use a different source.

Not sure I mind either way tbh.

cbaines:
> Oooh. Interesting. It would make this line in publisher look a bit weird/inconsistent though right? It'd be a raw www.gov.uk URL instead of using Plek. Maybe that's fine, but then it'd be harder to configure to use a different source.

My recommendation would be to configure Publisher with `PLEK_SERVICE_CALENDARS_URI=https://www.gov.uk/whatever`, if you indeed want to use the public endpoint. This way you keep using Plek, and its configurable through the interface that Plek provides.

h-lame:
If I get this right, then @cbaines you mean to replace this PR with one that sets a plek env var for calendars for the dev vm only?  And leave my PR for publisher alone?

cbaines:
> If I've got this right, then @cbaines you mean to replace this PR with one that sets a plek env var for calendars for the dev vm only? And leave my PR for publisher alone?

I don't have opinions on the govuk-puppet development-vm, I only commented to add that you can configure Publisher in the way that Tijmen described by using Plek.

PR #5886, opened 2017-05-08
Cope with structure.sql when resetting a Jenkins test db

h-lame:
Some rails apps use structure.sql instead of schema.rb to describe their
database.  The task for loading this is not `db:schema:load` but
`db:structure:load`.  Happily rails provides `db:reset` which not only
makes sure to run the correct schema vs. structure task but also does
the drop and create steps too.  It also runs `db:seed` which may or
may not be a problem for our test databases, however there aren't that
many of our apps that have things in their `db/seeds.rb` so we'll try
it out and see.

Encountered while [upgrading support-api to rails 5](https://github.com/alphagov/support-api/pull/109).

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -165,7 +165,7 @@ def buildProject(Map options = [:]) {
     lock("$repoName-$NODE_NAME-test") {
       if (hasDatabase()) {
         stage("Set up the database") {
-            runRakeTask("db:drop db:create db:schema:load")
+            runRakeTask("db:reset")
         }
       }
 


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -165,7 +165,7 @@ def buildProject(Map options = [:]) {
     lock("$repoName-$NODE_NAME-test") {
       if (hasDatabase()) {
         stage("Set up the database") {
-            runRakeTask("db:drop db:create db:schema:load")
+            runRakeTask("db:drop db:create db:schema:load_if_ruby db:structure:load_if_sql")
```
andrewgarner: Rails supplies `db:setup` which does these steps – https://github.com/rails/rails/blob/master/activerecord/lib/active_record/railties/databases.rake#L178-L179

Furthermore, it supplies `db:reset` which looks to do everything we need (though I haven't tested) – https://github.com/rails/rails/blob/master/activerecord/lib/active_record/railties/databases.rake#L143-L144

h-lame: They both also run `db:seed` which I'm not entirely sure we'll want to be run on our test databases.

andrewgarner: The [documentation](http://edgeguides.rubyonrails.org/active_record_migrations.html#migrations-and-seed-data) suggests that seeding is "especially useful when reloading the database frequently in development and test environments".

As a developer I would expect to be able to use seed data and not have to add a workaround to make it work in CI.

h-lame: I've always assumed `db:seed` to be used for bootstrap data in dev and prod, but not in test and managed that kind of thing on a per test basis with factories, or across the suite with fixtures.  I'm not entirely sure what we do here, but as we weren't previously using `db:seed` in this Jenkins script I assume it's more like that.

That said, we should probably get more eyes on this though as we don't want to go against the prevailing wind wrt. Rails conventions.

andrewgarner: 👍

kevindew: My vote would go for `db:reset`

h-lame: One thought I just had is that when you run the tests locally, you don't run `rake db:seed` before you run the tests, you just run the tests.  I'm not 100% convinced that having CI run the seeds automatically before running tests is a good thing in this case because it's different to what a dev would be doing when running the tests.  If we wanted the seeds to act like fixtures in the test dbs, it'd be worth making that explicit in each app that needs it so that everyone makes sure it's run.  What rails will run before each test run (if invoked via rake, not necessarily via rspec) is `rake db:test:prepare` which checks for missing migrations and then runs `db:schema:load` (or `db:structure:load` as appropriate).

h-lame: With two votes for `db:reset` I [had a look](https://github.com/search?p=1&q=org%3Aalphagov+db%3Aseed&type=Code&utf8=%E2%9C%93) and most of our rails apps have an empty `db:seed` so it's ultimately not going to affect them if we do run the seeds.  I'll change this to us `db:reset` and we can see if it affects those that do have seeds:

* [specialist publisher](https://github.com/alphagov/specialist-publisher/blob/master/db/seeds.rb) - has a guard against running in envs other than dev, so it almost certainly won't be a problem
* [contacts-admin](https://github.com/alphagov/contacts-admin/blob/master/db/seeds.rb) - this will make a call to the network to import organisations - not sure how well that'll work
* [release](https://github.com/alphagov/release/blob/master/db/seeds.rb)
* [imminence](https://github.com/alphagov/imminence/blob/master/db/seeds.rb) - will load anything in `db/seeds/*.rb` - but there aren't any
* [router-api](https://github.com/alphagov/router-api/blob/master/db/seeds.rb) - there are files in `db/seeds/*.rb` but they seem normal
* [publisher](https://github.com/alphagov/publisher/blob/master/db/seeds.rb) - there's one file in `db/seeds/*.rb` for business support facets which I suspect can maybe be deleted (that's what the only one in imminence was and it's been removed).

PR #5895, opened 2017-05-10
Fix mismatched brackets in govuk_jenkinslib.groovy

cbaines:


cbaines:
I've admin merged this, as it's attempting to fix the issue that is currently causing the CI process to fail.

PR #5900, opened 2017-05-10
Enable worker for Content Performance Manager

chao-xian:
Enable the sidekiq worker for Content Performance Manager, which is now
using the `govuk_sidekiq` gem.

Related: https://github.com/alphagov/content-performance-manager/pull/118

### DIFF
--- a/modules/govuk/manifests/apps/content_performance_manager.pp
+++ b/modules/govuk/manifests/apps/content_performance_manager.pp
@@ -54,6 +54,7 @@ class govuk::apps::content_performance_manager(
   $db_name = 'content_performance_manager_production',
   $db_password = undef,
   $db_username = 'content_performance_manager',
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/content_performance_manager.pp
```diff
@@ -76,6 +77,10 @@
     app    => $app_name,
   }
 
+  govuk::procfile::worker {'content-performance-manager':
```
cbaines: You could use `$app_name` here instead of `content-performance-manager` which is what the manifests for some other apps do.

chao-xian:
Thanks @cbaines - I've amended as you've suggested and fixed up.

cbaines:
> Thanks @cbaines - I've amended as you've suggested and fixed up.

:+1:

PR #5902, opened 2017-05-10
Allow pushing images to GCR

SamLR:
This adds a function that pushes docker images to GCR. It uses shell
commands rather than the docker plugin as I could find no way to
integrate the docker plugin with the GCR credentials file. There is a
plugin but it does not seem to support use in pipeline builds.


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -709,6 +698,24 @@ def pushDockerImage(imageName, tagName, asTag = null) {
   }
 }
 
+def pushDockerImageToGCR(imageName, tagName) {
+  gcrName = "gcr.io/govuk-test/${imageName}"
+  docker.build(gcrName)
```
surminus: Should this build with the tagName variable?

PR #5909, opened 2017-05-12
Add details about separately downloading DB data

timblair:
This info was previously hidden on a page about Troubleshooting Vagrant.


whoojemaflip:
Is this information definitely on the critical path for  a `getting started` guide? I tried to pare this guide down to just the basics, with links to reference materials if the user wanted a more advanced reference. I agree that `Troubleshooting Vagrant` maybe isn't the best place for this info, but is the `Getting Started Guide for new developers`?

timblair:
Yeah, I did consider moving it out to a separate page entirely ("How to Get Data") and just linking to that from the Getting Started guide. Thought I'd start with just moving this here as a first step, but happy to go all the way if that's preferred.

whoojemaflip:
It's a pretty good candidate for it's own page IMHO

PR #5910, opened 2017-05-12
Authenticate CI using GitHub teams

alexmuller:


surminus:
This code is much nicer than it is currently! 👏  

One thing I noted was that previously we had an "anonymouse" (🐭 ) user, and this enabled an ability to use the API without auth which was used for some monitoring. We might need to do some work to authenticate an API user for this purpose. 

Also I'm not sure if it's worth adding a spec test to ensure it populates the template as expected?

alexmuller:
Thanks! I've added a spec so fingers crossed it will pass :crossed_fingers:

alexmuller:
I wasn't aware that's what the anonymous user was used for, shall I add it back in? Should be quite straightforward.

surminus:
If you could that'd be good; I was wary over having full read access for an anonymous user, but I think it should be OK.

Spec test passed, thanks! :+1:

alexmuller:
Will do!

@jennyd had some useful concerns this morning and I wanted to be more explicit about the trade-off we're making here. As far as I can tell the 2 choices we have to authenticate Jenkins with GitHub are:

- A list of GitHub usernames (this is explicit, requires a commit and deploy to add but is a list of usernames which don't necessarily map to real names which might mean we miss some people on removal)
- A GitHub team (it's probably easier to accidentally or maliciously add a user to a team than it is to add them to a list in YAML, but when they leave GDS their access is revoked by removal from the alphagov organisation)

As far as I can tell there's no way to implement our ideal situation, which is that somebody needs to be in a YAML list but _also_ needs to be a member of the alphagov organisation.

surminus:
I think authing through a Github team is better than adding to a list of YAML. IMO having _more_ places to add people is more likely to cause a problem. Having one list in Github to be used with a variety of things makes it easier and less likely to have people hanging around.

We should just auth and add roles using our LDAP system for _everything_. One place and one team who manages all users and roles.

PR #5913, opened 2017-05-15
Add content performance manager job configuration

cbaines:


hieradata/production.yaml
```diff
@@ -102,6 +102,9 @@ govuk_jenkins::job::network_config_deploy::environments:
   - 'carrenza-production-dr'
   - 'skyscape-production'
 
+govuk_jenkins::job::content_performance_manager:
```
surminus: This should be `govuk_jenkins::job::content_performance_manager::rake_import_all_content_items_frequency: '0 2 * * *'` (a single line, otherwise it thinks it's a hash when it's really a parameter string)

cbaines: Thanks, splitting it was my attempt at avoiding a long line. I've changed it now to be as you posted.

pmanrubia:
I have rebased and force pushed in preparation for merging.

PR #5914, opened 2017-05-15
Add transition logs

surminus:
This commit adds a new class which enables running the code to push transition logs to the appropriate Github repositories.The code for doing this currently exists as it's own machine in UKCloud. This is the last machine that is still in UKCloud, and is not monitored.

The original code exists [here](https://github.com/alphagov/ci-puppet/blob/master/modules/ci_environment/manifests/transition_logs.pp)

---
How transition logs work:

  - Our CDN pushes logs for bouncer to a machine
  - These logs get rotated hourly by logrotate
  - A cronjob runs each hour that pulls two repositories from Github
  - Within one repository is a rake task that creates statistics based
  upon the logs
  - When the stats are compiled, they are committed to a git repo and
  pushed to master
  - Transition application then reads in the latest statistics
---

The Govuk_cdnlogs class already takes care of most of the work for the
actual logs, so this commit needs to provide user and scripts to run the
process.

The SSH private key will be added separately and to each Github repo.
Previously there was a different key for each repository, but this time
round I think it is OK to use the same key to authenticate with Github.

### DIFF
--- a/modules/govuk_cdnlogs/files/logs_processor_config
+++ b/modules/govuk_cdnlogs/files/logs_processor_config
@@ -1,0 +1,3 @@
+compression: gzip
+pattern: "cdn-bouncer.log-*.gz"
+format: redirector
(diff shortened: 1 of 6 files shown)


surminus:
I need to do a very careful managed roll out of this.

PR #5916, opened 2017-05-15
Set envvar for hmrc-manuals-api log request body

alexmuller:
We need to configure this using an environment variable rather than config from GitHub Enterprise.


alexmuller:
No longer needed apparently!

PR #5930, opened 2017-05-17
Run CPM migrations after data sync

tijmenb:
We sync the staging databases to integration every night. When a migration has run on integration but not production, this sometimes causes the application to error if it relies on the migrations being run.


cbaines:
Cool, thanks for this @tijmenb :)

PR #5931, opened 2017-05-17
Configure docker to log to syslog in production

SamLR:
Curently we only have a logit stack configured for integration. This
means that in staging and production docker logs (i.e. for gemstash)
aren't being collected.

This sets syslog as the log driver for production and staging. Ideally
we should aim to move away from this as shipping directly to logstash
via betas is a preferred solution to syslog.


surminus:
merge away

PR #5936, opened 2017-05-17
Add memcached container to frontend-docker 2

deanwilson:
We're going to experiment with a dockerised, central, memcached in
the frontend integration environment

### DIFF
--- a/modules/govuk/manifests/node/s_docker_frontend.pp
+++ b/modules/govuk/manifests/node/s_docker_frontend.pp
@@ -1,7 +1,19 @@
 # == Class: govuk::node::s_docker_frontend
 #
-class govuk::node::s_docker_frontend {
+# === Parameters
+#
+# [*enable_memcache*]
+#   Whether to install memcached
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/node/s_docker_frontend.pp
```diff
@@ -1,7 +1,19 @@
...
+# [*enable_memcache*]
+#   Whether to install memcached
+#
+class govuk::node::s_docker_frontend(
+  $enable_memcache = false,
+) {
 
   include ::govuk::node::s_base
 
+  if $enable_memcache == true {
```
surminus: Don't need the `== true`

modules/govuk/manifests/node/s_docker_frontend.pp
```diff
@@ -1,7 +1,19 @@
...
+    include ::govuk_docker
```
surminus: This should be enabled by default.

PR #5940, opened 2017-05-18
Backdrop transaction jenkins job

ambrozic:
Previous jenkins job was removed and needed to be replaced with new one on govuk_jenkins

### DIFF
--- a/modules/govuk_jenkins/manifests/job/backdrop_transactions_explorer_collector.pp
+++ b/modules/govuk_jenkins/manifests/job/backdrop_transactions_explorer_collector.pp
@@ -1,0 +1,17 @@
+# == Class: govuk_jenkins::job::backdrop_transactions_explorer_collector
+#
+# Backdrop collector for the Transactions Explorer Google Spreadsheet
+#
+class govuk_jenkins::job::backdrop_transactions_explorer_collector (
+  $google_credentials = undef,
+  $google_spreadsheet_key = undef,
+  $google_spreadsheet_worksheet = undef,
+  $backdrop_token = undef,
+) {
+
+  file { '/etc/jenkins_jobs/jobs/backdrop_transactions_explorer_collector.yaml':
+    ensure  => present,
+    content => template('govuk_jenkins/jobs/backdrop_transactions_explorer_collector.yaml.erb'),
+    notify  => Exec['jenkins_jobs_update'],
+  }
+}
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/manifests/job/backdrop_transactions_explorer_collector.pp
```diff
@@ -0,0 +1,33 @@
+# == Class: govuk_jenkins::job::backdrop_transactions_explorer_collector
+#
+# Backdrop collector for the Transactions Explorer Google Spreadsheet
+#
+class govuk_jenkins::job::backdrop_transactions_explorer_collector (
+  $app_domain = hiera('app_domain'),
```
surminus: I don't think is required.

ambrozic: removed

modules/govuk_jenkins/manifests/job/backdrop_transactions_explorer_collector.pp
```diff
@@ -0,0 +1,32 @@
...
+#
+class govuk_jenkins::job::backdrop_transactions_explorer_collector (
+  $google_credentials = undef,
+  $google_spreadsheet_key = undef,
+  $google_spreadsheet_worksheet = undef,
+  $backdrop_token = undef,
+) {
+
+  govuk::app::envvar {
+    "${title}-GOOGLE_CREDENTIALS":
```
surminus: Ah, sorry, I think I misunderstood slightly. You'd probably do this explicitly in the Jenkins shell part of the template rather than creating env vars to load in ie:
`export GOOGLE_CREDENTIALS=<%= @google_credentials %>`

modules/govuk_jenkins/templates/jobs/backdrop_transactions_explorer_collector.yaml.erb
```diff
@@ -0,0 +1,38 @@
...
+            set -x
+            ./jenkins-data.sh
+    wrappers:
+        - ansicolor:
+            colormap: xterm
+    parameters:
+        - choice:
+            name: ENVIRONMENT
+            choices:
+                - preview
```
surminus: this is now called "integration" rather than "preview"

PR #5967, opened 2017-05-24
Adding govuk_tagging_monitor job

carvil:
This job already exists in Jenkins, but not in code. This commit moves
the job configuration into puppet.

The existing job is here: https://deploy.publishing.service.gov.uk/job/govuk-tagging-monitor/

Trello: https://trello.com/c/NyNRDc78/148-add-govuk-tagging-monitor-tasks-to-puppet

### DIFF
--- a/modules/govuk_jenkins/manifests/job/govuk_tagging_monitor.pp
+++ b/modules/govuk_jenkins/manifests/job/govuk_tagging_monitor.pp
@@ -1,0 +1,23 @@
+# == Class: govuk_jenkins::job::govuk_tagging_monitor
+#
+# Monitor GOV.UK navigation pages to find invalid pages that we shouldn't
+# display to users. It also raises alerts when navigation pages exceed the
+# number of allowed links, so a content designer can act on those. The ruby
+# script handles slack notifications.
+#
+# === Parameters
+#
+# [*rate_limit_token*]
+#   Sets the header "Rate-Limit-Token" which ensures that the tagging monitor is
+#   whitelisted by the rate limiting function (receiving 429 status)
+#
+class govuk_jenkins::job::govuk_tagging_monitor (
+  $rate_limit_token = undef,
+) {
+
+  file { '/etc/jenkins_jobs/jobs/govuk_tagging_monitor.yaml':
+    ensure  => present,
+    content => template('govuk_jenkins/jobs/govuk_tagging_monitor.yaml.erb'),
+    notify  => Exec['jenkins_jobs_update'],
+  }
+}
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/manifests/job/govuk_tagging_monitor.pp
```diff
@@ -0,0 +1,23 @@
+# == Class: govuk_jenkins::job::govuk_tagging_monitor
+#
+# Monitor GOV.UK navigation pages to find invalid pages that we shouldn't
+# display to users. It also raises alerts when navigation pages exceed the
+# number of allowed links, so a content designer can act on those. The ruby
+# script handles slack notifications.
+#
+# === Parameters
```
alecgibson: While we're at it, should we parameterise the Sad Parrot Slack config?

modules/govuk_jenkins/templates/jobs/govuk_tagging_monitor.yaml.erb
```diff
@@ -0,0 +1,30 @@
...
+    project-type: freestyle
+    description: |
+      This job checks new navigation pages against a set of rules. This makes
+      sure we don't have invalid navigation pages being shown to users.
+    scm:
+      - govuk-tagging-monitor
+    logrotate:
+      numToKeep: 10
+    triggers:
+        - timed: '35 * * * *'
```
alecgibson: Are we meant to use hashes for cron jobs?

modules/govuk_jenkins/manifests/job/govuk_tagging_monitor.pp
carvil: It is a bit tricky because the slack plugin in jenkins is configured with the token for the 2nd line slack channel and not ours. I think it's fine to leave it for now if you agree.

modules/govuk_jenkins/templates/jobs/govuk_tagging_monitor.yaml.erb
carvil: I'm not sure I understand the question. I copied this from the existing job following some of the existing patterns: https://github.com/alphagov/govuk-puppet/blob/a6c51d887a6501f02766b7279127b60f02037a7f/modules/govuk_jenkins/templates/jobs/smokey_deploy.yaml.erb#L24

Would you like this to be done differently?

modules/govuk_jenkins/manifests/job/govuk_tagging_monitor.pp
alecgibson: Yeah fine by me

modules/govuk_jenkins/templates/jobs/govuk_tagging_monitor.yaml.erb
alecgibson: As discussed, using `H` will help to spread Jenkins' load.

PR #5971, opened 2017-05-24
Add govuk_tagging_monitor job to production

carvil:
This commit adds the `govuk_tagging_monitor` job to the production
manifest file for Jenkins. This makes sure the job will only run in
production, which is the expected behaviour for this monitoring tool.

Trello: https://trello.com/c/NyNRDc78/148-add-govuk-tagging-monitor-tasks-to-puppet


carvil:
This has been done already in https://github.com/alphagov/govuk-puppet/pull/5970

PR #5972, opened 2017-05-24
Added missing vagrant dependencies

1pretz1:
Vagrant dns --install missing from instructions & Vagrant dns --start also missing

### DIFF
--- a/development-vm/README.md
+++ b/development-vm/README.md
@@ -37,6 +37,8 @@ First, install:
 * [VirtualBox](https://www.virtualbox.org/)
 * [Vagrant](https://www.vagrantup.com/downloads.html)
 * vagrant-dns plugin: `vagrant plugin install vagrant-dns`
+* install vagrant: `vagrant dns --install`
+* start vagrant: `vagrant dns --start`
 
 ## 2. Create your GitHub accounts
 


development-vm/README.md
```diff
@@ -37,6 +37,8 @@ First, install:
 * [VirtualBox](https://www.virtualbox.org/)
 * [Vagrant](https://www.vagrantup.com/downloads.html)
 * vagrant-dns plugin: `vagrant plugin install vagrant-dns`
+* install vagrant: `vagrant dns --install`
```
davidbasalla: Great that we're improving the docs! 👍 

Could we indent these two new lines, so it's more obvious that they relate to the vagrant-dns plugin?
Also 'install vagrant' and 'start vagrant' do not describe these steps accurately. https://github.com/BerlinVagrant/vagrant-dns has some more accurate descriptions

davidbasalla:
A minor point, have a look at the govuk styleguides for git commit messages, the convention is to start with a verb in the present tense. 

https://github.com/alphagov/styleguides/blob/master/git.md#structure

development-vm/README.md
1pretz1: Okay, many thanks!! :)

PR #5980, opened 2017-05-26
Inline alphagov harden

deanwilson:
In a short while this repo will be the only one using the harden module so we're going to inline it to make future re-factoring easier.


surminus:
I think this looks good, I'm going to test in Vagrant to ensure that nothing changes functionally.

surminus:
```
==> jumpbox-1.management: Notice: Compiled catalog for jumpbox-1.management.dev.gov.uk in environment vagrant in 42.85 seconds
==> jumpbox-1.management: Notice: /Stage[main]/Base::Packages/Package[ruby-shadow]/ensure: created
==> jumpbox-1.management: Notice: /Stage[main]/Rkhunter::Package/File[/etc/cron.daily/rkhunter]/ensure: removed
==> jumpbox-1.management: Notice: /Stage[main]/Rsyslog::Config/File[/etc/rsyslog.d/20-ufw.conf]/ensure: removed
```

PR #5987, opened 2017-05-31
Upgrade Bundler version to 1.15.0 on ruby 2.3.0

carvil:
This is an attempt to fix the deployment of `bouncer`. At the moment, we
see the following error:

```
There was an error while trying to write to
/home/deploy/.bundle/cache/compact_index/rubygems.org.443.29b0360b937aa4d161703e6160654e47/info.
It is likely that you need to grant write permissions for that path.
There was an error while trying to write to
```

Further discussion in: https://github.com/bundler/bundler/issues/5691

If this works, we should roll-out the latest version of bundler to the
rest of the rubies.

### DIFF
--- a/modules/govuk_rbenv/manifests/all.pp
+++ b/modules/govuk_rbenv/manifests/all.pp
@@ -47,7 +47,7 @@ class govuk_rbenv::all (
   }
 
   rbenv::version { '2.3.0':
-    bundler_version => '1.14.5',
+    bundler_version => '1.15.0',
   }
   rbenv::version { '2.3.1':
     bundler_version => '1.14.5',


floehopper:
Was this upgrade successful? We have [an issue in Manuals Publisher](alphagov/manuals-publisher/issues/824) which is fixed by upgrading from Bundler v1.14.5 to v1.14.6. I'm happy to submit a PR, but wonder whether I should just upgrade Bundler in the version of Ruby used by Manuals Publisher or all the Rubies. Also whether just to upgrade to v1.14.6 or v1.15.0.

carvil:
@floehopper this by itself didn't fix my issue, but I definitely think we should upgrade all bundler versions to the latest version. There are loads of bugfixes between `1.14.5` and `1.15.0`, so it's definitely worth doing this.

floehopper:
@carvil: Cool. I've just submitted a PR to upgrade: #6103.

PR #5988, opened 2017-06-01
Don't attempt to run search tests when integration is off

boffbowsh:
Integration is generally not available when these tests run, so they
always fail. Allow configuration per environment and move the
integration schedule to a safer time


suzannehamilton:
🎉 Thanks for doing this!

Is integration switched off at the weekend? Perhaps we could we use `30 9 * * MON-FRI` and `0 10 * * MON-FRI` on integration?

boffbowsh:
Yes, good call. According to [this](https://stackoverflow.com/questions/15109443/jenkins-cron-format), Jenkins-speak for Mon-Fri is `1-5`.

PR #5996, opened 2017-06-01
Use private env-sync-and-backup repo

SamLR:
The env-sync-and-backup repository is now in github (rather than GHE).
This updates links to use its new location: https://github.com/alphagov/env-sync-and-backup

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/copy_data_to_integration.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/copy_data_to_integration.yaml.erb
@@ -3,7 +3,7 @@
     name: env-sync-and-backup_Copy_Data_to_Integration
     scm:
         - git:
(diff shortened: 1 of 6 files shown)


modules/govuk_jenkins/templates/jobs/copy_data_to_integration.yaml.erb
```diff
@@ -3,7 +3,7 @@
     name: env-sync-and-backup_Copy_Data_to_Integration
     scm:
         - git:
-            url: git@github.digital.cabinet-office.gov.uk:gds/env-sync-and-backup.git
+            url: git@github.com/alphagov/env-sync-and-backup.git
```
surminus: this should be `git@github.com:alphagov/...` (`:` instead of `/`) everywhere

SamLR: Good catch, done!

PR #6002, opened 2017-06-05
Automatically restart apps at memory warning

boffbowsh:
This used to be restarted at Critical, but this still leaves noise on
Icinga. The suggestion from Stephen Harker was to instead restart at
Warning level, then if it gets to Critical it’s something for us to
actually worry about and action.

### DIFF
--- a/modules/monitoring/files/usr/local/bin/event_handlers/govuk_app_high_memory.sh
+++ b/modules/monitoring/files/usr/local/bin/event_handlers/govuk_app_high_memory.sh
@@ -19,6 +19,7 @@ case "${SERVICESTATE}" in
         ;;
       HARD)
         logger --tag govuk_icinga_event_handler "Restarting app ${APPNAME} on ${HOSTADDRESS} because it's using too much memory"
+        echo -n "govuk.app.${APPNAME}.memory_restarts:1|c" > /dev/udp/localhost/8125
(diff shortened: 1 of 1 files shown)


surminus:
I agree with this change, however I would argue that if apps are being restarted at this level there should be a mechanism which lets us know if an app has restarted more than n times in an hour, for example, or at least don't restart if it hits that threshold.

Restarting at warning I think there is a higher chance we will miss apps endlessly restarting.

stephenharker:
It is a bit of a blunt instrument it's true. What would be better is that if apps trigger the warnings regularly and looking at the memory usage history shows that this is still normal operation for that app, we should bump up the alert thresholds for that app in Puppet. If this is done routinely by 2nd line, we can tune the monitoring better to suit the apps as they change over time. government-frontend, for instance, is now taking more and more frontend traffic from other deprecated frontends so it's not surprising if it uses a bit more memory if it's doing some internal caching or something.

boffbowsh:
For the monitoring, we could just get this script to increment a statsd counter on every restart.

surminus:
@boffbowsh sorry for tardy reply, statsd sounds like a sensible solution.

boffbowsh:
@surminus added. Once this is confirmed to be working we should add it to apps' deployment dashboard.

PR #6011, opened 2017-06-09
Update the validate_published_dns job

SamLR:
The govuk-dns-config repo has had its code move to the govuk-dns repo in
order to consolidate it. This reflects those changes.


surminus:
Is there a reason we use the build shell to clone the repo rather than doing in the native Jenkins way?

PR #6014, opened 2017-06-12
Retire content api

carvil:
In this PR, we remove most references to the Content API. The ones left will be removed once these changes have been deployed.

Trello: https://trello.com/c/2MF97xce/154-retire-content-api


tijmenb:
Do we need to remove parts of the content-api too?

https://github.com/alphagov/govuk-puppet/blob/1dbb291b130100c1de2e72f8d0aaf41bdf8259cb/modules/govuk/manifests/apps/publicapi.pp#L10

carvil:
@tijmenb there is still a few things to delete, but I think those have to be done once this PR is merged and deployed. I was following this guide: https://docs.publishing.service.gov.uk/manual/retiring-an-application.html

Does that make sense?

carvil:
@tijmenb I've added an extra commit based on your comment

PR #6020, opened 2017-06-13
Maslow set mongodb uri environment variable

cbaines:


cbaines:
Thanks for your review @issyl0 :)

PR #6021, opened 2017-06-13
Update ssh_config guidance

rubenarakelyan:
This commit updates the guidance on setting up and using `ssh_config` to point to the file in this repository rather than reiterating its contents (which is prone to becoming out-of-date).

### DIFF
--- a/development-vm/README.md
+++ b/development-vm/README.md
@@ -117,42 +117,11 @@ Your pull request from earlier will hopefully have been merged by now. It's time
 
 > If you're not in the office right now, you'll need to be connected to the Aviation House VPN for SSH access to Integration.
 
-While the applications are available directly via the public internet, SSH access to the environment is via a ‘jumpbox’. You’ll need to configure your machine to use this jumpbox, using this example SSH config file:
-
-```
-Host *
-  UseRoaming no
-
-Host *.integration
-    IdentityFile ~/.ssh/alphagov
-    User $USERNAME
-
-# Integration
-# -------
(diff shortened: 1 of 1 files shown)


development-vm/README.md
```diff
@@ -117,42 +117,11 @@ Your pull request from earlier will hopefully have been merged by now. It's time
...
-  Hostname jumpbox.integration.publishing.service.gov.uk
-  Port     1022
-  ProxyCommand none
-
-Host *.integration
-  ProxyCommand ssh -e none %r@jumpbox-1.management.integration -W $(echo %h | sed 's/\.integration$//'):%p
-```
-
-Copy that into the `~/.ssh/config` file on your mac, substituting your username for `$USERNAME`. You should then be able to ssh into any box in the Integration environment directly. Test that it works, by running:
+While the applications are available directly via the public internet, SSH access to the environment is via a ‘jumpbox’. You’ll need to configure your machine to use this jumpbox using the [example SSH config file](https://github.com/alphagov/govuk-puppet/blob/master/development-vm/ssh_config). Copy the file into the `~/.ssh/config` file on your mac. You should then be able to ssh into any box in the Integration environment directly. Test that it works, by running:
```
cbaines: `on your mac` is overly specific here.

cbaines: I think replacing `the environment` by `remote environments` would read better here.

rubenarakelyan: Fixed both

PR #6022, opened 2017-06-13
Remove Content API configuration

carvil:
Trello: https://trello.com/c/2MF97xce/154-retire-content-api


carvil:
@cbaines you're right, I should have added `ensure => absent` first and deploy that, and then remove this.

I will create that PR now and get it deployed. Thanks for pointing out.

PR #6026, opened 2017-06-14
Couple of transition logs updates

surminus:
After the migration there were a couple of things I wanted to ensure existed for transition logs:

 - monitoring
 - log rotation for cache files

Please see commits for more details.

Story: https://trello.com/c/EhorLoOS/575-post-migration-transition-logs

### DIFF
--- a/modules/govuk_cdnlogs/manifests/transition_logs.pp
+++ b/modules/govuk_cdnlogs/manifests/transition_logs.pp
@@ -91,6 +91,19 @@ class govuk_cdnlogs::transition_logs (
     source => 'puppet:///modules/govuk_cdnlogs/logs_processor_config',
   }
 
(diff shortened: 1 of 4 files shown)


modules/govuk_cdnlogs/manifests/transition_logs.pp
```diff
@@ -91,7 +91,31 @@
     source => 'puppet:///modules/govuk_cdnlogs/logs_processor_config',
   }
 
-  $process_script = '/usr/local/bin/process_transition_logs.sh'
+  # The transition logs processing script creates a cache of processed logs
+  file { "${log_dir}/cache":
```
deanwilson: Squashing this and the following resource in to an array is my only comment.

surminus:
@deanwilson I've updated, thank you!

PR #6030, opened 2017-06-16
Tidy up Procfile

Davidslv:
We have done this in separate commits in case anyone feels like reverting the change.

Paired with @Rosa-Fox

### DIFF
--- a/development-vm/Procfile
+++ b/development-vm/Procfile
@@ -10,7 +10,6 @@ calendars:             govuk_setenv calendars             ./run_in.sh ../../cale
 static:                govuk_setenv static                ./run_in.sh ../../static         bundle exec rails server -p 3013
 licencefinder:         govuk_setenv licencefinder         ./run_in.sh ../../licence-finder bundle exec rails server -p 3014
 signon:                govuk_setenv signon                ./run_in.sh ../../signon         bundle exec rails server -p 3016
-# tradetariff used port 3017
 # tradetariffapi used port 3018
 # efg used port 3019
 # reviewomatic used port 3021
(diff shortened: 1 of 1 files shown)


kevindew:
It'd be useful to keep the sidekiq monitoring ports so we know those port numbers are reserved

h-lame:
Some of the comments deleted refer to ports that are still used, for example [whitehall-admin is still running on port 3026](https://github.com/alphagov/govuk-puppet/blob/08cb80aee0e19b6dbd71b1b6976ce4c79f346ca7/hieradata/class/whitehall_backend.yaml#L6) - and I'm sure there are others (it might be a case of comments that say "uses" vs. comments that say "used").  I'd like a bit more info on each commit as to why we've removed each comment e.g. "we've removed the panopticon comment because the app has been switched off", "we've removed the whitehall_admin comment because ... " (I don't know why), etc...

Is the intention here to make people look into puppet to work out which ports are in use?  Comments in this file are likely to get out of date so this does seem like an admirable goal, but it is easier to look in one place for port numbers than it is to trawl through all the different files in puppet.  Unfortunately this file will always contain some port numbers (for those apps that don't have a startup.sh script and we run them directly from here with a port number) so I think it'll be confusing that there's a single place to look for ports for some apps, but somewhere else for others.

surminus:
I agree with @h-lame that this is a good a place as any to keep a record of all port numbers for apps and app-related things. If you need a file which is easier to parse I'd recommend looking at building some tooling around gathering the information you're after.

Davidslv:
Thanks for your comments, I'll close this.

PR #6031, opened 2017-06-16
Keep feedback paths in sync with app

rubenarakelyan:
This commit adds `assisted-digital-survey-feedback` to the list of paths served by the `feedback` application so that it is also subject to the request limiting in nginx.

See https://github.com/alphagov/feedback/blob/master/config/routes.rb#L17


h-lame:
Given I wrote both the request limiting stuff and the assisted digital endpoint in feedback, I should have remembered to do this.  Thanks for spotting it @rubenarakelyan !

PR #6038, opened 2017-06-21
Add govuk-nav-prototype application

Davidslv:
We need to access the draft-content-store in our prototype to display
draft taxonomies without publishing the taxonomy live (doing so could break user journeys and we don't want that).

Currently we can't access the draft-content-store from the "outside" (e.g: heroku) so we decided to add the prototype alongside other apps.

Trello: https://trello.com/c/9leIHFaX/230-deploy-prototype-to-govuk-infrastructure
Paired with @alecgibson


modules/govuk/manifests/apps/govuk_nav_prototype.pp
```diff
@@ -0,0 +1,32 @@
+# == Class: govuk::apps::govuk_nav_prototype
+#
+# This is a the navigation prototype application
+# Read more: https://github.com/alphagov/govuk-nav-prototype
+#
+# === Parameters
+#
+# [*port*]
+#   The port that govuk-nav-prototype is served on.
+#   Default: 3071
```
surminus: this needs updating

modules/govuk/manifests/apps/govuk_nav_prototype.pp
```diff
@@ -0,0 +1,32 @@
...
+#
+# [*enable_procfile_worker*]
```
surminus: this parameter doesn't exist

alecgibson:
@surminus I've addressed your comments - could you please re-review?

modules/govuk/manifests/apps/govuk_nav_prototype.pp
```diff
@@ -0,0 +1,28 @@
...
+#
+# === Parameters
+#
+# [*port*]
+#   The port that govuk-nav-prototype is served on.
+#   Default: 3211
+#
+class govuk::apps::govuk_nav_prototype (
+  $port = 3211,
+  $secret_key_base = undef,
```
surminus: We should add this to puppetdocs

alecgibson: @surminus what is puppetdocs?

surminus:
If you want to be able to deploy the application through Jenkins you'll need to add your application to yet another list [here](https://github.com/alphagov/govuk-puppet/blob/master/hieradata/common.yaml#L9). 

This will also create stuff in CI which will allow you to run tests in Jenkins and deploy automatically to Integration on master assuming you have a Jenkinsfile in the repo.

alecgibson:
@surminus thanks - I've added that  too.

modules/govuk/manifests/apps/govuk_nav_prototype.pp
surminus: Sorry, I meant just the docs above (puppetdocs is the format).

eg
```
[*secret_key_base*]
...
```

PR #6054, opened 2017-06-26
Remove NPM credentials from CI

36degrees:
My Puppet-foo is pretty weak, hope I've done this correctly.

All of the packages currently published under alphagov on the NPM registry either belong to the Service Manual, Patterns & Tools team (govuk-elements-sass, govuk_frontend_toolkit, govuk_template_ejs, govuk_template_jinja, govuk_template_mustache) or the Performance Platform (performanceplatform-client.js).

They have all been migrated to Travis and are now managed by their teams as part of the Services Programme, so there is no longer any reason for Jenkins to know the npm credentials.

### DIFF
--- a/modules/govuk_ci/manifests/credentials.pp
+++ b/modules/govuk_ci/manifests/credentials.pp
@@ -4,12 +4,6 @@
 #
 # === Parameters:
 #
-# [*npm_auth*]
-#   Auth code for npm
-#
-# [*npm_email*]
-#   Email for npn
-#
 # [*rubygems_api_key*]
 #   API key to authenticate with Rubygems
 #
@@ -29,8 +23,6 @@
 #   The user that has access to these credentials
 #
 class govuk_ci::credentials (
-  $npm_auth,
-  $npm_email,
   $rubygems_api_key,
   $pypi_username,
   $pypi_test_password,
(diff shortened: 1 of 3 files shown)


modules/govuk_ci/manifests/credentials.pp
```diff
@@ -61,9 +53,5 @@
   file {"${jenkins_home}/.pypirc":
     content => template('govuk_ci/pypirc.erb'),
   }
-
-  file { "${jenkins_home}/.npmrc":
-    content => template('govuk_ci/npmrc.erb'),
```
surminus: to ensure that the file is removed set it to absent with a `FIXME`, and then remove the resource later on down the line. For eg:
```
file { 'foo':
  ensure => 'absent',
}
```

surminus:
We'll need to move these credentials from the encrypted Puppet to the password store also.

36degrees:
Updated to remove the `.npmrc` file.

36degrees:
> We'll need to move these credentials from the encrypted Puppet to the password store also.

@surminus is this something you or someone on infra can take care of? I don't think I have access to the right things to do it.

surminus:
I'll open a PR now.

PR #6067, opened 2017-07-04
Make CI run on `vcloud-converter`

issyl0:


hieradata/common.yaml
```diff
@@ -631,6 +631,8 @@ govuk_ci::master::pipeline_jobs:
   transition-config: {}
   ubuntu_unused_kernels:
     repo_owner: 'gds-operations'
+  vcloud-converter:
+    repo-owner: 'alphagov'
```
surminus: this defaults to alphagov by default so it doesn't need to be specified.

PR #6068, opened 2017-07-04
Add asset_bom_removal-rails gem to jenkins

h-lame:
We've created a new gem called [`asset_bom_removal-rails`](https://github.com/alphagov/asset_bom_removal-rails) and want to add it to CI.  The gem hooks into rails asset-pipeline to remove the BOM from compiled CSS files to avoid a bug in firefox (< 52) when calculating SRI checksums.


h-lame:
@deanwilson oops! I've added a description, now, that look ok?

deanwilson:
That's clear and educational! thanks.

PR #6069, opened 2017-07-04
Add an early short circuit for AWS hosts

deanwilson:
We will be testing some of our Puppet code in AWS and this allows us
to be explicit about the class. There is a downside in that you could
manually change an existing hosts class and read its secrets but we're
accepting that for now.

### DIFF
--- a/modules/govuk/lib/puppet/parser/functions/govuk_node_class.rb
+++ b/modules/govuk/lib/puppet/parser/functions/govuk_node_class.rb
@@ -22,7 +22,8 @@ EOS
     ) unless args.size == 0
 
     # Explicit early return if in AWS.
-    return lookupvar('AWS_MIGRATION') unless lookupvar('AWS_MIGRATION').empty?
+    aws_migration = lookupvar('aws_migration')
+    return aws_migration unless (aws_migration.nil? || aws_migration.empty?)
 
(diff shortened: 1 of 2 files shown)


modules/govuk/lib/puppet/parser/functions/govuk_node_class.rb
```diff
@@ -21,6 +21,10 @@ module Puppet::Parser::Functions
       "govuk_node_class: Wrong number of arguments given #{args.size} for 0."
     ) unless args.size == 0
 
+    # Explicit early return if in AWS.
+    aws_migration = lookupvar('aws_migration')
+    return aws_migration unless (aws_migration.nil? or aws_migration.empty?)
```
issyl0: What about `||` instead of `or`? I'm never sure what the difference is, but I don't see `or` very often.

modules/govuk/spec/functions/govuk_node_class_spec.rb
```diff
@@ -11,8 +11,17 @@
     expect { scope.function_govuk_node_class([:gruffalo]) }.to raise_error(Puppet::ParseError, /given 1 for 0/)
   end
 
+  context 'aws_migration set' do
+    it 'should return aws_migration' do
+      aws_migration_name = 'chicken'
```
issyl0: 🐔

PR #6086, opened 2017-07-10
Link default certificate path to Let's Encrypt certificate

rubenarakelyan:
This commit links `default.crt` and `default.key` to the Let's Encrypt certificate since nginx expects this certificate to exist.


training-vm/provisioner/acme-certificates.sh
```diff
@@ -29,6 +29,9 @@ sudo certbot certonly --staging --cert-name ${MAIN_SITE} -m ${EMAIL} -n -d ${CS_
 # puppet and replace them with symlinks to the Let's Encrypt certificate
 rm -f ${NGINX_CONF_DIR}/ssl/*
 
+sudo ln -sf ${LETSENCRYPT_DIR}/live/${MAIN_SITE}/fullchain.pem ${NGINX_CONF_DIR}/ssl/default.crt
```
gpeng: The `NGINX_CONF_DIR` variable only gets used in this file with `/ssl` following it. Could we change it to `NGINX_SSL_DIR` and include the `/ssl`?

rubenarakelyan: Good point, done

PR #6098, opened 2017-07-12
Make sure /var/lib/jenkins/workspace exists

SamLR:
This directory is required but not always created. This ensures it's created before we try to populate it with jobs.


SamLR:
@surminus As far as I can see `/workspace` needs to exist because the `govuk_navigation_link_analysis` job adds a google credentials file to it in order to run. I suspect we may need to fix that but I'll merge this for the time being.

surminus:
Ooooh, OK that makes sense if something else is adding stuff to it.

PR #6102, opened 2017-07-13
Upgrade deploy to v2

surminus:
This branch is used to upgrade Jenkins v1 to v2 in place. The upgrade steps are as follows:

 - Lock Puppet on `jenkins-1`
 - Make a backup of working `/var/lib/jenkins` directory, excl. `bundles` and `workspace`
 - Moved `/var/lib/jenkins/config.xml` out the way
 - Manually moved the 2.46.2 version of Jenkins to `/usr/share/jenkins/jenkins.war`
 - Moved the current `/var/lib/jenkins/config.xml` out of the way
 - `sudo service jenkins restart`
 - When Jenkins has come back up with no security set, log in and update all the plugins. When it has rebooted, you may have to update plugins one further time.
 - To deploy Puppet, you will need to set the `ORGANISATION` global env
 - Deploy this branch of Puppet - I also had some warnings over the Slack plugin, but that can be removed temporarily from the configuration if required
 - Take backup of working `config.xml`
 - Run Puppet on `jenkins-1` and check changes seem sane
 - Attempt login to Jenkins

### DIFF
--- a/hieradata/class/ci_master.yaml
+++ b/hieradata/class/ci_master.yaml
@@ -74,8 +74,6 @@ govuk_jenkins::config::views:
 govuk_jenkins::config::create_agent_role: true
 govuk_jenkins::config::executors: '0'
 
-govuk_jenkins::version: '2.46.2'
-
 govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::job::integration_deploy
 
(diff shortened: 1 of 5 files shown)


surminus:
I've locked Puppet on the Jenkins machines in Staging and Production so this should be safe for merge. This is currently running in Integration.

PR #6103, opened 2017-07-13
Update rbenv Ruby and Bundler versions

floehopper:
* Add missing versions of Ruby to rbenv list
* Update rbenv aliases to appropriate Ruby version
* Upgrade bundler version to latest version (v1.15.1) across the board (mostly from v1.14.5)

I've tried to explain why I think it should be safe to make these changes in the relevant commit notes, but I'm willing to be persuaded the changes should be made in smaller increments.

### DIFF
--- a/modules/govuk_rbenv/manifests/all.pp
+++ b/modules/govuk_rbenv/manifests/all.pp
@@ -29,8 +29,11 @@ class govuk_rbenv::all (
   rbenv::version { '2.1.8':
     bundler_version => '1.14.5',
   }
+  rbenv::version { '2.1.10':
+    bundler_version => '1.14.5',
+  }
   rbenv::alias { '2.1':
-    to_version => '2.1.8',
+    to_version => '2.1.10',
   }
 
   rbenv::version { '2.2.2':
@@ -42,8 +45,11 @@ class govuk_rbenv::all (
   rbenv::version { '2.2.4':
     bundler_version => '1.14.5',
   }
+  rbenv::version { '2.2.7':
+    bundler_version => '1.14.5',
+  }
   rbenv::alias { '2.2':
-    to_version => '2.2.4',
+    to_version => '2.2.7',
   }
 
   rbenv::version { '2.3.0':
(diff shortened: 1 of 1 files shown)


carvil:
I agree with the bundler upgrade, it makes total sense to upgrade this regularly! Thanks for doing that.

I'm not 100% sure we should allow all those ruby versions, though. I think as much as possible we should limit the rubies we can use to the latest major versions (and within each major version, the latest minor and patch versions).

By quickly looking into all the repos I have checked out on my machine, these are the versions we have:

```
(carlosvilhena) ~/govuk » find . -type f -name '.ruby-version' | xargs cat | sort | uniq                                   ruby-
1.9.3
1.9.3-p550
2.1
2.1.4
2.1.8
2.2
2.2.2
2.2.3
2.2.4
2.3.0
2.3.1
2.3.3
2.4.0
system
```

I could see us, with not too much effort, reducing that list even further to a handful of ruby versions.

What do you think of this @floehopper ?

floehopper:
@carvil:

I agree that it would be good to reduce the number of Ruby versions, however that feels like a separate problem. I think the right way to tackle that would be to work through the existing apps which do not meet the Ruby version requirements you mention and upgrade them. Then as particular versions of Ruby are no longer used by any apps, they can be removed from here.

I'd be happy to change this PR to not add any new versions of Ruby other than the latest ones for each minor release, i.e. only add v2.1.10, v2.2.7, v2.3.4 & v2.4.1, if you'd prefer. It just seemed odd to miss out intermediate patch versions. What do you think?

Also, given that the Bundler upgrade doesn't seem so controversial, maybe I should split this PR into two PRs...?

carvil:
If you're ok with that, I would probably only add the latest of each minor release and then separately I will look into upgrading some of the ruby apps to reduce the number of rubies we use.

Happy for this to be on the same PR 👍

floehopper:
@carvil: No problem. Force-pushing now.

surminus:
If we're adding new Ruby versions someone will have to package them and make them available using [this repo](https://github.com/alphagov/packager). Puppet will most likely complain otherwise.

floehopper:
Just saw this. Sorry. My bad.

PR #6104, opened 2017-07-13
Allow aws ips in nrpe

SamLR:
This updates the `nrpe.cfg` to be reflect the version of nagios_nrpe_plugin that is installed.

This also adds the CIDR blocks that we will use in AWS to the `allowed_hosts`


deanwilson:
Can we move the file to be a template and split the sources in to the two hiera structures, current in the current one and the AWS ips in the newer AWS hiera structure please?

PR #6109, opened 2017-07-14
Remove the trade-tariff related entries from Procfile

cbaines:
These are no longer relevant, since trade-tariff is not managed
through govuk-puppet.

### DIFF
--- a/development-vm/Procfile
+++ b/development-vm/Procfile
@@ -14,8 +14,6 @@ static:                govuk_setenv static                ./run_in.sh ../../stat
 licencefinder:         govuk_setenv licencefinder         ./run_in.sh ../../licence-finder bundle exec rails server -p 3014
 # migratorator used port 3015
 signon:                govuk_setenv signon                ./run_in.sh ../../signon         bundle exec rails server -p 3016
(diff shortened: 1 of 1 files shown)


tijmenb:
We've always kept the ports in this file (as per https://docs.publishing.service.gov.uk/manual/retiring-an-application.html#10-update-development-scripts), although I'm not 100% sure why. Could we remove all the comments?

cbaines:
> We've always kept the ports in this file (as per https://docs.publishing.service.gov.uk/manual/retiring-an-application.html#10-update-development-scripts), although I'm not 100% sure why. Could we remove all the comments?

Ah ok. I've got no idea why it would be useful to keep them around, but I also don't use the Procfile and related utilities. I'll leave this open for now to see if someone comes along and points out why this is the wrong thing to do.

cbaines:
Ruben mentioned on slack about problems with running retired apps locally, if the port numbers are reused. This seems reasonable to me, so I've created a patch adding this to the docs [1], and I'll close this now.

1: https://github.com/alphagov/govuk-developer-docs/pull/334

PR #6110, opened 2017-07-15
Add AWS facts and AWS_DEFAULT_REGION to user profile

surminus:
Add a bunch of AWS specific facts if an instance lives in Amazon.

Set the AWS_DEFAULT_REGION default profile which is useful when working with the `aws` tool.


deanwilson:
Maybe whittle it down to 
```
 # availabilityzone
 # instanceid
 # instancetype
 # imageid
 # accountid
 # region
```

surminus:
@deanwilson I've amended the code to only include the useful ones, though the code may have got worse

surminus:
Thanks @issyl0 :)

PR #6112, opened 2017-07-17
[WIP] Manage guix-daemon setup on CI

cbaines:
This branch includes an additional module to install Guix on a machine, and then enables this on ci-agent-2.

Originally I wrote this back last year when I was using the govuk-puppet development-vm for local development. It came in useful with respect to the CI agents later last year when looking at running the Publishing E2E tests with Guix, as at that time, back before these CI agents were in regular use, an earlier version of these changes were deployed to some CI agents.

The guix-daemon service has been running ever since, however, its only within the last few months (from the end of April) that it's been used by jobs.

Currently the guix-daemon on CI agent 2 is enabling the govuk-guix-package-updates [1] job to run, which refreshes the package definitions within govuk-guix [2].

Merging this shouldn't change much, as the service is already running. 

1: https://ci.integration.publishing.service.gov.uk/job/govuk-guix-package-updates
2: https://github.com/alphagov/govuk-guix


surminus:
While I think the code is fine, I do have questions whether we want this in our codebase if it's not been widely adopted by GOV.UK. I think it would set a precedent we may not want to encourage.

I do appreciate however that this only relates to the CI environment, which does have more flexibility for engineers to enable better testing and development. I do wonder if there should be a wider conversation about if we want to branch in the methods in how we develop and test stuff.

As this is already running and entirely restricted to the CI environment, I can see the case for adding this in. I think we should get sign-off from The Powers That Be as to the future of using Guix around GOV.UK.

cbaines:
> While I think the code is fine, I do have questions whether we want this in our codebase if it's not been widely adopted by GOV.UK. I think it would set a precedent we may not want to encourage.
> 
> I do appreciate however that this only relates to the CI environment, which does have more flexibility for engineers to enable better testing and development. I do wonder if there should be a wider conversation about if we want to branch in the methods in how we develop and test stuff.
> 
> As this is already running and entirely restricted to the CI environment, I can see the case for adding this in. I think we should get sign-off from The Powers That Be as to the future of using Guix around GOV.UK.

Thanks for looking this over Laura. I'll mark this as WIP, so that the seal doesn't get upset with it.

stephenharker:
The install module should check the downloaded guix binary against the signature before installing.

stephenharker:
Also, I'd like to know a bit more about the expectations are for guix in the GOV.UK context. What jobs are now using it?

cbaines:
> The install module should check the downloaded guix binary against the signature before installing.

Yep, that's sensible, I've switched it to download over https (in d8a4b9e) and to check the signature (in cebbd48).

> Also, I'd like to know a bit more about the expectations are for guix in the GOV.UK context. What jobs are now using it?

In terms of Jenkins jobs, it's currently the package updates job (defined in govuk-guix/jenkins/[package-updates.groovy][]). Do you want to know about jobs in a more general way?

[package-updates.groovy]: https://github.com/alphagov/govuk-guix/blob/master/jenkins/package-updates.groovy

surminus:
Was this ever confirmed as an official developer tool for GOV.UK? I think that should be the condition for merge, as if it ever breaks things we need to ensure it's supported by more than a single person.

cbaines:
> Was this ever confirmed as an official developer tool for GOV.UK? I think that should be the condition for merge, as if it ever breaks things we need to ensure it's supported by more than a single person.

I'm guessing by "this" you mean [govuk-guix](https://github.com/alphagov/govuk-guix) rather than Guix itself, which I guess could count as a developer tool, but definitely isn't GOV.UK specific.

Regardless of that, I don't think either could be called "official" or that it would be listed anywhere as such, because before you mentioned it here, I didn't know such a thing (or set of things) existed. I've always thought of things like gds-boxen and the govuk-puppet development VM as supported, rather than in a unofficial/official sense.

surminus:
Yes sorry, there is no "official" list. My only concern is if it goes into the govuk-puppet codebase it sets a precedent that means other people will have to support it. 

Probably best to await the outcome [of this](https://github.com/alphagov/govuk-guix/issues/32).

tuzz:
Aren't we already using Puppet for package management? Won't it complicate things to have another tool that effectively does the same thing? I can see that guix provides some great features that GOV.UK could take advantage of, but perhaps it'd be better to propose that change through an RFC?

cbaines:
> Yes sorry, there is no "official" list. My only concern is if it goes into the govuk-puppet codebase it sets a precedent that means other people will have to support it.
> 
> Probably best to await the outcome of this.

Agreed, thanks for following up on this PR in the meantime anyway.

cbaines:
> Aren't we already using Puppet for package management? Won't it complicate things to have another tool that effectively does the same thing?

For package management as in "make this package be on this machine", then this is something that govuk-puppet, well, uses puppet to do. This PR doesn't do anything like that, it's just setting up the generic Guix build daemon.

> I can see that guix provides some great features that GOV.UK could take advantage of, but perhaps it'd be better to propose that change through an RFC?

This isn't very cleanly cut, as this service is already running on the machine. I deployed a older version of this code back before the current CI system was the current CI system, as part of a spike taking place at the time. I finally got around to making this pull request, not with the intent of suggesting a change (like a RFC), but just as fixing that disparity I caused between what govuk-puppet says about ci-agent-2 and the actual state of the machine.

I think an RFC is a very good suggestion, however, at the moment I don't actually have anything to request comments on... I am however trying to run monthly introductory meetings about Guix as a general topic, which I think will be helpful if there are any Guix of package/container related RFCs.

boffbowsh:
I'm going to close this based on the discussion in https://github.com/alphagov/govuk-guix/issues/32

PR #6117, opened 2017-07-17
Add govuk index worker to Rummager and RabbitMQ

brenetic:
We are creating a new govuk index for Elasticsearch. This eventually will consume
messages from the publishing api for all publishing events.

Once all the content has been migrated to this new index this will
replace the publishing-queue-listener which only consumes updates to links.

For now we are just listening for major publishing events while we test
things.

https://trello.com/c/dmBzBVLG/160-insert-data-into-govuk-index-from-rabbitmq-message

Mobbed with @binaryberry @suzannehamilton @dwhenry @Rosa-Fox

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -417,6 +417,7 @@ govuk::apps::rummager::nagios_memory_warning: 4600
 govuk::apps::rummager::nagios_memory_critical: 4900
 govuk::apps::rummager::enable_publishing_listener: true
 govuk::apps::rummager::rabbitmq::enable_publishing_listener: true
+govuk::apps::rummager::rabbitmq_user: 'rummager'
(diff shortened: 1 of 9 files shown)


hieradata/common.yaml
```diff
@@ -413,7 +413,9 @@ govuk::apps::release::github_username: "govuk-ci"
 govuk::apps::rummager::nagios_memory_warning: 4600
 govuk::apps::rummager::nagios_memory_critical: 4900
 govuk::apps::rummager::enable_publishing_listener: true
+govuk::apps::rummager::enable_govuk_index_worker: true
```
SamLR: If this is still in the testing phase would it be better to only enable this in integration?

Rosa-Fox:
@SamLR done. :)

PR #6118, opened 2017-07-17
Make search analytics job alerts non-critical

MatMoore:
Currently whenever a jenkins job fails, it raises a critical alert.
Allow this to be overriden so we can raise warning alerts instead.

The search-fetch-analytics job is known to be flaky, but it's fine to
leave it until the next day.

https://trello.com/c/xAUa1fQn/162-deprioritise-errors-when-search-reindex-fails

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/failure_passive_check.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/failure_passive_check.yaml.erb
@@ -8,7 +8,6 @@
     builders:
       - shell: |
           IPADDRESS=$(facter ipaddress)
-          NSCA_CODE=2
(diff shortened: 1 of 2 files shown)


MatMoore:
Yes, definitely.

suzannehamilton:
I've made a couple of changes because it looks like you can't set a default on a "choice" parameter:

- Remove the `default` setting
- Reorder the choices from 1,2,3 to 2,1,3 so that 2 is the default. That's a bit untidy, but I can't see an alternative other than installing the [extended choice plugin](https://wiki.jenkins.io/display/JENKINS/Extended+Choice+Parameter+plugin).

I've tested this by running `jenkins-jobs test failure_passive_check.yaml` on jenkins-1.management and looking at the XML output. I think this is now ready to merge once someone else has reviewed it.

brenetic:
🚀

MatMoore:
Thanks for merging! 👍

PR #6119, opened 2017-07-17
Update authorization strategy for AWS Jenkins

surminus:
Use github.com to authenticate into Jenkins in AWS.


SamLR:
That looks ok to me

PR #6120, opened 2017-07-17
Add MOTD for AWS instances

surminus:


surminus:
I just wanted to show how cool ASCII art is. We can do this properly tomorrow.

surminus:
yolo

PR #6131, opened 2017-07-19
Enable sass linting by default, as had been expected

fofr:
Reviewers: I don't know groovy and I am unsure how to test this. Help would be much appreciated.

The docs say that the sass linter is on by default:
https://github.com/alphagov/govuk-puppet/blob/f28348c1a9bc8fa5edd95e646df4bbd87f05e116/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy#L55

That's not the case though.
* If `options.sassLint` is omitted then checking `options.sassLint` is null and falsy.
* The default state was therefore no sass linting
* Explicitly check that it has been set to false so that if the option is not provided sassLint runs

I think this may have broken here:
https://github.com/alphagov/govuk-puppet/pull/5828

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -146,7 +146,7 @@ def buildProject(Map options = [:]) {
       echo "WARNING: You do not have Ruby linting turned on. Please install govuk-lint and enable."
     }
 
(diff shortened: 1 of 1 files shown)


fofr:
Related to https://github.com/alphagov/government-frontend/pull/402

tijmenb:
This seems good to me, but maybe @suzannehamilton has other ideas on how to set defaults in the options?

fofr:
Because we lint all SASS, when we merge this there will be build failures on unrelated PRs, there may be some clean-up required.

PR #6138, opened 2017-07-20
Begin basic node class for s_db_admin

surminus:
We're looking to provision our databases in RDS rather than provision EC2 machines with databases on them.

To create, administer and backup databases using Puppet, we need something to use that can do this in our current Puppet estate.

For now just install packages to used to work with our current versions MySQL and PostgreSQL.

### DIFF
--- a/modules/govuk/manifests/node/s_db_admin.pp
+++ b/modules/govuk/manifests/node/s_db_admin.pp
@@ -2,10 +2,40 @@
 #
 # This machine class is used to administer RDS instances.
 #
-class govuk::node::s_db_admin {
+# === Parameters
+#
+# [*mysql_db_host*]
+#  The database hostname
+#
+# [*mysql_db_password*]
+#  The database password
+#
+# [*mysql_db_user*]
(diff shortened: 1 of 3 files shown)


hieradata/common.yaml
```diff
@@ -1311,6 +1311,8 @@ hosts::production::licensify::hosts:
     ip: '10.5.0.18'
 
 hosts::production::management::hosts:
+  db-admin-1:
```
SamLR: Shouldn't this be in `hierdata_aws/common.yaml`?

surminus: Probably. I only added this to get tests to pass, and actually I think I'd rather add an exception in the tests.

SamLR: fair enough (we should probably look at rolling those tests out to cover stuff in `hieradata_aws` too but that's for another PR...)

surminus: I think you're right.

SamLR: otherwise LGTM : )

PR #6143, opened 2017-07-21
Use puppetlabs-lvm module

afda16:
We are using a fork of puppetlabs-lvm that is 4 years old. Update to latest
puppetlabs module, fixes several bugs.

https://github.com/alphagov/puppetlabs-lvm/compare/master...puppetlabs:master

Test carefully after merging


surminus:
It's failing because the package itself needs adding, I'll do that now.

surminus:
I've done the librarian shuffle.

PR #6145, opened 2017-07-24
Add class definition for exception_handler

SamLR:


hieradata_aws/class/exception_handler.yaml
```diff
@@ -0,0 +1,28 @@
...
+    mountoptions: 'defaults'
+  /var/lib/automongodbbackup:
+    disk: '/dev/mapper/backup-mongodb'
+    govuk_lvm: 'mongodb'
+    mountoptions: 'defaults'
+
+mongodb::firewall::enable_firewall_rules: false
+
+mongodb::server::replicaset_members:
+  'errbit.issam.integration.govuk-internal.digital':
```
surminus: Will this work? If this is doing a lookup of the ELB DNS name then will it try to resolve the IP of the ELB, rather than resolve the address of the local server? It will probably have issues trying to route out and then back in from it's own load balancer (unless AWS works differently to what I'm used to).

It may be easier to specify "localhost".

PR #6153, opened 2017-07-25
Update hostalias in AWS

surminus:
We updated the display_name so that the hosts looked friendlier in AWS (rather than just using the Amazon set hostname), but unfortunately Icinga does the annoying thing of sorting by hostname rather than the display_name.

I'm hoping updating the alias will mean that they get properly sorted into order, because otherwise it's very hard to parse the list to see what has come up.


surminus:
Unfortunately this won't do what I'd hoped it would do. Hosts are still sorted by "host" address, so their Amazon hostname. Until this is updated I'll just have to put up with - we could potentially update the host_name variable to use a friendlier name, but _so_ many things use this that it'd be a pain. Something to come back to when we really need it.

PR #6163, opened 2017-07-27
Add function to get full commit hash

emmabeynon:
This is now called when we run `runPublishingE2ETests`, solving the issue of
apps not setting the `publishingE2ETests` option to true.  We also standardise
the capitalization of `E2E` in the naming of the `publishingE2ETests` option.

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -61,7 +61,7 @@
  *          stage, such as environment variable configuration
  *        - overrideTestTask A closure containing commands to run to test the
  *          project. This will run instead of the default `bundle exec rake`
- *        - publishingE2eTests Whether or not to run the Publishing end-to-end
+ *        - publishingE2ETests Whether or not to run the Publishing end-to-end
 *           tests.  Default: false
  *        - afterTest A closure containing commands to run after the test stage,
  *          such as report publishing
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -61,7 +61,7 @@
  *          stage, such as environment variable configuration
  *        - overrideTestTask A closure containing commands to run to test the
  *          project. This will run instead of the default `bundle exec rake`
- *        - publishingE2eTests Whether or not to run the Publishing end-to-end
+ *        - publishingE2ETests Whether or not to run the Publishing end-to-end
```
kevindew: 🎉

modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -786,4 +783,11 @@ def runPublishingE2ETests(appCommitishName, testBranch, repo) {
   )
 }
 
+def setFullCommitHashEnvar() {
```
kevindew: I think rather than have this set an env var it should just return the full commit hash as a string. 

Setting the env var effectively just sets a global variable and we don't need that really.

emmabeynon:
@kevindew now updated 👍

emmabeynon:
Tested on CI integration for Travel Advice Publisher and Content Store, and both ran end-to-end tests 🎉

PR #6174, opened 2017-07-31
AWS configure collectd

afda16:
On AWS, do not use network plugin and use write_graphite instead. Update
fqdn_metrics to use aws facts.

### DIFF
--- a/modules/govuk/lib/facter/fqdn_metrics.rb
+++ b/modules/govuk/lib/facter/fqdn_metrics.rb
@@ -4,7 +4,12 @@
 
 Facter.add("fqdn_metrics") do
  setcode do
-    fqdn_metrics = Facter.value("fqdn_short").dup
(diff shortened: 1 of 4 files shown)


modules/govuk/lib/facter/fqdn_metrics.rb
```diff
@@ -4,7 +4,7 @@
 
 Facter.add("fqdn_metrics") do
  setcode do
-    fqdn_metrics = Facter.value("fqdn_short").dup
+    Facter.value("aws_migration").nil? ? fqdn_metrics = Facter.value("fqdn_short").dup : fqdn_metrics = "#{Facter.value(:aws_hostname)}.#{Facter.value(:aws_stackname)}.#{Facter.value(:aws_environment)}"
```
deanwilson: we should probably add a spec test for this.

afda16:
@deanwilson , updated

PR #6190, opened 2017-08-03
Store POST traffic to rummager servers using GOR

dwhenry:
This provides a simplified restore strategy as we can reload the index
from backup and then re-run the POST traffic. The big advantage here is
that we will not need to try and resend the changes from the individual
publishing apps.

https://trello.com/c/8AWCdA1Z/204-investigate-gor-traffic-replay

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -64,6 +64,8 @@ govuk_crawler::targets:
 
 govuk_elasticsearch::dump::run_es_dump_hour: '9'
 
+govuk_search::gor::enabled: true
+
 govuk_jenkins::config::banner_colour_background: '#ffbf47'
 govuk_jenkins::config::banner_colour_text: 'black'
 govuk_jenkins::config::banner_string: 'Carrenza INTEGRATION'
--- a/modules/govuk_search/manifests/gor.pp
+++ b/modules/govuk_search/manifests/gor.pp
@@ -1,0 +1,55 @@
+# == Class: govuk_search::gor
+#
+# Dump POST traffic to file so that it can be replayed. This means we have a simple
+# way to recover from failure as we can restore the last backup and then replay the
+# traffic (POST requests are for data insert/update only)
+#
+# Traffic can then be replayed by follow the guide at:
+# https://github.com/buger/goreplay/wiki/Capturing-and-replaying-traffic
+#
+# === Parameters
+#
+# [*enabled*]
+#   Boolean to determine if Search traffic will be saved; defaults to false.
+#
+# [*output_path*]
+#   File location for files to be saved to.
+#
+class govuk_search::gor (
+  $output_path = '/var/log/gor_dump',
+  $enabled = false
+) {
+
+  validate_bool($enabled)
+
+  if($enabled) {
+    validate_re($output_path, '^/.*')
+
+    file { $output_path:
+      ensure => 'directory',
+      owner  => root,
+      group  => root,
+      mode   => '0755',
+    }
+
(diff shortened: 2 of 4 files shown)


modules/govuk_search/manifests/gor.pp
```diff
@@ -0,0 +1,46 @@
...
+#
+# [*enabled*]
+#   Boolean to determine if Search traffic will be saved; defaults to false.
+#
+# [*output_path*]
+#   File location for files to be saved to.
+#
+class govuk_search::gor (
+  $output_path
+  $enabled = false,
```
brenetic: `,`

brenetic:
👍  when syntax stuff passes

modules/govuk_search/manifests/gor.pp
```diff
@@ -0,0 +1,48 @@
...
+#
+class govuk_search::gor (
+  $output_path = undef,
+  $enabled = false
+) {
+
+  validate_bool($enabled)
+
+  if($enabled) {
+    validate_re($output_file, '/.*')
```
MatMoore: should this regex be '^/.*' i.e. anchored to the beginning of the string ?

MatMoore: `$output_file` should be `$output_path`

dwhenry: Yes, thank you. I spent over an hour trying to work out why this didn't work.

dwhenry: @MatMoore added some tests and think this might now work, will test when I'm in the office tomorrow

hieradata/integration.yaml
```diff
@@ -64,6 +64,9 @@ govuk_crawler::targets:
 
 govuk_elasticsearch::dump::run_es_dump_hour: '9'
 
+govuk_search::gor::enabled: true
+govuk_search::gor::output_path: '/data/vhost/rummager.integration.publishing.service.gov.uk/gor_dump'
```
surminus: I would set this to a plainer location (something like `/etc/gor/dumps`) because otherwise it depends on the app resource for rummager being deployed, which could cause dependency errors.

modules/govuk_search/manifests/gor.pp
```diff
@@ -0,0 +1,55 @@
...
+# === Parameters
+#
+# [*enabled*]
+#   Boolean to determine if Search traffic will be saved; defaults to false.
+#
+# [*output_path*]
+#   File location for files to be saved to.
+#
+class govuk_search::gor (
+  $output_path = undef,
```
surminus: I don't think there's harm in setting a sane default.

modules/govuk_search/manifests/gor.pp
```diff
@@ -0,0 +1,55 @@
...
+  $enabled = false
+) {
+
+  validate_bool($enabled)
+
+  if($enabled) {
+    validate_re($output_path, '^/.*')
+
+    file { $output_path:
```
surminus: As my other comment, this should probably have a dependency.

modules/govuk_search/manifests/gor.pp
dwhenry: Set this to `/var/log/gor_dump`

hieradata/integration.yaml
dwhenry: removed as I can just use the default

modules/govuk_search/manifests/gor.pp
dwhenry: This should be fine now that the directory is not a dependency (I think)

PR #6193, opened 2017-08-03
Bugfix Elasticsearch 2.4 install

surminus:
- the versioncmp function works rather than the regex (which didn't
  work as expected)

  - only install plugins for the older versions of elasticsearch. The
  Puppet module we're using for Elasticsearch is too old to install
  plugins on ES >2, so we can either fork the module, or install plugins
  manually until we've upgraded the old versions of Elasticsearch we're
  running.

@dwhenry because we're using such an old version of the Puppet Elasticsearch module, the plugin install resource doesn't work for 2.4, which means you'll have to install plugins manually (or we'll have to fork the Puppet module).


dwhenry:
@surminus The manual plugin installation is fine. I will review which/if we need the existing plugins once we have everything else working. :)

PR #6200, opened 2017-08-03
Update govuk_node_list for AWS

surminus:
In AWS, we can find out the hostnames for our instances using tags. We still require PuppetDB to discover hosts which are assigned with
specific classes, but we can update the functionality for listing specific node classes by using boto.

This has been rewritten in a way that should not be visible to users or scripts so can be an in-place drop in. This should ensure we're able to continue using our current deployment scripts without impact.

### DIFF
--- a/modules/govuk_scripts/templates/govuk_node_list_aws.erb
+++ b/modules/govuk_scripts/templates/govuk_node_list_aws.erb
@@ -1,0 +1,118 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+
+# NB: This script uses optparse, not argparse, for Python 2.6 compatibility and
+# to avoid the need to depend on any other packages.
+from optparse import OptionParser
+import json
+import random
+import string
+import sys
+import urllib
+import urllib2
+import boto3
+
+stackname = '<%= @aws_stackname -%>'
+
+def commasep_callback(option, opt, value, parser):
+    setattr(parser.values, option.dest, value.split(','))
+
+def ec2_nodes(stackname, nodeclass=None):
+    client = boto3.client('ec2')
+
+    if nodeclass:
+        response = client.describe_instances(
+                   Filters=[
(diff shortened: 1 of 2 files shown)


modules/govuk_scripts/templates/govuk_node_list_aws.erb
```diff
@@ -0,0 +1,119 @@
...
+
+stackname = '<%= @aws_stackname -%>'
+
+def commasep_callback(option, opt, value, parser):
+    setattr(parser.values, option.dest, value.split(','))
+
+def ec2_nodes(stackname, nodeclass=None):
+    client = boto3.client('ec2')
+
+    if not nodeclass:
```
deanwilson: Swap the order of these so it becomes if `nodeclass else` ?

modules/govuk_scripts/templates/govuk_node_list_aws.erb
```diff
@@ -0,0 +1,119 @@
...
+                          {'Name': 'tag:aws_stackname', 'Values': [stackname]},
+                          {'Name': 'tag:aws_migration', 'Values': [nodeclass]},
+                          {'Name': 'instance-state-name', 'Values': ['running']}
+                   ]
+               )
+
+    nodes = response['Reservations']
+
+    hosts = []
+    for key in nodes:
```
deanwilson: Can key be replaced with something more descriptive? instance possibly?

modules/govuk_scripts/templates/govuk_node_list_aws.erb
```diff
@@ -0,0 +1,119 @@
...
+        hosts.append((key['Instances'][0]['NetworkInterfaces'][0]['PrivateDnsName']))
+
+    return(hosts)
+
+def main():
+    opts, args = parser.parse_args()
+
+    if opts.node_class:
+        if any('-' in c for c in opts.node_class):
```
deanwilson: I think this can just be `if '-' in`

PR #6201, opened 2017-08-04
Add AWS_DEFAULT_REGION to Jenkins

surminus:
This will allow app deployments when we do lookups based upon tags.

### DIFF
--- a/modules/govuk_jenkins/templates/config/config.xml.erb
+++ b/modules/govuk_jenkins/templates/config/config.xml.erb
@@ -163,7 +163,7 @@
             <comparator class="hudson.util.CaseInsensitiveComparator"/>
           </default>
           <%- if @aws_migration -%>
-          <int><%= @environment_variables.length + 1 -%></int>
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/templates/config/config.xml.erb
```diff
@@ -163,7 +163,7 @@
             <comparator class="hudson.util.CaseInsensitiveComparator"/>
           </default>
           <%- if @aws_migration -%>
-          <int><%= @environment_variables.length + 1 -%></int>
+          <int><%= @environment_variables.length + 2 -%></int>
```
deanwilson: It would be nice to have some context as to what and why we do this

surminus: It's an oddity of Jenkins config. You have to specify the number of environment variables you're setting, otherwise they won't appear or be set if there are any above the number you've specified. 

If you were adding through the UI it would do this automatically, sadly it's another one of those querks we've found trying to manage Jenkins config through Puppet.

PR #6203, opened 2017-08-04
Fork puppet-elasticsearch

surminus:
We run really old versions of both Elasticsearch and the Elasticsearch Puppet module. We have a requirement to install a later version of Elasticsearch on a separate instance. Unfortunately, the Puppet module we're using is not compatible for a couple of minor reasons.

For reference: https://github.com/alphagov/puppet-elasticsearch/commit/a0d8d3eae04a13408ec325fba5ffd36e3084f90a

To get around these issues, I have decided to fork the repo. I did attempt to upgrade to a slightly later version of the module, but that caused many more errors on the older versions of Elasticsearch than I
felt comfortable with.

The ambition would be to upgrade our older versions of Elasticsearch in line with upgrading the module itself. We may look at outsourcing our Elasticsearch clusters to a SaaS provider in the future.

We haven't updated this module for 3 years, so I don't see a problem with forking now as it's not like we're missing out on updates.

**Disclaimer: this PR makes me very unhappy.**


deanwilson:
We should remove the old vendored module tar

surminus:
@deanwilson updated.

PR #6204, opened 2017-08-04
Wrap upstream mysql::db

surminus:
In AWS we wish to use RDS to manage our databases. The current upstream MySQL module does not allow managing a remote database with the current defined types as it always requires the mysql::server class, which installs and sets up a MySQL server.

This is an attempt to copy the current code (and tests), except with added conditionals related to AWS which means resources do not require setting up the server.


surminus:
I have tested this as successfully working.

surminus:
This was successfully resolved by updating the MySQL Puppet module itself.

PR #6205, opened 2017-08-04
Add db_admin config for PostgreSQL

surminus:
Update db_admin class to manage an RDS instance. This unfortunately involves forking the puppetlabs-postgresql module.

See commits for more detail.


modules/govuk/manifests/node/s_db_admin.pp
```diff
@@ -34,6 +34,17 @@
   class { '::govuk::apps::signon::db': } ->
   class { '::govuk::apps::whitehall::db': }
 
+  # include all PostgreSQL classes that create databases and users
```
deanwilson: The class that configures the postgres client tools and creds should be at the start of this chain, I think.

hieradata_aws/class/db_admin.yaml
```diff
@@ -0,0 +1,2 @@
+---
+postgresql::server::package_ensure: false
```
deanwilson: Shouldn't we set this in common as we're RDS only in AWS?

PR #6217, opened 2017-08-07
We should point at the backend-lb, not the backend

deanwilson:


surminus:
I believe the backend_servers sets the upstream servers to send traffic _to_, so this change is incorrect as the backend-lb should be sending upstream traffic to the backend machines.

PR #6220, opened 2017-08-08
Setup Jenkins to fetch router-data from GitHub

cbaines:
For CI, rather than GitHub Enterprise.


cbaines:
> Looks good

Great, thanks for reviewing Sam :)

PR #6221, opened 2017-08-08
Record Sidekiq stats with collectd

alecgibson:
In order to help diagnose excessive memory usage of worker tasks,
this change collects Sidekiq process data with collectd, for monitoring
in Graphite.

worker-rss-usage

### DIFF
--- a/modules/govuk/manifests/procfile/worker.pp
+++ b/modules/govuk/manifests/procfile/worker.pp
@@ -39,6 +39,12 @@ define govuk::procfile::worker (
   include govuk::procfile
 
   $service_name = "${title}-procfile-worker"
+  $title_underscore = regsubst($title, '\.', '_', 'G')
+
+  collectd::plugin::process { "app-worker-${title_underscore}":
+    ensure => $ensure,
+    regex  => "sidekiq .* ${title}.*\\.gov\\.uk",
+  }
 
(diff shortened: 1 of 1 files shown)


alecgibson:
[Memory issues in Sidekiq jobs](https://trello.com/c/I5Nz2FeO/408-memory-issues-in-sidekiq-jobs)

modules/govuk/manifests/procfile/worker.pp
```diff
@@ -66,6 +67,10 @@
         service_description => "${title} procfile worker upstart up",
         host_name           => $::fqdn,
       }
+      collectd::plugin::process { "app-worker-${title_underscore}":
+        ensure => $ensure,
+        regex  => "sidekiq .* ${title}.*\\.gov\\.uk",
+      }
```
cbaines: If you put this block outside the `if $ensure == present {`, the use of `$ensure` in the block will mean that the `collectd::plugin::process` thing is removed if `$ensure` specifies it should be. At the moment, `$ensure` will always be `present` here, as that is the condition in the `if`.

alecgibson:
@cbaines does this look alright now?

cbaines:
> @cbaines does this look alright now?

Yep, I think that will work better.

PR #6237, opened 2017-08-09
Allow applications to set the SENTRY_DSN env var

tijmenb:
The `SENTRY_DSN` env var is used by the Sentry client to report exceptions. This commit allows us to set the var for applications. To use it we'll have to pass in the `sentry_dsn` variable via hieradata into `govuk::app`. The hieradata will be set in govuk-secrets.

The DSN for collections & content-tagger are added here because we'll be testing with them first. Once the test is successful we'll add the other applications as well.

https://trello.com/c/YsgDupoW

### DIFF
--- a/modules/govuk/manifests/apps/collections.pp
+++ b/modules/govuk/manifests/apps/collections.pp
@@ -17,6 +17,9 @@
 # [*secret_key_base*]
 #   The key for Rails to use when signing/encrypting sessions.
 #
+# [*sentry_dsn*]
+#   The URL used by Sentry to report exceptions
+#
 # [*nagios_memory_warning*]
 #   Memory use at which Nagios should generate a warning.
 #
@@ -28,6 +31,7 @@ class govuk::apps::collections(
   $port = '3070',
   $errbit_api_key = undef,
   $secret_key_base = undef,
+  $sentry_dsn = undef,
   $nagios_memory_warning = undef,
   $nagios_memory_critical = undef,
 ) {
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/apps/collections.pp
```diff
@@ -41,6 +45,7 @@
     vhost                  => $vhost,
     nagios_memory_warning  => $nagios_memory_warning,
     nagios_memory_critical => $nagios_memory_critical,
+    sentry_dsn             => $sentry_dsn,
```
cbaines: Given that this puppet thing uses `govuk::app::envvar`, I think it would be easier to understand if `govuk::app` wasn't used here, as all it is doing is passing this value through to `govuk:app:envvar`.

This would also match up with how Errbit is configured at the moment.

tijmenb: I think the way Errbit and other env vars are configured is way too tedious. I was actually hoping to avoid editing every app manifest, but that doesn't seem to be possible. In the mean time, I prefer to abstract away the setting of the env var to somewhere else. WDYT?

cbaines: > I think the way Errbit and other env vars are configured is way too tedious. I was actually hoping to avoid editing every app manifest, but that doesn't seem to be possible. In the mean time, I prefer to abstract away the setting of the env var to somewhere else. WDYT?

No objections from me, I don't have much knowledge on govuk-puppet code style.

PR #6243, opened 2017-08-10
Add head plugin to elasticsearch 2.4

dwhenry:


surminus:
The `install` command issue should have been taken of in the fork of the elastic/elasticsearch Puppet module I did. 

It would be better to update it there if we do:
https://github.com/alphagov/puppet-elasticsearch

dwhenry:
@surminus I thought about going with that approach, however, the plugin installation command if different for 1.7 v's 2.4 so we need a way to determine if we should use `-install` or `install` and I couldn't think of anything nice.

Let chat tomorrow to determine the best approach.

Gemfile
```diff
@@ -18,3 +18,4 @@ gem "sshkey", "1.7.0"
 gem "parallel_tests"
 gem "parallel"
 gem "librarian-puppet"
+gem 'pry'
```
deanwilson: Is this needed for anything else in the PR or are we sneaking it in? I'm not against including `pry` but we should do it in a separate PR if we're adding it.

dwhenry: no that was my bad, it has now been removed

dwhenry:
@surminus you were right the custom module was not required, this just works. :)

modules/govuk_elasticsearch/manifests/plugins.pp
```diff
@@ -33,11 +33,16 @@
 
   }
   # If the version is 2.4, then install the plugin the 2.4 way
-  elsif versioncmp($elasticsearch_version, '2.4') == 0 {
+  elsif versioncmp($elasticsearch_version, '2.4.6') == 0 {
```
surminus: is this the only version of 2.4 we're ever likely to use? if so this is fine :)

dwhenry: It should be unless we need to patch for security...

surminus: I thought the `versioncmp` function dealt with minor versions but I guess not. Might be safer to do a regex upon the versions (ie if we _did_ patch I wouldn't want Puppet to suddenly trash stuff because of this conditional)

PR #6244, opened 2017-08-10
AWS puppetserver support

afda16:
On AWS we are going to run the puppet master service from puppetserver
instead of Nginx + Unicorn, to help the transition to Puppet 4 and
support certificate extensions for node classification.

This requires a new puppetserver manifest and some adjustments to
install Openjdk and remove conflicting gems that are installed during
the Puppet master bootstrap. Nginx is still installed to enable access to
Puppetdb, but we should review this.

### DIFF
--- a/modules/puppet/manifests/puppetserver/package.pp
+++ b/modules/puppet/manifests/puppetserver/package.pp
@@ -1,0 +1,64 @@
+# == Class: puppet::puppetserver::package
+#
+# Install packages for a Puppetserver.
+#
+# === Parameters
+#
+# [*puppetdb_version*]
+#   Specify the version of puppetdb-terminus to install which should match
+#   the puppetdb installation. Passed in by parent class.
+#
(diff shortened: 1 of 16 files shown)


modules/puppet/manifests/puppetserver/package.pp
```diff
@@ -0,0 +1,71 @@
...
+  package { 'rack':
+    ensure   => '1.0.1',
+    provider => system_gem,
+  }
+
+  package { 'puppetdb-terminus':
+    ensure  => $puppetdb_version,
+  }
+
+  exec { 'puppetserver gem install hiera-eyaml-gpg':
```
surminus: I think you have to set a fully qualified path or search path when running `execs`

surminus:
One comment but I think this looks OK

afda16:
@surminus I updated the path

PR #6250, opened 2017-08-11
Setup a local proxy for the ES 2.4 instance on rummager boxes

dwhenry:
This is a temporary implementation while we are migrating.  Once the migration
is completed we should update port 9200 to point to the new cluster, we can
then revert the change in rummager so that it connects oin port 9200 again.

Once this is complete we can delete the proxy on 19200

https://trello.com/c/Ew3bOzy5/229-setup-rummager-to-work-with-es-24-on-integration

### DIFF
--- a/modules/govuk_elasticsearch/manifests/rummager_local_proxy.pp
+++ b/modules/govuk_elasticsearch/manifests/rummager_local_proxy.pp
@@ -1,0 +1,45 @@
+# == Class: govuk_elasticsearch::rummager_local_proxy
+#
+# This is a direct copy of the govuk_elasticsearch::local_proxy
+# class and has been added here to support the migration to
+# Elasticsearch 2.4. The only difference being the proxy port
+# which has been changed to 19200.
+#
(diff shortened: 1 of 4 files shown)


surminus:
In our migration work to AWS we're already not using the proxy because we're using ELBs 🎉

modules/govuk_elasticsearch/manifests/rummager_local_proxy.pp
```diff
@@ -0,0 +1,45 @@
...
+#   Nginx `proxy_read_timeout` value.
+#   Default: 60
+#
+# [*servers*]
+#   Array of servers to load balance requests to.
+#
+class govuk_elasticsearch::rummager_local_proxy(
+  $read_timeout = 60,
+  $port = 19200,
+  $servers
```
surminus: Can we add a `,` on the end for consistency?

dwhenry: done

PR #6254, opened 2017-08-11
[Don't merge] Allow SENTRY_DSN to be set for all applications

tijmenb:
This:

- adds the `sentry_dsn` option to all applications. This allows us to set the config in hieradata which will cause the `SENTRY_DSN` environment variable to be set.
- This sets the current environment for Sentry to the value currently used for errbit, so either integration/staging/production (getsentry/raven-ruby#736).

I _really_ dislike the mass-edit of all files here, but there doesn't seem to be a better way for now.

## Making sure that every app is configured

To make sure every app is correctly configured, I've run this script that checks that all documented apps were edited in this PR.

https://gist.github.com/tijmenb/354cbebe3e0a4666449ac414dc6f649c

Apps not included in this PR:

- collections # already done in https://github.com/alphagov/govuk-puppet/pull/6237
- content_tagger  # already done in https://github.com/alphagov/govuk-puppet/pull/6237
- design_principles # skipped, will be removed soon
- govuk_delivery # python app
- mapit # python app, open source
- metadata_api # go app, will be removed soon
- router # go app

https://trello.com/c/iTFRLCaM


modules/govuk/manifests/app/config.pp
```diff
@@ -125,7 +125,8 @@
         value   => "govuk.app.${title}.${::hostname}";
       "${title}-SENTRY_DSN":
         varname => 'SENTRY_DSN',
-        value   => $sentry_dsn;
+        value   => $sentry_dsn,
+        notify_service => false;
```
surminus: fat arrows should line up else linting might get upset

tijmenb:
After talking to @boffbowsh and @danielroseman we've decided that we'll accept the mass-unicorn restart probably caused by the change in environment variables (see https://github.com/alphagov/govuk-puppet/pull/6283 for more discussion). This PR can be reviewed again.

PR #6258, opened 2017-08-14
Fixed script mistake that doesn't cleanup old archives.

stephenharker:
The generated extension and the extension find was looking for were different and now they're the same. Without this, the find would never cleanup any files.


modules/govuk/files/usr/local/bin/govuk_backup_graphite_whisper_files
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-TIMESTAMP=`date +%Y-%m-%d_%Hh%Mm.%A`
+TIMESTAMP=`date +%Y-%m-%d_%Hh%Mm.%A`o
```
surminus: that looks like a typo (the `o` on the end)

stephenharker: Oh yeah.Grr.

PR #6273, opened 2017-08-15
Change permissions, add exclusions and cleanup for graphite data.

stephenharker:
A few changes. Delete ephemeral graphite interface data from the CI agents. Tweak permissions on the Graphite backup directory so that govuk-backup can read them. Use pigz to multi-thread compress the data. And exclude some large directories from being backed up that we don't care about (PP mostly).

### DIFF
--- a/modules/govuk/files/usr/local/bin/govuk_delete_ephemeral_interface_data
+++ b/modules/govuk/files/usr/local/bin/govuk_delete_ephemeral_interface_data
@@ -1,0 +1,8 @@
+#!/bin/bash
+# Expunge unneeded ephemeral interface data from the ci-agent-* boxes.
+
+for i in 1 2 3 4 5
+do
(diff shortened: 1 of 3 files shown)


modules/govuk/files/usr/local/bin/govuk_delete_ephemeral_interface_data
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Expunge unneeded ephemeral interface data from the ci-agent-* boxes.
+
+for i in 1 2 3 4 5
```
surminus: to ensure that we always delete data for all CI agents this could be:
```
for i in $(govuk_node_list -c ci_agent)`

stephenharker:
I'm running the backup script on Integration graphite-1 now. Then I'll manually kick off the backup of graphite-1 on backup-1 and see what happens.

PR #6280, opened 2017-08-15
Configure asset-manager to proxy to S3

chrisroos:
This config allows us to set HTTP headers in the asset-manager Rails app
to instruct nginx to proxy requests for assets to S3. This is going to
allow us to performance test this approach.

The `cloud-storage-proxy` location is marked as `internal` which means
that it can't be accessed from outside nginx. We use it by setting the
`X-Accel-Redirect` header to `/cloud-storage-proxy/` in our Rails app.
We additionally set the `X-Cloud-Storage-URL` and `X-Cloud-Storage-Host`
HTTP headers to pass the host and URL of the asset to nginx.

We're consciously not enabling caching of the assets from S3 at the
moment. We want to test the performance without it before deciding
whether it's necessary to add it.

### DIFF
--- a/modules/govuk/manifests/apps/asset_manager.pp
+++ b/modules/govuk/manifests/apps/asset_manager.pp
@@ -88,7 +88,46 @@ class govuk::apps::asset_manager(
       location ~ /raw/(.*) {
         internal;
         alias /var/apps/asset-manager/uploads/assets/$1;
-      }',
+      }
+
+      location ~ /cloud-storage-proxy/(.*) {
+        # Prevent requests to this location from outside nginx
+        internal;
+
+        # Construct download URL from:
+        # $1:       Host + path from regexp match in location
+        # $is_args: Optional ? delimiter
+        # $args:    Optional querystring params
+        set $download_url $1$is_args$args;
+
+        # The X-CLOUD-STORAGE-URL header contains a signed URL for the asset on
+        # S3. The signature of this URL is based in part on the request headers
+        # set in the asset-manager Rails app at the time the URL is generated.
+        # The headers we send now must match otherwise nginx will not be
+        # allowed to make the request. Since this location block inherits
+        # `proxy_set_header` directives from previous levels[1], we explicitly
+        # set the Host so that the inherited headers are over-written.
+        #
+        # [1] http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header
+        proxy_set_header Host $proxy_host;
+
+        # Remove S3 HTTP headers listed in:
+        # http://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html
+        # This keeps this HTTP response as similar as possible to the response
+        # sent when using Sendfile to serve files from NFS
+        proxy_hide_header x-amz-delete-marker;
+        proxy_hide_header x-amz-id-2;
+        proxy_hide_header x-amz-request-id;
+        proxy_hide_header x-amz-version-id;
+
+        # Add Google DNS server to avoid "no resolver defined to resolve"
+        # errors when trying to connect to S3
+        resolver 8.8.8.8;
+
+        # Download the file and send it to client
+        proxy_pass $download_url;
+      }
+      ',
     }
 
     govuk::app::envvar {


chrisroos:
Hi @deanwilson. I've requested your review based on GitHub's suggestion. Are you in a position to review this or is there someone more suitable that we should ask?

modules/govuk/manifests/apps/asset_manager.pp
```diff
@@ -88,7 +88,40 @@
...
+        # set in the asset manager rails app at the time the URL is generated.
+        # The headers we send now must match otherwise nginx will not be
+        # allowed to make the request. Since this location block inherits
+        # `proxy_set_header` directives from previous levels[1], we explicitly
+        # set the Host so that the inherited headers are over-written.
+        #
+        # [1] http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header
+        proxy_set_header Host $cloud_storage_host;
+
+        # Explicitly remove S3 HTTP headers
```
boffbowsh: What's the reasoning for removing the headers?

modules/govuk/manifests/apps/asset_manager.pp
```diff
@@ -88,7 +88,40 @@
...
-      }',
+      }
+
+      location = /cloud-storage-proxy/ {
+        # Prevent requests to this location from outside nginx
+        internal;
+
+        # Extract download host and URL from upstream HTTP headers
+        set $cloud_storage_host $upstream_http_x_cloud_storage_host;
+        set $download_url $upstream_http_x_cloud_storage_url;
```
boffbowsh: Sending this as part of the `X-Accel-Redirect` header might be more obvious for anyone needing to debug this later. You could then just `rewrite` out the initial `/cloud-storage-proxy`. I'm not sold on one method or the other, but it'd be good to note the reasons for this implementation style.

modules/govuk/manifests/apps/asset_manager.pp
```diff
@@ -88,7 +88,40 @@
...
+        proxy_set_header Host $cloud_storage_host;
+
+        # Explicitly remove S3 HTTP headers
+        proxy_hide_header x-amz-id-2;
+        proxy_hide_header x-amz-request-id;
+        proxy_hide_header x-amz-version-id;
+
+        # Add Google DNS server to avoid "no resolver defined to resolve"
+        # errors when trying to connect to S3
+        resolver 8.8.8.8;
```
floehopper: Minor: I think this is fine for the moment, but I wonder whether if we put this into use in production, we might want to (a) list multiple DNS servers; and/or (b) use GOV.UK DNS servers...?

modules/govuk/manifests/apps/asset_manager.pp
```diff
@@ -88,7 +88,40 @@
...
+        # `proxy_set_header` directives from previous levels[1], we explicitly
+        # set the Host so that the inherited headers are over-written.
+        #
+        # [1] http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header
+        proxy_set_header Host $cloud_storage_host;
+
+        # Explicitly remove S3 HTTP headers
+        proxy_hide_header x-amz-id-2;
+        proxy_hide_header x-amz-request-id;
+        proxy_hide_header x-amz-version-id;
```
floehopper: Minor: I think this is fine for the moment, but if there was a way to use a regular expression (i.e. headers starting with `x-amz-`) to identify S3 headers that would be nicer since it would be more robust to S3 changes.

modules/govuk/manifests/apps/asset_manager.pp
```diff
@@ -88,7 +88,40 @@
...
+        # Prevent requests to this location from outside nginx
+        internal;
+
+        # Extract download host and URL from upstream HTTP headers
+        set $cloud_storage_host $upstream_http_x_cloud_storage_host;
+        set $download_url $upstream_http_x_cloud_storage_url;
+
+        # The X-CLOUD-STORAGE-URL header contains a signed URL for the asset on
+        # S3. The signature of this URL is based in part on the request headers
+        # set in the asset manager rails app at the time the URL is generated.
```
floehopper: Very minor: I'd probably either hyphenate "asset-manager" and capitalise "Rails".

modules/govuk/manifests/apps/asset_manager.pp
```diff
@@ -88,7 +88,40 @@
...
+
+      location = /cloud-storage-proxy/ {
+        # Prevent requests to this location from outside nginx
+        internal;
+
+        # Extract download host and URL from upstream HTTP headers
+        set $cloud_storage_host $upstream_http_x_cloud_storage_host;
+        set $download_url $upstream_http_x_cloud_storage_url;
+
+        # The X-CLOUD-STORAGE-URL header contains a signed URL for the asset on
```
floehopper: Minor: I wonder whether this comment is too closely tied to what we're currently planning the Asset Manager to do, i.e. I think we could still use this mechanism if we were not using signed URLs, but in that case this comment would be out-of-date.

modules/govuk/manifests/apps/asset_manager.pp
chrisroos: We tried that initially. It worked well with public S3 URLs (i.e. URLs without querystrings) but we ran into some problems once we started sending presigned URLs (with querystrings). We'll revisit the problems we ran into to check whether this implementation really is warranted.

modules/govuk/manifests/apps/asset_manager.pp
chrisroos: Err, good question. I think it's mostly so that we're matching, as closely as possible, the standard responses from asset-manager. We'll update the comment to explain the reason for their removal.

modules/govuk/manifests/apps/asset_manager.pp
chrisroos: I've updated the nginx config to send the S3 URL as part of the `X-Accel-Redirect` header. I think the problems I ran into previously were due to my lack of nginx knowledge!

modules/govuk/manifests/apps/asset_manager.pp
chrisroos: I've updated the comment in the nginx config to explain why we're removing these headers.

modules/govuk/manifests/apps/asset_manager.pp
chrisroos: Done!

modules/govuk/manifests/apps/asset_manager.pp
chrisroos: I've not found a way of using a wildcard or regexp to match the x-amz-* headers unfortunately. In an attempt to improve things slightly I've added a link to the AWS docs that explain the HTTP headers we can expect.

modules/govuk/manifests/apps/asset_manager.pp
chrisroos: Very possibly. I'm using 8.8.8.8 as that's what we're also using in https://github.com/alphagov/govuk-puppet/blob/master/modules/bouncer/templates/www.mhra.gov.uk_nginx.conf.erb#L55 so maybe/hopefully it's good enough for now?

chrisroos:
@boffbowsh: I've updated the branch to address your feedback. Can you let me know whether you think this (sending the S3 URL in the `X-Accel-Redirect` header) is an improvement, please? If so I'll tidy up the commits (I'm planning to squash the together) and commit notes and ask for a review again.

modules/govuk/manifests/apps/asset_manager.pp
chrisroos: I agree that this comment is closely tied with the fact that we're sending presigned URLs and that the implementation still works if we send public URLs. I'm just not sure how best to convey that!

chrisroos:
Thanks @boffbowsh! I've squashed the commits and rebased on master in preparation for getting this merged.

PR #6283, opened 2017-08-15
[Discuss] Don't restart applications by default after env change

tijmenb:
Following a discussion in https://trello.com/c/PCkKYHnp, this changes the default behaviour to not restart applications when en environment variable changes. 

Applications would have to redeploy for the env variable to take effect. This is a better option than always restarting because redeploying is zero-downtime, while mass-restarting can cause errors to be served to users.

@surminus @danielroseman @SamLR @boffbowsh thoughts?

### DIFF
--- a/modules/govuk/manifests/app/envvar.pp
+++ b/modules/govuk/manifests/app/envvar.pp
@@ -20,7 +20,7 @@ define govuk::app::envvar (
   $ensure  = 'present',
   $envdir  = "/etc/govuk/${app}/env.d",
   $varname = $title,
-  $notify_service  = true,
+  $notify_service  = false,
 ) {
   validate_string($value)
 
--- a/modules/govuk/manifests/apps/content_store/enable_running_in_draft_mode.pp
+++ b/modules/govuk/manifests/apps/content_store/enable_running_in_draft_mode.pp
@@ -33,7 +33,6 @@ class govuk::apps::content_store::enable_running_in_draft_mode(
 
   Govuk::App::Envvar {
     app            => $app_name,
-    notify_service => false,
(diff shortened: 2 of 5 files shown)


boffbowsh:
I think I'm for this in theory. Things can opt into the notify behaviour as needed, although I can't instantly envisage a situation where you'd need this. 

In the case of adding new environment variables, you'd generally add the value then deploy the code that uses it so a restart on puppet rollout would have no effect. This is the case in the Sentry instance that has triggered this discussion.

danielroseman:
I am slightly concerned that it does leave applications in an unexpected state, and you would have to remember to restart all of them at some point. But balancing that is the fact that the current behaviour is surprising too.

I wonder if there is any way we could get Puppet to send `reload` rather than `restart`, to mimic the deploy behaviour? Separately, have we confirmed that reload does actually cause the workers to pick up the new env var?

boffbowsh:
`reload` wouldn't do as the env var is set by the upstart script. `reload` works by asking the unicorn master to spawn a new process, which means it will inherit the environment of the old master.

danielroseman:
But isn't that exactly the same behaviour as deploying? This is all managed by https://github.com/gds-operations/unicornherder, which intercepts the reload SIGHUP and manages the overlap of the unicorn workers. If reload doesn't read the new env, surely neither would redeploying, which means that this proposal can't work.

boffbowsh:
That's a very good point. And we're also then saying that `deploy:hard_restart` would also result in a similar number of 500s as just letting puppet restart things.

🤔

tijmenb:
Closing, we'll talk about this IRL.

Issue #5244, opened 2016-12-08
Email notifications


bruth:
I am not sure why, but I am getting email notifications from `webops+ci@digital.cabinet-office.gov.uk`.

alexmuller:
Hi - sorry about that. Is it possible that your email address is in the git commit history for this repository? If you can forward the email to me (alex.muller@digital.cabinet-office.gov.uk) that would be great. Thanks!

alexmuller:
Got some details from @bruth and I think I've figured out what happened - it's completely ridiculous but it's the only explanation I can come up with.

I think in March 2015 Matt was doing some work with the opsmanual which used [bruth/jekyll-docs-template](https://github.com/bruth/jekyll-docs-template/) which resulted in [this build](https://ci.dev.publishing.service.gov.uk/job/opsmanual_branches/205/).

Jenkins saw those commits and created a person on disk called `b` with @bruth's email address attached to it.

Recently @brenetic changed his email address to be `b@brenetic.com`. Every time @brenetic pushes a commit which fails tests, Jenkins decides that this person with a `b` email address is probably the same `b` that it already knows about and emails @bruth with the test results.

Argh! Sorry again for the problems.

PR #6292, opened 2017-08-16
Email users upstream of a failing Smokey build

alecgibson:
Prior to this change, Smokey would only notify a distribution list on
failure. If people aren’t watching that list, they will not be notified
of failures.

This change still notifies the distribution list, but on top of that it
tries to work out who kicked off the job upstream of Smokey causing it
to fail, and emails them.

In other words, if someone kicks off one of:
- Deploy App
- Deploy Puppet
- Deploy Licensify
- Network Config Deploy

...and then Smokey fails, then they will be notified by email, if that
email can be inferred.

Additionally, if changes were made to Smokey, and then it fails, then
the “culprits” who broke the build will also be notified.

In order to achieve this behaviour, we need to use the Jenkins
[`email-ext`](https://wiki.jenkins.io/display/JENKINS/Email-ext+plugin) plugin.

This plugin offers far more customisation than the plainer `mailer`
plugin, and - importantly - allows us to email addresses that are
unknown to Jenkins. After Jenkins’ security advisory
[`SECURITY-372`](https://jenkins.io/security/advisory/2017-03-20/), it was not possible to email addresses not associated with Jenkins users.

However, since we authenticate via Github, Jenkins does not actually
consider our email addresses to be associated with our projects.
`Email-ext` allows us to overcome this by running Jenkins with the
`-Dhudson.tasks.MailSender.SEND_TO_USERS_WITHOUT_READ=true` `JAVA_ARG`,
which is also added in this change.

### DIFF
--- a/hieradata/class/jenkins.yaml
+++ b/hieradata/class/jenkins.yaml
@@ -29,6 +29,11 @@ govuk_jenkins::config:
       --httpPort=$HTTP_PORT
       --ajp13Port=-1
       --sessionTimeout=1440
+  JAVA_ARGS:
+    value: >-
+      -Djava.awt.headless=true
+      -Djenkins.install.runSetupWizard=false
+      -Dhudson.tasks.MailSender.SEND_TO_USERS_WITHOUT_READ=true
 
 # FIXME: During the upgrade this stopped deploys being triggered from CI. When
 # the resolution has been configured remove this.
@@ -73,6 +78,8 @@ govuk_jenkins::plugins:
     version: '1.9'
   durable-task:
     version: '1.14'
+  email-ext:
+    version: '2.58'
   envinject-api:
     version: '1.2'
   envinject:
--- a/modules/govuk_jenkins/templates/jobs/smokey.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/smokey.yaml.erb
@@ -30,9 +30,33 @@
           auth-token: <%= @environment_variables['SECOND_LINE_SLACK_AUTH_TOKEN']%>
           build-server-url: <%= @slack_build_server_url %>
           room: <%= @slack_room %>
-      - email:
+      - email-ext:
           recipients: govuk-ci-notifications@digital.cabinet-office.gov.uk
-          send-to-individuals: true
+          failure: true
+          presend-script: |
+            import hudson.tasks.Mailer
+
+            def upstreamCause = build.getCause(Cause.UpstreamCause)
+
+            if (upstreamCause != null) {
+              def upstreamCauses = upstreamCause.getUpstreamCauses()
+              upstreamCauses.each() { cause ->
+                if (cause instanceof Cause.UserIdCause) {
+                  def user = User.get(cause.getUserId())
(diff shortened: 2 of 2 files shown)


alecgibson:
[Improve Smokey email notifications](https://trello.com/c/rgsTR6oy/719-improve-smokey-email-notifications)

hieradata/class/jenkins.yaml
```diff
@@ -29,6 +29,11 @@ govuk_jenkins::config:
       --httpPort=$HTTP_PORT
       --ajp13Port=-1
       --sessionTimeout=1440
+  JAVA_ARGS:
+    value: >-
+      -Djava.awt.headless=true
```
alecgibson: These first two lines are copied from whatever was in `JAVA_ARGS` before this change; is there any way to merge this change in, rather than simply override with a copy-and-paste? Or does it not matter?

modules/govuk_jenkins/templates/jobs/smokey.yaml.erb
```diff
@@ -30,9 +30,30 @@
           auth-token: <%= @environment_variables['SECOND_LINE_SLACK_AUTH_TOKEN']%>
           build-server-url: <%= @slack_build_server_url %>
           room: <%= @slack_room %>
-      - email:
+      - email-ext:
           recipients: govuk-ci-notifications@digital.cabinet-office.gov.uk
-          send-to-individuals: true
+          failure: true
+          presend-script: |
```
alecgibson: Is there a more sensible way to have this groovy script in a separate file, and then include that file, rather than have it in the config itself?

hieradata/class/jenkins.yaml
surminus: These are provided [by the upstream repo](https://github.com/jenkinsci/puppet-jenkins/blob/23df7764d58aed80b573621b102f82a702217a7a/manifests/params.pp#L28), so not sure if there is a good way to merge them together.

modules/govuk_jenkins/templates/jobs/smokey.yaml.erb
surminus: It may be possible to add it as a library like [we do here](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/manifests/pipeline.pp#L24-L30).

alecgibson: @surminus I'm not sure I understand how to get that into the YAML file? (This is where we reach the limit of my Puppet understanding).  Although I've noticed we're already writing raw shell script on lines 59-59 in this file as well anyway...

tijmenb:
Can we somehow restrict emails going out to cabinet-office email addresses? I'm worried about data being inadvertently sent to external people - see https://github.com/alphagov/govuk-puppet/issues/5244 for an example.

alecgibson:
@tijmenb that sounds fine to me. Although annoyingly an anecdotal quick glance through some Smokey builds shows that a few people are still using personal email addresses to commit 😞

alecgibson:
@tijmenb I've added a check for the email domain.
@surminus do you think it's worth investigating moving the Groovy script out of the YAML, or is this good to merge?

PR #6304, opened 2017-08-17
Install git from PPA

thomasleese:
This allows us to get a more recent version of git.


surminus:
You could add it as a mirror to our aptly repo: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/node/s_apt.pp#L35-L111

thomasleese:
@surminus Thanks, I've had another go, I think this makes more sense now!

thomasleese:
I'm not sure how to specify the version of git we want to install, or if we even need to do that.

thomasleese:
No longer required.

PR #6311, opened 2017-08-17
Extract graphite host url

deanwilson:
The location to icinga should check for graphite metrics changes in
AWS both in terms of name and http scheme.

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -986,6 +986,7 @@ icinga::client::config::allowed_hosts: "10.0.0.0/8"
 icinga::config::http_username: "%{hiera('http_username')}"
 icinga::config::http_password: "%{hiera('http_password')}"
 
+icinga::config::graphite_hostname: 'graphite'
 icinga::config::smokey::http_username: "%{hiera('http_username')}"
 icinga::config::smokey::http_password: "%{hiera('http_password')}"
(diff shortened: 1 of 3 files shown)


hieradata_aws/common.yaml
```diff
@@ -986,6 +986,7 @@ icinga::client::config::allowed_hosts: "10.0.0.0/8"
 icinga::config::http_username: "%{hiera('http_username')}"
 icinga::config::http_password: "%{hiera('http_password')}"
 
+icinga::config::graphite_host: 'graphite'
```
afda16: I think we've been using `_hostname` instead of `_host` to reference the AWS URL overrides

modules/icinga/templates/check_graphite_metric_args.cfg.erb
```diff
@@ -1,4 +1,4 @@
 define command {
   command_name check_graphite_metric_args
-  command_line <%= @check_graphite_command %> -H http://graphite.cluster/render -U '<%= @http_username %>' -P '<%= @http_password %>' -M '$ARG1$' -w $ARG2$ -c $ARG3$ $ARG4$
+  command_line <%= @check_graphite_command %> -H <%= $graphite_url %>/render -U '<%= @http_username %>' -P '<%= @http_password %>' -M '$ARG1$' -w $ARG2$ -c $ARG3$ $ARG4$
```
afda16: `@graphite_url` instead of `$graphite_url`

Issue #5942, opened 2017-05-18
Upgrade NodeJS


fofr:
We'd like to start linting our Javascript using `standard`, see:
https://standardjs.com/
https://github.com/alphagov/govuk-lint/issues/63
https://github.com/alphagov/govuk_frontend_toolkit/pull/334

To run standard linting on our VMs and on Jenkins we need a modern version of NodeJS (> 4). We currently run version 0.10, which hasn't been updated since 2015: https://github.com/alphagov/govuk-puppet/blob/a6c51d887a6501f02766b7279127b60f02037a7f/hieradata/common.yaml#L1150
https://github.com/alphagov/govuk-puppet/commit/009e0b1f06b0c6495cb1420b42701533bce149aa

The PR that updated Node mentioned StatsD and Performance Platform as dependencies. Is that still the case?

cc @boffbowsh

fofr:
This might be the time to include something like `nvm` too. https://github.com/creationix/nvm

PR #6314, opened 2017-08-18
Update node

thomasleese:
Update the version of node available on most of the machines, but leave the performance platform on the older version as they are still dependant on it.

Closes https://github.com/alphagov/govuk-puppet/issues/5942

[Trello Card](https://trello.com/c/YPhTen16/1041-2-automate-the-building-of-documentation) [Trello Card](https://trello.com/c/M8uRSjYq/113-spike-npm-integration-into-static-2)

### DIFF
--- a/modules/nodejs/manifests/init.pp
+++ b/modules/nodejs/manifests/init.pp
@@ -9,7 +9,7 @@
 #   Default: `undef`
 #
 class nodejs(
-  $version = undef
+  $version = undef,
 ) {
 
   if $version == undef {
@@ -18,7 +18,10 @@ class nodejs(
     $ensure = $version
   }
 
+  class { '::nodejs::repo': }
+
   package { 'nodejs':
     ensure  => $ensure,
+    require => Class['Nodejs::Repo'],
   }
 }
--- a/modules/nodejs/manifests/repo.pp
+++ b/modules/nodejs/manifests/repo.pp
@@ -1,0 +1,20 @@
+# == Class: nodejs::repo
+#
+# Use our own mirror of the nodejs repo. Should be used with `manage_repo`
+# disable of the upstream module.
+#
+# === Parameters
+#
+# [*apt_mirror_hostname*]
+#   Hostname to use for the APT mirror.
+#
+class nodejs::repo(
+  $apt_mirror_hostname = undef,
+) {
+  apt::source { 'nodejs':
+    location     => "http://${apt_mirror_hostname}/nodejs",
+    release      => $::lsbdistcodename,
+    architecture => $::architecture,
+    key          => '3803E444EB0235822AA36A66EC5FE1A937E3ACBB',
+  }
+}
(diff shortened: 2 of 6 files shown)


thomasleese:
Tested on integration, seems to be working:

```bash
❯ ssh jenkins-1.integration
Last login: Fri Aug 18 13:12:21 2017 from jumpbox-1.management.integration.publishing.service.gov.uk
thomasleese@integration-jenkins-1:~$ node --version
v6.11.2

❯ ssh performance-frontend-1.integration
Last login: Fri Aug 18 13:14:55 2017 from jumpbox-1.management.integration.publishing.service.gov.uk
thomasleese@integration-performance-frontend-1:~$ node --version
v0.10.37
```

cc @surminus @nickcolley

nickcolley:
I think the only place we current use Node.js on the main machines is in Sprockets JavaScript uglification, so we'd want to check the JavaScript output potentially for an application.

thomasleese:
Just checked the digests of the assets in integration and production, and they match.

thomasleese:
Plus performance platform still seems to be working fine in integration.

nickcolley:
Looks good to me 👍 

(Will await someone who knows how Puppet works :))

modules/nodejs/manifests/init.pp
```diff
@@ -18,6 +18,8 @@
     $ensure = $version
   }
 
+  class { '::nodejs::repo': }
+
   package { 'nodejs':
     ensure  => $ensure,
```
surminus: You should add a dependency for the repo:
`require => Class['Nodejs::Repo'],`

modules/nodejs/manifests/repo.pp
```diff
@@ -0,0 +1,20 @@
...
+#   Hostname to use for the APT mirror.
+#
+class nodejs::repo(
+  $apt_mirror_hostname = undef,
+) {
+  apt::source { 'nodejs':
+    location     => "http://${apt_mirror_hostname}/nodejs",
+    release      => $::lsbdistcodename,
+    architecture => $::architecture,
+    key          => '',
```
surminus: The key should be the same as all the others. Check across the codebase for where we have used `apt::source` previously.

surminus:
Couple of comments but it looks good! I think we install statsd with npm, so please check that's OK (which I'm sure it is).

thomasleese:
@surminus Thanks, I've addressed your comments. Regarding statsd, I saw that it installed correctly, and the following command returns something good:

```
thomasleese@integration-jenkins-1:~$ sudo service statsd status
statsd start/running, process 785
```

Anything more I could check?

surminus:
We use statsd for counters like number of status codes appearing in nginx for example, so I guess just check that graphite for these metrics don't disappear.

thomasleese:
@surminus Just checked in graphite, and there does appear to be data for the last hour in there.

PR #6316, opened 2017-08-18
Enable Content Audit allocation

andrewgarner:
Adds an environmental variable to enable content audit allocation in dev, integration and staging.

### DIFF
--- a/modules/govuk/manifests/apps/content_performance_manager.pp
+++ b/modules/govuk/manifests/apps/content_performance_manager.pp
@@ -22,6 +22,10 @@
 #   The view id of GOV.UK in Google Analytics
 #   Default: undef
 #
+# [*feature_auditing_allocation*]
+#   Whether the auditing allocation feature should be enabled/disabled.
+#   Default: false
+#
 # [*google_client_email*]
 #   Google authentication email
 #   Default: undef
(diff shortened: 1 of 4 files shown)


andrewgarner:
[Add feature toggle to disable allocation in production, and enable in staging and integration](https://trello.com/c/r3iKNHcw/449-add-feature-toggle-to-disable-allocation-in-production-and-enable-in-staging-and-integration)

modules/govuk/manifests/apps/content_performance_manager.pp
```diff
@@ -122,6 +122,14 @@
     port => $redis_port,
   }
 
+  if $::domain =~ /^.*\.(dev|integration\.publishing\.service|staging\.publishing\.service)\.gov\.uk/ {
```
alecgibson: Isn't the more normal thing to do here to add it to the enviornment-specific config files? eg https://github.com/alphagov/govuk-puppet/blob/master/hieradata/development.yaml#L97-L183

andrewgarner: Fair challenge. I've reworked the PR to adopt that approach 👍

modules/govuk/manifests/apps/content_performance_manager.pp
```diff
@@ -63,6 +63,7 @@
   $db_password = undef,
   $db_username = 'content_performance_manager',
   $enable_procfile_worker = true,
+  $feature_auditing_allocation = false,
```
alecgibson: Do we need to add a comment to the documentation at the top of this class?

andrewgarner: Added! 👍

PR #6317, opened 2017-08-18
Nginx limit rate alert on www and assets

afda16:
At the moment we monitor limit rate errors (HTTP 429) on the cache services
(www.govuk.uk and assets) only in Grafana. This change creates a Icinga alert,
similar to 5xx errors, to monitor on HTTP response code 429 during office
hours.

We should monitor the alert to check that the limits and metric window are
correct before enabling out of office hours.

Documentation in https://github.com/alphagov/govuk-developer-docs/pull/422


afda16:
The checks should be ready to test. We still need to add the document to developer docs

PR #6318, opened 2017-08-18
Add missing slack values

SamLR:
This adds missing values to the puppet class to stop job builder failing


deanwilson:
Is there a reason not to make these class params with the current values as defaults? Just in case we need to change any of them for testing.

issyl0:
Will this actually send notifications?

SamLR:
@deanwilson I don't think there's any particular reason not to but it's not been done anywhere else

@issyl0 to confirm, this can later have notifications re-enabled but right now, for testing, they're disabled.

PR #6321, opened 2017-08-18
Add resolver to nginx config in AWS

surminus:
On startup, nginx does DNS lookups for upstream hosts and caches the result. In AWS, when ELBs scale up or down, they are liable to change endpoints. We use DNS to reach the appropriate endpoints, but unless we restart nginx then upstream requests will timeout.

By adding a resolver directive, it means that if the upstream request times out, it will do a DNS search:

https://nginx.org/en/docs/http/ngx_http_core_module.html#resolver

The resolver value is for the VPC nameserver, which is always the same in a VPC based upon the network range:

http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_DHCP_Options.html#AmazonDNS

### DIFF
--- a/modules/nginx/manifests/config.pp
+++ b/modules/nginx/manifests/config.pp
@@ -13,10 +13,16 @@
 # [*denied_ip_addresses*]
 #   An array of IP addresses that Nginx should prevent from accessing this machine.
 #
+# [*stack_network_prefix*]
+#   This is only used in AWS. This adds a resolver so that nginx refreshes resolved addresses
+#   from DNS if an upstream lookup fails. This is the prefix address defined for a VPC.
+#
 class nginx::config (
   $server_names_hash_max_size,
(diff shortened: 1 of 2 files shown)


modules/loadbalancer/templates/nginx_balance_no_ssl.conf.erb
```diff
@@ -1,3 +1,6 @@
+<%- if scope.lookupvar('::aws_migration') %>
+resolver <%- scope.lookupvar('::stack_network_prefix') %>.2;
```
afda16: Should it be `-` or `=`?

surminus: `=` is only used when printing a `@variable`

surminus:
This doesn't work because the resolver can only be specified once.

PR #6325, opened 2017-08-18
Expose secrets added to hieradata from alphagov-deployment

boffbowsh:
These new environment variables were originally deployment secrets in
alphagov-deployment and are now exposed via hieradata.


modules/govuk/manifests/apps/router_api.pp
```diff
@@ -68,4 +74,6 @@
       value   => join($router_nodes, ','),
     }
   }
+
```
deanwilson: Can we remove one of these newlines please?

boffbowsh: Done.

PR #6330, opened 2017-08-21
[DO NOT MERGE] split puppet module for setting up RabbitMQ consumers

MatMoore:
As part of https://trello.com/c/5d4Qkt31/244-bulk-requeue-content-by-format, I'm looking at adding a separate queue to route requeued messages into, when we want to bulk-reindex content in search.

My reasoning for adding this is so that requeued messages won't get in the way of normal publishing. With a separate queue we can be less strict about persistence and easily purge messages if we don't want to wait for them to be processed. My plan is to make the new queue [transient](http://rubybunny.info/articles/durability.html) (non-durable).

Currently in puppet the `govuk_rabbitmq::consumer` class sets up the users and queues at the same time. This gets really messy when there are multiple queues per user, so I attempted to extract out the queue setup into its own thing.

On reflection I'm thinking this is a bit overkill for my immediate problem, so I'm going to first try having the consumer declare its own transient queue instead.

But I'd still like a review of this PR as some of the changes could still be merged, and I'd like someone to sanity check my approach.

The changes:

1. Remove amqp_queue_2, routing_key_2 in favour of explicitly creating a separate queue with `govuk_rabbitmq::queue_with_binding`

2. Rename `govuk_index` to `govuk_index_durable` - in contrast to the new queue which should be transient

3. Introduce extra parameters for the user permissions, so we can decouple the user/queue setup. This is the most awkward part of the PR since if it's wrong stuff will break. It also feels more complicated than it needs to be (see questions below).

4. Route messages with two words in (`*.*`) to `govuk_index_durable` instead of routing everything. My intention is to make the requeued messages use a longer routing key that doesn't match this pattern. An alternative would be to set up multiple bindings for the same queue (one for each of major, minor, links, unpublish), but this is not yet possible with the puppet module we're using.

## Questions
 I've attempted to preserve existing permissions for now, but I'm not sure the the defaults are optimal, and this adds complexity to the class.

1. Are there any downsides to granting full read permission to all consumers? This would mean consumers could read any exchange/queue/binding, but our messages are not sensitive.
2. Do consumers still need configure permissions by default? If puppet is supposed to declare the queues, then it seems like the consumers don't need to be doing that any more?
3. Do consumers need write permission at all?

Trello: https://trello.com/c/FFLNonSI/247-make-rummager-process-requeued-messages

### DIFF
--- a/modules/govuk/manifests/apps/rummager/rabbitmq.pp
+++ b/modules/govuk/manifests/apps/rummager/rabbitmq.pp
@@ -34,11 +34,14 @@ class govuk::apps::rummager::rabbitmq (
 
   # match everything on queue 2 while we test
   govuk_rabbitmq::consumer { 'rummager-v2':
-    ensure        => $toggled_ensure,
-    amqp_pass     => $password,
-    amqp_exchange => 'published_documents',
-    amqp_queue    => 'rummager_to_be_indexed',
-    routing_key   => '*.links',
+    ensure                      => $toggled_ensure,
+    amqp_pass                   => $password,
+    amqp_exchange               => 'published_documents',
+    amqp_queue                  => 'rummager_to_be_indexed',
+    routing_key                 => '*.links',
+    extra_read_permissions      => 'govuk_index_durable|govuk_index_transient',
+    extra_write_permissions     => 'govuk_index_durable|govuk_index_transient',
+    extra_configure_permissions => 'govuk_index_durable|govuk_index_transient',
   }
 
   # deprecated - we will remove this user once we have migrated to the new user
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/rummager/rabbitmq.pp
```diff
@@ -52,4 +53,13 @@
     routing_key   => '*.links',
     create_queue  => false,
   }
+
+  # This queue should replace `rummager_to_be_indexed`
+  govuk_rabbitmq::queue_with_binding { 'govuk_index_durable':
```
dwhenry: I believe the queue must exist before a user can be given permission to them.

modules/govuk/manifests/apps/rummager/rabbitmq.pp
```diff
@@ -34,13 +34,14 @@
...
-    amqp_pass     => $password,
-    amqp_exchange => 'published_documents',
-    amqp_queue    => 'rummager_to_be_indexed',
-    routing_key   => '*.links',
-    amqp_queue_2  => 'rummager_govuk_index',
-    routing_key_2 => '#',
+    ensure                      => $toggled_ensure,
+    amqp_pass                   => $password,
+    amqp_exchange               => 'published_documents',
+    amqp_queue                  => 'rummager_to_be_indexed',
```
dwhenry: do we even need this in the consumer, can we just create the queues and pass in the permission manually? this would simplify a lot of stuff IMO.

modules/govuk/manifests/apps/rummager/rabbitmq.pp
```diff
@@ -34,13 +34,14 @@
...
+    routing_key                 => '*.links',
+    extra_read_permissions      => 'govuk_index_durable|govuk_index_transient',
+    extra_write_permissions     => 'govuk_index_durable|govuk_index_transient',
+    extra_configure_permissions => 'govuk_index_durable|govuk_index_transient',
   }
```
dwhenry: You no longer seem to be creating `rummager_govuk_index` not sure if this was intentional?

modules/govuk/manifests/apps/rummager/rabbitmq.pp
```diff
@@ -34,13 +34,14 @@
 
   # match everything on queue 2 while we test
   govuk_rabbitmq::consumer { 'rummager-v2':
```
dwhenry: be aware that users can NOT be updated, hence the reason we have two users in here already (think we may be able to delete the `rummager` user as we are no longer using them, but would need to check.

modules/govuk/manifests/apps/rummager/rabbitmq.pp
MatMoore: When you say manually, do you mean call the new class from `rummager.pp`? If so,  I agree with this - this is just here as a stopgap for the moment.

modules/govuk/manifests/apps/rummager/rabbitmq.pp
MatMoore: Yes - the intention was to rename it (for clarity more than anything else) and to remove the old one by hand (in the rabbit interface) if we deploy this, since its quicker than rolling out puppet twice.

modules/govuk/manifests/apps/rummager/rabbitmq.pp
MatMoore: I was assuming it was the other way - you can define the pattern up front and it applies to anything create afterwards. I'll make sure to test this.

MatMoore:
Closing as this is no longer needed

PR #6334, opened 2017-08-22
Update Smokey email notification settings

alecgibson:
Prior to moving to the `email-ext` plugin, Smokey email notifications
used to:

- Notify users when the build was fixed. This allows users to track
  whether the build has been fixed, especially if they do not have
  access to the staging or production environments for Jenkins.
- View the log of the failed build. This gives a quick insight to the
  reason for failure, and is again useful for users without staging or
  production access.

This change updates our `email-ext` configuration to restore this
behaviour.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/smokey.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/smokey.yaml.erb
@@ -32,7 +32,9 @@
           room: <%= @slack_room %>
       - email-ext:
           recipients: govuk-ci-notifications@digital.cabinet-office.gov.uk
+          attach-build-log: true
           failure: true
+          fixed: true
           presend-script: |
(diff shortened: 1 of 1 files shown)


chrisroos:
This is great, @alecgibson. I think it'd be worth mentioning the reason for these changes in the commit note:

* Email notifications of the build being fixed allow us to see whether a failing build has been fixed without visiting the Jenkins interface. This is particularly useful for people that don't have access to Jenkins in staging and production.

* Email of the build log is particularly useful for people that don't have access to Jenkins in staging and production.

It's probably also worth mentioning that these changes are essentially restoring the behaviour that was in place when we were using the standard email notification mechanism.

alecgibson:
@chrisroos I've updated the commit message.

PR #6337, opened 2017-08-23
Build Content API docs from production

thomasleese:
Currently it builds from master, but it would be better for the
documentation to represent production.


modules/govuk_jenkins/templates/jobs/govuk_content_api_docs.yaml.erb
```diff
@@ -30,6 +30,7 @@
           rm -rf content-store
 
           git clone https://github.com/alphagov/content-store.git
+          git checkout deployed-to-production
```
tijmenb: This needs to go after the `cd`

PR #6338, opened 2017-08-23
Disable theme filtering in CPM

chao-xian:
Add env vars that toggle the feature in the Content Performance Manager.

See also https://github.com/alphagov/content-performance-manager/pull/296

### DIFF
--- a/hieradata/production.yaml
+++ b/hieradata/production.yaml
@@ -66,6 +66,8 @@ base::supported_kernel::enabled: true
 
 environment_ip_prefix: '10.3'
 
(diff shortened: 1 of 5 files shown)


hieradata/production.yaml
```diff
@@ -66,6 +66,8 @@ base::supported_kernel::enabled: true
 
 environment_ip_prefix: '10.3'
 
+govuk::apps::content_performance_manager::feature_auditing_allocation: true
```
alecgibson: Are we definitely enabling this in production?

chao-xian: :man_facepalming:

PR #6353, opened 2017-08-25
Add an uptime collecting service

thomasleese:
This sends the uptime data it collects to statsd which allows it
to be viewed in Graphite.

We decided to go down the route of a service rather than taking
advantage of the Icinga checks because we wanted to have control
over when the /healthcheck endpoint will be polled.

[Trello Card](https://trello.com/c/zveUxj8x/943-spike-uptime-and-availability-metrics-timebox-3-days)

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -245,6 +245,7 @@ monitoring::checks::enable_support_check: false
 monitoring::checks::pingdom::enable: false
 monitoring::checks::ses::region: eu-west-1
 monitoring::checks::smokey::environment: 'integration'
+monitoring::uptime_collector::environment: 'integration'
 
 postfix::smarthost:
   - 'email-smtp.eu-west-1.amazonaws.com:587'
--- a/modules/grafana/files/dashboards/application_uptime.json
+++ b/modules/grafana/files/dashboards/application_uptime.json
@@ -1,0 +1,97 @@
+{
+  "id": null,
+  "title": "Application Uptime",
+  "tags": [],
+  "style": "dark",
+  "timezone": "browser",
+  "editable": true,
+  "hideControls": false,
+  "sharedCrosshair": false,
+  "rows": [
+    {
+      "collapse": false,
+      "editable": true,
+      "height": "650px",
+      "panels": [
+        {
+          "columns": [],
+          "datasource": "Graphite",
+          "editable": true,
+          "error": false,
+          "fontSize": "100%",
+          "id": 4,
+          "isNew": true,
+          "links": [],
+          "pageSize": null,
+          "scroll": true,
+          "showHeader": true,
+          "sort": {
+            "col": 0,
+            "desc": true
+          },
+          "span": 12,
+          "styles": [
+            {
+              "dateFormat": "YYYY-MM-DD",
+              "pattern": "Time",
+              "type": "date"
+            },
+            {
+              "colorMode": "cell",
+              "colors": [
+                "rgba(245, 54, 54, 0.9)",
+                "rgba(237, 129, 40, 0.89)",
+                "rgba(50, 172, 45, 0.97)"
+              ],
+              "decimals": 2,
+              "pattern": "/.*/",
+              "thresholds": ["0.9931", "1"],
+              "type": "number",
+              "unit": "percentunit"
+            }
+          ],
+          "targets": [
+            {
+              "hide": false,
+              "refId": "A",
+              "target": "aliasByNode(summarize(stats.gauges.uptime.*, '1d', 'avg'), 3)",
+              "textEditor": false
+            }
+          ],
+          "title": "Uptimes",
+          "transform": "timeseries_to_columns",
+          "type": "table"
+        }
+      ],
+      "title": "Uptimes"
+    }
+  ],
+  "time": {
+    "from": "now-6M",
+    "to": "now"
+  },
+  "timepicker": {
+    "refresh_intervals": [
+      "5s",
+      "10s",
+      "30s",
+      "1m",
+      "5m",
+      "15m",
+      "30m",
+      "1h",
+      "2h",
+      "1d"
+    ]
+  },
+  "templating": {
+    "list": []
+  },
+  "annotations": {
+    "list": []
+  },
+  "schemaVersion": 12,
+  "version": 1,
+  "links": [],
+  "gnetId": null
+}
--- a/modules/monitoring/files/usr/local/bin/govuk_uptime_collector
+++ b/modules/monitoring/files/usr/local/bin/govuk_uptime_collector
@@ -1,0 +1,71 @@
+#!/usr/bin/env ruby
+
+# Usage: govuk_uptime_collector <environment> <service ...>
+
+require "csv"
+require "fileutils"
+require "net/http"
+require "time"
+require "statsd"
+
+class Collector
+  def initialize(service_name, environment)
+    @statsd = Statsd.new("127.0.0.1")
+    @service_name = service_name
+    @environment = environment
+  end
+
+  def call
+    while true
+      send_to_statsd(check_service_up)
+      sleep 5
+    end
+  end
+
+private
+
+  attr_reader :statsd, :service_name, :environment
+
+  def healthcheck_uri
+    @healthcheck_uri ||= begin
+      prefix = environment == "production" ? "#{service_name}" : "#{service_name}.#{environment}"
+      URI("https://#{prefix}.publishing.service.gov.uk/healthcheck")
+    end
+  end
+
+  def check_service_up
+    http = Net::HTTP.new(healthcheck_uri.host, healthcheck_uri.port)
+    http.use_ssl = true
+    http.read_timeout = 20
+
+    request = Net::HTTP::Head.new(healthcheck_uri.request_uri)
+    res = http.request(request)
+
+    status_code = res.code.to_i
+    (200..299).include?(res.code.to_i)
+  end
+
+  def statsd_key
+    @statsd_key ||= "uptime.#{service_name}"
+  end
+
+  def send_to_statsd(status)
+    statsd.gauge(statsd_key, status ? 1 : 0)
+  end
+end
+
+def main
+  threads = []
+
+  environment = ARGV[0]
+
+  ARGV[1..-1].each do |service_name|
+    threads << Thread.new do
+      Collector.new(service_name, environment).call
+    end
+  end
+
+  threads.each(&:join)
+end
+
+main
--- a/modules/monitoring/templates/govuk-uptime-collector.conf.erb
+++ b/modules/monitoring/templates/govuk-uptime-collector.conf.erb
@@ -1,0 +1,9 @@
+description "govuk-uptime-collector"
+
+start on runlevel [2345]
+respawn
+
+env RBENV_VERSION=2.3
+env PATH=/usr/lib/rbenv/shims:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
+exec govuk_uptime_collector <%= @environment %> publishing-api content-store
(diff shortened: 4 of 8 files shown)


thomasleese:
Tested and working on integration.

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env ruby
+
+# Usage: ./collect.rb <environment> <service ...>
```
kevindew: Looks like this usage is out of date

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -0,0 +1,81 @@
...
+
+require "csv"
+require "fileutils"
+require "net/http"
+require "time"
+require "statsd"
+
+class Array
+  def each_after(n)
```
kevindew: You'd probably want to approach this with a [refinement](https://ruby-doc.org/core-2.1.1/doc/syntax/refinements_rdoc.html) rather than a monkey patch, I don't think it's actually necessary though

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -0,0 +1,81 @@
...
+private
+
+  attr_reader :statsd, :service, :environment
+
+  def healthcheck_uri
+    @healthcheck_uri ||= (if environment == "production"
+      URI("https://#{service}.publishing.service.gov.uk/healthcheck")
+    else
+      URI("https://#{service}.#{environment}.publishing.service.gov.uk/healthcheck")
+    end)
```
kevindew: This might be a bit cleaner as:
```
@healthcheck_uri ||= begin 
  prefix = service
  prefix << ".#{environment} unless environment == "production"
  URI("https://#{prefix}.publishing.service.gov.uk/healthcheck")
end
```

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -0,0 +1,81 @@
...
+    each_with_index do |elem, i|
+      yield elem if i >= n
+    end
+  end
+end
+
+class Collector
+  def initialize(service, environment)
+    @statsd = Statsd.new("127.0.0.1")
+    @service = service
```
kevindew: Not sure if this is a good suggestion or not, but service seems like such a loaded word to me that I think `service_name` might be easier to follow?

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -0,0 +1,81 @@
...
+  def send_to_statsd(status)
+    statsd.gauge("uptime.#{service}", status ? 1 : 0)
+  end
+end
+
+# def append_to_csv(filename, service, status)
+#   CSV.open(filename, "ab") do |csv|
+#     csv << [service, Time.now.utc.iso8601, status]
+#   end
+# end
```
kevindew: Probably want to delete this

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -0,0 +1,81 @@
...
+
+  def check_status
+    http = Net::HTTP.new(healthcheck_uri.host, healthcheck_uri.port)
+    http.use_ssl = true
+
+    request = Net::HTTP::Head.new(healthcheck_uri.request_uri)
+    res = http.request(request)
+
+    status_code = res.code.to_i
+    status_code >= 200 && status_code <= 299
```
kevindew: this could be `(200..299).include?(res.code.to_i)`

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -0,0 +1,81 @@
...
+
+  def healthcheck_uri
+    @healthcheck_uri ||= (if environment == "production"
+      URI("https://#{service}.publishing.service.gov.uk/healthcheck")
+    else
+      URI("https://#{service}.#{environment}.publishing.service.gov.uk/healthcheck")
+    end)
+  end
+
+  def check_status
```
kevindew: I'd maybe change the status in this to be something like `check_service_up` or something along those lines - `status` doesn't really feel like something that's a boolean whereas whether something is up is more of a true or false.

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -0,0 +1,81 @@
...
+#     csv << [service, Time.now.utc.iso8601, status]
+#   end
+# end
+
+def main
+  threads = []
+
+  environment = ARGV[0]
+
+  ARGV.each_after(1) do |service|
```
kevindew: You could do ARGV[1..-1].each here to avoid the need for the monkey patched method

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -0,0 +1,81 @@
...
+    threads << Thread.new do
+      Collector.new(service, environment).call
+    end
+  end
+
+  threads.each { |thread| thread.join }
```
kevindew: code golf but could be `threads.each(&:join)`

modules/monitoring/templates/govuk-uptime-collector.conf.erb
```diff
@@ -0,0 +1,5 @@
+description "govuk-uptime-collector"
+
+start on runlevel [2345]
+
+exec /usr/lib/rbenv/versions/2.3.1/bin/ruby /usr/local/bin/govuk_uptime_collector <%= @environment %> publishing-api content-store specialist-publisher travel-advice-publisher publisher whitehall
```
kevindew: Is it normal for us to tie to a particular ruby version like this? I just wonder that this might get missed if ruby 2.3 support is dropped. I wonder if there's a way we can execute rbenv and just tell it latest version of ruby 2?

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -0,0 +1,81 @@
...
+      URI("https://#{service}.#{environment}.publishing.service.gov.uk/healthcheck")
+    end)
+  end
+
+  def check_status
+    http = Net::HTTP.new(healthcheck_uri.host, healthcheck_uri.port)
+    http.use_ssl = true
+
+    request = Net::HTTP::Head.new(healthcheck_uri.request_uri)
+    res = http.request(request)
```
kevindew: Do you know if this has a timeout on it? I guess we want it die if a healthcheck takes more than a couple of seconds.

modules/monitoring/templates/govuk-uptime-collector.conf.erb
```diff
@@ -0,0 +1,5 @@
+description "govuk-uptime-collector"
+
+start on runlevel [2345]
+
+exec /usr/lib/rbenv/versions/2.3.1/bin/ruby /usr/local/bin/govuk_uptime_collector <%= @environment %> publishing-api content-store
```
kevindew: Do other processes specify ruby version like this? It feels a bit precarious as to when we drop 2.3.1 from stack. Is there anyway we can tell rbenv to run it against a 2.* version of Ruby?

modules/monitoring/templates/govuk-uptime-collector.conf.erb
```diff
@@ -0,0 +1,6 @@
+description "govuk-uptime-collector"
+
+start on runlevel [2345]
+respawn
+
+exec /usr/lib/rbenv/versions/2.3.1/bin/ruby /usr/local/bin/govuk_uptime_collector <%= @environment %> publishing-api content-store
```
kevindew: Do other processes specify a ruby version like this? it feels a bit precarious for when we end usage of 2.3.1. 

I wonder if there's a way we can execute through rbenv to specify we just want a 2.x version of Ruby to run it?

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
thomasleese: I was hoping Ruby supported `200 <= status_code <= 299` à la Python but sadly not!

kevindew: haha added to the long list of Rubys failings against the mighty Python?

thomasleese: It's so long now I've lost count

modules/monitoring/templates/govuk-uptime-collector.conf.erb
thomasleese: I couldn't figure out a way of pinning to any version of Ruby 2.x, but I was able to use rbenv to select 2.3 instead of 2.3.1 specifically.

thomasleese:
screen shot 2017-08-29 at 09 59 08

hieradata/integration.yaml
```diff
@@ -243,6 +243,8 @@ monitoring::checks::enable_support_check: false
 monitoring::checks::pingdom::enable: false
 monitoring::checks::ses::region: eu-west-1
 monitoring::checks::smokey::environment: 'integration'
+monitoring::uptime_collector::enabled: true
```
surminus: if this is `true` everywhere can we just set that to the default?

thomasleese: Oh yes... in my mind this was here to stop the uptime collecting service going on anything other than the monitoring box but I realise that doesn't make sense.

surminus:
Other than one comment this looks fine

modules/grafana/files/dashboards/application_uptime.json
```diff
@@ -0,0 +1,108 @@
...
+      "5m",
+      "15m",
+      "1h",
+      "6h",
+      "12h",
+      "24h",
+      "2d",
+      "7d",
+      "30d"
+    ]
```
kevindew: Do we want these to be in the days region rather than seconds / minutes? as I guess the data only applies once you're beyond a day?

I say this though being kind of aware that the time periods are a real pain on this stuff as it seems to always be at 1 week.

modules/monitoring/templates/govuk-uptime-collector.conf.erb
```diff
@@ -0,0 +1,9 @@
+description "govuk-uptime-collector"
+
+start on runlevel [2345]
+respawn
+
+env RBENV_VERSION=2.3
+env PATH=/usr/lib/rbenv/shims:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
+exec govuk_uptime_collector <%= @environment %> publishing-api content-store
```
kevindew: Out of my own ignorance, do we know what happens if this fails? are we notified at all - or does something monitor this (monitor inception)

thomasleese: If it fails, the service will be respawned automatically. I figured we would notice if the uptime stopped being reported, but perhaps we should add something more explicit here?

kevindew: Sure, I'm not sure we would notice - can't imagine we'd look at it regularly guessing someone will see it in weeks or months rather than hourly.

thomasleese: I've added an Icinga check for it.

kevindew: Great! will that link to the docs if there's a problem?

modules/grafana/files/dashboards/application_uptime.json
```diff
@@ -0,0 +1,104 @@
...
+      "1h",
+      "2h",
+      "1d"
+    ],
+    "time_options": [
+      "1d",
+      "2d",
+      "7d",
+      "14d",
+      "30d"
```
kevindew: Should this carry on until 6m or longer since it defaults to showing 6m?

thomasleese:
@kevindew Cool, those should be resolved now!

PR #6356, opened 2017-08-25
Bump max_connections to postgres

tijmenb:
Bouncer and transition are seeing a lot of `PG::ConnectionBad: remaining connection slots are reserved for non-replication superuser connections` errors. It means there are no more database connections for the app to use.

https://sentry.io/govuk/app-bouncer/issues/335728174/events

This happens only on integration and staging, not production. I think this is because the `max_connections` in those environments is set to 120, while it's 200 in production.

Note that it was bumped from 100 to 120 last month: https://github.com/alphagov/govuk-puppet/pull/6128. I think that PR intended to solve the same problem.

I'm wondering if this is the correct solution, or that this is actually an application issue that should be solved there (perhaps connected to https://github.com/alphagov/bouncer/pull/143/files). Opening this to see if anyone has some more insight.

### DIFF
--- a/hieradata/class/staging/postgresql_primary.yaml
+++ b/hieradata/class/staging/postgresql_primary.yaml
@@ -1,1 +1,0 @@
-govuk_postgresql::server::max_connections: 120
--- a/hieradata/class/production/postgresql_primary.yaml
+++ b/hieradata/class/production/postgresql_primary.yaml
@@ -11,5 +11,3 @@ lv:
   data:
     pv: '/dev/sdd1'
     vg: 'postgresql'
-
-govuk_postgresql::server::max_connections: 200
--- a/hieradata/class/production/postgresql_standby.yaml
+++ b/hieradata/class/production/postgresql_standby.yaml
@@ -1,1 +1,0 @@
-govuk_postgresql::server::max_connections: 200
--- a/hieradata/class/staging/postgresql_standby.yaml
+++ b/hieradata/class/staging/postgresql_standby.yaml
@@ -1,1 +1,0 @@
-govuk_postgresql::server::max_connections: 120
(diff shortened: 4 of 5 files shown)


cbaines:
I did once encounter this issue in production, and I wrote a specific dashboard that covers some useful metrics [1].

My best guess is that this is something due to the way data is replicated from production to staging and integration, and I don't think increasing the maximum connections will help, but I'm not against trying just to see what it does.

1: https://grafana.integration.publishing.service.gov.uk/dashboard/file/bouncer.json?from=now-7d&to=now

hieradata/class/staging/postgresql_primary.yaml
```diff
@@ -1 +1 @@
-govuk_postgresql::server::max_connections: 120
+govuk_postgresql::server::max_connections: 200
```
surminus: If we're setting this everywhere we may as well just set it as the default in the manifest itself.

surminus:
Context: this was originally done in Production after incidents where we ran out of spare connections. We'd got to the point that we had enough apps (~30), running on 3 backend boxes, connecting to the postgres server. If they keep their connections open, then we had 90 permanent connections, and when we tried to do backups or something it would run out of available connections. 

Production was the only one bumped to 200 because it was the only one that had it's memory increased. 

I'm not against setting 200 as the default, and if we see Staging or Integration run out of memory then we can deal with it.

Equally I agree that if it's an application causing the number of connections to increase then that problem should be fixed, but it may just be a natural threshold which needs raising.

tijmenb:
@surminus updated PR to set the default to 200.

surminus:
Try repushing to fix failing travis tests.

PR #6362, opened 2017-08-29
Use response headers from Asset Manager when proxying to S3

floehopper:
This is a follow-up to [the PR in which we made it possible to configure the Asset Manager to proxy asset requests to S3 via Nginx][1]. Since then we've merged [an Asset Manager PR][2] which allows us to divert a percentage of asset request traffic to be proxied to S3. We plan to roll this out gradually while carefully monitoring the load on and performance of the production stack.

Unfortunately Nginx and S3 use different algorithms to generate an ETag response header and, since the S3 copy of each asset has a different last modified time, the Last-Modified response header values are different too. So in order to avoid unnecessary cache invalidation and hence confound our monitoring of load & performance, we've made [a change to the Asset Manager Rails app to generate Nginx-esque ETag & Last-Modified response headers][3].

The changes to the Asset Manager Nginx config in this commit store the response header values from Rails and then add them to the response from S3 before the response is sent back to the user thus avoiding the unnecessary cache invalidation mentioned above.

We're also about to make [another change to the Asset manager Rails app to send Cache-Control, Content-Disposition & Content-Type headers when instructing Nginx to proxy requests to S3][4]. The idea behind this is to avoid duplicate Cache-Control headers ending up in the response to the user and to make the Rails app the canonical source for response headers. Fortunately Nginx copies the Cache-Control & Content-Disposition headers into the proxied response by default, but we need to explicitly store and set the Content-Type header in the same way as we are doing for ETag & Last-Modified.

[1]: https://github.com/alphagov/govuk-puppet/pull/6280
[2]: https://github.com/alphagov/asset-manager/pull/152
[3]: https://github.com/alphagov/asset-manager/pull/157
[4]: https://github.com/alphagov/asset-manager/pull/164

### DIFF
--- a/modules/govuk/manifests/apps/asset_manager.pp
+++ b/modules/govuk/manifests/apps/asset_manager.pp
@@ -100,6 +100,12 @@ class govuk::apps::asset_manager(
         alias /var/apps/asset-manager/uploads/assets/$1;
       }
 
+      # Store values from Rails response headers for use in the
+      # cloud-storage-proxy location block below.
+      set $etag_from_rails $upstream_http_etag;
+      set $last_modified_from_rails $upstream_http_last_modified;
+      set $content_type_from_rails $upstream_http_content_type;
+
       location ~ /cloud-storage-proxy/(.*) {
         # Prevent requests to this location from outside nginx
         internal;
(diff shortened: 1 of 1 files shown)


chrisroos:
I've tested these changes in combination with https://github.com/alphagov/asset-manager/pull/164#pullrequestreview-59514854.

PR #6363, opened 2017-08-30
Update assets routing in AWS

surminus:
In AWS we're terminating SSL on the ELBs. We disable SSL on the nginx config, and the router template for assets had some logic which meant if SSL is disabled, then it would also set the upstream path to HTTP rather than the required HTTPS.

This adds a full conditional block for that ensures that all upstream paths are set to HTTPS in AWS.

### DIFF
--- a/modules/router/templates/assets_origin.conf.erb
+++ b/modules/router/templates/assets_origin.conf.erb
@@ -41,13 +41,13 @@ server {
   <%- @asset_routes.each do |alias_path, vhost_name| -%>
   location <%= alias_path %> {
     proxy_set_header Host <%= vhost_name %>.<%= @app_domain %>;
-    proxy_pass <%= @enable_ssl ? 'https' : 'http' %>://<%= vhost_name %>.<%= @app_domain %>;
+    proxy_pass <%= @upstream_ssl ? 'https' : 'http' %>://<%= vhost_name %>.<%= @app_domain %>;
   }
 
   <%- end -%>
 
   location / {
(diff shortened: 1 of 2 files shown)


modules/router/templates/assets_origin.conf.erb
```diff
@@ -38,16 +38,32 @@ server {
...
-    proxy_pass <%= @enable_ssl ? 'https' : 'http' %>://<%= vhost_name %>.<%= @app_domain %>;
-  }
-
-  <%- end -%>
-
-  location / {
-    proxy_set_header Host static.<%= @app_domain %>;
-    proxy_pass <%= @enable_ssl ? 'https' : 'http' %>://static.<%= @app_domain %>;
-  }
+  <%- unless scope.lookupvar('::aws_migration') %>
```
SamLR: is there a reason your doing this as an `unless` rather than swapping the if/else blocks?

modules/router/templates/assets_origin.conf.erb
```diff
@@ -38,16 +38,32 @@ server {
...
+    <%- @asset_routes.each do |alias_path, vhost_name| -%>
+    location <%= alias_path %> {
+      proxy_set_header Host <%= vhost_name %>.<%= @app_domain %>;
+      proxy_pass <%= @enable_ssl ? 'https' : 'http' %>://<%= vhost_name %>.<%= @app_domain %>;
```
SamLR: could we move the `::aws_migration` look up to just wrap these proxy_pass lines to make it clearer what the difference is?

PR #6371, opened 2017-08-30
Add Dennis Tsiang to manifest

DennisTsiang:


modules/users/manifests/dennistsiang.pp
```diff
@@ -0,0 +1,8 @@
+# Create the dennistsiang user
+class users::dennistsiang {
+  govuk_user { ‘dennistsiang’:
```
h-lame: Looks like we have "fancy" quotes in here and the puppet parser doesn't like them.  Canyou replace with "normal" single quotes.

PR #6377, opened 2017-08-31
Escape commas for Emergency Banner Deploy task

alecgibson:
Rake is not very good at handling arguments with commas in. It has
surprising behaviour, such as:

`rake test[‘foo, bar’]` will give you two arguments: `foo` and `bar`.

If you escape the comma, then they are provided as a single argument
with a comma: `foo, bar`.

This change escapes the commas in the provided arguments, in order
to reduce human error when manually escaping. Since the command is
executed inside speech marks, this change also escapes quotes.

The new task has been tested with arguments whose commas and
quotes were not manually escaped:

banner-deploy-quotes-commas


leenagupte:
We had a trello card for this: https://trello.com/c/xAbDe1EK/232-the-emergency-banner-does-not-display-correctly-if-the-details-contain-a-comma

PR #6384, opened 2017-08-31
Set special Sentry environment name for AWS

tijmenb:
On AWS, the Sentry environment will be either `integration-blue-aws` or `integration-green-aws`.


tijmenb:
@surminus @SamLR this doesn't seem to have worked:

https://sentry.io/govuk/app-router-api/issues/343939049 still shows `integration`. Can we check if the `SENTRY_CURRENT_ENV` is set correctly on AWS? Not sure where the problem would be.

SamLR:
@tijmenb I'll have a look at this now

surminus:
@tijmenb I haven't deployed that Puppet change into AWS yet. It doesn't happen automagically yet.

PR #6385, opened 2017-08-31
Add X-Frame-Options header when proxying requests to S3

chrisroos:
This ensures that we're sending the `X-Frame-Options` header with a value of `DENY` if `$deny_framing` is set to `true`; irrespective of whether we serve the asset file from the filesystem or from S3.

This address [issue 166 in asset-manager][1].

[1]: https://github.com/alphagov/asset-manager/issues/166

### DIFF
--- a/modules/govuk/manifests/apps/asset_manager.pp
+++ b/modules/govuk/manifests/apps/asset_manager.pp
@@ -76,6 +76,8 @@ class govuk::apps::asset_manager(
       app => $app_name,
     }
 
+    $deny_framing = true
+
     $nginx_extra_config = inline_template('
       client_max_body_size 500m;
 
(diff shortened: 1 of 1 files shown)


chrisroos:
@floehopper has just mentioned that we might want to toggle this option based the `@deny_framing` flag so that it matches the behaviour of the header being set in the outer context.

chrisroos:
@floehopper: I've updated this branch so that setting the `X-Frame-Options` header is based on the value of the `$deny_framing` variable. This matches the default behaviour in proxy-vhost.conf. Although your PR approval appears to remain valid, I wondered whether you could take a look to make sure the changes make sense?

I've tried these changes on the development VM and can confirm the presence of the `X-Frame-Options` header when `$deny_framing` is `true`, and the absence of the header when `$deny_framing` is `false`.

floehopper:
@chrisroos: The changes in the commits make sense to me. Also I've just run `govuk_puppet` on this branch on my development VM and it correctly adds the header in both places in the Nginx config. When set the `$deny_framing` variable to false in `asset_manager.pp` and run `govuk_puppet` again, the header is not added in either place as expected. So I'm still happy for this to be merged.

chrisroos:
Thanks @floehopper. I've rebased on master and force pushed in preparation for merging.

PR #6395, opened 2017-09-01
Increase Postgres Primary disk space

kevindew:
Postgres Primary is running out of space when performing backups in the
production environment since there is not enough disk space available
for the uncompressed backups.

At current time we have a 126G partition on integration and production
and a 156G one on staging. So this commit only adds extra drives to the
integration and production environments.

### DIFF
--- a/hieradata/class/integration/postgresql_primary.yaml
+++ b/hieradata/class/integration/postgresql_primary.yaml
@@ -7,6 +7,7 @@ lv:
       - '/dev/sdc1'
       - '/dev/sde1'
       - '/dev/sdf1'
+      - '/dev/sdg1'
     vg: 'backups'
   data:
     pv: '/dev/sdd1'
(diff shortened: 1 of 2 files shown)


surminus:
Do we need to increase space in all environments?

kevindew:
> Do we need to increase space in all environments?

I thought it was just production, but looking it would probably be prudent to do it for all environments.

We currently have:

production: `/dev/mapper/backups-postgresql  126G  104G   23G  83% /var/lib/autopostgresqlbackup`
staging: `/dev/mapper/backups-postgresql  158G  121G   37G  77% /var/lib/autopostgresqlbackup`
integration: `/dev/mapper/backups-postgresql  126G   99G   28G  79% /var/lib/autopostgresqlbackup`

I wonder if a mistake happened for staging to be bigger? 

Do you think it'd make sense to drop the `lv:` definition from `production/postgresql_primary.yaml` and use this as an opportunity to give them all the same drives. It's quite confusing to my brain how it's set up and overridden in multiple files.

kevindew:
On second thoughts looking at how it's configured staging will have data on `pv` of: `/dev/sdc1`whereas integration and production have it on `/dev/sdd1` so I don't think it'd be possible to easily make them consistent. I'm just going to add an extra one for integration too then, since staging is looking good space wise.

kevindew:
@surminus updated to just address production and integration.

PR #6398, opened 2017-09-01
[WIP] Test internal Plek routing

surminus:
In AWS we wish to route things internally using DNS records.

Applications use Plek to find services, and quite often we find that things fail because they're requesting the external name. Internal services should have internal service records and internal ELBs, so we want to test setting the Plek root domain to the internal domain name and see what breaks.

### DIFF
--- a/modules/govuk/manifests/deploy/config.pp
+++ b/modules/govuk/manifests/deploy/config.pp
@@ -76,7 +76,12 @@ class govuk::deploy::config(
     require => File['/etc/govuk'],
   }
 
-  $app_domain = hiera('app_domain')
+  # In AWS we want to route things using the internal DNS name
+  if $::aws_migration {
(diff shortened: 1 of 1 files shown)


cbaines:
With govuk-guix, I attempted to configure all apps explicitly with PLEK_SERVICE_..._URI, and one thing I came across when getting things working is that Whitehall (and possibly others...) use Plek to lookup themselves [1].

So, if you have internal and external names, and assuming that you wouldn't be able to access these as a regular user, you might want to by default explicitly set the PLEK_SERVICE_..._URI environment variable for the service itself to the external address, to cope with the case where the application uses this in links.

1: https://github.com/alphagov/whitehall/blob/638dd41447f093683a8c0a6ab2df79b17542cb86/app/controllers/admin/documents_controller.rb#L4

surminus:
Close for now and we can revisit this later.

PR #6400, opened 2017-09-01
Set hardcoded name for gor logrotate

dwhenry:
https://trello.com/c/2Way8sKg/254-fully-implement-a-gor-solution


modules/govuk_search/manifests/gor.pp
```diff
@@ -32,7 +32,7 @@
       mode   => '0755',
     }
 
-    @logrotate::conf { "govuk-${title}":
+    @logrotate::conf { "govuk_search__gor__logs":
```
surminus: failed tests because these need to be single quotes

PR #6401, opened 2017-09-03
Various updates

surminus:
A few changes to ensure applications work in AWS. See commits for detail. 

https://trello.com/c/DS5stwo8/730-fix-frontend-flows


hieradata/common.yaml
```diff
@@ -2,6 +2,7 @@
 HIERA_SAFETY_CHECK: true
 
 app_domain: 'dev.gov.uk'
+app_domain_internal: 'dev.govuk-internal.digital'
```
afda16: `digital` doesn't have any meaning on non-AWS, maybe we should set app_domain_internal to `dev.gov.uk`?

surminus: Sure.

PR #6402, opened 2017-09-04
Assets shouldn't be synced to aws

SamLR:


issyl0:
Why not?

SamLR:
Because there's a lot of them and we hope that they'll be run off of S3 soon

PR #6407, opened 2017-09-04
Add router_nodes_file for router-api

surminus:
One of the functions of router-api is to send a reload routes request to the router.

Traditionally this has been achieved by hardcoding the hostnames of the router nodes, which hasn't been a problem because we set hostnames of all our machines.

In a dynamic environment this is more problematic because we don't know when machines will be respun and what hostname they'll use.

This adds a basic function that writes out the current nodes to a file, which router-api will then address when it reloads the routes.

Ref: https://github.com/alphagov/router-api/commit/27637862590867f315eea74d0722da827c48fed7

Ideally in the future this kind of function should be solved using a key/value store which is more in tune with dynamic environments, but this solves the problem for the short term.

### DIFF
--- a/modules/govuk/manifests/apps/router_api.pp
+++ b/modules/govuk/manifests/apps/router_api.pp
@@ -31,15 +31,20 @@
 # [*sentry_dsn*]
 #   The URL used by Sentry to report exceptions
 #
+# [*router_nodes_class*]
+#   The class or classes of machine that run router that require reloading
+#   after app deployment. Acceptable formats are "cache" or "cache,draft_cache"
+
 class govuk::apps::router_api(
   $port = '3056',
   $mongodb_name,
   $mongodb_nodes,
-  $router_nodes,
+  $router_nodes = [],
   $vhost = 'router-api',
   $secret_key_base = undef,
   $sentry_dsn = undef,
   $errbit_api_key = undef,
+  $router_nodes_class = undef,
 ) {
   $app_name = 'router-api'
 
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/router_api.pp
```diff
@@ -79,4 +84,25 @@
...
+      $router_nodes_file = '/etc/router_nodes'
+
+      cron::crondotdee { 'update-router-nodes':
+        command => "/usr/local/bin/govuk_node_list -c ${router_nodes_class} | sed 's/$/:${router_port}/g' > ${router_nodes_file}",
+        hour    => '*',
+        minute  => '*/5',
+        mailto  => '""',
+      }
+
+      govuk::app::envvar { "${title}-ROUTER_NODES_FILE":
```
afda16: Does the application get refreshed if this file changes?

surminus: The application reads in the location of the file rather than the values of the file, and then reads the file to gather the values each time it needs to reload router, so it _should_ read in the newest values (at least that's the theory).

cbaines: I don't think its going to work like that yet. The nodes are stored in a variable associated with the class [1], which is set when the Rails initialisers run.

The RouterReloader is used from the controller here [2], and it just calls the reload function. The reload function then reads the value in `router_reload_urls`.

1: https://github.com/alphagov/router-api/blob/b1be434b7e55e587ecdb8f2e4165eecb41293120/lib/router_reloader.rb#L9
2: https://github.com/alphagov/router-api/blob/master/app/controllers/routes_controller.rb#L41

PR #6410, opened 2017-09-05
Add asset manager header to emit location of files

chrisroos:
The addition of this informational header will allow us to see whether an individual asset-manager request was served by an asset hosted on S3 or on NFS. Writing the value of the header to the `json_event` nginx log will mean that it ends up in Kibana and we'll be able to use this to monitor the proportion of requests being served by assets hosted on S3 vs NFS. This is all related to the work we're doing to host assets on S3.

I've tested this change on the Developer VM and can confirm that the HTTP header is added as expected and that the value of the header is written to the log file.

I'm slightly uncomfortable with the header being written to nginx log files for _all_ apps but there seems to be some precedent for this in PR #5577.

### DIFF
--- a/modules/govuk/manifests/apps/asset_manager.pp
+++ b/modules/govuk/manifests/apps/asset_manager.pp
@@ -89,6 +89,7 @@ class govuk::apps::asset_manager(
       # publicly as it is an internal path mapping.
       location ~ /raw/(.*) {
         internal;
+        add_header GOVUK-Asset-Manager-File-Store NFS;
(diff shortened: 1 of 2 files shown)


chrisroos:
@boffbowsh - I've added you as a reviewer as suggested by GitHub.

@tijmenb - I've requested your review as we chatted about this on Slack yesterday.

chrisroos:
Thanks @floehopper. I'm going to get this merged as I'm keen to use it to monitor our rollout of serving assets from S3. We can always revert the change if it turns out that it's causing a problem.

PR #6411, opened 2017-09-05
Remove metadata api

elenatanasoiu:
For: https://trello.com/c/pKthUgvD/199-retire-metadata-api

This removes the metadata-api application. 

The last part of removing metadata-api config is done here: https://github.com/alphagov/govuk-puppet/pull/6412, which will be merged after this current PR is deployed and tested.

PR #6412, opened 2017-09-05
Remove last of metadata api

elenatanasoiu:
For: https://trello.com/c/pKthUgvD/199-retire-metadata-api

NB: Depends on https://github.com/alphagov/govuk-puppet/pull/6411 being merged and deployed first. 

This removes the last of the config files for retiring of metadata-api.


development-vm/Procfile
```diff
@@ -89,8 +89,6 @@ service-manual:        govuk_setenv service-manual        ./run_in.sh ../../gove
 # collections-api used port3084
 info-frontend:         govuk_setenv info-frontend         ./run_in.sh ../../info-frontend bundle exec rails server -p 3085
 # sidekiq-monitoring for transition uses port 3086
-# metadata-api uses port 3087
```
h-lame: Maybe it's worth leaving this in (but changing `uses` to `used` like on line 89 about collections-api above?  Not sure what the pattern is though.

elenatanasoiu: There's description of the pattern on step 9 here: https://docs.publishing.service.gov.uk/manual/retiring-an-application.html

>Leave a comment in the Procfile indicating that the port used to be used by this app, to avoid port conflicts causing a problem running this app locally in the future (eg https://github.digital.cabinet-office.gov.uk/gds/development/pull/149).

PR #6413, opened 2017-09-05
Enable search GOR for all environments

dwhenry:
https://trello.com/c/2Way8sKg/254-fully-implement-a-gor-solution


hieradata/production.yaml
```diff
@@ -200,6 +200,8 @@ govuk::node::s_transition_postgresql_standby::redirector_ip_range: "%{hiera('gov
 govuk_postgresql::backup::auto_postgresql_backup_hour: 5
 govuk_postgresql::backup::auto_postgresql_backup_minute: 0
 
+govuk_search::gor::enabled: true
```
surminus: Set this in `common.yaml`, and set it to `false` in `development.yaml`

PR #6414, opened 2017-09-05
Remove references to metadata-api

boffbowsh:
It is no longer in use


tijmenb:
I think this duplicates https://github.com/alphagov/govuk-puppet/pull/6412.

boffbowsh:
Heh, so it does. I really need to pay more attention to github notifications

PR #6416, opened 2017-09-05
Add Puppet for Asset Master in AWS

surminus:
In AWS we're making use of EFS.

Traditionally we created a large disk on the asset-master and then exported that to our backend machines.

In AWS we're creating an EFS disk, and mounting it onto each server that requires it. Asset master can do all the virus scan stuff it needs to do, and the backend machines can access and upload to the same EFS mount as required.

https://trello.com/c/YGAQ2KDD/732-aws-migration-asset-master-asset-slave


surminus:
Need to check ordering of file creation and mount point.

PR #6417, opened 2017-09-06
Adding myself as a new user

maxgds:
Following through the documentation at https://docs.publishing.service.gov.uk/manual/get-started.html and looking to add myself as a new user here.

### DIFF
--- a/modules/users/manifests/maxlehmann.pp
+++ b/modules/users/manifests/maxlehmann.pp
@@ -1,0 +1,8 @@
+# Creates the maxlehmann user
+class users::maxlehmann {
+  govuk_user { 'maxlehmann':
+    fullname => 'Max Lehmann',
+    email    => 'max.lehmann@digital.cabinet-office.gov.uk',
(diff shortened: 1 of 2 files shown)


modules/users/manifests/maxlehmann.pp
```diff
@@ -0,0 +1,8 @@
+# Creates the maxlehmann user
+class users::maxlehmann {
+  govuk_user { 'maxlehmann':
+    fullname => Max Lehmann'',
```
cbaines: You've got a couple of rouge quotes here `''`.

maxgds: Thanks, now fixed.

nickcolley:
@maxgds sorry I didn't introduce you to how we write commits (https://github.com/alphagov/styleguides/blob/master/git.md#example-1)

In this case we'd do something like 'Add Max Lehmann to manifest'.

We also tend to squash commits together if they're just to fix a small error in a previous commit.

fofr:
Closing in favour of #6422

PR #6419, opened 2017-09-06
Proxy 50% of asset-manager requests to S3

chrisroos:
We've uploaded all assets to S3 in integration. Setting this option to
50% will allow us to test that asset requests are being served by both
NFS and S3 before we roll this out in production.

I've tested this by setting the same option in development.yaml, running
`govuk_puppet` and checking that
/etc/govuk/asset-manager/env.d/PROXY_PERCENTAGE_OF_ASSET_REQUESTS_TO_S3_VIA_NGINX
contains '50'.


chrisroos:
The Travis build is taking too long (I'm impatient) so I'm going to merge this given that it's succeeded in Jenkins.

PR #6423, opened 2017-09-07
Add Nginx proxy entry for docs.content-api.publishing...

thomasleese:
This will be hosted by GitHub Pages, the DNS entry will come later.

[Trello Card](https://trello.com/c/09DCw1RQ/1080-2-create-api-documentation-domain)


surminus:
Is there a reason why this is being routed through our stack and not just straight to Github Pages?

thomasleese:
@surminus I think the main reason was so that we can use the publishing.service.gov.uk SSL certificate, perhaps we don't need to route it through to get that?

PR #6424, opened 2017-09-07
Update errbit env name in AWS

surminus:
This was updated in 3fa1a2292f, but I've realised that this wasn't relevant, and needs to be updated here.


surminus:
The sentry config uses this parameter: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/deploy/config.pp#L88

PR #6425, opened 2017-09-07
Remove Need API

elenatanasoiu:
This removes Need API

For: https://trello.com/c/ClCFMBlG/217-retire-need-api

NB: https://github.com/alphagov/govuk-puppet/pull/6430 depends on this PR being merged first


modules/govuk/manifests/apps/publisher.pp
```diff
@@ -145,9 +140,6 @@
     "${title}-PUBLISHING_API_BEARER_TOKEN":
       varname => 'PUBLISHING_API_BEARER_TOKEN',
       value   => $publishing_api_bearer_token;
-    "${title}-NEED_API_BEARER_TOKEN":
```
h-lame: Does this (and the similar one for whitehall) need to go through the "ensure absent in PR 1, then remove in PR 2 dance" so we make sure we don't leave the env var lying around?  (This is not a leading question, I genuinely don't know)

surminus: I'd suspect they'd be purged, but check the module.

elenatanasoiu: @surminus I'm looking at the envvar module but I can't see an explicit removal of previous vars: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_envvar/manifests/init.pp

elenatanasoiu: I've added `ensure: absent` for all envvars instead of removing them outright. You can see those changes here https://github.com/alphagov/govuk-puppet/pull/6425/commits/d0eaf7b7e633f0cce572344d7a494ddc43f4b8c4

h-lame: yay!

PR #6426, opened 2017-09-07
Add Elasticsearch URI to licencefinder

surminus:
As per the change here:

https://github.com/alphagov/licence-finder/commit/84bd22f519f670d1efb8d49ca052428ececdff5f


modules/govuk/manifests/apps/licencefinder.pp
```diff
@@ -81,4 +81,12 @@
     varname => 'PUBLISHING_API_BEARER_TOKEN',
     value   => $publishing_api_bearer_token,
   }
+
+  if $aws_migration {
```
deanwilson: This should be `$::aws_migration`

surminus: I've typed it a million times, I'm not sure how I didn't do that!

PR #6432, opened 2017-09-11
Proxy 100% of asset-manager-requests to S3 in production

chrisroos:
We've been proxying 20% of requests to S3 since Thu 7 Sep. We're
confident that everything is working as expected and that the change
hasn't had any negative effects (see
https://github.com/alphagov/asset-manager/issues/153 for more info) so
we're now ramping up to 100%.


chrisroos:
Thanks @floehopper! Merging.

PR #6436, opened 2017-09-12
Update the docs on creating a new user account

issyl0:
- These were horribly out of date - we now use GitHub teams to manage
  access, not YAML files with usernames.


tijmenb:
There's also https://docs.publishing.service.gov.uk/manual/access-ci.html, which looks out of date. Maybe we can cross-link too?

PR #6438, opened 2017-09-12
Added user suthagarkathirkamathamby

suthagarht:
Added user suthagarkathirkamathamby


hieradata/integration.yaml
```diff
@@ -314,6 +314,7 @@ users::usernames:
   - stephenharker
   - stevelaing
   - suzannehamilton
+  - suthagarkathirkamathamby
```
surminus: Minor comment, can this be sorted into alphabetical order?

suthagarht: Hi @surminus , I have now placed my name in the correct order. Apologies for that.

PR #6442, opened 2017-09-12
Simplify consumer class by removing exchanges

MatMoore:
RabbitMQ exchanges are only relevant to the publisher of the messages, so its a
bit weird for the consumer stuff to create these.

They're only used in one place for creating test infrastructure.

Simplifying the consumer makes it easier to change the search infrastructure.
At the moment this is difficult because there is a lot of logic for
constructing permission strings.

The ultimate goal is to separate queue creation so we can set up multiple
queues per consumer.

Trello: https://trello.com/c/k2L67c4a/272-cleanup-puppet-module-for-declaring-message-queue-consumers


MatMoore:
Closing in favour of https://github.com/alphagov/govuk-puppet/pull/6445

PR #6445, opened 2017-09-12
Separate permission logic, queues, test exchange creation from RabbitMQ consumer class

MatMoore:
This PR supersedes https://github.com/alphagov/govuk-puppet/pull/6442

Simplifying the consumer makes it easier to change the search infrastructure.
At the moment this is difficult because there is a lot of logic for
constructing permission strings.

By separating out the queue creation we can set up multiple queues per consumer. Rummager has two queues at the moment and I want to add a separate one for bulk reindexing.

This is best reviewed commit by commit.

Things I don't want to address in this PR:
- Implementing the new queue
- Changing the permissions used for each consumer
- Removing the unused RabbitMQ user for rummager

This is a refactor so I want it to have no effect when run.

Trello: https://trello.com/c/k2L67c4a/272-cleanup-puppet-module-for-declaring-message-queue-consumers

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_service/rabbitmq.pp
+++ b/modules/govuk/manifests/apps/email_alert_service/rabbitmq.pp
@@ -29,9 +29,12 @@ class govuk::apps::email_alert_service::rabbitmq (
 ) {
 
   govuk_rabbitmq::consumer { $amqp_user:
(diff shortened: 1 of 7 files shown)


dwhenry:
Seems to make sense to me, but might be worth getting someone with more puppet experience to do a check as well.

PR #6448, opened 2017-09-13
Make postgres connections consistent across environments

tijmenb:
This sets the max_connections for postgres to 200 in staging and integration, like production. It's intended to solve a problem in bouncer:

https://sentry.io/govuk/app-bouncer/issues/335728174/

A previous PR for this was reverted because it changed the default value for other postgresses as well.

### DIFF
--- a/hieradata/class/integration/transition_postgresql_master.yaml
+++ b/hieradata/class/integration/transition_postgresql_master.yaml
@@ -1,0 +1,1 @@
+govuk_postgresql::server::max_connections: 200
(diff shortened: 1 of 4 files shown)


surminus:
I've just had a thought (my brain isn't with it today, sorry), we could add this to`hieradata/class/postgresql_primary.yaml` and `hieradata/class/postgresql_standby.yaml` and it would ensure all environments are consistent and not affect Transition PostgresSQL.

surminus:
However, since it's bouncer that is throwing the errors we should update:

`hieradata/class/integration/transition_postgresql_master.yaml`
`hieradata/class/integration/transition_postgresql_slave.yaml`
`hieradata/class/staging/transition_postgresql_master.yaml`
`hieradata/class/staging/transition_postgresql_slave.yaml`

PR #6450, opened 2017-09-13
[WIP] Try updating to use internal app domain globally

surminus:
I'm removing both frontend-lb and backend-lb so it's probably time to try setting everything to use internal again.

I suspect this will break things, so we may need to be specific with Plek services.

Equally I'm not against apps hitting back against the external load balancer for backend.

https://trello.com/c/goRJMaPL/712-remove-aws-backend-lb-machines


surminus:
This is no longer relevant as we are specifically setting overrides for services.

PR #6453, opened 2017-09-13
Update dev vm replication to use ES 2.4 server

dwhenry:
https://trello.com/c/KMIQEH5p/281-cleanup-tasks

### DIFF
--- a/development-vm/fix-elasticsearch-2.4.sh
+++ b/development-vm/fix-elasticsearch-2.4.sh
@@ -19,7 +19,7 @@ if [ -d "/usr/share/elasticsearch/plugins/migration" ]; then
 fi
 
 echo "Removing Elasticsearch data directory"
-echo "This is required as the data needs to be rebuilt for ES 2.4"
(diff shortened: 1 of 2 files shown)


development-vm/fix-elasticsearch-2.4.sh
```diff
@@ -19,7 +19,7 @@ if [ -d "/usr/share/elasticsearch/plugins/migration" ]; then
 fi
 
 echo "Removing Elasticsearch data directory"
-echo "This is required as the data needs to be rebuilt for ES 2.4"
+echo "This is require as the data needs to be rebuilt for ES 2.4"
```
suzannehamilton: If you're fixing typos, there's an "elastiocsearch" on line 4 that it'd be good to catch. 🙂

dwhenry: didn't do this as I forgot I wanted to merge this PR before I sent the note out and waiting for this to rebuild might have caused problems. I will try and remembner to do this next time I have a puppet PR (later today)

PR #6454, opened 2017-09-13
[DISCUSS] add durable option for published_documents exchange

MatMoore:
https://www.rabbitmq.com/tutorials/amqp-concepts.html

> Exchanges can be durable or transient. Durable exchanges survive broker restart whereas transient exchanges do not (they have to be redeclared when broker comes back online). Not all scenarios and use cases require exchanges to be durable.

All of our queues have the durable bit set, but the exchange doesn't, so if the
cluster went down, we'd have to wait for puppet to run again to re-declare the
exchange.

Merging this PR as is doesn't actually change the exchange to set the durable flag.

As far as I can tell, you can't alter an already declared exchange, only create a new one.

```
vagrant@development:/var/govuk/govuk-puppet$ rabbitmqadmin --username=root --password=root declare exchange name=published_documents  type=topic durable=true
*** 406 PRECONDITION_FAILED - inequivalent arg 'durable' for exchange 'published_documents' in vhost '/': received 'true' but current is 'false'
```

So I'm not sure how to change this on production without interruption.

### DIFF
--- a/modules/govuk/manifests/apps/publishing_api/rabbitmq.pp
+++ b/modules/govuk/manifests/apps/publishing_api/rabbitmq.pp
@@ -43,6 +43,7 @@ class govuk::apps::publishing_api::rabbitmq (
 
   govuk_rabbitmq::exchange { "${amqp_exchange}@/":
     type     => 'topic',
+    durable  => true,
   }
 
   if $configure_test_exchange {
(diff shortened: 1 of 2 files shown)


dwhenry:
I think you would need to do a dance. create a new exchange which is durable, change the publishers to use that exchange, wait for the consumers to drain the old exchange and then update them to read from the new exchange. FUN

MatMoore:
☹️

dwhenry:
@MatMoore are we still planning to do this? or should we close the PR for now?

MatMoore:
Let's close for now. This needs some more thought and the PR isn't useful by itself.

PR #6456, opened 2017-09-13
revert to upstream logspout

philandstuff:
(this is untested; I'm not sure how to test it 😕 )

We had this wrapper docker image because logit had a configuration
issue where they weren't properly presenting their intermediate CA
certificate in the TLS handshake, so the certificate chain wouldn't
validate.  They have now fixed this so we don't need to download their
intermediate cert into our docker image any more.

(I don't fully understand all of these files or what they are consumed
by; there don't seem to be references in puppet code to the files in
modules/govuk_docker/files/* but might they be consumed somewhere
else?)


philandstuff:
I'll check to see in logit that things start coming through with `docker.image:"gliderlabs/logspout:latest"`

PR #6457, opened 2017-09-14
Control filebeat from hiera, not conditional logic

philandstuff:
This PR removes conditional logic for controlling filebeat stuff:

    if $::domain =~ /^.*\.(dev|integration\.publishing\.service)\.gov\.uk/ {
      # blah
    }

and replaces it with a `govuk_beat::enable` hiera value, which is disabled by default.

The idea is that we want to add more filebeat::prospectors but we don't want to add a conditional every time we want to add a prospector.

This should be a pure refactoring -- there shouldn't be any behaviour changes here.

I've copied the filebeat::prospector commit message here too for visibility:

## collect prospectors in the filebeat class

We only want filebeat::prospector instances to have an effect on nodes
which actually have filebeat installed.

Previously, this was achieved in nginx::log by having an `if`
statement conditional on $::domain.  However, I don't really want to
duplicate that conditional logic everywhere we might want to have a
prospector.

This commit makes filebeat::prospector create virtual file resources,
which only get realized in the filebeat class.  This means we can
remove the conditional logic from where we declare prospectors.

(For this, I take advantage of puppet's automatic tagging: the file
created in filebeat::prospector is automatically tagged with
filebeat::prospector).

We use virtual resources elsewhere with things like `@collectd::plugin`,
so there is prior art for this kind of thing.

I wasn't sure whether to make the `file` resource virtual, or the
`filebeat::prospector` resource.


surminus:
This looks good, but wondering whether making the `filebeat::prospector` resource virtual makes more sense when reading the code?

philandstuff:
@surminus yeah I think you're probably right.  Happy to do that

philandstuff:
@surminus updated

PR #6460, opened 2017-09-14
remove logit intermediate cert workaround

philandstuff:
This removes the logit intermediate certificate workaround, which is
no longer needed.

I also added a comment in the Dockerfile to explain how build.sh and
modules.go get consumed because i found that really confusing the
first time around.

Build on docker hub was successful: https://hub.docker.com/r/govuk/logspout-alpine/builds/buhapzmzpvh84j3yz5vttjn/


philandstuff:
I don't think this will actually have an effect on machine which have already downloaded the image

philandstuff:
Tested the image manually on docker-management-1 in integration. it works 🎉

PR #6467, opened 2017-09-14
Add MapIt env var

sihugh:
We want to use this instead of having stuff in alphagov-deployment.  This should match the env variable in alphagov/mapit#42.

The env var name is the same as that in the MapIt yaml configuration ([general.yml](https://github.com/alphagov/mapit/blob/a55bddf61ba2bb25cb185eff69ebc2fe69f54600/conf/general.yml-example#L10) etc) to make it clear what it replaces.

### DIFF
--- a/modules/govuk/manifests/apps/mapit.pp
+++ b/modules/govuk/manifests/apps/mapit.pp
@@ -58,6 +58,9 @@ class govuk::apps::mapit (
   }
 
   govuk::app::envvar {
+    "${title}-DB_PASSWORD":
+        varname => 'MAPIT_DB_PASS',
+        value   => $db_password;
     "${title}-DJANGO_SECRET_KEY":
         varname => 'DJANGO_SECRET_KEY',
         value   => $django_secret_key;


modules/govuk/manifests/apps/mapit.pp
```diff
@@ -58,6 +58,9 @@
   }
 
   govuk::app::envvar {
+    "${title}-DB_PASSWORD":
+        varname => 'DB_PASSWORD',
```
cbaines: This would need to be `MAPIT_DB_PASS` to match up with your mapit change [1], is there something I'm missing?

1: https://github.com/alphagov/mapit/pull/42/commits/995c7d1d4478e450c6a675fc298094436710df15#diff-9ca8db372ff7b61122729b2518307c66R86

sihugh: You're right of course :-). I'll change this one, thanks

cbaines: Looks good 👍

PR #6470, opened 2017-09-14
WIP - Adding File Prospector

suthagarht:


modules/govuk/manifests/app.pp
```diff
@@ -359,6 +373,13 @@
       fields  => {'application' => $title},
     }
 
+    @filebeat::prospector { "${title}-app-err":
+      ensure => $ensure,
+      paths  => "/var/log/${title}/app.err.log",
+      tags   => ['stderr', 'app'],
```
philandstuff: we should pass a `json` parameter here (and elsewhere in this PR) to match the `govuk_logging::logstream` instance

modules/govuk/manifests/apps/publishing_api.pp
```diff
@@ -218,4 +218,9 @@
     json    => true,
   }
 
+  @filebeat::prospector { 'publishing_api_sidekiq_json_log':
+    paths  => '/var/apps/publishing-api/log/sidekiq.json.log',
+    fields => {'application' => 'publishing-api-sidekiq'},
```
philandstuff: in cases like this where json is hardcoded to true, we can definitely just pass `json => {add_error_key => true}` inline to `filebeat::prospector`

modules/govuk/manifests/app.pp
```diff
@@ -334,13 +334,27 @@
     fields  => {'application' => $title},
   }
 
+  @filebeat::prospector { "${title}-upstart-out":
+    ensure => $ensure,
+    paths  => "/var/log/${title}/upstart.out.log",
```
philandstuff: shouldn't this be a list? ie `paths  => ["/var/log/${title}/upstart.out.log"]`?

modules/govuk/manifests/node/s_base.pp
```diff
@@ -105,6 +105,31 @@
...
+      fields => {'application' => 'apt'};
+    'dpkg':
+      paths  => '/var/log/dpkg.log',
+      fields => {'application' => 'dpkg'};
+    'unattended-upgrades':
+      paths  => '/var/log/unattended-upgrades/unattended-upgrades.log',
+      tags   => ['unattended'],
+      fields => {'application' => 'apt'};
+    'unattended-upgrades-shutdown':
+      paths  => '/var/log/unattended-upgrades/unattended-upgrades-shutdown.log',
```
philandstuff: this could be combined with `unattended-upgrades` above by passing both filenames to the `paths` list

modules/govuk_rabbitmq/manifests/logging.pp
```diff
@@ -5,23 +5,48 @@
...
+    fields => {'application' => 'rabbitmq'},
+  }
+
   govuk_logging::logstream { 'rabbitmq_startup_error':
     logfile => '/var/log/rabbitmq/startup_error',
     fields  => {'application' => 'rabbitmq'},
   }
 
+  @filebeat::prospector { 'rabbitmq_startup_error':
+    paths  => '/var/log/rabbitmq/startup_error',
```
philandstuff: similarly I think these could all be combined into a single `filebeat::prospector` by passing all the filenames to a single `paths` list

modules/icinga/manifests/service.pp
```diff
@@ -15,4 +15,11 @@
     require => Service['icinga'],
   }
 
+  @filebeat::prospector { 'icinga_server':
+    ensure => 'present',
+    fields => {'application' => 'icinga'},
+    paths  => '/var/log/icinga/icinga.log',
+    tags   => ['monitoring', 'icinga'],
```
philandstuff: should we replicate the `require => Service['icinga']` here too?

modules/govuk/manifests/app.pp
```diff
@@ -398,18 +427,31 @@
                           value => '@fields.view'}],
     }
 
+    # @filebeat::prospector { "${title}-app-out":
+    #   ensure => $ensure,
+    #   paths  => "/var/log/${title}/app.out.log",
+    #   tags   => ['application'],
+    #   fields => {'application' => $title},
+    # }
```
surminus: Need to remove these commented lines

modules/govuk/manifests/node/s_base.pp
```diff
@@ -103,8 +103,7 @@
     'rkhunter':
       logfile => '/var/log/rkhunter.log',
       fields  => {'application' => 'rkhunter'};
-  }
-
+    }
```
surminus: Why has this changed?

modules/govuk/manifests/app.pp
suthagarht: Hi @surminus , Will do now. Thank you.

surminus:
This is fine, but can we remove "WIP" from the title of the commit? 

It might be nice to mention in the commit message that we are intentionally doing this in parallel with the logstream while we test.

PR #6473, opened 2017-09-14
Add a separate queue for bulk search indexing

MatMoore:
Enabled on development only for the moment.

This will allow us to bulk reindex content without republishing new editions.

Haven't tested end to end, because my elasticsearch is a bit broken, but I've verified using the rabbitmq console that messages published by `rake queue:requeue_document_type[answer]` in publishing api get picked up the rummager consumer.

Corresponding rummager change: https://github.com/alphagov/rummager/pull/924

https://trello.com/c/nPiykYmS/270-get-rummager-reindex-working-in-dev

### DIFF
--- a/modules/govuk/manifests/apps/rummager/rabbitmq.pp
+++ b/modules/govuk/manifests/apps/rummager/rabbitmq.pp
@@ -16,10 +16,14 @@
 # [*enable_govuk_index_listener*]
 #   Whether or not to configure the queue for the rummager govuk indexer
 #
+# [*enable_bulk_reindex_listener*]
+#   Whether or not to configure the queue for the rummager govuk bulk indexer
+#
 class govuk::apps::rummager::rabbitmq (
   $password  = 'rummager',
   $enable_govuk_index_listener = false,
   $enable_publishing_listener = false,
+  $enable_bulk_reindex_listener = false,
 ) {
 
   $amqp_exchange = 'published_documents'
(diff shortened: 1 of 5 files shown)


modules/govuk/manifests/apps/rummager/rabbitmq.pp
```diff
@@ -34,23 +38,32 @@
     amqp_queue    => 'rummager_to_be_indexed',
     routing_key   => '*.links',
     durable       => true,
-  } ->
```
suzannehamilton: Just to check - did you remove the `->` because the resources don't actually need to be created in a specific order?

modules/govuk/manifests/apps/rummager/rabbitmq.pp
```diff
@@ -34,23 +38,32 @@
...
+  }
 
   govuk_rabbitmq::queue_with_binding { 'rummager_govuk_index':
     amqp_exchange => $amqp_exchange,
     amqp_queue    => 'rummager_govuk_index',
     routing_key   => '*.*',
```
suzannehamilton: Will the bulk reindexing messages also be added to the existing queue, and is that OK? (IIRC we don't do any extra filtering by the routing key in rummager.)

modules/govuk/manifests/apps/rummager/rabbitmq.pp
MatMoore: I don't think they do, no. I actually introduced those arrows a couple of days ago when refactoring. The user permissions reference the queues, but they're just regular expressions so the queues shouldn't need to exist to create the users.

modules/govuk/manifests/apps/rummager/rabbitmq.pp
MatMoore: Nope, I deliberately made the routing key of the messages 3 parts to avoid this: eg `answer.bulk.reindex`.

dwhenry: This is something that it might be worth documenting with a comment so that future generations can understand the process.

PR #6479, opened 2017-09-15
Add the Groovy plugin

boffbowsh:
Allow us to write build steps for non-pipeline jobs in Groovy

https://wiki.jenkins.io/display/JENKINS/Groovy+plugin?focusedCommentId=18153476


surminus:
If we're going to start writing more groovy I think it'd be nice to create a seperate repo for our libraries, such as "govuk-jenkinslib", which we can then import into Jenkins rather than have this code sitting in our Puppet repo (more portable and shareable and stuff).

PR #6430, opened 2017-09-11
Remove last of Need API

elenatanasoiu:
Need API has been removed.

For: https://trello.com/c/ClCFMBlG/217-retire-need-api

NB: Depends on https://github.com/alphagov/govuk-puppet/pull/6425 being merged first after which point the current PR will be rebased off master.

PR #6482, opened 2017-09-18
[DO NOT MERGE] Remove govuk delivery

gpeng:
Depends on https://github.com/alphagov/whitehall/pull/3446 being merged to avoid Whitehall errors.

Whitehall is now sending email via the publishing API rabbit queue -> email-alert-service -> email-alert-api and no longer uses `govuk_delivery`.

This PR removes it from our infrastructure.

### DIFF
--- a/development-vm/Procfile
+++ b/development-vm/Procfile
@@ -42,8 +42,7 @@ asset-manager:         govuk_setenv asset-manager         ./run_in.sh ../../asse
 asset-manager-worker:  govuk_setenv asset-manager         ./run_in.sh ../../asset-manager  bundle exec rake jobs:work
 # limelight used port 3040
 # transaction_wrappers used port 3041
-govuk-delivery:        govuk_setenv govuk-delivery        ./run_in.sh ../../govuk-delivery ./startup.sh # govuk-delivery uses port 3042
-govuk-delivery-worker: govuk_setenv govuk-delivery        ./run_in.sh ../../govuk-delivery ./venv/bin/celery worker -A service
+# govuk-delivery used port 3042
 # tariff_demo_api uses port 3043 in integration only
 transition:            govuk_setenv transition            ./run_in.sh ../../transition     bundle exec rails server -p 3044
 # tariff_demo uses port 3045 in integration only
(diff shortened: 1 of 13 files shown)


tuzz:
Looks good to me, but I'll wait for someone with more Puppet expertise to approve. 👍

development-vm/Procfile
```diff
@@ -42,8 +42,6 @@ asset-manager:         govuk_setenv asset-manager         ./run_in.sh ../../asse
 asset-manager-worker:  govuk_setenv asset-manager         ./run_in.sh ../../asset-manager  bundle exec rake jobs:work
 # limelight used port 3040
 # transaction_wrappers used port 3041
-govuk-delivery:        govuk_setenv govuk-delivery        ./run_in.sh ../../govuk-delivery ./startup.sh # govuk-delivery uses port 3042
-govuk-delivery-worker: govuk_setenv govuk-delivery        ./run_in.sh ../../govuk-delivery ./venv/bin/celery worker -A service
```
issyl0: Rather than removing entries completely from this file, previously we've just said `# _app_name_ used port _n_`. I don't remember why this is, but it still seems to be convention (`need_api` was recently removed and has done this).

gpeng: Thanks @issyl0 . Fixed in fe1d8f0be

modules/govuk/manifests/apps/govuk_delivery.pp
```diff
@@ -43,66 +43,17 @@
...
     vhost_ssl_only     => true,
     health_check_path  => '/_status',
     log_format_is_json => true,
   }
-
-  govuk::procfile::worker { 'govuk-delivery':
-    enable_service => $enable_procfile_worker,
-  }
-
-  Govuk::App::Envvar {
```
afda16: I'm not sure if we need to set these resources to `ensure => absent` to make sure they get removed (rather than stop managing the resource by removing the code), and then remove the entire class after applying puppet with the `ensure => absent` setting. That's what was done in https://github.com/alphagov/govuk-puppet/pull/6425/files and https://github.com/alphagov/govuk-puppet/pull/6430/files

Maybe @surminus can help to clarify? or the people involved in removing need-api

gpeng: Thanks @afda16 I've updated this in 716526ee0

PR #6487, opened 2017-09-18
Switch link-checker-api to redis2

elenatanasoiu:
For: https://trello.com/c/7WBKFT7W/231-investigate-support-api-timeouts

Link-checker-api needs more memory in order to perform, as it's currently causing timeouts for support-api every time it runs.

Moving it to its own redis will alleviate the load on the original redis in order to handle support-api requests successfully.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -285,7 +285,7 @@ govuk::apps::hmrc_manuals_api::redis_port: "%{hiera('sidekiq_port')}"
 
 govuk::apps::link_checker_api::db_hostname: "postgresql-primary-1.backend"
 govuk::apps::link_checker_api::db::backend_ip_range: "%{hiera('environment_ip_prefix')}.3.0/24"
-govuk::apps::link_checker_api::redis_host: "%{hiera('sidekiq_host')}"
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/link_checker_api.pp
```diff
@@ -64,7 +64,7 @@
   $sentry_dsn = undef,
   $google_api_key = undef,
   $port = 3208,
-  $redis_host = undef,
+  $redis_host = 'redis-2.backend',
```
sihugh: I think this is going to get overwritten by https://github.com/alphagov/govuk-puppet/blob/21f35fcafe1727c7c67840ffd7297436124dbaff/hieradata/common.yaml#L288.  

I'd change it in the common.yaml anyway as it makes it easier to see what's pointing where.

elenatanasoiu:
Moving to redis-2 from redis-1 won't mean that we lose any data on redis-1 that is already enquequed because:

1. `local-links-manager` is the only application that uses the `link-checker-api`
2. there is a job scheduled in `link-checker-api` which will check for incomplete tasks and finish them off. (https://github.com/alphagov/link-checker-api/issues/57)

In which case moving from redis-1 to redis-2 is safe.

PR #6493, opened 2017-09-18
Mount /var/lib/postgresql on AWS for MapIt machines

issyl0:
- In 7d9cd35, the "mount /var/lib/postgresql' line was made to not be
  executed on AWS (reasons unknown). However, in trying to make Puppet
  run successfully for MapIt machines, we realised we needed this (and
  accompanying volumes and mountpoints in `mapit.yaml`), a) to match the
  existing and b) because the data is mounted into this directory.
  ElasticSearch on AWS works in this way with volumes and data, so
  there's prior art for this approach.

(Tested this by deploying a branch of puppet, and Puppet ran successfully.)

### DIFF
--- a/hieradata_aws/class/mapit.yaml
+++ b/hieradata_aws/class/mapit.yaml
@@ -3,4 +3,15 @@
 govuk::node::s_base::apps:
   - mapit
 
+lv:
+  data:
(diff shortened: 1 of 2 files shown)


issyl0:
Closing this in favour of reverting via #6496.

PR #6496, opened 2017-09-18
Revert "Remove LV and manual mounting in the AWS mapit"

issyl0:
Reverts alphagov/govuk-puppet#6375

- In the commit this reverts, the `mount /var/lib/postgresql ...` line was made to not be
executed on AWS (reasons unknown). However, in trying to make Puppet
run successfully for MapIt machines, we realised we needed this (and
accompanying volumes and mountpoints in mapit.yaml), a) to match the
existing and b) because the data is mounted into this directory.
ElasticSearch on AWS works in this way with volumes and data, so
there's prior art for this approach.

(Tested this revert on the MapIt machines in AWS by deploying a branch of Puppet and it ran successfully.)

### DIFF
--- a/hieradata_aws/class/mapit.yaml
+++ b/hieradata_aws/class/mapit.yaml
@@ -3,4 +3,15 @@
 govuk::node::s_base::apps:
   - mapit
 
+lv:
+  data:
+    pv: '/dev/xvdf'
+    vg: 'postgresql'
+
+mount:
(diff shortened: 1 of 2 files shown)


issyl0:
I did have #6493, then I realised that this commit did exactly the same thing as I was doing, so this revert directly makes history easier to refer to.

PR #6507, opened 2017-09-19
Use common node runtime for performance platform

leelongmore:
The Performance Platform Spotlight app has been upgraded to v6.11.2 so we need to remove the use of node 0.10.37 on the PP frontend and backend boxes.


leelongmore:
Thanks @surminus 👍

PR #6509, opened 2017-09-19
Remove rkhunter

surminus:
rkhunter has a vulnerability where it fetches it's definition updates from sourceforge over http, which means it's ripe for a MITM attack. Reference: http://www.cvedetails.com/cve/CVE-2017-7480/

Remove rkhunter.

/cc @davbo @stephenharker


davbo:
Thanks @surminus this looks good to me but I'll leave it to someone who knows govuk-puppet to review the code changes.

PR #6512, opened 2017-09-19
Allow logstream to only ship statsd metrics

surminus:
We are looking to migrate pushing our logs to The GDS Way (tm) sanctioned SaaS ELK provider. We are accomplishing this using Filebeat (fb624f2fc351e9bffa81b5e76733b80cc538fc2f) rather than Logship.

We use Logship to also ship statsd metrics and timers for applications based upon application log files.

During our migration phase, we do not want to disrupt sending of the metrics, but we do want to turn off any infrastructure related to our current ELK stack.

This commit adds a conditional that can be set globally to override the defined type, and adds a template which only ships statsd metrics.

It also adds tests which rely on setting some additional hiera config in the spec tests.


surminus:
I've tested this in AWS, and it seems to do the right thing 🎉

PR #6516, opened 2017-09-20
Add redis host for email-alert-service

surminus:
We're looking to move values from hardcoded deployment scripts into Puppet and environment variables. This moves the redis host for email-alert-service.

Required: https://github.com/alphagov/email-alert-service/pull/81

https://trello.com/c/fenlgsMp/745-make-icinga-in-aws-green

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_service.pp
+++ b/modules/govuk/manifests/apps/email_alert_service.pp
@@ -33,6 +33,7 @@ class govuk::apps::email_alert_service(
   $rabbitmq_password = 'email_alert_service',
   $sentry_dsn = undef,
   $errbit_api_key = undef,
+  $redis_host = undef,
 ) {
   govuk::app { 'email-alert-service':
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/email_alert_service.pp
```diff
@@ -51,6 +52,10 @@
     password => $rabbitmq_password,
   }
 
+  govuk::app::envvar::redis { 'email-alert-service':
+    host => $redis_host,
```
gpeng: Do we need the port too?

surminus: It gets [set as the default](https://github.com/alphagov/govuk-puppet/blob/260f0efb5b0ab82b4e0b38fd99f20b84972a298b/modules/govuk/manifests/app/envvar/redis.pp#L25) in the upstream, so I didn't think there was required.

PR #6517, opened 2017-09-20
Add more data to rummager deployment dashboard

suzannehamilton:
Add extra graphs:
- Sidekiq queue size and error count
- Request durations (the existing graph does not work for rummager
  because it depends on stats for Rails controllers, and rummager is a
  sinatra app)
- 5xx errors in apps which depend on rummager, such as frontend
- Stats for the Elasticsearch machines, such as memory usage

These graphs have been useful when debugging recent errors caused by the overnight popularity updates.

screen shot 2017-09-20 at 17 25 39
screen shot 2017-09-20 at 17 25 32

https://trello.com/c/saDruTlG/307-investigate-1-day-intermittent-rummager-timeouts

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -903,8 +903,16 @@ grafana::dashboards::deployment_applications:
   rummager:
     # @fields.controller is blank and rummager is a sinatra app
     show_controller_errors: false
+    show_elasticsearch_stats: true
+    show_response_times: true
+    show_sidekiq_graphs: true
     show_slow_requests: false
     has_workers: true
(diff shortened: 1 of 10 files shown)


suzannehamilton:
I've just noticed from the screenshots that there's something wrong with the Sidekiq graphs. I'll take a look at it tomorrow.

suzannehamilton:
I've fixed the Sidekiq graphs. Those panels had the same IDs as some existing panels, and I suspect it was confusing the data rendering.

Here's how that section of the dashboard looks:

screen shot 2017-09-21 at 12 36 15

PR #6521, opened 2017-09-21
Add a second Content Performance Manager Sidekiq process

alecgibson:
Content Performance Manager requires two different queues with their
own concurrency settings. According to [the documentation][sidekiq],
the best way to achieve this is to have two separate queues, each with
their own concurrency, and then have two separate processes that own
those queues.

This change adds a second Sidekiq process.

## Dependencies

- [x] https://github.com/alphagov/content-performance-manager/pull/358

[sidekiq]: https://github.com/mperham/sidekiq/wiki/Advanced-Options#reserved-queues


alecgibson:
[Use different concurrency sidekiq policy for GA and for PublishingAPI](https://trello.com/c/0TgXtx7w/549-use-different-concurrency-sidekiq-policy-for-ga-and-for-publishingapi)

PR #6522, opened 2017-09-21
More govuk delivery removal

gpeng:
We are trying to remove govuk_delivery as it has been retired. The resources have been marked `ensure => 'absent'` within the govuk_delivery app but because it was also removed from the backend class that file was not being executed.

### DIFF
--- a/modules/govuk/manifests/apps/govuk_delivery.pp
+++ b/modules/govuk/manifests/apps/govuk_delivery.pp
@@ -73,20 +73,17 @@ class govuk::apps::govuk_delivery(
     enable_service => $enable_procfile_worker,
   }
 
-  Govuk::App::Envvar {
-    app => $app_name,
+  Govuk::App::Envvar { app => $app_name,
   }
 
   govuk::app::envvar::mongodb_uri { $app_name:
-    ensure   => 'absent',
     hosts    => $mongodb_hosts,
     database => $mongodb_database,
   }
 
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/apps/govuk_delivery.pp
```diff
@@ -78,15 +78,13 @@
   }
 
   govuk::app::envvar::mongodb_uri { $app_name:
-    ensure   => 'absent',
```
afda16: I think we can leave this line, this is only to configure the env.d file

modules/govuk/manifests/apps/govuk_delivery.pp
```diff
@@ -78,15 +78,13 @@
...
     hosts    => $mongodb_hosts,
     database => $mongodb_database,
   }
 
   govuk::app::envvar::redis { $app_name:
-    ensure => 'absent',
```
afda16: I think we can leave this line, this is only to configure the env.d file

modules/govuk/manifests/apps/govuk_delivery.pp
gpeng: CI fails with this in here (and the redis one below) e.g. https://ci.integration.publishing.service.gov.uk/job/govuk-puppet/job/more-govuk-delivery-removal/7/console

PR #6526, opened 2017-09-22
Add second Content Performance Manager Sidekiq process

alecgibson:
A previous commit added a second Sidekiq process to the development VM,
but did not update this Puppet config to also run the new process.

This change runs the new processes, as defined in the corresponding
Content Performance Manager update.

## Dependencies

- [x] https://github.com/alphagov/content-performance-manager/pull/361

### DIFF
--- a/modules/govuk/manifests/apps/content_performance_manager.pp
+++ b/modules/govuk/manifests/apps/content_performance_manager.pp
@@ -103,8 +103,16 @@ class govuk::apps::content_performance_manager(
     app    => $app_name,
   }
 
-  govuk::procfile::worker {$app_name:
+  govuk::procfile::worker { "${app_name}-google-analytics-worker":
     enable_service => $enable_procfile_worker,
+    setenv_as      => $app_name,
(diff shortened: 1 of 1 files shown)


alecgibson:
[Use different concurrency sidekiq policy for GA and for PublishingAPI](https://trello.com/c/0TgXtx7w/549-use-different-concurrency-sidekiq-policy-for-ga-and-for-publishingapi)

modules/govuk/manifests/apps/content_performance_manager.pp
```diff
@@ -103,8 +103,16 @@
     app    => $app_name,
   }
 
-  govuk::procfile::worker {$app_name:
+  govuk::procfile::worker { 'content-performance-manager-google-analytics-worker':
```
andrewgarner: Can you not do `"${app_name}-google-analytics-worker"`?

alecgibson: @andrewgarner done.

PR #6527, opened 2017-09-22
Remove govuk_delivery mongo env

gpeng:
As part of my ongoing struggle to get rid of govuk_delivery. This commit removes the mongodb env. Setting the app itself to absent removes the app's directory from env.d so the mongo bit causes errors when it tries to write to a directory that no longer exists.

Once this is deployed I think we'll be able to remove this file altogether, finally... (fingers crossed).


surminus:
You also need to remove the related hieradata.

gpeng:
Thanks @surminus. I was going to clear all that up in the next one. Just trying to get it to run without errors. I've removed it now in fd6fd9a

PR #6529, opened 2017-09-22
AWS Plek overrides

afda16:
On AWS, draft-content-store, static and draft-static are only available internally.
Override the Plek variable to point to the right domain.

### DIFF
--- a/modules/govuk/manifests/node/s_draft_frontend.pp
+++ b/modules/govuk/manifests/node/s_draft_frontend.pp
@@ -25,4 +25,12 @@ class govuk::node::s_draft_frontend() inherits govuk::node::s_base {
       'PLEK_SERVICE_SEARCH_URI': value => "https://search.${app_domain}";
     }
   }
+
+  if $::aws_migration {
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/node/s_draft_frontend.pp
```diff
@@ -25,4 +25,12 @@
       'PLEK_SERVICE_SEARCH_URI': value => "https://search.${app_domain}";
     }
   }
+
+  if $::aws_migration {
+    $app_domain_internal = hiera('app_domain_internal')
+
+    govuk_envvar {
+      'PLEK_SERVICE_CONTENT_STORE_URI': value => "https://content-store.${app_domain_internal}";
```
surminus: Should this speak to draft-content-store?

afda16: There is a host prefix that gets appended internally

surminus: Ah, yes of course! This is all so clear.

PR #6530, opened 2017-09-22
Fully remove rkhunter

surminus:
I almost removed rkhunter, but some of the monitoring was left hanging around, so remove that, too.


boffbowsh:
I might be misunderstanding hieradata, but I think this is still set as `present`?

PR #6536, opened 2017-09-22
Add rate limit IP addresses to MapIt env vars

sihugh:
The rate limit IPs are the last remaining environment specific configuration left in [MapIt's configuration files][1].  (The other bits are env vars already, or have been superseded by us using Sentry for error reporting)

Once this is deployed and MapIt has been updated to read the config from env vars rather than general.yml, we can move the configuration file into the repo and rid ourselves of the bits in alphagov-deployment.

[1]: https://github.com/alphagov/alphagov-deployment/blob/1ecf707fd59ced975bac2dd9aa8bfb9f3b9809a8/mapit/settings/integration/general.yml"

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -364,6 +364,10 @@ govuk::apps::local_links_manager::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::local_links_manager::redis_port: "%{hiera('sidekiq_port')}"
 
 govuk::apps::mapit::enabled: true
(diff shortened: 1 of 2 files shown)


surminus:
Could I be annoying and ask you to add the changes to `hieradata_aws/common.yaml`, too? 

I probably need to do a diff at some point and find what we've missed.

sihugh:
You sure can, I really meant to do this but got distracted by a squirrel.

sihugh:
Closing as the load balancers don't exist on AWS.  Will have a think

PR #6538, opened 2017-09-22
Remove govuk delivery

gpeng:
Removes the remnants of govuk_delivery which has been retired. Previous commits have been deployed and the app is no longer present on our infrastructure.


tuzz:
⭐️

PR #6540, opened 2017-09-22
Remove Tom Gladhill

whoojemaflip:


modules/users/manifests/tomgladhill.pp
```diff
@@ -1,10 +0,0 @@
-# Create the tomgladhill user
-class users::tomgladhill {
-  govuk_user { 'tomgladhill':
```
cbaines: So, standard puppet comment... I think deleting this will leave some configuration around on the machine. The cleanest way for the infrastructure might actually be to leave this, but change `ensure` to `absent` [1].

1: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_user/manifests/init.pp#L42

PR #6545, opened 2017-09-25
Attempt to make SharePoint stop crawling as frequently

boffbowsh:
We added this bot previously, however
https://support.microsoft.com/en-us/help/3019711/the-sharepoint-server-crawler-ignores-directives-in-robots-txt suggests that the bot
recognises this User-Agent instead.

If this doesn’t work - we should contact Microsoft’s abuse mailbox.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1588,7 +1588,9 @@ router::nginx::robotstxt: |
   Disallow: /
   # Complaints of 429 'Too many requests' seem to be coming from SharePoint servers
   # (https://social.msdn.microsoft.com/Forums/en-US/3ea268ed-58a6-4166-ab40-d3f4fc55fef4)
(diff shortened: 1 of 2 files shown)


hieradata/common.yaml
```diff
@@ -1588,7 +1588,9 @@ router::nginx::robotstxt: |
   Disallow: /
   # Complaints of 429 'Too many requests' seem to be coming from SharePoint servers
   # (https://social.msdn.microsoft.com/Forums/en-US/3ea268ed-58a6-4166-ab40-d3f4fc55fef4)
```
JonathanHallam: Do we still need this comment?

boffbowsh: I think so - it's still that bot, but both links combine to show the actual issue.

PR #6551, opened 2017-09-25
Automated app deployments into AWS

surminus:
We're finding apps are getting out of date because we're not involved in deployments, so an automated deployment mechanism is now required to keep us up to date.

### DIFF
--- a/modules/govuk_jenkins/manifests/job/integration_deploy.pp
+++ b/modules/govuk_jenkins/manifests/job/integration_deploy.pp
@@ -13,6 +13,16 @@
 # [*jenkins_integration_api_password*]
 #   The password used to authenticate with the deployment Jenkins in Integration.
 #
+# [*jenkins_integration_aws_api_user*]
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/manifests/job/integration_deploy.pp
```diff
@@ -13,13 +13,26 @@
...
 # [*puppet_auth_token*]
 #   This token is used to authenticate with the deploy job in Integration.
 #
 class govuk_jenkins::job::integration_deploy (
   $applications = undef,
   $jenkins_integration_api_user = undef,
   $jenkins_integration_api_password = undef,
+  $jenkins_integration_aws_api_user = undef,
+  $jenkins_integration_aws_api_password = undef,
+  $aws_deploy_url = 'deploy.blue.integration.govuk.digital'
```
deanwilson: Missing comma at end of line.

PR #6554, opened 2017-09-26
[DO NOT MERGE] Initial Prometheus install

afda16:
Test initial Prometheus server installation:
- Add Prometheus Puppet module to librarian and update dependencies
(puppet-archive 1.3.0). The original Puppet module doesn't support
Puppet 3 (https://github.com/voxpupuli/puppet-prometheus), so we are
testing the installation with a forked repository.
- Add monitoringtest instance class and exclude in the list of govuk_hosts.
The instance class installs the Prometheus server with a few default options.


afda16:
This was a POC

PR #6559, opened 2017-09-26
Remove vestiges of delayed_job from asset-manager

chrisroos:
We've switched to using Sidekiq instead of DelayedJob in Asset Manager
(see https://github.com/alphagov/asset-manager/pull/232).

The presence of `govuk::delayed_job::worker` meant that the
asset-manager machines were trying to start and monitor the `jobs:work`
Rake task that is no longer available in the code.

### DIFF
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -91,7 +91,6 @@ govuk::node::s_development::apps:
 govuk_app_enable_capistrano_layout: false
 govuk_app_enable_services: false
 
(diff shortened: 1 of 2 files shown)


chrisroos:
This is dependent on us first removing https://github.com/alphagov/govuk-app-deployment/blob/master/asset-manager/config/deploy.rb#L19-L22.

chrisroos:
@chrislo has updated govuk-app-deployment in https://github.com/alphagov/govuk-app-deployment/pull/193.

PR #6560, opened 2017-09-26
Ikenna is no longer with us.

deanwilson:


afda16:
Do we need to remove from `users::usernames:` too?

PR #6561, opened 2017-09-27
Increase Xtrabackup incremental threshold

boffbowsh:
As this job only runs hourly, the freshness threshold is too short and
causes false alarms. The lock is set for 30 minutes, so 90 minutes
seems a sensible threshold.

### DIFF
--- a/modules/govuk_mysql/manifests/xtrabackup/backup.pp
+++ b/modules/govuk_mysql/manifests/xtrabackup/backup.pp
@@ -26,8 +26,8 @@
 # [*base_backup_cron_hour*]
 #   The hour(s) that the daily base backup(s) run(s).
 #
-# [*incremental_backup_cron_minute*]
-#   The minute(s) that the incremental backup(s) run(s).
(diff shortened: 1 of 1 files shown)


surminus:
Hmm, it should run every 15 minutes: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_mysql/manifests/xtrabackup/backup.pp#L47

Maybe it's taking too long to run? Potentially the incremental data towards the end of the data may get too large to run that often, and it also won't run while the base backup is being taken which could trigger the alert. I'd be interested to see when that "freshhold" is being triggered.

boffbowsh:
Ah yes, I missed that bit. I think it is taking too long to run, then. I was seeing it a lot this morning while monitoring alerts during reboots. 

I wonder if we should set the freshhold (:clap: btw) to be the same as the interval, then if there's warnings we actually should investigate?

surminus:
Sure, sounds sensible to me 👍

PR #6563, opened 2017-09-27
Add dependency management docs

rubenarakelyan:
This commit moves "dependency management" from the style guides since it is specific to this repo.

### DIFF
--- a/docs/dependency-management.md
+++ b/docs/dependency-management.md
@@ -1,0 +1,105 @@
+# Dependency management
+
+-   Prefer `require` to `before`.
+-   Do not use `->` and `~>`, use `require`, `notify` or `subscribe`
+    instead.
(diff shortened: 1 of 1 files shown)


docs/dependency-management.md
```diff
@@ -0,0 +1,105 @@
+# Dependency management
+
+-   Prefer `require` to `before`.
+-   Do not use `->` and `~>`, use `require`, `notify` or `subscribe`
+    instead.
+-   *Never* use `<-` or `<~`. (Lint will stop you, anyway.)
+-   Don't break encapsulation. In particular, a resource within one module
+    not create a dependency (require or notify) to a resource deep within
```
gpeng: should there be a 'should' in here or 'must' maybe?

rubenarakelyan: Yup, fixed

PR #6570, opened 2017-09-27
This code isn't used anymore.

deanwilson:
Asset manager no longer uses delayed job worker.


floehopper:
I've renamed this PR to better reflect it's purpose.

PR #6584, opened 2017-09-28
Prevent Sharepoint crawling site

tomnatt:
We have previously slowed the crawl, however it is still hitting our thresholds.
For the last few hours we have been hitting warnings and it has just jumped to
Critical levels.


tomnatt:
For reference: https://github.com/alphagov/govuk-puppet/pull/6545

boffbowsh:
:+1: I'm of the opinion that block them and wait for someone to complain if it's a legit usage.

PR #6587, opened 2017-09-28
GitHub Enterprise is no more, so remove the VPN module and certificates

issyl0:
- All our stuff has moved off it, so we no longer need a VPN to connect
  to it.


hieradata/vagrant_credentials.yaml
```diff
@@ -127,9 +127,6 @@ govuk_ci::master::jenkins_api_token: 'thisisatoken'
 
 govuk_ci::agent::master_ssh_key: 'key'
 
-govuk_ci::vpn::ghe_vpn_username: 'bella'
```
surminus: Now I have to find somewhere else to sneak her into the codebase :(

issyl0:
The `openconnect` package needs to be `ensure => absent` before removing this stuff, hence the WIP while we work out how to do this.

surminus:
Related: https://github.com/gds-operations/puppet-openconnect/pull/18

PR #6588, opened 2017-09-28
Add ASSET_MANAGER_BEARER_TOKEN env var to Whitehall

chrisroos:
In preparation for us being able to save Whitehall assets to Asset
Manager.

I've tested this change in the Dev VM and confirmed that the environment
variable is written as expected.

We'll also need to ask someone to:

1. Add an application token for Asset Manager to the Whitehall API user
in signon.

2. Store the token generated in step 1 in the govuk-secrets repository.

### DIFF
--- a/modules/govuk/manifests/apps/whitehall.pp
+++ b/modules/govuk/manifests/apps/whitehall.pp
@@ -334,8 +334,9 @@ class govuk::apps::whitehall(
           value   => '0';
     }
 
-    # Protocol relative URL so assets in admin are on the same domain but work
-    # in production and development. (This is needed for IE8)
+    # NOTE. The GOVUK_ASSET_ROOT environment variable uses a protocol relative
+    # URL so assets in admin are on the same domain but work in production and
+    # development. (This is needed for IE8)
     govuk::app::envvar {
       "${title}-ASSET_MANAGER_BEARER_TOKEN":
         varname => 'ASSET_MANAGER_BEARER_TOKEN',
--- a/modules/govuk/manifests/apps/whitehall.pp
+++ b/modules/govuk/manifests/apps/whitehall.pp
@@ -22,6 +22,10 @@
 #   Rack limit for how many form parameters it will parse.
 #   Default: undef
 #
+# [*asset_manager_bearer_token*]
+#   The bearer token to use when communicating with Asset Manager.
+#   Default: undef
+#
 # [*basic_auth_credentials*]
 #   Basic auth credentials (necessary for LinkChecker config) used
 #   by some environments.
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/apps/whitehall.pp
```diff
@@ -332,6 +337,9 @@
     # Protocol relative URL so assets in admin are on the same domain but work
     # in production and development. (This is needed for IE8)
     govuk::app::envvar {
+      "${title}-ASSET_MANAGER_BEARER_TOKEN":
+        varname => 'ASSET_MANAGER_BEARER_TOKEN',
+        value   => $asset_manager_bearer_token;
```
floehopper: Minor: It looks as if the comment beginning "Protocol relative URL..." belongs with the definition of `GOVUK_ASSET_ROOT`. By adding the definition of `ASSET_MANAGER_BEARER_TOKEN` here, I think we're making things worse by further separating the comment from its related code.

floehopper: Minor: Did you consider putting the `` definition in the other `govuk::app::envvar` block further down which includes the definition of `PUBLISHING_API_BEARER_TOKEN`? Is this definitely the right place?

modules/govuk/manifests/apps/whitehall.pp
```diff
@@ -91,6 +91,10 @@
 #   The secret used to encode JWT authentication tokens. This value needs to be
 #   shared with authenticating-proxy which decodes the tokens.
 #
+# [*asset_manager_bearer_token*]
+#   The bearer token to use when communicating with Asset Manager.
+#   Default: undef
+#
```
floehopper: Minor: It seems slightly odd not to put this next to the documentation for `publishing_api_bearer_token`. Similarly in the rest of the file, I would've expected to see the two bearer token definitions next to each other unless there's a good reason not to do that.

chrisroos: I've moved this option so that the list of all options is alphabetically ordered as, apart from `sentry_dsn` and `jwt_auth_secret`, this seems to be the convention in the file.

modules/govuk/manifests/apps/whitehall.pp
chrisroos: > It looks as if the comment beginning "Protocol relative URL..." belongs with the definition of `GOVUK_ASSET_ROOT`...

I've added a new commit to make it clear that the comment relates to `GOVUK_ASSET_ROOT`. I've left the `ASSET_MANAGER_BEARER_TOKEN` where it is as the list of environment variables in that block are in alphabetical order.

chrisroos:
Thanks for reviewing, @floehopper. I've addressed your feedback and rebased on master in preparation for merging.

floehopper:
Good work! 👍

PR #6595, opened 2017-10-02
Add Crumb to trigger_data_sync_complete

surminus:


modules/govuk_jenkins/templates/jobs/trigger_data_sync_complete.yaml.erb
```diff
@@ -8,8 +8,11 @@
         numToKeep: 10
     builders:
         - shell: |
+            echo "Getting crumb..."
+            CRUMB=$(curl https://ci_alphagov:$API_KEY@$HOSTNAME/crumbIssuer/api/json | jq '. | .crumb' | sed 's/"//g')
```
cbaines: Adding `--raw-output` to `jq` would probably allow you to get rid of `sed` here.

surminus: Thanks, really useful!

PR #6597, opened 2017-10-02
Default to present.

deanwilson:
Remove the conditional and use the value passed in to the class. The
default is present so there should be no functional change.

### DIFF
--- a/modules/nodejs/manifests/init.pp
+++ b/modules/nodejs/manifests/init.pp
@@ -9,19 +9,13 @@
 #   Default: `undef`
 #
 class nodejs(
-  $version = undef,
+  $version = 'present',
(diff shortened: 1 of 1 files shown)


modules/nodejs/manifests/init.pp
```diff
@@ -9,15 +9,9 @@
 #   Default: `undef`
 #
 class nodejs(
-  $version = undef,
+  $version = 'present',
```
surminus: The code below needs to be changed to `$version` rather than `$ensure`

deanwilson: Nice catch. Fixed and pushed. Functional testing would catch this ;)

PR #6607, opened 2017-10-02
Remove config for errbit

tijmenb:
In https://github.com/alphagov/govuk-puppet/pull/6575 we removed most of the errbit config. That's been deployed so puppet should have now removed the files from the servers. This PR removes the last code.

https://trello.com/c/h7Rt76jU


surminus:
Can we also get rid of anything related to `exception-handler`? It's only purpose is to run errbit.

tijmenb:
@surminus don't we need to do this in order https://docs.publishing.service.gov.uk/manual/remove-machines.html?

surminus:
Oh yeah, that's true. When I'm back in the office we can deactivate, delete and celebrate.

PR #6609, opened 2017-10-02
Update monitoring URL for Logit Kibana

surminus:
Create a new function that processes that unbelievably gross Kibana 5 URL syntax for a query. Seriously, there must be a different way that I can't find documentation on.

Add a conditional that we will only enable in Integration now for testing.

Eventually we can remove the rest of the code related to the older, rather pleasant, syntax.

https://trello.com/c/YrTOpoq7/746-logit-migration

Will require users to check this branch of the documentation while we're testing:

https://github.com/alphagov/govuk-developer-docs/pull/480

### DIFF
--- a/modules/govuk/lib/puppet/parser/functions/kibana5_url.rb
+++ b/modules/govuk/lib/puppet/parser/functions/kibana5_url.rb
@@ -1,0 +1,41 @@
+module Puppet::Parser::Functions
+  newfunction(:kibana5_url, :type => :rvalue, :doc => <<-EOS
+Craft a URL for a Kibana5 search. Positional arguments:
+
+1. BaseURL of your Kibana instance
+2. Search hash: containing `query` and (optionally) `from` period.
+    EOS
+  ) do |args|
+
+    raise(ArgumentError, "kibana5_url: Wrong number of arguments " +
+      "given #{args.size} for 2") unless args.size == 2
+
+    base_url = args[0]
+    search_hash = args[1]
+    logstash_dash_path = "/app/kibana#/discover"
+
+    raise(ArgumentError, "kibana5_url: No 'query' key found in hash") \
+      unless search_hash.has_key?('query')
+
+    # https://github.com/elasticsearch/kibana/issues/1284
+    raise(ArgumentError, "kibana5_url: Query must use single quotes instead of double") \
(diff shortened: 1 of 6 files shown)


modules/varnish/manifests/monitoring.pp
```diff
@@ -29,11 +31,20 @@
...
 
   $app_domain = hiera('app_domain')
-  $kibana_url = "https://kibana.${app_domain}"
   $hostname_prefix = split($::hostname, '-')
-  $kibana_search = {
-    'query' => "@source_host:${hostname_prefix[0]}* AND status:[500 TO 599]",
-    'from'  => '4h',
+
+  if $logit_kibana {
+    $kibana_url = 'https://logit.kibana.io'
```
issyl0: You've got kibana and logit the wrong way around here.

surminus: Good catch!

modules/govuk/lib/puppet/parser/functions/kibana5_url.rb
```diff
@@ -0,0 +1,41 @@
...
+    query_middle = URI.encode("&_a=(columns:!(_source),index:'*-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:'")
+    query_suffix = URI.encode("')),sort:!('@timestamp',desc))")
+
+    base_url.chomp!('/')
+    query_params = search_hash['query']
+    # Convert spaces because not application/x-www-form-urlencoded
+    #query_params.gsub!('+', '%20')
+    query_params.gsub!(' ', '+')
+
+    return "#{base_url}#{logstash_dash_path}?#{query_prefix}#{query_middle}#{query_params}#{query_suffix}"
```
Davidslv: You don't need the return, in ruby the last line will be the returning value 👍

modules/govuk/spec/functions/kibana5_url_spec.rb
```diff
@@ -0,0 +1,46 @@
+require_relative '../../../../spec_helper'
```
Davidslv: You could do `require './spec_helper'` instead of going backwards

surminus:
@Davidslv @issyl0 thanks both, updated!

PR #6610, opened 2017-10-03
Move firewall from its own module to part of base

deanwilson:
Everything should have the firewall configured so it should
be part of the base module.


surminus:
This is failing because of the upstream module. I'm not sure why moving the module into another class has caused it to fail.

deanwilson:
This work has been deprioritised so closing the PR.

PR #6615, opened 2017-10-03
Fix Kibana link

surminus:
I forgot to change the function I was calling. They process queries differently so we must call the correct function within the `icinga::passive_check` defined type.

### DIFF
--- a/modules/varnish/manifests/monitoring.pp
+++ b/modules/varnish/manifests/monitoring.pp
@@ -39,12 +39,14 @@ class varnish::monitoring (
       'query' => "host:${hostname_prefix[0]}* AND @fields.status:[500 TO 599]",
       'from'  => '4h',
     }
+    $action_url = "kibana5_url(${kibana_url}, ${kibana_search})"
     } else {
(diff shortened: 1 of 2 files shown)


modules/varnish/manifests/monitoring.pp
```diff
@@ -39,12 +39,14 @@
...
       'from'  => '4h',
     }
+    $action_url = "kibana5_url(${kibana_url}, ${kibana_search})"
     } else {
     $kibana_url = "https://kibana.${app_domain}"
     $kibana_search = {
       'query' => "@source_host:${hostname_prefix[0]}* AND status:[500 TO 599]",
       'from'  => '4h',
     }
+    $action_url = "kibana5_url(${kibana_url}, ${kibana_search})"
```
issyl0: Did you mean this to be `kibana3_url`, and the LogIt one only to be `kibana5_url`?

surminus: Ugh, yes. Time to go home 😭

PR #6617, opened 2017-10-03
Rework logic for adding internal domains

surminus:
This errored because the `$name` of the defined type was "graphite", which did not match the regex.

However, we pass an alias "graphite.*", so we were trying to add `$internal_aliases` when the variable was never created.

This means we have to add a bunch more logic into the template itself as Puppet is too limited to construct a check against an array.

If we pass `foo.external` as the `$title`, along with `bar.external` as an alias, we should create:

`server_name foo.external foo.internal bar.external bar.internal;`

If we pass `foo` as the `$title`, along with `bar` as an alias, we should create:

`server_name foo bar bar.internal`

However, if we pass `foo` as the `$title`, along with `foo.*` as an alias, we should only set:

`server_name foo foo.*`


surminus:
I tested this in AWS and it seems to do the correct behaviour on both `graphite` and `search` instances which both do different things based upon the aliases.

PR #6627, opened 2017-10-04
[DO NOT MERGE] Upgrade Grafana via Docker

surminus:
GOV.UK have recently undertaken work to remove the old ELK stack that was running Elasticsearch 1.4. As I understand it, running the older version caused problems when trying to upgrade Grafana as we use Elasticsearch as a data source. We're now using Logit, and I wish to test the latest version of Grafana against the Logit backend (to see if this is possible).

I will probably try to test this change in AWS to begin with, and then if successful, move on to our current live environments.

Upgrade Grafana by running it in a container rather than installing as a package.

This is officially documented here:
http://docs.grafana.org/installation/docker/

All configuration options can be passed in as environment variables, which makes it straight forward to pass in any additional arguments in the future.

To upgrade we would need to update the image tag we use, which should make life a little easier (in theory).

I considered moving all the dashboards across and removing the grafana module, but decided it was clearer to leave them where they were. I moved across the grafana.ini and vhost config files to be with the container.


surminus:
Going to close this off for the time-being.

PR #6628, opened 2017-10-05
Inline xvfb

deanwilson:


modules/govuk_testing_tools/manifests/xvfb.pp
```diff
@@ -1,17 +1,20 @@
...
   package { 'xvfb':
-    ensure => 'present',
+    ensure => $ensure,
   }
 
   file { '/etc/init/xvfb.conf':
-    ensure => present,
-    owner  => root,
-    group  => root,
-    source => 'puppet:///modules/xvfb/xvfb.conf',
```
issyl0: Just for my own purposes: why does this no longer work?

deanwilson: Make a note of it and when we move this to test-kitchen we can work through it and look at the failures.

PR #6629, opened 2017-10-05
Redirect URLs containing uppercase characters

tuzz:
The W3C standard for URLs says:

>URLs in general are case-sensitive (with the exception of machine
>names). There may be URLs, or parts of URLs, where case doesn't matter,
>but identifying these may not be easy. Users should always consider
>that URLs are case-sensitive.

Although this is 'generally' true, in GOV.UK's specific case, we don't
really have a user need to treat URLs differently depending on case.
We had a Zendesk ticket come through because we were handling `/world`
taxon pages differently depending on case. We could cater for this at
an application level, but I'm wondering if it's better to handle this
all in one place?

I apprecicate this change might be contentious. Feedback would be
great. I'd also like to see if there's an Nginx module to do this,
but I thought I'd open the discussion as early as possible.

https://trello.com/c/v8557oJ2/188-fix-case-sensitivity-of-world-location-pages

### DIFF
--- a/modules/router/templates/router_include.conf.erb
+++ b/modules/router/templates/router_include.conf.erb
@@ -29,7 +29,7 @@ error_log  /var/log/nginx/lb-error.log;
 # If slug contains no lowercase letters then lowercase it
 # eg www.gov.uk/GOVERNMENT/GUIDANCE -> www.gov.uk/government/guidance
 # eg WWW.GOV.UK/GOVERNMENT/GUIDANCE -> www.gov.uk/government/guidance
-location ~ ^\/[A-Z]+[A-Z\W\d]+$ {
+location ~ [A-Z] {
   rewrite ^(.*)$ $scheme://$host$uri_lowercase permanent;
 }
 


rjw1:
This has certainly been contentious the last 2 or 3 times someone has suggested it.
Also, uploads may not be all lowercased.

rubenarakelyan:
For me, case sensitivity only makes sense when the thing you are requesting is something directly represented on the underlying file system (such as an uploaded file). This would generally have been the case when the W3C document was written but we now have routing and the majority of our paths are virtual.

From a user's point-of-view, I don't see any case in which, for example, www.gov.uk/Education should be any different to www.gov.uk/education. The differences can be either:

1. The first URL returns a 404
2. The first URL returns a different piece of content to the second

In both cases, there is great potential for confusion, especially for users who don't understand the concept of case sensitivity or don't have the mental model that it should apply here.

We already redirect paths that are fully uppercase to their lowercase equivalent so this is an extension of that functionality.

It perhaps also merits mentioning that RFC 2181 states the mailbox part of an email address (the part before the @ sign) is case sensitive, but I have yet to come across a mail server that enforces that rule.

rubenarakelyan:
Also, since this will affect all of GOV.UK, we should probably go down the RFC route to get some comments about the approach.

rubenarakelyan:
https://github.com/alphagov/govuk-rfcs/pull/83

kevindew:
@tuzz is it worth closing this while the RFC runs?

rubenarakelyan:
Closing this until we get the result of the RFC

rubenarakelyan:
We're going to be doing this work in the router instead following the RFC.

PR #6630, opened 2017-10-05
Update query for logit use

surminus:
The only index that needs changing is "@fields.application", so I did wonder if it would be easier to add Logstash configuration that made a duplicate field.

However, in the future we ideally want to move away from using the `@fields` prefix (something I'm still trying to work out to do), so I think it's best to put in a conditional until we've entirely migrated to Logit.

I've tested the datasource and it correctly works, but it does require Logit specifically enabling it for us via a support ticket.

EDIT:

As suspected, I've tested this out on our current Grafana 3 -> ES5.4, and it's not supported. There is an [open issue](https://github.com/grafana/grafana/issues/6238) regarding this as we receive the following error:

```
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "No search type for [count]"
      }
    ],
    "type": "illegal_argument_exception",
    "reason": "No search type for [count]"
  },
  "status": 400
}
```

Which means we probably want to organise the upgrade of Grafana, which I [already have an open PR about](https://github.com/alphagov/govuk-puppet/pull/6627).

/cc @sihugh

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -222,6 +222,7 @@ govuk_sudo::sudo_conf:
 govuk_ppa::path: 'preview'
 
 grafana::dashboards::machine_suffix_metrics: '_integration'
+grafana::dashboards::logit_only: true
 grafana::version: '4.5.2'
 
 hosts::production::ip_api_lb: '10.1.4.254'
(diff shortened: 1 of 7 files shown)


surminus:
I have upgraded Grafana to 4.5.2 in Integration!

PR #6633, opened 2017-10-05
Revert "Remove remnants of delayed job files and service"

floehopper:
The 2 files and the service relating to delayed job seem to have *already* been removed from integration, staging & production. It's not obvious to me how this was achieved and they were definitely there fairly recently, because they were causing exceptions, but now the changes introduced in the reverted commit are causing the following Puppet error & warning:

```
Error: /Stage[main]/Govuk::Apps::Asset_manager/Govuk::Delayed_job::Worker[asset-manager]/Service[asset-manager-delayed-job-worker]: Could not evaluate: Could not find init script or upstart conf file for 'asset-manager-delayed-job-worker'
Notice: /Stage[main]/Govuk::Apps::Asset_manager/Govuk::Delayed_job::Worker[asset-manager]/File[/etc/init/asset-manager-delayed-job-worker.conf]: Dependency Service[asset-manager-delayed-job-worker] has failures: true
Warning: /Stage[main]/Govuk::Apps::Asset_manager/Govuk::Delayed_job::Worker[asset-manager]/File[/etc/init/asset-manager-delayed-job-worker.conf]: Skipping because of failed dependencies
```

Reverting the commit should fix these, because Puppet will no longer try to stop the non-existent service.

This reverts commit 681ebc1b26a78b0b548ba972bed2f74e47d30641.


floehopper:
@thomasleese:

> I think this happened because puppet was deployed a couple of hours ago.

Ah, OK. That makes sense. I guess my Puppet changes weren't resilient against running multiple times. Sorry!

floehopper:
@thomasleese has confirmed that this has fixed the problem (at least in integration).

PR #6634, opened 2017-10-05
[WIP] Initial Puppetserver 5 provisioning

afda16:
- New puppetserver node class to install puppetserver from containers
- Disable hiera-eyaml-gpg check in site.pp, which is not supported in Puppet5
- Disable puppet and unused_kernel classes in base class, that need to be updated
for Puppet5 clients
- Remove ruby-shadow gem from Hieradata, as the Gem provider didn't work on Puppet5
- Restore default logging driver on AWS containers (json-file), as syslog is not
logging messages.
- Add elastic-beat apt repo key that was missing and failing on Puppet 5
- Add container classes and test configuration for puppetserver, puppetdb and
puppetdb-postgres

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -110,9 +110,10 @@ apt::sources:
 
 backup::repo::apt_mirror_hostname: "%{hiera('apt_mirror_hostname')}"
 
-base::packages::gems:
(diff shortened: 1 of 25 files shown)


deanwilson:
Closing but not deleting as we might need this.

PR #6647, opened 2017-10-06
[WIP] Add Terraboard

surminus:
Very much a work in progress. Start of the manifest to add terraboard.

Need to add a bunch of enviornment variables, and also have a postgres
database somewhere (potentially running as a container).


surminus:
Requires authentication so becomes more complicated, removing and will look back at doing this later.

PR #6649, opened 2017-10-09
Add mtail class to Nginx manifest

afda16:
Add Mtail to Nginx manifest to monitor `/var/log/nginx/*-access.log`
log files. Metrics are sent to Statsd running on localhost.


hieradata/common.yaml
```diff
@@ -781,6 +781,14 @@ govuk_jenkins::config::user_permissions:
 
 govuk::node::s_logging::apt_mirror_hostname: "%{hiera('apt_mirror_hostname')}"
 
+govuk_mtail::enabled: true
+govuk_mtail::extra_args: "-log_dir /var/log/mtail -graphite_prefix %{::fqdn_metrics}."
+govuk_mtail::graphite_host_port: 'graphite:2003'
```
surminus: is this correct for Carrenza? should it be `graphite.cluster`?

hieradata/common.yaml
```diff
@@ -781,6 +781,14 @@ govuk_jenkins::config::user_permissions:
...
+govuk_mtail::logs:
+  - '/var/log/nginx/*-access.log'
+govuk_mtail::manage_repo_class: true
```
surminus: if the default is `true` in common.yaml, make it true by default in the manifest?

modules/govuk_mtail/manifests/init.pp
```diff
@@ -0,0 +1,76 @@
...
+#   Host:port to statsd server to write metrics to.
+#
+# [*metric_push_interval*]
+#   Interval between metric pushes, in seconds (default 60)
+#
+# [*extra_args*]
+#   Mtail program extra arguments (default -log_dir /var/log/mtail)
+#
+class govuk_mtail(
+  $logs = [],
```
surminus: we could add a `validate_array` function for this

modules/govuk_mtail/manifests/init.pp
```diff
@@ -0,0 +1,76 @@
...
+
+  if $manage_repo_class {
+    $repo_class = 'govuk_mtail::repo'
+  } else {
+    $repo_class = 'govuk_ppa'
+  }
+
+  include $repo_class
+
+  $logs_ = hiera_array('govuk_mtail::logs', $logs)
```
surminus: Why do we do this, rather than just passing in the `$logs` parameter?

afda16: To make sure we do a recursive merge

surminus:
I think in the future this would be a nice module to be able to break out when we can do tests for later versions of Puppet so that other people can use it.

hieradata/common.yaml
afda16: Updated

hieradata/common.yaml
afda16: I made it true directly when we implement the 'mtail' class

modules/govuk_mtail/manifests/init.pp
afda16: I removed the parameter because we are doing a lookup with the `hiera_array` function. If we don't want to do a recursive merge, we should remove that function and put the variable back here

surminus: I think it's clearer to understand if we use a parameter rather than `hiera_array` lookup as it makes it more portable as a class. I think we should also do checks incase it's an empty value, and avoid adding mtail if there are no logs to parse.

afda16: I thought in doing a merge so each application can declare an individual log in its Hieradata file without having to use a common variable in Hieradata for the entire infrastructure. The `hiera_array` function is already passing a default so it'll never pass an empty variable to the mtail class.

What you suggest is in line with the way in which we are doing things right now so I'll change this code, but maybe when we start moving to version 5 we could consider other options?

PR #6650, opened 2017-10-09
Add a fields_prefix in deployment dashboards

surminus:
Previously when a dashboard did not work because of a missing `@fields` prefix, we removed the dashboard entirely (53510b39199c2fbc99c9d35a6c5d257fd592fecf).

We have narrowed down the change in behaviour for log outputs to the different versions of [logstasher](https://github.com/shadabahmed/logstasher).

It seems that anything below v1 adds a `@fields` hash which contains data such as request, duration etc. Anything above v1 does not contain this hash, and outputs this information as a top level object in the JSON output.

This means that it gets parsed in this manner straight into Kibana, and the queries we set on the deployment dashboards do not work.

In the future we would like to entirely remove the `@fields` prefix, so we are adding a new parameter in the dashboard defined type which can
set to help migrate to newer versions of logstasher (or at least a unified logging format across all apps).

It is difficult to remove `@fields` using logstash config due to the nested hash within the JSON object, and I would prefer not to be bandaid
it in this way to encourage movement to a consistent logging approach.

It's worth noting that the `@fields.application` is *not* affected by this, although in new world logging (ie using Filebeat), this is logged out as plain "application".

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -826,10 +826,11 @@ grafana::dashboards::machine_suffix_metrics: ''
 grafana::dashboards::deployment_applications:
   asset-manager: {}
   calculators:
-    # Status, duration, controller missing @fields. prefix
-    show_controller_errors: false
-    show_slow_requests: false
-  calendars: {}
+    # logstasher >1.x
+    fields_prefix: ''
+  calendars:
+    # logstasher >1.x
+    fields_prefix: ''
   collections: {}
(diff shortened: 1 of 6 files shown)


surminus:
There are only a handful of apps on the newer logstasher, so I think for the time being it's better to specify the edge cases, then when stuff gets updated the pendulum can swing the other way and we can taunt any app using an old version of logstasher 😄

PR #6652, opened 2017-10-11
Remove bob walker's access to GOV.UK

rjw1:
bob has stopped being on call for GOV.UK so no longer needs access to the
machines that GOV.UK runs on.


surminus:
I'm going to let @rjw1 press the big green button

PR #6654, opened 2017-10-11
Add an additional disk to asset-master and asset-slave

tomnatt:
Both mounts are close to full and need additional capacity. This change
triggered by alerts in production environment.


hieradata/node/asset-slave-1.backend.staging.publishing.service.gov.uk.yaml
```diff
@@ -4,6 +4,8 @@ lv:
     pv:
       - '/dev/sdc1'
       - '/dev/sdd1'
+      - '/dev/sde1'
+      - '/dev/sdf1'
```
surminus: seem to be adding two disks here

tomnatt: Future-proofing?

Removed.

PR #6658, opened 2017-10-11
Remove fields prefix from nginx config

surminus:
This goes by the way we do our naming as defined in [The GDS Way](https://gds-way.cloudapps.digital/manuals/logging.html#naming-conventions).

### To do

 - [x] Update the rest of the fields to match the naming convention
 - [x] Check documentation

### DIFF
--- a/modules/nginx/templates/etc/nginx/nginx.conf.erb
+++ b/modules/nginx/templates/etc/nginx/nginx.conf.erb
@@ -38,7 +38,6 @@ http {
       '$ssl_protocol $ssl_cipher';
 
     log_format json_event '{ "@timestamp": "$time_iso8601", '
-                         '"@fields": { '
                          '"remote_addr": "$remote_addr", '
                          '"remote_user": "$remote_user", '
                          '"body_bytes_sent": $body_bytes_sent, '
@@ -64,7 +63,7 @@ http {
                          '"govuk_abtest_educationnavigation": "$http_govuk_abtest_educationnavigation", '
                          '"varnish_id": "$http_x_varnish", '
                          '"ssl_protocol": "$ssl_protocol", '
-                         '"ssl_cipher": "$ssl_cipher" } }';
+                         '"ssl_cipher": "$ssl_cipher" }';
 
     access_log  /var/log/nginx/access.log timed_combined;
     access_log  /var/log/nginx/json.event.access.log json_event;


philandstuff:
this PR is a prime candidate for the [`?w=1` query string hack](https://github.com/alphagov/govuk-puppet/pull/6658/files?w=1)

modules/nginx/templates/etc/nginx/nginx.conf.erb
```diff
@@ -38,33 +38,35 @@ http {
...
-                         '"http_user_agent": "$http_user_agent", '
-                         '"govuk_request_id": "$http_govuk_request_id", '
-                         '"govuk_original_url": "$http_govuk_original_url", '
-                         '"govuk_dependency_resolution_source_content_id": "$http_govuk_dependency_resolution_source_content_id", '
-                         '"govuk_abtest_educationnavigation": "$http_govuk_abtest_educationnavigation", '
-                         '"govuk_asset_manager_file_store": "$sent_http_govuk_asset_manager_file_store", '
-                         '"varnish_id": "$http_x_varnish", '
-                         '"ssl_protocol": "$ssl_protocol", '
-                         '"ssl_cipher": "$ssl_cipher" } }';
+                         '"access": {',
```
philandstuff: oops, that trailing comma shouldn't be there

philandstuff:
amended my commit, hopefully it's not broken now

surminus:
I have rebased this. I had a look through Puppet code and found a few things in documentation to [raise a PR on](https://github.com/alphagov/govuk-developer-docs/pull/657). Other I don't believe there is much else to change.

boffbowsh:
I think there's a few dashboards-as-code that would be affected by this:
https://github.com/alphagov/govuk-puppet/blob/659cc49111c3ff61f8ee9baab5ac9f41434a7c1a/modules/grafana/templates/dashboards/publishing_api_overview.json.erb https://github.com/alphagov/govuk-puppet/tree/19ee89d6c9ec430e70eab19d8b57c137f9c522af/modules/grafana/files/dashboards

surminus:
I can't work if these would be affected or not, because mostly they are linked to the logs which are sent by the application rather than NGINX, which is why it has that `@fields_prefix` conditional all over the place.

sihugh:
Does [this mean](https://gds-way.cloudapps.digital/manuals/logging.html#http-fields) we should be moving our app logging towards the `access.<foo>` model?  Our use of govuk_app_config seems to have us working towards having no prefix for `status` etc.  My cheese keeps getting moved! 🧀

surminus:
I'm not really sure anymore. If you check my first commit on this PR I removed that prefix, then it was followed up with best practice that added a prefix back in. I think the idea is that these relate to access logs: https://gds-way.cloudapps.digital/manuals/logging.html#http-fields

In this sense I suspect that having application logs without that prefix makes sense.

sihugh:
I'd rather stuff was consistent, but I guess that's an argument for another day.  Fortunately we're getting to the point where we could bump govuk_app_config and let dependabot sort it all out.

MatMoore:
@sihugh is there anything stopping us from updating `govuk_app_config` now to be consistent with this PR? I'd like to move towards the standard format as soon as possible, as the inconsistency makes kibana so much harder to use.

I'm happy to raise PRs if this is helpful...

sihugh:
@MatMoore I've had a quick look at [logstasher](https://github.com/shadabahmed/logstasher) and I'm not sure it's easy for us to do this in the apps.  There was a [PR merged to remove the `@fields` prefix](https://github.com/shadabahmed/logstasher/pull/64) in logstasher a while back.  Effectively we'd be wanting to put this back in (but call it `access`), and rename a few fields.  

If there's an easy way I'm all for it 👍

surminus:
Filebeat may be able to do this. Have a look at the prospector values. 

The easiest change to begin with would be entirely remove the prefix IMO, and "head towards the recommended guidelines in the future".

MatMoore:
I really want to emphasise the value of keeping fields consistent within a stack.

A single request to www.gov.uk goes through all these things before something starts rendering a page:

- the CDN
- cache-*.router nginx
- cache-*.router varnish
- cache-*.router router
- frontend-lb-*.frontend nginx
- frontend-*.frontend nginx
- frontend-*.frontend application

This alone is really confusing for anyone who isn't really familiar with the architecture of GOV.UK.

Each of these has one or more log files that are shipped to logit. To be able to use kibana I need to be able to easily isolate the part of stack I'm interested in, so that I can look at single events instead of the same events represented in 20 different ways.

*If I can trust field names to be consistent*, I should be able to change a `tags`/`application`/`host` filter and the same query will work in a different context. If I'm writing a query from scratch, I should just be able to pick any one of those things and be able to use filters like `status`, `path` etc. and not be met with instant failure.

If there are two competing conventions (as there are now, with or without this PR) changing the filter suddenly breaks my query. Queries that work fine in one context don't work in another context, and I lose trust in kibana as a source of information.

sihugh:
Should we then revert this PR to be just @surminus 's original commit as it helps us converge again?

MatMoore:
@sihugh I think that makes more sense for now, unless we can figure out how to change the application ones.

I had a quick scan through the filebeat docs but all I found was the `fields_under_root` flag. I didn't see a way to introduce a namespace, but I may be missing something as I don't know much about how it works.

surminus:
Let me know if you want me to remove the 2nd commit from this PR.

surminus:
@MatMoore @sihugh I've removed the 2nd commit, so if you're happy merge when ready.

I've pushed the 2nd commit into it's own branch (`remove-fields-prefix-from-nginx-config`) so GOV.UK as a programme can revisit this later if they wish to.

PR #6672, opened 2017-10-16
[WIP] Enable logspout everywhere but dev

surminus:
I didn't realise this was gated and not enabled. Since we're using Logit everywhere now, we should ship all Docker logs across by default.


surminus:
We wish to look at other options than using logspout. For example, using the syslog driver and appropriately configuring rsyslog to output logs to a directory for each container.

surminus:
Closing this for now.

PR #6673, opened 2017-10-16
Show draft flows env variable for smart answers

kevindew:
This is to move smart answers towards being a 12 factor application.
Currently this configuration is made by the deployment of different
files at deploy time. This commit changes this to be be defined by an
environment variable.

### DIFF
--- a/modules/govuk/manifests/apps/smartanswers.pp
+++ b/modules/govuk/manifests/apps/smartanswers.pp
@@ -13,6 +13,11 @@
 #   available alongside other formats (e.g. html, json)
 #   Default: false
 #
+# [*show_draft_flows*]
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/apps/smartanswers.pp
```diff
@@ -51,6 +57,14 @@
     }
   }
 
+  if $flow_show_drafts {
```
thomasleese: Should this be `if $show_drafts_flows`?

kevindew: Yes - whoops. Thanks!

kevindew:
Thanks @thomasleese 👍

PR #6675, opened 2017-10-16
Remove apps that can't deploy

kevindew:
With the retiring of [alphagov-deployment][] these apps will no longer
be safe to deploy as they won't have files automatically copied into
them. Therefore to avoid accidental incorrect deployments they are
removed from the list of deployable applications.

If someone was in a bind it'd still be easy to deploy an app as you'd
just have to edit the jenkins job to add the application to the list.

[alphagov-deployment]: https://github.com/alphagov/alphagov-deployment

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -10,10 +10,6 @@ app_domain_internal: 'dev.gov.uk'
 deployable_applications: &deployable_applications
   asset-manager: {}
   authenticating-proxy: {}
-  backdrop-read:
-    repository: 'backdrop'
-  backdrop-write:
-    repository: 'backdrop'
   bouncer: {}
   calculators: {}
   calendars: {}
@@ -48,7 +44,6 @@ deployable_applications: &deployable_applications
   manuals-publisher: {}
   mapit: {}
   maslow: {}
-  performanceplatform-admin: {}
   performanceplatform-big-screen-view: {}
   policy-publisher: {}
   publisher: {}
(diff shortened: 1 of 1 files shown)


tijmenb:
Maybe I'm confused, but I think still need to deploy these apps because not all of them are moving to the PaaS?

kevindew:
@tijmenb @tomnatt did some digging today and concluded that even those not on PAAS shouldn't have more deployments. I've pinged @tomnatt to sanity check this first but current plan is to retire alphagov-deployment tomorrow so I don't want it to be easy to do bad deploys.

tomnatt:
Performance Platform, Spotlight and Stagecraft are all migrated. Backdrop is not, but they have no intention of deploying again. As long as we have an idea how to make this happen in the event of disaster, we should be ok.

tijmenb:
> As long as we have an idea how to make this happen in the event of disaster, we should be ok.

If alphagov-deployment is deleted, that would be quite hard. I'm not sure how to?

tomnatt:
I'd suggest something ugly using manual intervention. 

1. Deploy
2. Put the required files in place by hand

This is only in the case of disaster and is only something we need for the next few months (backdrop should be migrated by Christmas). We don't need to delete the alphagov-deployment, only remove it from the automated pipeline. 

Given this, it might be sensible to leave backdrop in the list of applications which can be deployed. We know it will be an incomplete deploy and can fix by hand if needs be.

Horrible, but simple I think?

kevindew:
I'm going to close this now so this can be pondered over a bit more.

PR #6682, opened 2017-10-19
Add streaming RDS backup scripts

surminus:
Add backup scripts that do not do anything too clever: take a list of databases, excluding some specific ones, take a dump, and stream the output to a file in an S3 bucket. All backups will go into the same bucket, which we'll enable with expiration dates to manage storage.

Ensure there are passive checks that alert us if a backup fails.

In theory these scripts could also be run on demand if developers wanted more up to date data to import.

We are relying on the connection during transfer being encrypted (which is by default with gof3r), and the bucket being encrypted at rest.

We will use IAM roles to allow access to the bucket from the instance.

https://trello.com/c/F6UQeeiN/777-design-rds-replicas-and-backups

### DIFF
--- a/modules/govuk/manifests/node/s_db_admin.pp
+++ b/modules/govuk/manifests/node/s_db_admin.pp
@@ -14,15 +14,37 @@
 #  The database user to connect to the remote database as
 #
 class govuk::node::s_db_admin(
-  $mysql_db_host     = undef,
-  $mysql_db_password = undef,
-  $mysql_db_user     = undef,
-  $postgres_host     = undef,
-  $postgres_user     = undef,
-  $postgres_password = undef,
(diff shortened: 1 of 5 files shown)


modules/govuk/manifests/node/s_db_admin.pp
```diff
@@ -14,16 +14,38 @@
...
+    architecture => $::architecture,
+    key          => '3803E444EB0235822AA36A66EC5FE1A937E3ACBB',
+  }
+
+  package { 'gof3r':
+    ensure  => '0.5.0',
+    require => Apt::Source['gof3r'],
+  }
+
+  $alert_hostname = 'monitoring'
```
afda16: I think we made the hostname `alert`

PR #6689, opened 2017-10-24
Fix hard-coded proxy pass for Asset Manager

chrislo:
9b3b712714 attempted to change the configuration so that existing
headers were only added to a subset of requests proxied to the Rails
app. The `proxy_pass` URL was, however, hard-coded to
`http://asset-manager.dev.gov.uk-proxy`. This is not correct for all
environments.

This commit uses `@app_domain` to make sure it is correctly set.


chrislo:
Thanks @suzannehamilton. Merging to master.

floehopper:
This was deployed to staging at ~12:45 today. 30 mins later I ran some checks that Asset Manager was still serving requests OK and @suzannehamilton confirmed that the syntax error in the Nginx config had gone away and Nginx could be restarted OK.

It was then deployed to production at 14:44 today. 30 mins later @suzannehamilton & I ran the same checks and were happy that all was well.

/cc @chrislo, @chrisroos

PR #6692, opened 2017-10-24
MongoDB S3 backups in AWS

surminus:
https://trello.com/c/FQYnkupU/779-design-backup-method-to-s3-for-mongo

We already have scripts and manifests that create backups for MongoDB to an S3 bucket. However, these scripts are heavily based on a few assumptions:

 - We want to encrypt using GPG
 - We are not in an AWS environment
 - Our hosts are static

Rather than try to work around all the code that does the above, I created a new class and template which excludes anything to do with GPG and AWS API settings.

Rather than also having to set a load of hieradata to specifically include the class of a specific node, I just include it on all nodes, but only run it if the instance is the primary instance.

We are unable to specify based upon hostname in AWS, and this also ensures that if primary goes away in our cluster, when the new primary steps up, backups continue to happen.

This means that a Nagios passive check is included on all machines, but they should always report back as "OK" if they're not a primary so should not cause any alarms.

### DIFF
--- a/modules/mongodb/templates/mongodump-to-s3.erb
+++ b/modules/mongodb/templates/mongodump-to-s3.erb
@@ -1,0 +1,58 @@
+#!/bin/bash
+#
+# This script takes a mongodump, copies it to an S3 bucket,
+# and cleans up afterward.
+#
+# At a set time it will push the backup to a different location
+# so we can set shorter expiration on the regular backups, and
+# keep one backup everyday for longer.
+#
+set -e
+
+exec 1> >(/usr/bin/logger -s -t $(basename $0)) 2>&1
+
+NAGIOS_CODE=2
+NAGIOS_MESSAGE='CRITICAL: MongoDB backup to S3 failed'
+
+# Triggered whenever this script exits, successful or otherwise. The values
+# # of CODE/MESSAGE will be taken from that point in time.
+function nagios_passive () {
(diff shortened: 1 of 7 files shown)


modules/mongodb/templates/mongodump-to-s3.erb
```diff
@@ -0,0 +1,58 @@
...
+if [[ "$(date +%H:%M)" == "<%= @daily_time -%>" ]]; then
+  BUCKET_KEY='mongodb/daily/<%= @aws_migration -%>'
+else
+  BUCKET_KEY='mongodb/regular/<%= @aws_migration -%>'
+fi
+
+# Only run on the current primary
+if /usr/bin/mongo --quiet --eval "db.isMaster().primary === db.isMaster().me" | grep -q true; then
+  echo "Starting backup"
+  /usr/bin/mongodump -o <%= @backup_dir -%>/mongodump
```
timmow: Not a problem, but was there a reason for writing to disk rather than compressing on the fly?

surminus: I wasn't entirely sure if the mongodump function allowed compression on the fly. In later versions of Mongo it is possible to set the output to STDOUT which allow more flexibility, but wasn't sure if it was possible in the version of Mongo we (mostly) run (2.4).

PR #6683, opened 2017-10-23
Only configure Sendfile for public asset requests from Asset Manager

floehopper:
Previously we were setting the `X-Sendfile-Type` & `X-Accel-Mapping` request headers in *all* requests proxied from Nginx to the Rails app. However, these headers only need to be set for the `/media/*` (Mainstream) and `/government/uploads/*` (Whitehall) URL paths, because these are the only routes which ever make use of the Sendfile behaviour.

In the not too distant future, we're planning to change the Rails app to proxy all these public asset requests to S3, like public Mainstream asset requests already are in production. At this point, it will be possible to remove both this new Nginx `location` block and the related internal `location` block.

My initial motivation for doing this was to solve a problem with alphagov/asset-manager#267 in which the Rails app proxies public asset requests to fake S3 URLs served by the Rails app. Having the Sendfile-related headers set for the latter requests was problematic for serving the fake S3 URLs.

Note that I have introduced a bit of duplication in the new Nginx `location` block (in the `proxy_pass` statement). I did try referencing the existing named `location` block `@app`, but I couldn't get this to work. However, given that the amount of duplication is minimal, there is already duplication like this in the Nginx config, and the `location` block should be going away in the not too distant future, I'm happy to leave it like this.

PR #6693, opened 2017-10-24
[Do Not Merge] Fail fast if Nginx config contains a syntax error

floehopper:
Previously if the Nginx config changed (e.g. in an individual virtual host config), but it contained a syntax error, the following would occur when running `govuk_puppet`:

1. Puppet notices the config has changed and triggers a restart of the Nginx service
2. Puppet runs `/etc/init.d/nginx reload` to restart the Nginx service
3. The syntax error prevents the new config from being loaded
4. Nginx continues to run with the old config
5. `/etc/init.d/nginx reload` returns a zero (success) error code (it seems to do this whether or not the reload succeeds!)
6. Puppet assumes the restart has completed successfully and continues without any warnings or errors
7. If the Nginx process is subsequently restarted (e.g. on reboot), it will stop and then fail to start, because of the syntax error

After this change, Puppet's restart of the Nginx service first checks that there are no syntax errors in the config and fails fast with a useful error message if there is a syntax error. Otherwise it continues as before.

Ideally we'd also roll back the change to the Nginx configuration, but I'm not sure how to do that. However, I still think that this change is an improvement, because (assuming we have alerts setup for the Puppet cron job failing), we'll find out immediately if there is a syntax error in the Nginx config.

The motivation behind this change was [a real case where a syntax error in the Nginx config was accidentally introduced][1] and Nginx silently failed to reload the config. Later when the (integration) machine was rebooted, the OS attempted to start Nginx, but it failed to start, meaning that all requests to any Nginx virtual hosts failed.

I'm no Puppet expert and I'm unsure whether this is the best approach, but it feels as if there's something no quite right about the current behaviour.

Here's some sample output when a syntax error occurs with the change from this PR in place:

```
Notice: /Stage[main]/Govuk::Apps::Asset_manager/Govuk::App[asset-manager]/Govuk::App::Config[asset-manager]/Govuk::App::Nginx_vhost[asset-manager]/Nginx::Config::Vhost::Proxy[asset-manager.dev.gov.uk]/Nginx::Config::Site[asset-manager.dev.gov.uk]/File[/etc/nginx/sites-available/asset-manager.dev.gov.uk]/content: content changed '{md5}b1d58c15d80d37cd43de5e57a0b4ff49' to '{md5}6d3942333b7e7d4a57b1a70b1d9f4bc9'
Error: /Stage[main]/Nginx::Service/Service[nginx]: Failed to call refresh: Could not restart Service[nginx]: Execution of 'nginx -t && /etc/init.d/nginx reload' returned 1: nginx: [emerg] host not found in upstream "asset-manager.yyydev.gov.uk-proxy" in /etc/nginx/sites-enabled/asset-manager.dev.gov.uk:56
nginx: configuration file /etc/nginx/nginx.conf test failed
Error: /Stage[main]/Nginx::Service/Service[nginx]: Could not restart Service[nginx]: Execution of 'nginx -t && /etc/init.d/nginx reload' returned 1: nginx: [emerg] host not found in upstream "asset-manager.yyydev.gov.uk-proxy" in /etc/nginx/sites-enabled/asset-manager.dev.gov.uk:56
nginx: configuration file /etc/nginx/nginx.conf test failed
Notice: Finished catalog run in 56.68 seconds
```

[1]: https://github.com/alphagov/govuk-puppet/pull/6683

### DIFF
--- a/modules/govuk/manifests/apps/asset_manager.pp
+++ b/modules/govuk/manifests/apps/asset_manager.pp
@@ -100,7 +100,7 @@ class govuk::apps::asset_manager(
          proxy_set_header X-Sendfile-Type X-Accel-Redirect;
          proxy_set_header X-Accel-Mapping /var/apps/asset-manager/uploads/assets/=/raw/;
 
-         proxy_pass http://asset-manager.<%= @app_domain %>-proxy;
+         proxy_pass http://asset-manager.intentionally-broken-<%= @app_domain %>-proxy;
        }
 
       # /raw/(.*) is the path mapping sent from the rails application to
--- a/modules/nginx/manifests/service.pp
+++ b/modules/nginx/manifests/service.pp
@@ -4,7 +4,7 @@ class nginx::service {
   service { 'nginx':
     ensure    => running,
     hasstatus => true,
-    restart   => '/etc/init.d/nginx reload',
+    restart   => 'nginx -t && /etc/init.d/nginx reload',
   }
 
 }


floehopper:
@suzannehamilton I hope you don't mind me asking you to review this - I thought it would make sense given that you experienced the problem! However, feel free to suggest someone else if you think it would be useful.

chrisroos:
This looks good to me, @floehopper!

It might be worth mentioning in the commit note that this is similar to the config check that happens in the `nginx::restart` class: https://github.com/alphagov/govuk-puppet/blob/d15b19c76fabffdf59c1b7f87c490de4c4ba5f71/modules/nginx/manifests/restart.pp#L6

As a slight aside, I don't think the `onlyif` command in `nginx::restart` will work as expected. It seems that, at least on my dev-vm, `/etc/init.d/nginx configtest` always exits with a code of 0 irrespective of whether the nginx config is valid. I'm going to open another PR to explain this is more detail.

I've checked this change in my dev-vm and am happy that it does what we want. I introduced a syntax error in my nginx config, confirmed that I saw "Error" messages when running `govuk_puppet` and confirmed that the nginx workers had not been restarted. I then removed the syntax error from the nginx config, confirmed that there were no "Error" messages when running `govuk_puppet` and that the nginx workers had been restarted.

I think it'd be good to test our assumption about there being monitoring of puppet errors. How about, after merging this branch, creating a new branch that introduces a syntax error in the nginx config, deploying that branch to integration and asking in 2ndline whether they see any alerts when `govuk_puppet` runs?

chrisroos:
I've opened PR #6696 to use `nginx -t` when checking whether nginx can be restarted.

floehopper:
@chrisroos:

Thanks for reviewing.

> It might be worth mentioning in the commit note that this is similar to the config check that happens in the `nginx::restart` class

I'll do that now and force-push the branch.

> As a slight aside, I don't think the `onlyif` command in `nginx::restart` will work as expected

Oh, yes. I noticed that and then forgot to do anything about it. Well spotted!

floehopper:
@suzannehamilton:

Thanks for reviewing.

> There is definitely some puppet monitoring - we see Icinga alerts on 2nd line about puppet failing on specific machines.

Great. That's good to know and gives me more confidence to merge this.

> It would be great to test this on a branch as Chris suggested, though, to see how easy it would be to debug if it happened in the wild.

Sounds like a good plan. 👍

floehopper:
Rebasing against master and force-pushing in preparation for testing on integration and then merging.

floehopper:
I've added a temporary commit to intentionally break the Nginx config so I can test this change in integration.

floehopper:
Rebasing against master and force-pushing in preparation for testing on integration.

floehopper:
I deployed this branch to integration this morning including the temporary commit which intentionally causes a syntax error in the Nginx config. I saw the following output from the `govuk_puppet` cron job on one of the backend machines:

```
Nov  9 12:36:01 backend-2.backend CRON[16770]: (root) CMD (/usr/local/bin/govuk_puppet)
Nov  9 12:38:37 backend-2.backend puppet-agent[16778]: (/Stage[main]/Ssh/Sshkey[ci-agent-4.ci.integration.publishing.service.gov.uk]/host_aliases) host_aliases changed ['ci-agent-4.ci', '172.17.0.1'] to 'ci-agent-4.ci 172.18.0.1'
Nov  9 12:38:52 backend-2.backend puppet-agent[16778]: (/Stage[main]/Govuk::Apps::Asset_manager/Govuk::App[asset-manager]/Govuk::App::Config[asset-manager]/Govuk::App::Nginx_vhost[asset-manager]/Nginx::Config::Vhost::Proxy[asset-manager.integration.publishing.service.gov.uk]/Nginx::Config::Site[asset-manager.integration.publishing.service.gov.uk]/File[/etc/nginx/sites-available/asset-manager.integration.publishing.service.gov.uk]/content) content changed '{md5}8e64946d6357e3202c174b29d05babc8' to '{md5}dd1181adc4cb7909694a9e1a934182b4'
Nov  9 12:38:53 backend-2.backend puppet-agent[16778]: (/Stage[main]/Nginx::Service/Service[nginx]) Failed to call refresh: Could not restart Service[nginx]: Execution of 'nginx -t && /etc/init.d/nginx reload' returned 1: nginx: [emerg] host not found in upstream "asset-manager.intentionally-broken-integration.publishing.service.gov.uk-proxy" in /etc/nginx/sites-enabled/asset-manager.integration.publishing.service.gov.uk:64
Nov  9 12:38:53 backend-2.backend puppet-agent[16778]: (/Stage[main]/Nginx::Service/Service[nginx]) nginx: configuration file /etc/nginx/nginx.conf test failed
Nov  9 12:38:53 backend-2.backend puppet-agent[16778]: (/Stage[main]/Nginx::Service/Service[nginx]) Could not restart Service[nginx]: Execution of 'nginx -t && /etc/init.d/nginx reload' returned 1: nginx: [emerg] host not found in upstream "asset-manager.intentionally-broken-integration.publishing.service.gov.uk-proxy" in /etc/nginx/sites-enabled/asset-manager.integration.publishing.service.gov.uk:64
Nov  9 12:38:53 backend-2.backend puppet-agent[16778]: (/Stage[main]/Nginx::Service/Service[nginx]) nginx: configuration file /etc/nginx/nginx.conf test failed
Nov  9 12:38:55 backend-2.backend puppet-agent[16778]: Finished catalog run in 50.94 seconds
```

Thus the syntax error was detected but this did not seem to cause an Icinga alert, presumably because the `govuk_puppet` script did not return an error code. This means that although no attempt will have been made to restart/reload Nginx, the Nginx config will have become invalid and if the machine was re-booted Nginx would not come up again. Thus this PR needs a re-think.

@chrisroos: I wonder whether this issue might be relevant to #6696 too. And I also wonder whether it might make sense to combine that PR with this one...?

floehopper:
Here are a couple of links suggested by @sihugh which show how some Rake tasks trigger Icinga alerts:

* https://github.com/alphagov/local-links-manager/blob/master/lib/services.rb#L34
* https://github.com/alphagov/govuk-puppet/search?utf8=%E2%9C%93&q=passive_check&type=

chrisroos:
We've just seen another problem where we weren't alerted to invalid nginx config - https://github.com/alphagov/asset-manager/issues/412.

thomasleese:
Closing this to reduce number of open PRs, we may come back to this in the future.

PR #6695, opened 2017-10-24
Configure Bouncer in AWS

issyl0:
EDIT (2017-11-02): This is finally working - Bouncer can read from the Transition database on the RDS replica (transition-postgresql-standby).

- This required a rework of the `postgresql_role` class to not use raw SQL commands but module classes instead (after an interlude where we tried to bodge it into working by specifying usernames, passwords and hosts in `aws_migration` conditionals), which reduces the amount of code and makes it a lot neater.

### DIFF
--- a/hieradata_aws/class/bouncer.yaml
+++ b/hieradata_aws/class/bouncer.yaml
@@ -1,0 +1,4 @@
+---
+
+govuk::node::s_base::apps:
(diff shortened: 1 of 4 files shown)


issyl0:
@surminus Yeah, I was going to suggest we re-review this!

PR #6696, opened 2017-10-25
Update nginx config check in restart class

chrisroos:
I investigated the difference between `/etc/init.d/nginx configtest` and
`nginx -t` while reviewing PR #6693.

On my dev-vm, and therefore presumably in our other environments,
`/etc/init.d/nginx configtest` always appears to exit with a status code
of 0 irrespective of whether there are errors in the config. Puppet uses
the exit code of this `onlyif` command to determine whether to run the
main command[1]. Because it was always exiting with 0, an invalid nginx
config wouldn't actually prevent the restart command from being run.

I've confirmed that `nginx -t` does exit with a non-zero code when it
encounters an error in the config so I think this is a better command to
use when checking whether we can restart.

While I'm fairly confident this change is good, I've not been able to
test it on my dev-vm. The only things that trigger `nginx::restart` are
an install/upgrade of nginx-common and nginx-full so I think I'd have to
somehow change the version being installed and it's not clear how I'd do
that.

This snippets below illustrate the differences between `nginx -t` and
`configtest` having first updated the nginx config to contain a syntax
error:

== Using `configtest`

```
$ sudo /etc/init.d/nginx configtest
 * Testing nginx configuration
   ...fail!

$ echo $?
0

$ sudo /etc/init.d/nginx configtest && echo "nginx config ok"
 * Testing nginx configuration
   ...fail!
nginx config ok
```

== Using `nginx -t`

```
$ sudo nginx -t
nginx: [emerg] unknown directive "foo" in /etc/nginx/nginx.conf:3
nginx: configuration file /etc/nginx/nginx.conf test failed

$ echo $?
1

$ sudo nginx -t && echo "nginx config ok"
nginx: [emerg] unknown directive "foo" in /etc/nginx/nginx.conf:3
nginx: configuration file /etc/nginx/nginx.conf test failed
```

[1]: https://puppet.com/docs/puppet/5.2/types/exec.html#exec-attribute-onlyif

### DIFF
--- a/modules/nginx/manifests/restart.pp
+++ b/modules/nginx/manifests/restart.pp
@@ -3,7 +3,7 @@ class nginx::restart {
 
   exec { '/etc/init.d/nginx restart':
     refreshonly => true,
-    onlyif      => '/etc/init.d/nginx configtest',
+    onlyif      => 'nginx -t',
   }
 
 }


floehopper:
@chrisroos:

> While I'm fairly confident this change is good, I've not been able to
test it on my dev-vm. The only things that trigger nginx::restart are
an install/upgrade of nginx-common and nginx-full so I think I'd have to
somehow change the version being installed and it's not clear how I'd do
that.

I haven't really thought it through properly or tried it, but could you add a new trigger to restart Nginx when a file changes? Something like this:

```
file { "/tmp/nginx-restart.txt":
  content => 'nginx-restart',
  notify  => Class['nginx::service'],
}
```

floehopper:
@chrisroos It might also be good to add something to the commit note / PR description to explain what currently happens if `/etc/init.d/nginx restart` is run when there is a syntax error in the config. Presumably the service stops and then can't be restarted...?

chrisroos:
> It might also be good to add something to the commit note / PR description to explain what currently happens if `/etc/init.d/nginx restart` is run when there is a syntax error in the config. Presumably the service stops and then can't be restarted...?

I've just check this, @floehopper. I was surprised to find that it prints an error message and exits with a 0 status code! I've updated the commit message with the following information:

> As with PR #6693, I think the previous implementation left us open to
> deploying invalid nginx config changes and only finding out about them
> when restarting a machine or force restarting the nginx process. This is
> because if there's an error in the nginx config when executing
> `/etc/init.d/nginx restart` then it'll print an error message before
> exiting with a status code of 0:
> 
>     # Check configuration before stopping nginx
>     if ! test_nginx_config; then
>       log_end_msg 1 # Configuration error
>       exit 0
>     fi
> 
> It's not obvious that we'd notice these error messages if they appeared
> in the Puppet output so I think that using `nginx -t` to fail fast make
> sense.

chrisroos:
We've just seen another problem where we weren't alerted to invalid nginx config - https://github.com/alphagov/asset-manager/issues/412.

thomasleese:
Closing this to reduce number of open PRs, we may come back to this in the future.

PR #6706, opened 2017-10-25
Ensure Mtail service is stopped

afda16:
We are not going to continue working on Mtail for now,
there are some problems with exporting the metrics to
Graphite/Statsd. Ensure the service is stopped.

### DIFF
--- a/modules/govuk_mtail/manifests/init.pp
+++ b/modules/govuk_mtail/manifests/init.pp
@@ -75,6 +75,7 @@ class govuk_mtail(
   }
 
   @@icinga::check { "check_mtail_up_${::hostname}":
+    ensure              => 'absent',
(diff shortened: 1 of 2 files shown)


modules/govuk_mtail/manifests/init.pp
```diff
@@ -74,9 +74,9 @@
     notify  => Service['mtail'],
   }
 
-  @@icinga::check { "check_mtail_up_${::hostname}":
-    check_command       => 'check_nrpe!check_proc_running!mtail',
-    service_description => 'mtail not running',
-    host_name           => $::fqdn,
-  }
+#  @@icinga::check { "check_mtail_up_${::hostname}":
```
surminus: this can be set `ensure => 'absent'`

PR #6707, opened 2017-10-25
Define Asset Manager as a dependency of Whitehall

chrisroos:
Whitehall is now uploading organisation logos to Asset Manager so it's
useful to have Asset Manager started as part of starting Whitehall in
development.


chrisroos:
Thanks for the review, @chrislo.

PR #6720, opened 2017-10-27
Remove Stagecraft

surminus:
This includes removing the api-postgresql instances as Stagecraft was the only application using them.

Otherwise set the application to absent where possible, with the intention of removing the code once everything has been deleted from the app servers.


pcraig3:
I was looking at this yesterday as well and also ended up with a failing branch, but I can't see why it was failing.

surminus:
My failures were linting errors.

PR #6723, opened 2017-10-27
Mirror repository for elasticsearch 5 releases

MatMoore:
We're aiming to upgrade to this version before 2.4 reaches its end of life
early next year.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/deb.html#deb-repo


surminus:
Can we remove anything related to 1.x while you're there?

MatMoore:
@surminus I had a look and there's some left over stuff related to api-elasticsearch, which I'm assuming we can delete now.

As far as I can tell the AWS and training environments are still set to use 1.7, so I'm not sure if its safe to delete those repos yet.

surminus:
Ah OK, no worries, I will check next week. :+1:

MatMoore:
Just pushed one more change to deploy the migration plugin. This'll let us enable some deprecation logging on integration.

PR #6727, opened 2017-10-30
Amend RDS backups

surminus:
This updates some logic to allow protection against Puppet errors, and removes the
placeholder bucket name in hieradata.

https://trello.com/c/F6UQeeiN/777-design-rds-replicas-and-backups

### DIFF
--- a/modules/govuk/manifests/node/s_db_admin.pp
+++ b/modules/govuk/manifests/node/s_db_admin.pp
@@ -30,7 +30,14 @@ class govuk::node::s_db_admin(
 ) {
   include ::govuk::node::s_base
 
+  if $backup_s3_bucket {
+    $ensure = 'present'
+  } else {
+    $ensure = 'absent'
+  }
+
   apt::source { 'gof3r':
+    ensure       => $ensure,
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/node/s_db_admin.pp
```diff
@@ -30,15 +30,22 @@
...
+    ensure       => $ensure,
     location     => "http://${apt_mirror_hostname}/gof3r",
     release      => $::lsbdistcodename,
     architecture => $::architecture,
     key          => '3803E444EB0235822AA36A66EC5FE1A937E3ACBB',
   }
 
   package { 'gof3r':
-    ensure  => '0.5.0',
+    ensure  => $ensure,
```
deanwilson: Are we ok with the version of this changing? I assume because we maintain our own package 0.5.0 is the newest and we'll get the version we want but this does loosen the requirement a little.

surminus: I think I am happy not pinning this. There hasn't been a new version released in two years, and is the only one we have available.

PR #6732, opened 2017-10-31
Remove stagecraft rabbit user permissions

gpeng:
The stagecraft rabbit mq user is `ensure => absent` so the permission config is now causing puppet run errors.

This commit removes the offending lines to allow puppet to run successfully.

### DIFF
--- a/modules/govuk/manifests/apps/stagecraft/rabbitmq.pp
+++ b/modules/govuk/manifests/apps/stagecraft/rabbitmq.pp
@@ -39,14 +39,4 @@ class govuk::apps::stagecraft::rabbitmq (
     admin    => false,
     password => $amqp_pass,
   }
-
-  rabbitmq_user_permissions { "${amqp_user}@/${amqp_vhost}":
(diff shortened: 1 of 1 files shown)


surminus:
Is this change in Production? If so, then the PR could be expanded to remove anything related to Stagecraft. If not, then we should ensure Puppet is deployed to Production before merging this.

gpeng:
It's not in production yet. I want to do a deploy but the current release on integration is erroring on the rabbit boxes. This seems to fix it so I was planning to merge it and deploy. There are a few other open PRs removing all the other stuff I think.

gpeng:
We could merge #6718 and #6714 too I think and I could deploy it all.

surminus:
Before we remove https://github.com/alphagov/govuk-puppet/pull/6714 we need to deploy https://github.com/alphagov/govuk-puppet/pull/6713 probably.

gpeng:
Oh yeah.. that's the one I meant. I'll merge them in :+1:

PR #6734, opened 2017-10-31
Big tidy of Elasticsearch to S3 snapshots

surminus:
This simplifies the manifest and script for creating Elasticsearch snapshots in S3.

 - Remove the user variable and just run it as root as there is no reason to create a separate user for this task
 - Rather than creating a separate "awspasswd" file, pass the credentials straight into the script.
 - Set an "ensure" parameter
 - Add a conditional if in AWS to set a base_path, but not set AWS access keys
 - Only create the scripts and cron if an S3 bucket is defined to ensure no errors

I will raise another PR to add the hieradata which will specify the indices and create the backups.

https://trello.com/c/wl3p8Y9D/787-investigate-elasticsearch-backups


modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -39,71 +34,53 @@
...
-      owner   => $user,
-      group   => $user,
-      mode    => '0640',
-      content => template('govuk_elasticsearch/awspasswd.erb'),
-    }
+  $threshold_secs = 28 * 3600
+  $service_desc = 'Elasticsearch-Index_Backup'
+  $json_es_indices = join(regsubst($es_indices, '(.*)', '"\1"'), ',')
 
+  if $s3_bucket {
```
afda16: If all the resources of the class require `s3_bucket` being defined, why don't we make the parameter mandatory in the class definition?

modules/govuk_elasticsearch/manifests/backup.pp
```diff
@@ -39,71 +34,53 @@
...
-      path    => '/usr/local/bin/es-restore-s3',
-      owner   => 'root',
-      group   => 'root',
+    file { '/usr/local/bin/es-restore-s3':
+      ensure  => $ensure,
       mode    => '0755',
       content => template('govuk_elasticsearch/es-restore-s3.erb'),
     }
 
+    class { 'govuk_elasticsearch::housekeeping':
```
afda16: Why do we need a different `govuk_elasticsearch::housekeeping` class if only `govuk_elasticsearch::backup` is going to include it?

modules/govuk_elasticsearch/manifests/backup.pp
surminus: Good point.

modules/govuk_elasticsearch/manifests/backup.pp
surminus: Will change.

modules/govuk_elasticsearch/manifests/housekeeping.pp
```diff
@@ -2,28 +2,22 @@
...
-class govuk_elasticsearch::housekeeping(
+class govuk_elasticsearch::housekeeping (
+  $ensure = 'present',
   $es_repo = [],
-  $es_snapshot_limit = 5,
-  $user = 'govuk-backup',
+  $es_snapshot_limit = 7,
   ) {
 
   if $es_repo == undef {
```
afda16: Can  `$es_repo == undef` be true if we are setting `$es_repo = []` as default?

modules/govuk_elasticsearch/spec/classes/govuk_elasticsearch__backup_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    let(:params){{
+        :es_indices => [
+            'cagney',
+            'lacey'
+        ],
+        :s3_bucket => 'mybucket',
+        :aws_secret_access_key => 'foo',
+    }}
+
+    it { is_expected.to contain_file('/usr/local/bin/es-backup-s3').with_ensure('present').with_content(/"cagney","lacey"/).without_content(/"access_key": "bar",\n    "secr    et_key": "foo"/) }
```
afda16: There are extra white spaces in "secr    et_key"

PR #6736, opened 2017-11-01
Proxy Whitehall asset requests to S3 in production

chrisroos:
We're planning to serve Whitehall organisation logos from Asset Manager
in the near future. We'll need to serve these from S3 in order to
preserve the custom `ETag` and `Last-Modified` values that are stored
when the assets are created (these custom values aren't used when assets
are served from NFS).

Note that we'll need to be regularly copying assets from Production to
Staging and Integration before we can start proxying these requests to
S3 in Staging and Integration.


chrisroos:
Thanks, @floehopper. I've rebased on master in preparation for merging.

PR #6742, opened 2017-11-03
Retire stagecraft

pcraig3:
This pull request removes whatever was left of stagecraft and its associated services. All of the `*.pp` files I've deleted have an `ensure => 'absent'` key-value so I'm assuming they've already been deleted by now. 

Do let me know if you spot any problems.


surminus:
I'm not sure if this needs to go in 8 commits, but otherwise it looks fine 👍 . Do you have anyone who can remove all Stagecraft references in encrypted hieradata?

pcraig3:
Squished. 

About the encrypted data, I can ask @leelongmore when he's back in. I wouldn't be able to unencrypt it.

PR #6743, opened 2017-11-03
Remove performance_frontend_* references

pcraig3:
There are no longer any performance platform frontend apps being deployed through puppet so we should be okay to get rid of this.

We're still running backdrop so I didn't touch any of the backend stuff though.


surminus:
We should also turn off the related machines.

pcraig3:
@surminus, is that something else to be done in this PR or does that happen afterwards/somewhere else?

surminus:
I'll turn them off/delete them today. Feel free to merge.

PR #6744, opened 2017-11-03
Enable ElasticSearch backups to S3

issyl0:
- Only run the backup script on the designated ES machine in AWS. Report OK to Nagios even if it doesn't run.

This needs a `?w=1` for review on GitHub as there are a load of indentation changes.

### DIFF
--- a/modules/govuk_elasticsearch/templates/es-backup-s3.erb
+++ b/modules/govuk_elasticsearch/templates/es-backup-s3.erb
@@ -15,6 +15,19 @@ function nagios_passive () {
 
 trap nagios_passive EXIT
 
+<%- if scope.lookupvar('::aws_migration') %>
+  # Get the instance ID for the specific machine.
+  INSTANCE_ID=$(ec2metadata --instance-id)
+  # Which instance ID is enabled for backups?
+  INSTANCE_WITH_BACKUPS_ENABLED=$(aws ec2 describe-tags --filters "Name=key,Values=backups_enabled" "Name=value,Values=1" | jq -r '.Tags[] | [.ResourceId] | @csv' | tr -d '"' )
+
+  if [ "$INSTANCE_WITH_BACKUPS_ENABLED" != "$INSTANCE_ID" ]; then
+    NAGIOS_CODE=0
+    NAGIOS_MESSAGE="OK: This machine does not run ES backups"
+    exit 0
+  fi
+<%- end -%>
+
 ES_SNAPSHOT=snapshot_$(date +%F)
 
 # PARAMETERS FOR ELASTICSEARCH SNAPSHOT REPOSITORY


modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -2,57 +2,71 @@
...
-}
-EOD
-)
-
-/usr/bin/curl --connect-timeout 10 -sS -XPUT "http://127.0.0.1:9200/_snapshot/<%= @es_repo -%>?wait_for_completion=true&pretty" -d "$REPO_DATA"
-
-if [ $? -eq 0 ]; then
-  /usr/bin/curl --connect-timeout 10 -sS -XPUT "http://127.0.0.1:9200/_snapshot/<%= @es_repo -%>/${ES_SNAPSHOT}?wait_for_completion=true&pretty" -d "$SNAPSHOT_DATA" &> /dev/null
-fi
+if [<%= @aws_migration %>]; then
```
surminus: Make this entire line `<%- if scope.lookupvar('::aws_migration') %>`

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -15,6 +15,19 @@ function nagios_passive () {
 
 trap nagios_passive EXIT
 
+<%- if scope.lookupvar('::aws_migration') %>
+  # Get the instance ID for the specific machine.
+  INSTANCE_ID=ec2metadata --instance-id
```
afda16: We have the instance ID in a facter too, we could use it in the template. Anyway, the exec needs to be wrapped in $() or /` /`

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -15,6 +15,19 @@ function nagios_passive () {
...
+  # Which instance ID is enabled for backups?
+  INSTANCE_WITH_BACKUPS_ENABLED=aws ec2 describe-tags --filters "Name=key,Values=backups_enabled" "Name=value,Values=1" | grep ResourceId | grep "i\-"
```
afda16: The exec needs to be wrapped in $() or \` `

Instead of `grep`, you can use `jq` to parse the JSON result, same as we do in the user_data. It would look like:

```
INSTANCE_WITH_BACKUPS_ENABLED=$(aws ec2 describe-tags --filters "Name=key,Values=backups_enabled" "Name=value,Values=1" | jq -r '.Tags[] | [.ResourceId] | @csv' | tr -d '"' )
```

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -15,6 +15,19 @@ function nagios_passive () {
...
+
+  if [INSTANCE_WITH_BACKUPS_ENABLED != INSTANCE_ID]; then
```
afda16: The variable needs `$`

modules/govuk_elasticsearch/templates/es-backup-s3.erb
issyl0: That's clever! Thanks.

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -15,6 +15,19 @@ function nagios_passive () {
 
 trap nagios_passive EXIT
 
+<%- if scope.lookupvar('::aws_migration') %>
+  # Get the instance ID for the specific machine.
+  INSTANCE_ID=$(ec2metadata --instance-id)
+  # Which instance ID is enabled for backups?
+  INSTANCE_WITH_BACKUPS_ENABLED=$(aws ec2 describe-tags --filters "Name=key,Values=backups_enabled" "Name=value,Values=1" | jq -r '.Tags[] | [.ResourceId] | @csv' | tr -d '"' ))
```
afda16: Remove one `)`

modules/govuk_elasticsearch/templates/es-backup-s3.erb
```diff
@@ -15,6 +15,19 @@ function nagios_passive () {
...
+
+  if [$INSTANCE_WITH_BACKUPS_ENABLED != $INSTANCE_ID]; then
```
afda16: The conditional also needs spaces after `[` and before `]`, you could also add quotes around the variable

issyl0: Quotes instead of `$`?

issyl0: Oh, as well as. Amended. Bash isn't my forte.

PR #6766, opened 2017-11-09
Remove some more machines

surminus:
Remove performance-frontend and api-postgresql machines as they have gone to a special
infrastructure in the sky.


issyl0:
"Special infrastructure in the sky". <3

PR #6776, opened 2017-11-13
Serve requests to consultation response form assets from Asset Manager

chrislo:
We have changed Whitehall to save consultation
response forms in Asset Manager[1] and have migrated all existing data
to Asset Manager[2].

This commit changes `nginx` to serve requests for these assets from
Asset Manager instead of Whitehall.

We made an equivalent change for Organisation logos in
17e2e1d4dbf. That commit has details about the considerations around
cache control headers which also apply here.

[1] https://github.com/alphagov/whitehall/pull/3523
[2] https://github.com/alphagov/asset-manager/issues/297#issuecomment-343971912

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1627,6 +1627,7 @@ router::assets_origin::asset_routes:
   '/government/placeholder': "whitehall-frontend"
   '/government/uploads/': "whitehall-frontend"
   '/government/uploads/system/uploads/organisation/logo/': "static"
+  '/government/uploads/system/uploads/consultation_response_form_data/file/': "static"
   '/government-frontend/': "government-frontend"
   '/info-frontend/': "info-frontend"
   '/manuals-frontend/': "manuals-frontend"
--- a/modules/govuk/templates/static_extra_nginx_config.conf.erb
+++ b/modules/govuk/templates/static_extra_nginx_config.conf.erb
@@ -18,7 +18,8 @@ location /robots.txt {
 
 <% [
   '/media/',
-  '/government/uploads/system/uploads/organisation/logo/'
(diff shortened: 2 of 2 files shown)


chrisroos:
I'm taking a look at this now.

chrisroos:
I've updated the title to indicate that we shouldn't merge this based on my comment in https://github.com/alphagov/asset-manager/issues/297#issuecomment-346404083.

chrisroos:
I've rebased this on master and force-pushed.

chrisroos:
I was hoping to get this merged today. Before doing so, because it's been open for quite some time, I compared the number of consultation response form assets on the Whitehall NFS mount to the corresponding number of assets in the Asset Manager database. Unfortunately, there was a discrepancy of 2 which were due to us not migrating hidden files when we uploaded existing consultation response forms from Whitehall to Asset Manager. I've captured this in https://github.com/alphagov/asset-manager/issues/368. I'm not sure whether we should address that issue before merging this change.

chrisroos:
I've rebased on master and force pushed again in preparation for merging.

chrisroos:
@dwhenry has checked the consultation response form data files in production and confirmed that the number of files on disk matches the number of files in the database:

```
irb(main):001:0> WhitehallAsset.where(legacy_url_path: %r(/government/uploads/system/uploads/consultation_response_form_data/file)).count
=> 741

WhitehallAsset.deleted.where(legacy_url_path: %r(/government/uploads/system/uploads/consultation_response_form_data/file)).count
=> 4

find /data/uploads/whitehall/clean/system/uploads/consultation_response_form_data/file/ -type f | wc -l
741
```

As such I'm happy to merge this change.

PR #6779, opened 2017-11-14
Add credentials for Google Management API

elenatanasoiu:
For: https://trello.com/c/SYgBUl4q/258-develop-an-upload-to-ga-of-link-checker-data

This is going to enable us to upload custom files to Google Analytics using the Google Management API. Previously we would do this manually.

In order to do this we need some credentials to be set for a service account that was created to do the upload. Once this is set up we can get authorization to perform our upload as a scheduled task.

### DIFF
--- a/modules/govuk/manifests/apps/local_links_manager.pp
+++ b/modules/govuk/manifests/apps/local_links_manager.pp
@@ -88,10 +88,13 @@ class govuk::apps::local_links_manager(
   $redis_port = undef,
   $local_links_manager_passive_checks = false,
   $google_analytics_govuk_view_id = undef,
-  $google_client_email = undef,
-  $google_private_key = undef,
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/local_links_manager.pp
```diff
@@ -88,10 +88,13 @@
...
   $google_analytics_govuk_view_id = undef,
-  $google_client_email = undef,
-  $google_private_key = undef,
   $publishing_api_bearer_token = undef,
   $link_checker_api_secret_token = undef,
+  $google_client_email = undef,
+  $google_private_key = undef,
+  $google_export_account_id = undef,
+  $google_export_custom_data_import_source_id = undef,
+  $google_export_tracker_id = undef
```
surminus: This needs a comma on the end

PR #6781, opened 2017-11-14
Ensure design principles is absent

h-lame:
For: https://trello.com/c/M45WyUGZ/261-retire-design-principles

Start the process of removing the design principles app.

Flagged as DNM because it should not be merged until alphagov/short-url-manager#104 has been deployed (or some alternative to it) as that PR handles the redirects that allow us to remove this app.


h-lame:
Removed DNM flag as the PR that redirects all the design principles routes has been deployed to production.

PR #6783, opened 2017-11-14
Enable Gor for AWS on cache-1 staging

afda16:
Start replaying traffic on AWS Integration from cache-2 on Staging.
Add a new parameter to router::gor to disable adding the target IP
to the hosts file for the AWS endpoint (it's behind an ELB so we need
to resolve). On Carrenza we keep the current configuration, adding the
IP to the hosts file, to avoid making DNS calls.

### DIFF
--- a/hieradata/node/cache-2.router.staging.publishing.service.gov.uk.yaml
+++ b/hieradata/node/cache-2.router.staging.publishing.service.gov.uk.yaml
@@ -1,0 +1,5 @@
+---
+
+router::gor::add_hosts: false
+router::gor::replay_targets:
+  'www-origin.blue.integration.govuk.digital': {}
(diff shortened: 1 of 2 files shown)


hieradata/node/cache-2.router.staging.publishing.service.gov.uk.yaml
```diff
@@ -0,0 +1,6 @@
+---
+
+router::gor::add_hosts: false
+router::gor::replay_targets:
+  'www-origin.blue.integration.govuk.digital':
+    ip: 'www-origin.blue.integration.govuk.digital'
```
surminus: I wonder since we're only using the key we could just set it to:
```
router::gor::replay_targets:
  'www-origin.blue.integration.govuk.digital': {}
```

afda16: you are right, updating

PR #6789, opened 2017-11-15
Alert on growth or decline of search indexes

MatMoore:
I'm using graphite to store a [statsd gauge](https://github.com/etsy/statsd/blob/master/docs/metric_types.md#gauges) which measures the number of documents in each elasticsearch index over time. Icinga should alert if the current value is suddenly very different from a previous value. 

I've made it compare to 3 days ago so that on monday morning, it will compare to friday, instead of the weekend. If the alert triggers out of hours, we don't want that value to become the new baseline until we've had a chance to see it.

If statsd data is missing (a value of None), no alert will be raised. This is not a huge problem. As long as statsd is running it keeps gauges at their previous value.

Trello: https://trello.com/c/XvzOqrHB/432-trigger-an-alert-if-there-is-high-growth-reduction-in-index-size

### DIFF
--- a/modules/monitoring/manifests/checks.pp
+++ b/modules/monitoring/manifests/checks.pp
@@ -110,23 +110,49 @@ class monitoring::checks (
   # END signon
 
   # START search
-  # TODO: increase the comparison period to a week so alerts don't dissapear if they happen over the weekend
-  # and extend checks to other indices
-  icinga::check::graphite { 'check_rummager_govuk_index_size_increased':
-    target    => 'divideSeries(stats.gauges.govuk.app.rummager.govuk_index.docs.count, timeShift(stats.gauges.govuk.app.rummager.govuk_index.docs.count, "1d"))',
-    warning   => 110,
-    critical  => 125,
-    desc      => 'rummager govuk index size has increased by more than 10% over the last 3 days',
-    host_name => $::fqdn,
+
+  # On average 326 new documents were created per day in 2017, in total.
+  # The govuk index will only receive a fraction of these until we start
+  # indexing whitehall content in it.
+  icinga::check::graphite { 'check_rummager_govuk_index_size_changed':
+    target              => 'absolute(diffSeries(stats.gauges.govuk.app.rummager.govuk_index.docs.count, timeShift(stats.gauges.govuk.app.rummager.govuk_index.docs.count, "3d"))',
+    warning             => 1500,
+    critical            => 15000,
+    desc                => 'rummager govuk index size has significantly increased/decreased over the last 3 days',
+    host_name           => $::fqdn,
+    notification_period => 'inoffice',
   }
 
-  icinga::check::graphite { 'check_rummager_govuk_index_size_decreased':
-    target    => 'transformNull(divideSeries(timeShift(stats.gauges.govuk.app.rummager.govuk_index.docs.count, "1d"), stats.gauges.govuk.app.rummager.govuk_index.docs.count), 10000000)',
-    warning   => 101,
-    critical  => 101,
-    desc      => 'govuk index document count from 1 week ago exceeds current count',
-    host_name => $::fqdn,
+  # Mainstream is comparable to the govuk index.
+  icinga::check::graphite { 'check_rummager_mainstream_index_size_changed':
+    target              => 'absolute(diffSeries(stats.gauges.mainstream.app.rummager.mainstream_index.docs.count, timeShift(stats.gauges.mainstream.app.rummager.mainstream_index.docs.count, "3d"))',
+    warning             => 1500,
+    critical            => 15000,
+    desc                => 'rummager mainstream index size has significantly increased/decreased over the last 3 days',
+    host_name           => $::fqdn,
+    notification_period => 'inoffice',
   }
+
+  # Government is comparable to the govuk index.
+  icinga::check::graphite { 'check_rummager_government_index_size_changed':
+    target              => 'absolute(diffSeries(stats.gauges.government.app.rummager.government_index.docs.count, timeShift(stats.gauges.government.app.rummager.government_index.docs.count, "3d"))',
+    warning             => 1500,
+    critical            => 15000,
+    desc                => 'rummager government index size has significantly increased/decreased over the last 3 days',
+    host_name           => $::fqdn,
+    notification_period => 'inoffice',
+  }
+
+  # Detailed is smaller than the other indexes (about 4500 documents)
+  icinga::check::graphite { 'check_rummager_detailed_index_size_changed':
+    target              => 'absolute(diffSeries(stats.gauges.detailed.app.rummager.detailed_index.docs.count, timeShift(stats.gauges.detailed.app.rummager.detailed_index.docs.count, "3d"))',
+    warning             => 200,
+    critical            => 400,
+    desc                => 'rummager detailed index size has significantly increased/decreased over the last 3 days',
+    host_name           => $::fqdn,
+    notification_period => 'inoffice',
+  }
+
   # END search
 
   if $enable_support_check {
--- a/modules/monitoring/manifests/checks.pp
+++ b/modules/monitoring/manifests/checks.pp
@@ -109,6 +109,26 @@ class monitoring::checks (
   }
   # END signon
 
+  # START search
+  # TODO: increase the comparison period to a week so alerts don't dissapear if they happen over the weekend
+  # and extend checks to other indices
+  icinga::check::graphite { 'check_rummager_govuk_index_size_increased':
+    target    => 'divideSeries(stats.gauges.govuk.app.rummager.govuk_index.docs.count, timeShift(stats.gauges.govuk.app.rummager.govuk_index.docs.count, "1d"))',
+    warning   => 110,
+    critical  => 125,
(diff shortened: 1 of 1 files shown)


modules/monitoring/manifests/checks.pp
```diff
@@ -109,6 +109,26 @@
...
+  icinga::check::graphite { 'check_rummager_govuk_index_size_increased':
+    target    => 'divideSeries(stats.gauges.govuk.app.rummager.govuk_index.docs.count, timeShift(stats.gauges.govuk.app.rummager.govuk_index.docs.count, "1d"))',
+    warning   => 110,
+    critical  => 125,
+    desc      => 'rummager govuk index size has increased by more than 10% over the last 3 days',
+    host_name => $::fqdn,
+  }
+
+  icinga::check::graphite { 'check_rummager_govuk_index_size_decreased':
+    target    => 'transformNull(divideSeries(timeShift(stats.gauges.govuk.app.rummager.govuk_index.docs.count, "1d"), stats.gauges.govuk.app.rummager.govuk_index.docs.count), 10000000)',
```
suzannehamilton: What's the significance of 10000000? And why doesn't the other check need `transformNull`?

suzannehamilton: Sorry, I should read the whole of the PR description...

MatMoore: There is a definite case I want this check to handle where the count goes from some number to zero. In this case I put 10000000 in as an arbitrary value.

The other case is checking for an increase, so the only divide by zero is if the data was wrong in the past. In this case we can't tell if something's wrong so I'm just ignoring it.

I'm not really happy with this though because there's also the case where graphite has null data - this will currently also trigger the "decreased" alert.

MatMoore:
I've simplified this a bit, and tested the queries in grafana. When this is on integration we can test it by deleting things.

modules/monitoring/manifests/checks.pp
```diff
@@ -109,6 +109,52 @@
...
 
+  # START search
+
+  # On average 326 new documents were created per day in 2017, in total.
+  # The govuk index will only receive a fraction of these until we start
+  # indexing whitehall content in it.
+  icinga::check::graphite { 'check_rummager_govuk_index_size_changed':
+    target              => 'absolute(diffSeries(stats.gauges.govuk.app.rummager.govuk_index.docs.count, timeShift(stats.gauges.govuk.app.rummager.govuk_index.docs.count, "3d"))',
+    warning             => 1500,
+    critical            => 15000,
```
dwhenry: what do these values mean, are the change in the number of documents? or percentage? also, the critical warning feels too big compared to the warning.

modules/monitoring/manifests/checks.pp
```diff
@@ -109,6 +109,52 @@
...
+    desc                => 'rummager government index size has significantly increased/decreased over the last 3 days',
+    host_name           => $::fqdn,
+    notification_period => 'inoffice',
+  }
+
+  # Detailed is smaller than the other indexes (about 4500 documents)
+  icinga::check::graphite { 'check_rummager_detailed_index_size_changed':
+    target              => 'absolute(diffSeries(stats.gauges.detailed.app.rummager.detailed_index.docs.count, timeShift(stats.gauges.detailed.app.rummager.detailed_index.docs.count, "3d"))',
+    warning             => 200,
+    critical            => 400,
```
dwhenry: It feels strange this has a different number but government is the same as mainstream/govuk given they are also quite different in size.

modules/monitoring/manifests/checks.pp
```diff
@@ -109,6 +109,52 @@
   }
   # END signon
 
+  # START search
+
+  # On average 326 new documents were created per day in 2017, in total.
```
dwhenry: I would be more interested in know what the largest change was for a three day period? how often would we have gotten an alert with the below settings?

modules/monitoring/manifests/checks.pp
MatMoore: These are changes in the number of documents.

modules/monitoring/manifests/checks.pp
MatMoore: We don't have the historical data in graphite for the actual indexes, so we don't know how many times it would legitimately fire.

With these thresholds there are only 4 days that could have been false positives (if my query works)

```
{"end_day"=>"2017-01-26", "total_3d"=>"1536"}
{"end_day"=>"2017-01-27", "total_3d"=>"1635"}
{"end_day"=>"2017-02-13", "total_3d"=>"4252"}
{"end_day"=>"2017-02-14", "total_3d"=>"4430"}
{"end_day"=>"2017-02-15", "total_3d"=>"4701"}
{"end_day"=>"2017-02-16", "total_3d"=>"5039"}
{"end_day"=>"2017-02-23", "total_3d"=>"2068"}
{"end_day"=>"2017-02-24", "total_3d"=>"2609"}
{"end_day"=>"2017-02-25", "total_3d"=>"2228"}
{"end_day"=>"2017-02-26", "total_3d"=>"1832"}
{"end_day"=>"2017-03-09", "total_3d"=>"1778"}
{"end_day"=>"2017-03-10", "total_3d"=>"1778"}
{"end_day"=>"2017-03-16", "total_3d"=>"1665"}
{"end_day"=>"2017-03-17", "total_3d"=>"1637"}
{"end_day"=>"2017-03-23", "total_3d"=>"1545"}
{"end_day"=>"2017-03-30", "total_3d"=>"1737"}
{"end_day"=>"2017-03-31", "total_3d"=>"1725"}
{"end_day"=>"2017-04-05", "total_3d"=>"1521"}
{"end_day"=>"2017-04-06", "total_3d"=>"1966"}
{"end_day"=>"2017-04-07", "total_3d"=>"1951"}
{"end_day"=>"2017-04-21", "total_3d"=>"1661"}
{"end_day"=>"2017-06-28", "total_3d"=>"2202"}
{"end_day"=>"2017-06-29", "total_3d"=>"5892"}
{"end_day"=>"2017-06-30", "total_3d"=>"5977"}
{"end_day"=>"2017-07-01", "total_3d"=>"5350"}
{"end_day"=>"2017-07-02", "total_3d"=>"4012"}
{"end_day"=>"2017-07-07", "total_3d"=>"2313"}
{"end_day"=>"2017-07-08", "total_3d"=>"2081"}
{"end_day"=>"2017-07-09", "total_3d"=>"1720"}
{"end_day"=>"2017-07-10", "total_3d"=>"1625"}
{"end_day"=>"2017-07-14", "total_3d"=>"1501"}
{"end_day"=>"2017-07-20", "total_3d"=>"1850"}
{"end_day"=>"2017-07-21", "total_3d"=>"1834"}
{"end_day"=>"2017-07-26", "total_3d"=>"1530"}
{"end_day"=>"2017-07-27", "total_3d"=>"1864"}
{"end_day"=>"2017-07-28", "total_3d"=>"1846"}
{"end_day"=>"2017-08-02", "total_3d"=>"1516"}
{"end_day"=>"2017-08-03", "total_3d"=>"1726"}
{"end_day"=>"2017-08-07", "total_3d"=>"16419"}
{"end_day"=>"2017-08-08", "total_3d"=>"16459"}
{"end_day"=>"2017-08-09", "total_3d"=>"16648"}
{"end_day"=>"2017-08-10", "total_3d"=>"17465"}
{"end_day"=>"2017-08-11", "total_3d"=>"1729"}
{"end_day"=>"2017-08-12", "total_3d"=>"1534"}
{"end_day"=>"2017-09-07", "total_3d"=>"1578"}
{"end_day"=>"2017-09-08", "total_3d"=>"1570"}
{"end_day"=>"2017-09-14", "total_3d"=>"1509"}
{"end_day"=>"2017-09-20", "total_3d"=>"1715"}
{"end_day"=>"2017-09-21", "total_3d"=>"2050"}
{"end_day"=>"2017-10-12", "total_3d"=>"1516"}
{"end_day"=>"2017-10-20", "total_3d"=>"1599"}
{"end_day"=>"2017-11-08", "total_3d"=>"3232"}
{"end_day"=>"2017-11-09", "total_3d"=>"3846"}
{"end_day"=>"2017-11-10", "total_3d"=>"4140"}
{"end_day"=>"2017-11-11", "total_3d"=>"3855"}
```

This is counting documents being created rather than published, and doesn't include unpublishing, so it's an overestimate.

Query:
```sql
results = ActiveRecord::Base.connection.execute("
with days as (
  SELECT date_trunc('day', d)::date as day
  FROM generate_series(CURRENT_DATE-365, CURRENT_DATE-1, '1 day'::interval) d ),

counts as (
  select days.day, count(*) as total
  from days left join documents on (documents.created_at::date = days.day)
  group by days.day
),
alert_periods as (
  select
      day as end_day,
      sum(total) over (order by day ROWS BETWEEN 3 PRECEDING AND CURRENT ROW) as total_3d
  from counts

)
select * from alert_periods where total_3d > 1500 order by end_day
"
)
```

It does look like the warnings are a bit low.

modules/monitoring/manifests/checks.pp
MatMoore: These numbers should be revisited anyway as govuk will get bigger. I made most of them the same because it's fairly arbitrary at the moment.

I made an exception for detailed because significantly smaller (just one format).

PR #6790, opened 2017-11-15
Update Grafana dashboards for AWS

afda16:
Our Grafana dashboards are using hostname regular expressions that
don't work on AWS. On Carrenza the metric fqdn is created including the
VPC name, which is not part of the AWS infrastructure (on AWS we use
`stackname_environment` as part of the metric fqdn), so the VPC part
needs to be removed from the expression.

All the `*_frontend` expressions need to be updated to explicity exclude efg
and licensing (if we use `*frontend*` we would be including efg and
licensing on the Carrenza dashboards, which is not bad but currently
are not included).

### DIFF
--- a/modules/grafana/templates/dashboards/whitehall_health.json.erb
+++ b/modules/grafana/templates/dashboards/whitehall_health.json.erb
@@ -57,7 +57,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "target": "groupByNode(stats.whitehall-frontend-?_frontend*.nginx_logs.whitehall-frontend_*publishing_service_gov_uk.http_5xx,4,\"sumSeries\")",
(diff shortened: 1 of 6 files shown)


modules/grafana/templates/dashboards/whitehall_health.json.erb
```diff
@@ -445,7 +445,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "target": "aliasSub(groupByNode({elasticsearch,support}-?_backend*.cpu-*.cpu-user,0,\"sumSeries\"),\"_backend\",\"\")",
+              "target": "aliasSub(groupByNode({elasticsearch,support}-*.cpu-*.cpu-user,0,\"sumSeries\"),\"_backend\",\"\")",
```
surminus: Are these both names of nodes? If so, then then the only Elasticsearch nodes we have are called `rummager-elasticsearch`, and I'm not sure what `support` is.

afda16: Nice catch, I updated

PR #6793, opened 2017-11-16
Add more jobs to Jenkins AWS

surminus:
In starting to think about how 2ndline might start to use the environment, add a few more jobs that might be used, and see if they work in the way we expect.

### DIFF
--- a/hieradata_aws/class/jenkins.yaml
+++ b/hieradata_aws/class/jenkins.yaml
@@ -54,10 +54,17 @@ govuk_jenkins::config::banner_string: 'AWS Deploy'
 
 govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::job::deploy_app
+  - govuk_jenkins::job::deploy_emergency_banner
   - govuk_jenkins::job::deploy_puppet
   - govuk_jenkins::job::passive_checks
+  - govuk_jenkins::job::remove_emergency_banner
+  - govuk_jenkins::job::run_rake_task
+  - govuk_jenkins::job::search_benchmark
+  - govuk_jenkins::job::search_index_checks
+  - govuk_jenkins::job::search_test_spelling_suggestions
   - govuk_jenkins::job::smokey
   - govuk_jenkins::job::smokey_deploy
+  - govuk_jenkins::job::user_monitor
 
 govuk_jenkins::plugins:
   ace-editor:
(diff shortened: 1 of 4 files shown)


hieradata_aws/class/jenkins.yaml
```diff
@@ -54,10 +54,18 @@ govuk_jenkins::config::banner_string: 'AWS Deploy'
 
 govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::job::deploy_app
+  - govuk_jenkins::job::deploy_emergency_banner
   - govuk_jenkins::job::deploy_puppet
+  - govuk_jenkins::job::deploy_router_data
```
afda16: Do we need to change the PLEK values of the job to work on AWS?

hieradata_aws/class/jenkins.yaml
```diff
@@ -54,10 +54,18 @@ govuk_jenkins::config::banner_string: 'AWS Deploy'
...
   - govuk_jenkins::job::passive_checks
+  - govuk_jenkins::job::remove_emergency_banner
+  - govuk_jenkins::job::run_rake_task
+  - govuk_jenkins::job::search_benchmark
```
afda16: In the configuration it uses: www-origin.integration.publishing.service.gov.uk, do we want to change this for AWS?

hieradata_aws/class/jenkins.yaml
```diff
@@ -54,10 +54,18 @@ govuk_jenkins::config::banner_string: 'AWS Deploy'
...
+  - govuk_jenkins::job::search_index_checks
+  - govuk_jenkins::job::search_test_spelling_suggestions
```
afda16: It uses www-origin.integration.publishing.service.gov.uk

hieradata_aws/class/jenkins.yaml
surminus: Probably :) I just blindly added some jobs, good stuff in actually doing in the hard work and checking them!

hieradata_aws/class/jenkins.yaml
surminus: This job is rarely used, so I'm going to remove for now.

surminus:
Updated to address all comments.

PR #6797, opened 2017-11-16
[AWS] Create jobs that deploy all apps for a given node

surminus:
This is to allow easy deployment of all apps for a given node class. The context behind this is that if a node goes away in AWS, it will be rebuilt by the autoscaling group. If this occurs, then the applications will not be automatically deployed, and having to log in to deploy them all is problematic to do by hand.

This is a very short term solution that creates jobs for each type of node class, that triggers downstream jobs for each app that is on that node class. It will use the "deployed-to-integration" branch which is set whenever an app is deployed to determine what version of the application is running in each environment.

It will also deploy using the task "deploy:without_migrations", as we just want to deploy the code automatically rather than running any tasks against the database.

This does also kick off the Smokey job after each successful deployment of an app, and this takes ~6 minutes to run, but I don't think this is too problematic.

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -90,10 +90,6 @@ node_class: &node_class
       - manuals-frontend
       - service-manual-frontend
       - static
-  frontend_lb:
-    apps:
-      - publicapi
-      - public-event-store
   logs_cdn:
     apps:
       - govuk-cdn-logs-monitor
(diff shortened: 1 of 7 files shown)


surminus:
I have tested this as working in AWS.

afda16:
Do we need to update any developer documentation on how to add new applications? We already have deployable_apps info, it's going to be a little bit confusing why we need to add the application name again in a different parameter

surminus:
I'm going to try to think of a different method to gather up which apps live on which nodes.

surminus:
I'm going to refactor first and then we can come back to this.

surminus:
I've tried refactoring it here: https://github.com/alphagov/govuk-puppet/pull/6822

Once that has gone, I will recode this PR.

PR #6800, opened 2017-11-17
Delete graphite backup files older than 2 days

dwhenry:
`find * -mtime 2` only finds files which are exactly 2 days old, this means if
the process fails the backup for that day will never be deleted. Changing it to
`find * -mtime +2` will find all files older than 2 days avoiding this buildup.


surminus:
Good spot!

PR #6805, opened 2017-11-17
Add a separate jenkins job for recreating the search indexes

MatMoore:
This process
- uses rummager's configuration to generate the settings/mappings for a
  new set of search indexes
- creates the new indexes
- triggers a reindex of content from the old index to the new one
- while this is happening, a write lock is placed on the old index so
  data isn't lost
- after completing the reindex, it runs some checks on the new index, and
  switches over to it if successful (by moving aliases)

This used to be part of the `search_fetch_analytics` job, and ran every
night but we split it out, and now only need to run it when changing
the mappings or updating elasticsearch.

This is a very slow process, so we don't want to hook it into our CI
pipeline, but we do want to test it periodically on integration.

https://trello.com/c/9nZLUhqi/439-regularly-test-reindex-task

### DIFF
--- a/modules/govuk_jenkins/manifests/job/search_reindex_with_new_schema.pp
+++ b/modules/govuk_jenkins/manifests/job/search_reindex_with_new_schema.pp
@@ -21,7 +21,6 @@ class govuk_jenkins::job::search_reindex_with_new_schema (
     host_name           => $::fqdn,
     freshness_threshold => 619200,
     action_url          => $job_url,
-    # TODO: add this to the developer docs
-    # notes_url           => monitoring_docs_url(search-reindex-failed),
+    notes_url           => monitoring_docs_url(search-reindex-failed),
   }
 }
--- a/modules/govuk_jenkins/manifests/job/search_reindex_with_new_schema.pp
+++ b/modules/govuk_jenkins/manifests/job/search_reindex_with_new_schema.pp
@@ -4,6 +4,8 @@
 #
 class govuk_jenkins::job::search_reindex_with_new_schema (
   $app_domain = hiera('app_domain'),
+  $icinga_check_enabled = false,
+  $cron_schedule = undef,
 ) {
 
   $check_name = 'search-reindex-with-new-schema'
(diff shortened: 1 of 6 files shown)


dwhenry:
This needs to ensure that old indices are deleted to avoid build up.

suzannehamilton:
The new cron commit looks fine to me. 👍 

Does the rest of it still need to be reviewed? There's still a TODO in the code, and I don't know if you're going to add anything to clean up the indices like you mentioned above?

dwhenry:
Good spot @suzannehamilton I didn't see the TODO, I will do that now before merging this PR. 

In term of cleanup, this happens as part of the "copy to integration" job, that means after this finishes running we will have two indices, the one that was imported from production and the new one we generated, we want both incase we want to do any comparison. 

As a result, I think we can ignore my above comment.

dwhenry:
documentation has been added: https://github.com/alphagov/govuk-developer-docs/pull/593

modules/govuk_jenkins/manifests/job/search_reindex_with_new_schema.pp
```diff
@@ -0,0 +1,27 @@
...
+  file { '/etc/jenkins_jobs/jobs/search_reindex_with_new_schema.yaml':
+    ensure  => present,
+    content => template('govuk_jenkins/jobs/search_reindex_with_new_schema.yaml.erb'),
+    notify  => Exec['jenkins_jobs_update'],
+  }
+
+  @@icinga::passive_check { "${check_name}_${::hostname}":
+    service_description => $service_description,
+    host_name           => $::fqdn,
+    freshness_threshold => 619200,
```
suzannehamilton: Should we configure this passive check to only run on integration? The job won't be run regularly on staging and production, which I think will trigger (incorrect) passive alerts.

dwhenry: I blame @MatMoore which is super easy to do as he isn't here. I have updated this with a flag in the hieradata.

PR #6811, opened 2017-11-20
Configure the rabbitmq hosts used by publishing_api

deanwilson:
In our current provider we use a set of servers, addressed by name.
In AWS we use an ELB endpoint with an ASG beind it.

### DIFF
--- a/modules/govuk/manifests/apps/publishing_api.pp
+++ b/modules/govuk/manifests/apps/publishing_api.pp
@@ -63,6 +63,13 @@
 # [*oauth_secret*]
 #   Sets the OAuth Secret Key
 #
+# [*rabbitmq_hosts*]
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/apps/publishing_api.pp
```diff
@@ -143,6 +152,12 @@
       port => $redis_port,
     }
 
+    govuk::app::envvar::rabbitmq { $app_name:
+      hosts    => $rabbitmq_hosts,
+      user     => $rabbitmq_user,
+      password => $rabbitmq_password,
```
surminus: Where was this set previously, as the parameter already existed?

deanwilson: It was specified below, which I've now removed. Nice catch.

PR #6814, opened 2017-11-21
Add environment filter to sentry link on deployment dashboard

h-lame:
Otherwise we link to all errors for the project.  Sentry doesn't make it
obvious via the UI how to filter to only see the errors for a given
environment so we might mean people don't check for errors because it's
too noisy, or worry about errors on integration when they're deploying
to production.

### DIFF
--- a/modules/grafana/manifests/dashboards/deployment_dashboard.pp
+++ b/modules/grafana/manifests/dashboards/deployment_dashboard.pp
@@ -45,6 +45,8 @@ define grafana::dashboards::deployment_dashboard (
   $dependent_app_5xx_errors = undef,
   $show_elasticsearch_stats = false,
   $fields_prefix = '@fields.',
+  $sentry_environment = $::govuk::deploy::config::errbit_environment_name,
+
 ) {
   if $has_workers {
     $worker_row = [['worker_failures', 'worker_successes']]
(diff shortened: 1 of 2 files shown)


modules/grafana/templates/dashboards/deployment_panels/_links.json.erb
```diff
@@ -1,5 +1,5 @@
...
+  "content": "<ul style=\"font-size: 20px;\">\n<li><a target=\"_blank\" href=\"https://sentry.io/govuk/app-<%= @app_name %>/?query=is%3Aunresolved+environment%3A<%= hiera('govuk::deploy::config::errbit_environment_name') %>\">Errors for <%= @app_name %> (Sentry)</a></li>\n<li><a target=\"_blank\" href=\"https://kibana.logit.io\">Kibana (Logit)</a></li>\n<li><a target=\"_blank\" href=\"https://alert.<%= @app_domain %>/\">Icinga</a></li>\n<li><a target=\"_blank\" href=\"https://docs.publishing.service.gov.uk/apps/<%= @docs_name %>.html\"><%= @app_name %> documentation</a></li>\n<li><a target=\"_blank\" href=\"https://govuk.zendesk.com/agent/filters/135081425\">Dashboard feedback</a></li>\n</ul>\n",
```
h-lame: Is this even how you do this?

h-lame: I'm using that var because it's the one that is used to set [SENTRY_CURRENT_ENV env vars in common.yaml](https://github.com/alphagov/govuk-puppet/blob/7a1aeee8bd675a741b7293d3a8a25beba93bbbf0/hieradata/common.yaml#L633)

h-lame: Turns out, no it wasn't.  So instead of trying to use `hiera` in the template, I'm now trying to set a var in the manifest and use that in the template.

modules/grafana/manifests/dashboards/deployment_dashboard.pp
```diff
@@ -117,6 +117,8 @@
     $elasticsearch_stats
   )
 
+  $sentry_environment = hiera('govuk::deploy::config::errbit_environment_name')
```
surminus: We do not allow direct hiera lookups in this way (other than a few special snowflakes).

Set it as a parameter at the top of the manifest, and either:

1. Reference the variable in hieradata.
2. Reference the variable in the parameter, ie `$sentry_environment = $::govuk::deploy::config::errbit_environment_name`

h-lame: I think I've done what you suggest (option 2) and it passes lint.  Thanks @surminus!

PR #6816, opened 2017-11-21
[AWS] Always run the Data_Sync_Complete job after the data sync

issyl0:
- This job wasn't running, causing alerts.

### DIFF
--- a/hieradata_aws/class/jenkins.yaml
+++ b/hieradata_aws/class/jenkins.yaml
@@ -53,6 +53,7 @@ govuk_jenkins::config::user_permissions:
 govuk_jenkins::config::banner_string: 'AWS Deploy'
 
 govuk_jenkins::job_builder::jobs:
+  - govuk_jenkins::job::data_sync_complete_integration
   - govuk_jenkins::job::deploy_app
   - govuk_jenkins::job::deploy_puppet
   - govuk_jenkins::job::passive_checks
--- a/modules/govuk_jenkins/templates/jobs/data_sync_complete_integration.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/data_sync_complete_integration.yaml.erb
@@ -12,9 +12,13 @@
     builders:
        - shell: |
            # Clear the Sidekiq queues to stop any residual 'stale' jobs running
+           <%- if scope.lookupvar('::aws_migration') %>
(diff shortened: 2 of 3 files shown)


hieradata_aws/class/jenkins.yaml
```diff
@@ -53,6 +53,7 @@ govuk_jenkins::config::user_permissions:
 govuk_jenkins::config::banner_string: 'AWS Deploy'
 
 govuk_jenkins::job_builder::jobs:
+  - govuk_jenkins::job::data_sync_complete
```
surminus: I think this will need to be `govuk_jenkins::job::data_sync_complete_integration`

issyl0: Ah!

issyl0:
@surminus Your PR https://github.com/alphagov/govuk-puppet/pull/6793 adds `run_rake_task` to the list of jobs, so that should be merged first?

surminus:
Merged!

modules/govuk_jenkins/templates/jobs/data_sync_complete_integration.yaml.erb
```diff
@@ -12,9 +12,15 @@
     builders:
        - shell: |
            # Clear the Sidekiq queues to stop any residual 'stale' jobs running
+           <%- if scope.lookupvar('::aws_migration') %>
+           for host in $(govuk_node_list -c backend); do
```
surminus: I think we only need to run this once, so I would add `--single-node` to the `govuk_node_list` command.

surminus:
One small comment, but otherwise looks OK!

modules/govuk_jenkins/templates/jobs/data_sync_complete_integration.yaml.erb
issyl0: In Carrenza on the redis hosts it did `for host in ...`, and there were multiple hosts... I assumed we'd need to run it on every host, but now you mention it if it's connecting to the central Redis then it's fine.

(I discovered yesterday that `redis-cli` isn't on backend machines in AWS, which makes running it harder (!), and I can't find anywhere that installs it in Puppet, so we'll have to fix that.)

surminus: Yup, we only have essentially single Redis endpoint, whereas in Carrenza we (confusingly) have two separate instances.

PR #6818, opened 2017-11-22
Update the support app queue metric

cbaines:
As it doesn't contain recent data. The replacement metric seems
consistent with other apps, and does contain up to date data.

### DIFF
--- a/modules/monitoring/manifests/checks.pp
+++ b/modules/monitoring/manifests/checks.pp
@@ -157,7 +157,7 @@ class monitoring::checks (
 
   if $enable_support_check {
     icinga::check::graphite { 'check_support_default_queue_size':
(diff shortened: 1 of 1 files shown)


maxgds:
Just an FYI: There's a reference to this PR on an icinga alert that keeps recurring in production. It seems to keep appearing and settling so probably isn't anything to worry about, but I thought it worth a mention. 
screen shot 2017-12-14 at 09 18 30

cbaines:
> Just an FYI: There's a reference to this PR on an icinga alert that keeps recurring in production. It seems to keep appearing and settling so probably isn't anything to worry about, but I thought it worth a mention.

Sorry, didn't see this comment at the time. As it happens, I did put in a "fix" for this, which reduce the number of false positives https://github.com/alphagov/govuk-puppet/pull/6963

PR #6822, opened 2017-11-22
Refactor how we include apps on hosts

surminus:
This change rather large refactor on how we include app classes on hosts has come about from another change in which we wish to automate the deployment of applications based upon the node class.

Having the source of truth in two places would make it more complicated, and liable to mistakes. Equally, having all the apps listed under their node class in hieradata makes it easier to understand what is included where, rather than being split into separate hiera files based upon class.

The logic should only be applied if we specify, and the node class will need to match the hostname or aws_migration fact depending on the environment.


modules/govuk/manifests/node/s_base.pp
```diff
@@ -34,8 +34,19 @@
     include hosts
   }
 
-  $app_classes = regsubst($apps, '^', 'govuk::apps::')
-  include $app_classes
+  if $::aws_migration {
+    $node_class = $::aws_migration
+  } else {
+    $node_class = regsubst($::fqdn_short, '-.*$', "")
```
afda16: Should we use the `govuk_node_class` ?

surminus: I chose not to use this because this is used within the main manifest and thought it might be misleading.

PR #6823, opened 2017-11-22
Don't hardcode `publishing.service.gov.uk` in govuk_uptime_collector

issyl0:
- Adjust this for AWS, with a new parameter to run the script with (defaults to
  `false`) and some hieradata.

### DIFF
--- a/modules/monitoring/files/usr/local/bin/govuk_uptime_collector
+++ b/modules/monitoring/files/usr/local/bin/govuk_uptime_collector
@@ -9,10 +9,11 @@ require "time"
 require "statsd"
 
 class Collector
-  def initialize(service_name, environment)
+  def initialize(service_name, environment, aws)
     @statsd = Statsd.new("127.0.0.1")
     @service_name = service_name
     @environment = environment
+    @aws = aws
   end
 
   def call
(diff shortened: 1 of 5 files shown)


modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -58,10 +63,11 @@ def main
   threads = []
 
   environment = ARGV[0]
+  aws = ARGV[1]
```
thomasleese: Suspect this will have to be `ARGV[1] == 'true' ? true : false` otherwise the string `"false"` will be truthy.

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -58,10 +63,11 @@ def main
...
 
   ARGV[1..-1].each do |service_name|
```
thomasleese: This will have to be `ARGV[2..-1]` now.

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
issyl0: Will it ever be a string? It's in Puppet and Hieradata as a boolean. But OK, I've amended.

thomasleese: Yeah, annoyingly it will be a string as it comes from https://github.com/alphagov/govuk-puppet/pull/6823/files#diff-3c14b0dbebef6ce25a9e337b66b257fdR9 which passes it in as a normal command line argument.

modules/monitoring/files/usr/local/bin/govuk_uptime_collector
```diff
@@ -9,10 +9,11 @@ require "time"
 require "statsd"
 
 class Collector
-  def initialize(service_name, environment)
+  def initialize(service_name, environment, aws=false)
```
issyl0: Given `false` is set by default in https://github.com/alphagov/govuk-puppet/pull/6823/files#diff-ef3bfb7435b4b29de7e6bf80fd302bd3R16, we don't need to set this `=false` here, do we?

thomasleese: Yep, that's true. In fact probably worth removing it.

issyl0: My first attempt at doing this was before I realised I had to edit the Puppet, so it's a hangover from that.

PR #6824, opened 2017-11-22
On AWS, use the internal domain for MapIt checks

issyl0:
- We were seeing alerts for "not responding to postcode query - unable to open TCP socket".

### DIFF
--- a/modules/icinga/templates/check_mapit.cfg.erb
+++ b/modules/icinga/templates/check_mapit.cfg.erb
@@ -1,5 +1,10 @@
 define command {
   command_name check_mapit
+  <%- if scope.lookupvar('::aws_migration') %>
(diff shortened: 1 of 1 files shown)


issyl0:
Before:

```
/usr/lib/nagios/plugins/check_http -H mapit.STACKNAME.integration.govuk.digital -u /postcode/W54XA -w1 -c2 -s 'E14000676' --ssl
Name or service not known
HTTP CRITICAL - Unable to open TCP socket
```

After:

```
/usr/lib/nagios/plugins/check_http -H mapit.STACKNAME.integration.govuk-internal.digital -u /postcode/W54XA -w1 -c2 -s 'E14000676' --ssl
HTTP OK: HTTP/1.1 200 OK - 2190 bytes in 0.077 second response time |time=0.076645s;1.000000;2.000000;0.000000 size=2190B;;;0```

PR #6826, opened 2017-11-22
Call initializeParameters in buildProject

h-lame:
This makes sure that build parameters are always available as environment
variables, and have their default values set if they've not been provided.
This definitely feels like something that should be handled by Jenkins
already, but it's not.

To avoid having to set separate default values as arugments we iterate
over the existing parameterDefinitions (which may be augmented by
extraParameters).  The issue here is that we can have two separate types
of parameter definitions:

1. those defined as simple map objects, for example:
    
        [$class: 'StringParameterDefinition',
          name: 'SCHEMA_BRANCH',
          defaultValue: 'deployed-to-production',
          description: 'The branch of govuk-content-schemas to test against']

2. those defined using the DSL, for example:

        stringParam(
          name: 'SCHEMA_BRANCH',
          defaultValue: 'deployed-to-production',
          description: 'The branch of govuk-content-schemas to test against']
        )

Those defined in the first way can have their details read via simple
accessors, whereas those defined the second way need to be converted
into a map first as they are (at this point anyway) objects of type
`org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable`
(see: https://github.com/jenkinsci/structs-plugin/blob/master/plugin/src/main/java/org/jenkinsci/plugins/structs/describable/UninstantiatedDescribable.java).

This also requires that we approve using the `toMap` method of
`org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable`
in [the script approval area of jenkins](https://jenkins.io/doc/book/managing/script-approval/).

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -115,6 +115,20 @@ def buildProject(Map options = [:]) {
     [$class: 'ParametersDefinitionProperty', parameterDefinitions: parameterDefinitions],
   ])
 
+  def defaultParameterValuesMap = [:]
+  parameterDefinitions.each {
+    // to handle params defined with the xxxParam(...) DSL instead of
+    // [$class: ... ] style because we can't call .name / .defaultValue
+    // on them directly
+    if (it.class == org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable) {
+      def mapVersionOfIt = it.toMap()
+      defaultParameterValuesMap[mapVersionOfIt.name] = mapVersionOfIt.defaultValue
+    } else {
+      defaultParameterValuesMap[it.name] = it.defaultValue
+    }
+  }
+  initializeParameters(defaultParameterValuesMap)
+
   try {
     if (!isAllowedBranchBuild(env.BRANCH_NAME)) {
       return


h-lame:
This is in response to [a failure on publishing-api](https://ci.integration.publishing.service.gov.uk/job/publishing-api/job/use_rabbitmq_envvar/3/console) where new build branches failed on first build because env vars defined weren't set.  We've used the jenkins replay feature to [make this build pass using this change](https://ci.integration.publishing.service.gov.uk/job/publishing-api/job/use_rabbitmq_envvar/15/console).  Note that the publishing-api used to call `initializeParameters` before it was moved to use `buildProject`, but we decided making all projects do this wasn't a bad idea.

See: https://github.com/alphagov/publishing-api/pull/1065#pullrequestreview-78434865 for discussion

h-lame:
It's not clear to me if we can make changes to the script approval config of jenkins via puppet which means this is something we'd have to do manually in all environments 😞

modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -115,6 +115,20 @@ def buildProject(Map options = [:]) {
     [$class: 'ParametersDefinitionProperty', parameterDefinitions: parameterDefinitions],
   ])
 
+  def defaultParameterValuesMap = [:]
+  parameterDefinitions.each {
+    // to handle params defined with the xxxParam(...) DSL instead of
+    // [$class: ... ] style because we can't call .name / .defaultValue
+    // on them directly
+    if (it.class == org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable) {
+      def mapVersionOfIt = it.toMap()
```
thomasleese: Instead of using `toMap` (if it means we have to enable something in Jenkins), can we use [`getArguments`](http://javadoc.jenkins-ci.org/structs/org/jenkinsci/plugins/structs/describable/UninstantiatedDescribable.html#getArguments--) or [`instantiate`](http://javadoc.jenkins-ci.org/structs/org/jenkinsci/plugins/structs/describable/UninstantiatedDescribable.html#instantiate--) ?

h-lame: Oh, good idea, I'll try it and see what happens, but I think the groovy sandbox is very limited.

h-lame: Alas, same problem with `getArguments`:

[`org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use method org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable getArguments`](https://ci.integration.publishing.service.gov.uk/job/publishing-api/job/use_rabbitmq_envvar/16/console)

and `instantiate`:

[`org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use method org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable instantiate`](https://ci.integration.publishing.service.gov.uk/job/publishing-api/job/use_rabbitmq_envvar/17/console).

Shame.

h-lame:
Merging this to help unblock the publishing-api builds, will investigate puppetising the script approval config in a separate PR.  The impact of of it being manual for now seems low as we don't have that many CI jenkins servers.

PR #6827, opened 2017-11-22
[AWS] Enable authenticating proxy on draft-origin

issyl0:
- This was set to the wrong node class (`s_cache`, not `s_draft_cache`), so the
  unauthenticated redirect to Signon wasn't working (and causing Smokey to fail).

### DIFF
--- a/hieradata_aws/class/draft_cache.yaml
+++ b/hieradata_aws/class/draft_cache.yaml
@@ -18,7 +18,7 @@ govuk::apps::router_api::vhost: 'draft-router-api'
 govuk::node::s_base::apps:
   - router_api
 
-govuk::node::s_cache::enable_authenticating_proxy: true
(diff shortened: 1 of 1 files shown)


surminus:
I think there may have been an issue with this. Poor review from myself.

The parameter only confusingly lives in `govuk::node::s_cache`, yet is only enabled on `govuk::node::s_draft_cache`, I think because it inherits the code from `cache`.

We will probably have to revert this. I had to terminate an instance because it wasn't responding, and suspect this may have been the issue, or we should look to see if we can refactor the code to make it less confusing.

issyl0:
That is confusing. I'd expect the class name to match the one it's being applied to in this case, but apparently not. Thanks for investigating and making it so it doesn't break things with #6828 - though I'm not sure I quite understand there either. I'll revert this PR though in favour of that.

PR #6828, opened 2017-11-22
Make enable_authenticating_proxy parameter clearer

surminus:
When checking this I found it extremely confusing. We inherit a class, and set a parameter in the hieradata for a different node type.

It would be clearer to call the class so we can understand what is going on here.

This change will make it more obvious that authenticating proxy is disabled on s_cache, but enabled on s_draft_cache.

### DIFF
--- a/modules/govuk/manifests/node/s_draft_cache.pp
+++ b/modules/govuk/manifests/node/s_draft_cache.pp
@@ -5,8 +5,12 @@
 #
 # Includes `govuk::node::s_cache`.
 #
-class govuk::node::s_draft_cache() {
-  include govuk::node::s_cache
+class govuk::node::s_draft_cache(
+  $enable_authenticating_proxy = true,
+) {
+  class { 'govuk::node::s_cache':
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/node/s_draft_cache.pp
```diff
@@ -5,8 +5,12 @@
 #
 # Includes `govuk::node::s_cache`.
 #
-class govuk::node::s_draft_cache() {
-  include govuk::node::s_cache
+class govuk::node::s_draft_cache(
+  $enable_authenticating_proxy = true,
+) {
+  class { 'govuk::node::s_cache':
+    enable_authenticating_proxy => $enable_authenticating_proxy,
```
issyl0: Is this the only thing we need from `govuk::node::s_cache`?

issyl0: Ah, `class` = `include` with no specific params. Thanks!

PR #6830, opened 2017-11-23
Add Slack notification to CDN dictionary deployment

suzannehamilton:
Notify the govuk-deploy channel when the Update_CDN_Dictionaries Jenkins job finishes.

The CDN deployment job already does this, but it's useful to know about dictionary updates too because they could also cause errors if misconfigured.

### DIFF
--- a/modules/govuk_jenkins/manifests/job/update_cdn_dictionaries.pp
+++ b/modules/govuk_jenkins/manifests/job/update_cdn_dictionaries.pp
@@ -2,7 +2,14 @@
 #
 # Create a file on disk that can be parsed by jenkins-job-builder
 #
-class govuk_jenkins::job::update_cdn_dictionaries {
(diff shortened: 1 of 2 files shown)


suzannehamilton:
I've just noticed that the deploy_cdn job is configured to post to 2ndline:
https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/manifests/job/deploy_cdn.pp#L10

But it actually posts to govuk-deploy. I can't spot anything in puppet which overrides the channel setting, and the Jenkins job config doesn't show the details of the Slack configuration. Anyone know what's going on? 🤔

suzannehamilton:
@deanwilson I've updated the config as we discussed, with the room configured in the properties section. Let's see if this posts updates to the correct channel.

PR #6832, opened 2017-11-23
Add Link Checker API secret token env var

camillavk:
This commit adds an env var for the Link Checker API secret token for Whitehall

Link Checker API integration: https://github.com/alphagov/whitehall/pull/3557

### DIFF
--- a/modules/govuk/manifests/apps/whitehall.pp
+++ b/modules/govuk/manifests/apps/whitehall.pp
@@ -101,6 +101,10 @@
 #   email to the content second line group in GDS.
 #   Default: undef
 #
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/whitehall.pp
```diff
@@ -133,6 +137,7 @@
   $vhost_protected,
   $jwt_auth_secret = undef,
   $co_nss_watchkeeper_email_address = undef,
+  $link_checker_api_secret_token = undef,
```
thomasleese: You'll also need to add it to the `govuk::app::envvar` section below.

PR #6834, opened 2017-11-23
[WIP] Introduce statsd repeating into graphite.cluster

kevindew:
We have identified a problem with our statsd configuration that affects statistics that can be sent with the same key from multiple hosts. Graphite will only store the last value sent within a window (this is currently configured to 5 seconds). If two hosts send a statistic with the same key within this window only the last one will be recorded and may mean for some rapidly changing statistics our data represents a 1/2 or 1/3 of the actual value.

This PR introduces a plugin to statsd - statsd_smart_repeater - which is used to send the aggregated statistics of a host to another statsd instance which then sends this data to graphite. This second statsd instance will do the aggregation between hosts.

The scope of this PR is to introduce this functionality in addition to current behaviour (with all the stats behind a namespace of smart_repeater) to allow comparison with existing behaviour.

The script for statsd_smart_repeater is from https://github.com/lucidsoftware/statsd-smart-repeater this was used instead of the repeater packaged with statsd as that does not support
aggregating stats. A slightly more popular alternative seems to be https://github.com/wanelo/gossip_girl however this was rejected due to it not supporting TCP (and it's a little less feature rich e.g. namespaces).

There's a couple of things I'm not sure about with this PR though

- [ ] Whether I'd need to open up statsd port on graphite.cluster
- [ ] Whether repeating all statsd stats in graphite for comparison may cause some sort of disk space catasrophe

### DIFF
--- a/hieradata/class/graphite.yaml
+++ b/hieradata/class/graphite.yaml
@@ -6,6 +6,8 @@ govuk_safe_to_reboot::reason: 'We lose stats and some alerting when this machine
 icinga::client::checks::disk_time_warn: 280 # milliseconds
 icinga::client::checks::disk_time_critical: 400 # milliseconds
 
+statsd::use_smart_repeater: false
+
 lv:
   data:
     pv:
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1511,6 +1511,9 @@ sidekiq_port: '6379'
 
 ssh::config::allow_users_enable: true
 
+statsd::use_smart_repeater: true
+statsd::graphite_hostname: "graphite.cluster"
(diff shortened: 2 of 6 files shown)


kevindew:
@afda16 I was talking to @surminus about this PR on Thursday and she suggested it might be good if I worked with you on this one a bit, as she said you were good with statistics. Are you able to help at all?

kevindew:
I'm closing this in favour of using the hostname namespaced approach you get by using [GOVUK_STATSD_PREFIX](https://github.com/alphagov/govuk-puppet/blob/cfa5460d709423115ea93ea57c9ac8e4f612f25f/modules/govuk/manifests/app/config.pp#L124-L125)

PR #6837, opened 2017-11-24
[AWS] Fix rummager checks

surminus:
In AWS we can don't need to go in via the cache machines, we can just call the Rummager API by going to the DNS record for the Search machines.


surminus:
It already works in Carrenza, and the change I made was specifically under a conditional for AWS. I tested the URL in AWS and it works.

stephenharker:
Ah ok sorry, missed that.

stephenharker:
How slow IS Travis anyway?

PR #6838, opened 2017-11-24
Update docker compose version

kevindew:
This is to update docker-compose so we can make use of all the [shiny features](https://docs.docker.com/compose/compose-file/#healthcheck) that are available now we are using an up-to-date docker version. To update the docker-compose version we merely have to specify the version number in the include.

However once we update past Docker 1.9.0 we hit an issue where the dependency for [docker-py conflicts with docker-compose](https://github.com/docker/compose/issues/4569) and then we hit an even thornier issue when we try use package to do `pip install docker` as puppet won't allow us to have another package named `docker`. This uses exec to get around that constraint.

Couple of things I need to check before merging this:

- Is there an example of collectd docker output? it'd be useful to check this before and after doing a test deploy to integration as I'm not sure how to test it out locally.
- Do I need any extra checks in the exec pip install as per: https://github.com/alphagov/govuk-puppet/blob/master/modules/monitoring/manifests/uptime_collector.pp#L18-L24  ?

### DIFF
--- a/modules/collectd/manifests/plugin/docker.pp
+++ b/modules/collectd/manifests/plugin/docker.pp
@@ -6,11 +6,23 @@ class collectd::plugin::docker {
 
   $dependencies = [
     'py-dateutil',
-    'docker-py',
   ]
 
   package { $dependencies:
     ensure   => 'present',
+    provider => 'pip',
+  }
+
+  # This has to be installed via exec rather than a package as the namespace
+  # conflicts with the different (but same name) docker package. This issue
+  # seems to be fixed in puppet 4: https://tickets.puppetlabs.com/browse/PUP-1073
+  exec { 'pip install docker':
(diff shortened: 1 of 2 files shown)


afda16:
Maybe use 'unless' , something like `pip list | grep -w docker-py`

kevindew:
> Maybe use 'unless' , something like `pip list | grep -w docker-py`

@afda16  Yeah I wondered about that but I didn't know one where only one thing matches e.g. 

```
vm  govuk pip list | grep -w docker
docker (2.6.1)
docker-pycreds (0.2.1)
```

It also seemed pretty unconsequential to run pip install again:

```
vm  govuk pip install docker
Requirement already satisfied (use --upgrade to upgrade): docker in /usr/local/lib/python2.7/dist-packages
Cleaning up...
```

kevindew:
Everything seems to have gone well testing this so I'm going to merge it :crossed_fingers:

PR #6840, opened 2017-11-24
[AWS] Add STS headers

surminus:
This is normally included when we include SSL, but this is offloaded to ELBs in AWS land.


surminus:
While this fixes the alert, we're wondering if we need to set this header at all, and if we should make the check more encompassing. We have created two stories to investigate:

https://trello.com/c/db8Zp1h7/236-investigate-http-https-redirects
https://trello.com/c/11W8aH3k/259-investigate-if-the-hsts-icinga-check-is-suitable

I have acknowledged the alert, and am closing this PR for now.

PR #6846, opened 2017-11-24
Fix ENVIRONMENT variable ordering for search jenkins jobs

dwhenry:
search_fetch_analytics_data needs to set the var before calling the script
search_reindex_with_new_schema need to set the var after the rake task is called

https://trello.com/c/3VkbdYrB/456-test-nightly-popularity-rebuild-on-integration-and-staging


dwhenry:
@cbaines Types have been fixed (I did it in the PR then forgot to update the commit). I have added a comment about the extra ENV variable. :)

PR #6849, opened 2017-11-24
[AWS] Update jenkins-cli and include with app machines

surminus:
This would allow a machine to interact with Jenkins using the jenkins-cli tool. In AWS,
we wish for machines to trigger a job which deploys all apps that they should run
when they are first created at boot.

Including the jenkins-cli tool allows an easy way to trigger a job, and already
includes the credentials neccessary to do this.

We may want to create a specific user for this however, as the current API user
has quite far reaching permissions in Jenkins.


surminus:
After discussing this with @timmow we've decided that the permissions would be too risky to implement this change.

Instead we may be able to set an authentication token for the job which can be triggered by hitting a Jenkins URL.

PR #6850, opened 2017-11-24
Grafana dashboard for Email Alert API Metrics

kevindew:
Trello: https://trello.com/c/u20s96bS/353-set-up-statistics-monitoring-to-inform-product-decisions

This dashboard is to assist the Email team in understanding key metrics
that affect performance of the Email Alert API as it grows in scope.
This iteration of this dashboard only contains statistics that are
related with the Notify integration, however the intention is that this
will continue into more data that cover wider concerns.

This dashboard was built through the Grafana UI and this file was
created by exporting that data.

When populated the dashboard looks like this:
screen shot 2017-11-24 at 18 15 55

### DIFF
--- a/modules/grafana/files/dashboards/email_alert_api.json
+++ b/modules/grafana/files/dashboards/email_alert_api.json
@@ -1,0 +1,466 @@
+{
+  "annotations": {
+    "list": []
+  },
+  "editable": true,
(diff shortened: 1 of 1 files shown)


thomasleese:
Looks awesome!

PR #6852, opened 2017-11-27
[AWS] Don't run router in a container

issyl0:
- This has been a snowflake. Remove the `govuk_containers` router manifest.
  We're going to revert to doing this normally, apart from pulling the binary
  from S3.

Reverts #6259 apart from the hieradata.
https://trello.com/c/IbnMEmul/793-router-move-away-from-container-deployment


surminus:
I don't think it's worth removing `modules/govuk_containers/manifests/apps/router.pp` entirely, as it makes a good backup way to run Router if we want to.

issyl0:
I've added it back, but it won't run unless we reinstate the `include` on `s_cache`.

PR #6853, opened 2017-11-27
[DO NOT MERGE] Enable sysdig everywhere

surminus:
Sysdig is an extremely useful tool for diagnosing any OS level issues.

It is something that should be a default tool at our disposable in our systems. A reboot would be required to enable it, but we should organically let it be enabled over time as instances are rebooted.

https://trello.com/c/3y6WEkin/821-check-sysdig-repository-and-add-back-in


surminus:
There seems to be an issue with the snapshot on `apt-1`. This will have to be resolved before proceeding with this.

It's also worth bearing in mind that there is a race condition with enabling this on `apt-1` before the repository works.

surminus:
I'm going to close this for now and we can come back to it.

PR #6855, opened 2017-11-27
Include trigger_data_sync_complete_aws on Carrenza integration for AWS

issyl0:


modules/govuk_jenkins/templates/jobs/trigger_data_sync_complete_aws.yaml.erb
```diff
@@ -0,0 +1,30 @@
...
+    project-type: freestyle
+    description: "This job triggers the Data_Sync_Complete jobs on a remote Jenkins machine in AWS"
+    logrotate:
+        numToKeep: 10
+    builders:
+        - shell: |
+            # Trigger on AWS
+            echo "Triggering post-sync job on $HOSTNAME in AWS"
+            CRUMB=$(curl https://<%= @jenkins_integration_aws_api_user %>:<%= @jenkins_integration_aws_api_password %>@<%= @aws_deploy_url %>/crumbIssuer/api/json | jq --raw-output '. | .crumb')
+            curl -f -H "Jenkins-Crumb:$CRUMB" -XPOST https://<%= @jenkins_integration_aws_api_user %>:<%= @jenkins_integration_aws_api_password %>@<%= @aws_deploy_url %>/job/Data_Sync_Complete/build --data-urlencode json="{}"
```
surminus: I think this will work, as it will pull the parameters `@jenkins_integration_aws_api_user` and `@jenkins_integration_aws_api_password` from another manifest, but ideally we should include them in the manifest that uses the template.

issyl0: Added!

PR #6860, opened 2017-11-28
Proxy email-status-updates.* to email-alert-api/status_updates

tuzz:
https://trello.com/c/7pDM2swr/372-proxy-requests-from-a-public-url-to-the-statusupdates-endpoint

To do:

- ensure dns changes have propogated
- document the puppet module
- re-use email-alert-api's port
- add bearer token checking to /status_updates
- (optional) validate request for presence of fields before enqueueing job?


modules/govuk/manifests/apps/email_status_updates.pp
```diff
@@ -0,0 +1,30 @@
...
+
+        ssl_certificate /etc/nginx/ssl/email-alert-api.${app_domain}.crt;
+        ssl_certificate_key /etc/nginx/ssl/email-alert-api.${app_domain}.key;
+        include /etc/nginx/ssl.conf;
+
+        add_header Access-Control-Allow-Origin *;
+        add_header Access-Control-Allow-Methods POST;
+
+        location / {
+          proxy_pass http://email-status-updates.${app_domain}-proxy/status_updates;
```
rubenarakelyan: Should this be proxying to `email-alert-api`?

tuzz: It is, but it's a bit confusing.

The `-proxy` upstream is defined above (lines 7 through 9), which sends traffic to `locahost:3088`.

We might be able to re-use `email-alert-api` as the upstreams as it's defined by `email_alert_api.pp`, but we'd be slightly coupled across Nginx config.

rubenarakelyan: Ah right yes, I always get confused by these things that look like URLs but are actually references to a proxy config...

kevindew:
I'm closing this in favour of https://github.com/alphagov/govuk-puppet/pull/6876 as discussed

PR #6861, opened 2017-11-28
Collect sidekiq logs consistently

sihugh:
I would like to collect logs for Publisher's workers, but feel it might be
useful to collect them for everything (and hopefully get closer to a consistent
logging experience across our apps).

Some apps log to sidekiq.json.log and some to sidekiq.log but it all seems to
be in json format so I think we should be OK.

If there are multiple files that the path matches then Filebeat will
[set up a harvester for each](https://www.elastic.co/guide/en/beats/filebeat/master/configuration-filebeat-options.html#prospector-paths)

I think it follows that if it doesn't find anything the, that's OK too.

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -370,6 +370,12 @@ define govuk::app (
     fields => {'application' => $title},
   }
 
+  @filebeat::prospector { "${title}_sidekiq_json_log":
+    paths  => ["/var/apps/${title}/log/sidekiq*.log"],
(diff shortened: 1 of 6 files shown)


tijmenb:
Is this something we can push up so every app catches the sidekiq logs by default? I'm wondering how content-taggers logs get into kibana (or if they even do).

sihugh:
I've updated this now.  My update to publisher.pp looks a little incongruous now that I'm not actually doing anything with it now :-/

PR #6871, opened 2017-12-01
Link Rummager size change alert to grafana graphs

dwhenry:
When the index size change alert is triggered these graphs should
help to determine the size/impact of the change. In addition they
should help identify when the alert is triggered as a result of a
historic change (i.e. the data from a week ago had a spike)

https://trello.com/c/MlnXML1q/480-better-documentation-for-index-size-alerts

### DIFF
--- a/modules/grafana/manifests/dashboards.pp
+++ b/modules/grafana/manifests/dashboards.pp
@@ -24,6 +24,8 @@ class grafana::dashboards (
 
   $app_domain_metrics = regsubst($app_domain, '\.', '_', 'G')
 
+  $index_names = ['govuk', 'mainstream', 'government', 'detailed']
+
   file { $dashboard_directory:
     ensure  => directory,
     recurse => true,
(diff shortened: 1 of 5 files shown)


modules/grafana/manifests/dashboards/rummager_index_size_dashboard.pp
```diff
@@ -0,0 +1,21 @@
...
+  $dashboard_directory = '/etc/grafana/dashboards',
+  $index_names = ['govuk', 'mainstream', 'government', 'detailed'],
+) {
+
+  file {
+    "${dashboard_directory}/rummager_index_size.json":
+    notify  => Service['grafana-server'],
+    content => template('grafana/dashboards/rummager_index_size.json.erb');
+  }
+}
```
cbaines: I don't think you should need this class, could you add it to this list? https://github.com/alphagov/govuk-puppet/blob/master/modules/grafana/manifests/dashboards.pp#L33-L41

cbaines:
I was changing a dashboard [1] when I was on 2ndline, and that uses templating to show data for all/any of the indexes, which might remove the need for doing any dashboard creation in ERB?

1: https://grafana.publishing.service.gov.uk/dashboard/db/gov-uk-search-indexes?refresh=1m&orgId=1&var-index=All

cbaines:
Ah, I've just noticed that this involves a 2nd index size related dashboard, so my comments might not apply.

modules/grafana/manifests/dashboards/rummager_index_size_dashboard.pp
dwhenry: Good call, have simplified this now.

PR #6875, opened 2017-12-01
Add basic spec tests for the jenkins jobs

deanwilson:
We should at least check they compile


surminus:
Is there anyway we could do this more automagically? We always want to test against jobs within `govuk_jenkins::job`, so is it possible to look across every job in that class?

deanwilson:
We can make a clever spec that loads all the classes and dynamically creates tests. Depends on if you'd like explicit and lots of boilerplate or implicit and more code generation

PR #6876, opened 2017-12-01
Enable email-alert-api-public proxy

kevindew:
Trello: https://trello.com/c/7pDM2swr/372-proxy-requests-from-a-public-url-to-the-statusupdates-endpoint

This adds a hostname, email-alert-api-public, that is exposed
publicly which is used as a means for external services to route
requests in to email-alert-api.

We decided to go with the unconventional name of email-alert-api-public
as it wasn't clear about all the usages of this hostname might be and
this seemed to be the clearest about the intent of it.

It proxies the endpoint status-updates and will accept any subroutes
that start with that segment. Any other requests will receive a 404.

This is currently only enabled on integration until we have completed
authentication work.

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -33,6 +33,7 @@ govuk::apps::content_performance_manager::feature_auditing_theme_filtering: true
 govuk::apps::content_store::performance_platform_big_screen_view_url: 'https://performance-platform-big-screen-view-staging.cloudapps.digital'
 govuk::apps::content_store::performance_platform_spotlight_url: 'https://performance-platform-spotlight-staging.cloudapps.digital'
 govuk::apps::email_alert_api::allow_govdelivery_topic_syncing: true
+govuk::apps::email_alert_api::enable_public_proxy: true
 govuk::apps::email_alert_api::govdelivery_account_code: 'UKGOVUKDUP'
 govuk::apps::email_alert_api::govdelivery_hostname: 'stage-api.govdelivery.com'
 govuk::apps::email_alert_api::govdelivery_public_hostname: 'stage-public.govdelivery.com'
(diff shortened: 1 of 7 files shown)


hieradata/integration.yaml
```diff
@@ -33,10 +33,11 @@ govuk::apps::content_performance_manager::feature_auditing_theme_filtering: true
 govuk::apps::content_store::performance_platform_big_screen_view_url: 'https://performance-platform-big-screen-view-staging.cloudapps.digital'
 govuk::apps::content_store::performance_platform_spotlight_url: 'https://performance-platform-spotlight-staging.cloudapps.digital'
 govuk::apps::email_alert_api::allow_govdelivery_topic_syncing: true
+govuk::apps::email_alert_api::enable_callbacks_proxy: true
```
rubenarakelyan: Should this be `enable_public_proxy`?

kevindew: Dang! missed that one - thanks 👍

kevindew:
@thomasleese would you mind reviewing this again? It's changed a fair bit since you ticked it. Also has https://github.com/alphagov/govuk-secrets/pull/136

modules/govuk/manifests/apps/email_alert_api.pp
```diff
@@ -55,15 +63,16 @@
 #   Number of requests per seconds that the notify delivery worker
 #   is allowed to make (currently 50 requests/s / 6 workers = 8 (ish)
 #
-# [*secret_key_base*]
```
thomasleese: Did you mean to lose the documentation for the `secret_key_base` field?

kevindew: I did not - thanks for catching. I'm getting sloppy 😟

PR #6879, opened 2017-12-04
Add one spec for all of the Jenkins jobs manifests

issyl0:
- In PR 6875, it was decided that we should make one spec test for all
  the Jenkins jobs, rather than forgetting to add new spec tests for
  new jobs, and rather than increasing the amount of Puppet code to do
  exactly the same tests for every job.
- This is an attempt at that. `Dir.entries` ouputs the filename with the
  extension, hence `split` then being used to get rid of the `.pp`
  extension.

The specs themselves are extremely unhappy, but the same happened with #6875, so ¯\_(ツ)_/¯.

https://trello.com/c/KeLujIgw/855-add-specs-for-jenkins-jobs


surminus:
Ouch, shows how much we needed the tests!

issyl0:
Closing in favour of #6959.

PR #6881, opened 2017-12-04
Rename singular `job` to plural `jobs` everywhere for govuk_jenkins

issyl0:
- This was getting confusing with the templates and manifests "job(s)" directories when we tried to make a test that tested all of the `govuk_jenkins` manifests automatically, so update `govuk_jenkins::job` to `govuk_jenkins::jobs` everywhere for consistency.

There's another PR for govuk-secrets incoming.

The first commit is from #6879, 'cause I branched off there. That's why this is a DISCUSS PR.


issyl0:
The secrets PR is at https://github.com/alphagov/govuk-secrets/pull/146.

PR #6884, opened 2017-12-04
Overhaul server_name lookups in AWS

surminus:
We normally statically set the server_name paramater in our NGINX config based upon the globally defined "app_domain" parameter.

This is problematic in AWS, particularly during our migration period, as we are going to use several different DNS records:

 - The external AWS top level DNS name (integration.govuk.digital)
 - The external AWS stack specific name (stack.integration.govuk.digital)
 - The internal AWS stack specific name (stack.integration.govuk-internal.digital)
 - The main service domain name (integration.publishing.service.gov.uk)

Having to work all these into our code was going to be problematic and require lots of logic that does not fit into the current way that NGINX is configured with Puppet.

We have made the decision to loosen the requirements, and only apply this to AWS.

Eventually most configuration is going to be applied on the Application Load Balancers so the server_name directive will be less of a thing in The New World.

This enables us to test both the service domain and internal/external names to help us migrate smoothly when the time comes.

### DIFF
--- a/modules/router/templates/assets_origin.conf.erb
+++ b/modules/router/templates/assets_origin.conf.erb
@@ -1,5 +1,9 @@
 server {
+  <%- if scope.lookupvar('::aws_migration') %>
+  server_name <%= @vhost_name -%> <%= @vhost_name -%>.* <%= @vhost_aliases.join(' ') -%>;
+  <%- else %>
   server_name <%= @vhost_name -%> <%= @vhost_aliases.join(' ') -%>;
+  <%- end %>
   listen 80;
 <% if @enable_ssl -%>
   rewrite ^/(.*) https://$host/$1 permanent;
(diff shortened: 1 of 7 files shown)


modules/router/templates/assets_origin.conf.erb
```diff
@@ -6,8 +6,11 @@ server {
 }
 
 server {
+  <%- if scope.lookupvar('::aws_migration') %>
+  server_name <%= @vhost_name -%> <%= @vhost_name -%>.* <%= @vhost_aliases.join(' ') -%>;
+  <%- else %>
   server_name <%= @vhost_name -%> <%= @vhost_aliases.join(' ') -%>;
-
+  <%- end %>
```
afda16: You'll be removing a white line in the current configs

surminus: Hmm, these conditionals are a bit strange anyway. I think they may do stuff I don't expect.

surminus: I've updated this so the conditional is in the right place

PR #6885, opened 2017-12-04
Add default healthcheck for cache

surminus:
Related to 9996b3781963c715192b9d34a99a2009607c9d20

The cache machines have their own default config, so also change it here.

### DIFF
--- a/modules/router/templates/default-vhost.conf.erb
+++ b/modules/router/templates/default-vhost.conf.erb
@@ -6,6 +6,12 @@ server {
     proxy_pass http://varnish;
   }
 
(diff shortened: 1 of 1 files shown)


modules/router/templates/default-vhost.conf.erb
```diff
@@ -5,7 +5,12 @@ server {
     # This is the varnish upstream defined in the www.gov.uk vhost
     proxy_pass http://varnish;
   }
-
```
afda16: This line will be removed, and all Nginx processes reloaded

surminus: Updated. Good point about everything reloading.

PR #6895, opened 2017-12-06
Setup statsd to not send previous values for gauges

cbaines:
Currently, when statsd receives a value for a gauge, it will continue
to send this each time metrics are flushed to graphite. As many of
the gauges used within GOV.UK could end up being sent from multiple
machines, this could be causing some contention over metrics in
graphite, between different statsd instances on different machines,
possibly with different values.

To avoid this, setup statsd not to repeat the last value for
gauges. The empty regions can then be filled in by using the
keepLastValue function in graphite.

### DIFF
--- a/modules/statsd/templates/etc/statsd.conf.erb
+++ b/modules/statsd/templates/etc/statsd.conf.erb
@@ -44,4 +44,5 @@ Optional Variables:
 , graphiteHost: "<%= @graphite_hostname -%>"
 , port: 8125
 , flushInterval: 5000
+, deleteGauges: true
 }


cbaines:
> This seems reasonable to me 👍 Just suggest caution in checking it won't break any existing monitoring

Good point, I'll keep an eye out :)

surminus:
> don't send values to graphite for inactive gauges, as opposed to sending the previous value

Seems sensible to me.

cbaines:
> Seems sensible to me.

Awesome, let's try this :shipit:

PR #6898, opened 2017-12-06
Configure staging email-alert-api to use pseudo email sending provider

thomasleese:
This is so that we can do loading testing in staging.

[Trello Card](https://trello.com/c/7zoZKZkS/281-agree-approach-for-load-testing-with-notify)


deanwilson:
Can we get a few words about what "PSEUDO" is please?

deanwilson:
That's perfect, thank you.

PR #6899, opened 2017-12-06
Add an ssh config for jenkins

timmow:
That set StrictHostKeyChecking to no.

This is similar to
https://github.com/alphagov/govuk-app-deployment/pull/212, but slightly
different. App deployments also run rysnc as part of the deployments,
- this does not use the capistrano ssh config, but uses the ssh config
files on the system

A new machine coming up requires puppet to run to add the known host to
`/etc/ssh/ssh_known_hosts` - if this is not there the deploy will
currently fail because there is no `usr/bin/ssh-askpass` program
available - it would fail requiring interactive confirmation that we
trust the host key. If we set StrictHostKeyChecking to yes, it would
fail until puppet is run.

As we are ignoring host keys for capistrano deployments, although we
could fix this rsync issue by making sure puppet was run before
deployments kick off, it wouldn't give us any benefits over simply
ignoring the host key verification, which we do for capistrano
operations in the deployment anyway.

### DIFF
--- a/modules/govuk_jenkins/files/ssh-config
+++ b/modules/govuk_jenkins/files/ssh-config
@@ -1,0 +1,2 @@
+Host *
+  StrictHostKeyChecking no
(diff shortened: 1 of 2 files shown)


stephenharker:
Although it would be nice if we didn't need to do this :/

PR #6900, opened 2017-12-06
Handle intermittent gauge metrics in monitoring targets

cbaines:
Now that statsd isn't repeating the last know values, use the
keepLastValue function in graphite to smooth out the series for the
monitoring targets.


cbaines:
> 👍 Keeping the last value sounds like a good solution to the spurious alerts.

Awesome, thanks for reviewing Suzanne :)

PR #6901, opened 2017-12-06
Update Docker tag names

danielburnley:
Docker tag does not allow forward slashes in the tag names, and for
repos with both this and dependabot enabled, the tests were failing due
to dependabots branch naming containing forward slashes.

To get around this, we have replaced the forward slashes with
underscores in the tag name, and created a method for if this tag name
will be used elsewhere within the future


danielburnley:
Closed for #6905

PR #6905, opened 2017-12-06
Strip slashes from docker image tags

alecgibson:
We pass branch names in to the Jenkins `buildDockerImage` function to
be used as a tag for the Docker image. However, a branch name might
contain a slash (eg `feature/awesome-new-feature`), which is not valid
in the Docker tag name.

In this case, the build will fail.

This change adds a method for replacing these slashes with underscores,
which are allowed in the tag name.

An example of this running successfully is here: https://ci.integration.publishing.service.gov.uk/job/content-performance-manager/job/feature%252Fcontent-preview/9/console


surminus:
Feels similar to https://github.com/alphagov/govuk-puppet/pull/6901 😄

alecgibson:
@surminus ha wow. Pick whichever you think is more appropriate; just the functionality merged 😃

danielburnley:
Haha - I'd merge this one. It's picked up a couple extra places I missed out on, i'll close mine

PR #6906, opened 2017-12-06
Set LINK_CHECKER_API_CALLBACK_HOST

SebAshton:
This PR adds an env var to use for the URL host when building a Callback URL for Link Checker API requests from Whitehall scheduled workers alphagov/whitehall#3574


kevindew:
@SebAshton This feels like something that [plek](https://github.com/alphagov/plek) should be able to take care of and shouldn't need an env var. Ignore me though if you've already explained this to @thomasleese.

PR #6913, opened 2017-12-07
Add a feature flag to use email alert frontend for email collection

tuzz:
https://trello.com/c/bwRGOWZQ/409-add-feature-to-apps-that-serve-email-signup-pages-to-redirect-to-email-alert-frontend-for-email-address-collection

We want to be able to switch users over to a new email signup journey
that uses email-alert-frontend to collect the user's email address
rather than govdelivery which is currently how it works.


thomasleese:
Would be good to get some documentation on the field.

tuzz:
@thomasleese: updated

PR #6919, opened 2017-12-07
Increase disk size for postgres primary in staging

brenetic:
postgres-primary in staging had 100GB less space than in prod. This should bring them to parity.

### DIFF
--- a/hieradata/class/staging/postgresql_primary.yaml
+++ b/hieradata/class/staging/postgresql_primary.yaml
@@ -1,1 +1,15 @@
+lv:
+  postgresql:
+    pv:
+      - '/dev/sda1'
+      - '/dev/sdc1'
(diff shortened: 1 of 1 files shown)


surminus:
If the configuration is the same as Production, I would put the disk configuration in `hieradata/class/postgresql_primary.yaml`, and leave the differences in their own files.

brenetic:
@surminus Have updated 👍 

I am still unsure whether the config is correct though. Don't know what `pv` refers to for example, but it differs from the class to the environment config ...

brenetic:
@surminus As discussed, I have swapped back to adding the postgres primary production details into the staging one. Hope this is ok?

surminus:
Feel free to merge, BTW :)

thomasleese:
This was done in #7005.

PR #6922, opened 2017-12-07
Update bouncer healthcheck in AWS

surminus:
In fcad11ccc9a31750ff85aa7e8f7eac6521db0e9d I added the healthcheck config in the wrong vhost.

Bouncer has a default vhost as it accepts all hosts due to the number of domains that it handles.

Add configuration that specifically listens for the healthcheck URL.


modules/govuk/manifests/apps/bouncer.pp
```diff
@@ -63,6 +63,18 @@
 
   if $::aws_migration {
     $vhost = 'bouncer'
+
+    nginx::config::site {
```
afda16: Does it need a title?

PR #6926, opened 2017-12-08
Set up extra CI hosts for Publishing end-to-end tests

kevindew:
This configures the 3 additional CI Agents introduced in https://github.com/alphagov/govuk-provisioning/pull/33 so that there are 4 CI agents dedicated to running just end-to-end tests, this will isolate these from the rest of CI. 

When adding these there are a few future expectations:

1. We should resolve issues that stop us running end-to-end tests in parallel, so hopefully they can make better usage of the parallel builds - when these come we can up the executors.
2. There's a good amount of low hanging fruit surrounding end-to-end test performance that we can pick off so they should get faster
3. More applications are going to be added to run end-to-end tests so there will eventually be more of them
4. With the introduction of [Dependabot](https://dependabot.com/) to our stack there is likely to be an increase in builds and thus end-to-end tests.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -727,19 +727,19 @@ govuk_ci::master::pipeline_jobs:
 govuk_ci::master::ci_agents:
   ci-agent-1:
     agent_hostname: 'ci-agent-1.ci'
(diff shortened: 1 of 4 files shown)


kevindew:
@surminus Thanks - I've had a look and not found anything. I will keep an eye out.

PR #6928, opened 2017-12-08
Explicitly prevent S3's Content-Type header reaching the client

boffbowsh:
This hopefully fixes https://github.com/alphagov/asset-manager/issues/321.

By explicitly telling nginx not to forward certain headers from S3 to
the client, we should prevent a potential race condition or undefined
ordering behaviour in nginx where the empty `Content-Type` field in S3
takes precedence over the one set by Rails.

This causes the issue as all(?) assets in S3 have a blank `Content-Type`
field, which S3 returns as an empty field.


chrisroos:
@boffbowsh - Have you tried this change in the dev-vm/integration? I've only glanced at the change but it looks like it might be removing the `Content-Type` header from all responses.

boffbowsh:
Yeah I've tried it in VM, integration and staging. It only hides it from S3, not from the Rails backend.

chrisroos:
OK, great! Good work.

PR #6932, opened 2017-12-11
Include openjdk-7-jdk on CI agents

surminus:
ci-agent-6 specifically failed to join the Jenkins master as an agent
because a JDK was not installed. This was because it had been explicitly
excluded from installing Elasticsearch (probably from when we were using
it to test Xenial).

If we're not including Elasticsearch then we should include a JDK
instead.

### DIFF
--- a/modules/govuk_ci/manifests/agent.pp
+++ b/modules/govuk_ci/manifests/agent.pp
@@ -20,9 +20,6 @@ class govuk_ci::agent(
   include ::clamav::cron_freshclam
   include ::govuk_ci::agent::redis
   include ::govuk_ci::agent::docker
-  if $elasticsearch_enabled {
-    include ::govuk_ci::agent::elasticsearch
-  }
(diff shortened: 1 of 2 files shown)


modules/govuk_ci/manifests/agent.pp
```diff
@@ -31,16 +31,21 @@
...
   include ::govuk_jenkins::user
   include ::govuk_rbenv::all
   include ::govuk_sysdig
   include ::govuk_testing_tools
 
+  if $licensing_tests {
+    include ::govuk_java::oracle8
+  } else {
+    package { 'openjdk-7-jdk':
+      ensure => 'present',
```
deanwilson: Should this include `govuk_java::openjdk7::jdk` instead?

surminus: Oh, probably :)

PR #6938, opened 2017-12-12
[AWS] Update app_domain and app_domain_internal

surminus:
This is a change we are making pre-migration to test the functionality
of the site.

We update the main app_domain to ensure users are presented with the
correct domain when browsing the site.

We update the internal domain to the top level domain, which then
CNAMEs/attaches to a stack level set of instances.

WARNING: Everything may break (in AWS)


issyl0:
:crossed_fingers:

PR #6948, opened 2017-12-13
Ignore veth and br interfaces in Collectd

kevindew:
Docker creates lots of virtual networks with bridges, which quickly fill
up lots of disk space for graphite. Since these are short lived these
stats are also effectively useless.

In ignoring these I'm somewhat concerned that there could be a network
bridge we do monitor that this regex will accidentally catch.

### DIFF
--- a/modules/collectd/files/etc/collectd/conf.d/default.conf
+++ b/modules/collectd/files/etc/collectd/conf.d/default.conf
@@ -9,6 +9,13 @@ LoadPlugin df
 LoadPlugin disk
 LoadPlugin entropy
 LoadPlugin interface
+
+<Plugin interface>
+  Interface "/^veth/"
+  Interface "/^br/"
+  IgnoreSelected true
+</Plugin>
+
 LoadPlugin load
 LoadPlugin memory
 LoadPlugin swap


stephenharker:
There are some other machines that also use Docker which this PR will also affect:

apt-1
docker-frontend-1
docker-backend-1
docker-backend-2
docker-management-1
mirrorer-1

but if we're similarly not especially interested in the interface metrics for these either it's probably fine...

There is a very basic script on graphite-1 that periodically just deletes all the veth and br interface data:

/etc/cron.d/delete_ephemeral_interface_data_from_ci_agents

but this only knows about agents 1-5. I can fix that script to get the number of agents from Puppet rather than hard-coded which would solve the immediate space issue..?

kevindew:
@stephenharker Ah sure, I didn't realise there was prior art. Feels far less disruptive to update that script than introduce this.

deanwilson:
@kevindew your change would still be worth doing so we're not sending a pile of traffic, writing it to disk and then removing it.

kevindew:
FYI this has seemed to work fine running manually on ci-agent-8 so I'm (mostly) confident it works. 

Since I've never used the veth or br data I'm not sure I'm best placed to decided whether this is a route to use or not. @stephenharker or @deanwilson any chance you can agree on how to proceed.

I'm also opening up https://github.com/alphagov/govuk-puppet/pull/6950 so we have the other route too.

deanwilson:
We've spoken and we'll make some more changes to this to restrict it to the agents. Thank you for your initial work on this.

rubenarakelyan:
Closed in favour of https://github.com/alphagov/govuk-puppet/pull/6949

PR #6950, opened 2017-12-13
Include CI agents 6 - 8 in delete ephemeral interface data script

kevindew:
We now have additional CI agents so we're quickly building up junk in
our graphite stats from veth and br interface databases.


surminus:
Let's just do https://github.com/alphagov/govuk-puppet/pull/6949 instead

kevindew:
@surminus yup that looks better 👍

PR #6952, opened 2017-12-14
Remove redundant Asset Manager environment variables

floehopper:
These variables were made redundant by alphagov/asset-manager#328 in which proxying asset requests to S3 via Nginx becomes the default for both Mainstream & Whitehall assets.

We will need to follow this change up with another PR to actually remove the `govuk::app::envvar` statements altogether.

This can safely be merged & applied once alphagov/asset-manager#328 has been deployed to production.

PR #6955, opened 2017-12-14
Remove redundant Asset Manager environment variable definitions

floehopper:
These variables were made redundant by alphagov/asset-manager#328 in which proxying asset requests to S3 via Nginx became the default for both Mainstream & Whitehall assets.

This is a follow-up to #6952 which actually removed these variables from `/etc/govuk/asset-manager/env.d/`.

This PR can safely be merged & applied once #6952 has been merged & applied.

Note that this PR currently contains the commit from #6952, so ideally it will need to be rebased against master and force-pushed once that PR has been merged.


floehopper:
I've just merged #6952 and so I'm going to rebase this PR against `master` and force-push.

PR #6957, opened 2017-12-15
Add LINK_CHECKER_API_SECRET_TOKEN to manuals-publisher

SebAshton:
This PR adds the secret token env variable required for alphagov/manuals-publisher#1226

### DIFF
--- a/modules/govuk/manifests/apps/manuals_publisher.pp
+++ b/modules/govuk/manifests/apps/manuals_publisher.pp
@@ -49,6 +49,10 @@
 #   Redis port for sidekiq.
 #   Default: 6379
 #
+# [*link_checker_api_secret_token*]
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/manuals_publisher.pp
```diff
@@ -63,6 +63,7 @@
   $redis_host = undef,
   $redis_port = undef,
   $secret_token = undef,
+  $link_checker_api_secret_token = undef,
```
thomasleese: Could we add a line of documentation for this?

SebAshton: Updated @thomasleese

thomasleese:
The secret has been added now.

PR #6959, opened 2017-12-15
Generate jenkins job puppet class specs

deanwilson:


issyl0:
See commit messages for context as to why.

PR #6967, opened 2017-12-18
[AWS] Public API fix ups

surminus:
A couple of other fixes I've noticed since we changed the `app_domain` to the publishing domain.

### DIFF
--- a/modules/govuk/manifests/apps/publicapi.pp
+++ b/modules/govuk/manifests/apps/publicapi.pp
@@ -16,8 +16,7 @@ class govuk::apps::publicapi (
 
     $whitehallapi = "whitehall-frontend.${app_domain_internal}"
     $rummager_api = "search.${app_domain_internal}"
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/publicapi.pp
```diff
@@ -15,12 +16,15 @@
 
     $whitehallapi = "whitehall-frontend.${app_domain_internal}"
     $rummager_api = "search.${app_domain_internal}"
-    # Content Store is always public
-    $content_store_api = "content-store.${app_domain}"
+    $content_store_api = "content-store.${app_domain_internal}"
```
issyl0: The content store is not "always public" anymore then.

issyl0: Oh, wait, now it matches the others in this file (whitehall and rummager), so it's OK.

PR #6969, opened 2017-12-18
Increase the window size for fetching Signon sidekiq data

cbaines:
This is showing as Unknown on Staging, which is a false positive, as
the metrics are still valid, it's just that nothing has happened, so
the Sidekiq middleware hasn't sent any data.

A very large window isn't ideal, as this could hide a lack of data, so
just increase it a little bit, to hopefully reduce the false positives
on Staging.

### DIFF
--- a/modules/monitoring/manifests/checks/sidekiq.pp
+++ b/modules/monitoring/manifests/checks/sidekiq.pp
@@ -32,13 +32,13 @@ class monitoring::checks::sidekiq (
     critical  => 50,
     desc      => 'signon background worker queue size unexpectedly large',
     host_name => $::fqdn,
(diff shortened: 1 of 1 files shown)


modules/monitoring/manifests/checks/sidekiq.pp
```diff
@@ -38,7 +38,7 @@
     # many false positives from not much activity, while also ensuring
     # that if there is no data for longer than 24 hours, the alert
```
rubenarakelyan: Should this also be "5 hours"?

cbaines: Yep, good spot. I've updated the comment now.

cbaines:
Thanks Ruben :)

PR #6972, opened 2017-12-18
Revert DNS resolver changes

surminus:
This change was not required, and moving back to the Google DNS servers means that we are not dependent on Carrenza resolvers (which are not accessible from all instances).


tijmenb:
@surminus Out of interest, how did this ever work on AWS? Is there a VPN between Carrenza and AWS?

surminus:
We use whatever resolver is set by the VPC itself, rather than on the instances.

PR #6976, opened 2017-12-18
Add RATE_LIMIT_BYPASS_TOKEN to link-checker-api

SebAshton:
This PR adds a token to the `link-checker-api` `ENV` to bypass the rate limiter within nginx

### DIFF
--- a/modules/govuk/manifests/apps/link_checker_api.pp
+++ b/modules/govuk/manifests/apps/link_checker_api.pp
@@ -49,6 +49,10 @@
 #   start.
 #   Default: undef
 #
+# [*rate_limit_token*]
+#   The token used to bypass the rate limiting on GOV.uk
+#   Default: undef
+#
 class govuk::apps::link_checker_api (
   $db_hostname = undef,
   $db_username = 'link_checker_api',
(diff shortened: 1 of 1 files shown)


SebAshton:
Thanks @issyl0, I've made a request to 2ndLine to get this added.

surminus:
This will create an environment variable that the application can use, but a rate-limit token will need to be added to NGINX to allow pass through.

You will need to add the token in `router::nginx::rate_limit_tokens` in govuk-secrets. You will see similar ones added.

modules/govuk/manifests/apps/link_checker_api.pp
```diff
@@ -49,6 +49,10 @@
 #   start.
 #   Default: undef
 #
+# [*rate_limit_bypass_token*]
```
thomasleese: Can we rename this to `rate_limit_token` for consistency with the other apps?

SebAshton: Sure @thomasleese, have updated.

thomasleese:
Secrets have been merged in now.

PR #6980, opened 2017-12-19
Update the Sidekiq Grafana dashboard

cbaines:
To support the metrics generated by version 3.0.0 of the govuk_sidekiq
Gem.

Now metrics are reported per machine, so aggregate the metrics by
looking for the maximum value in the time interval.


cbaines:
Thanks for reviewing Tom ⭐️

PR #6981, opened 2017-12-19
Move sidekiq monitoring to a new port and enable on dev vm

h-lame:
While trying to view sidekiq monitoring for support-api and following [the docs](https://docs.publishing.service.gov.uk/manual/monitor-sidekiq-workers.html) I found out that all I got was a:

```json
{
  "error": "Fell through to default vhost"
}
```

error from nginx.  It seems that 60a89ae9a7a30406225a3ff37cf17928941fd427 removed the `localhost` definition from the nginx config that made these instructions work.  We can't really put it back because it'll never work on AWS, so instead we give sidekiq-monitoring it's own port.

We also make it available on the dev VM so you don't have to tunnel to see the monitoring on there.

If this PR is accepted, I'll make a PR to the docs repo to change to reflect the new reality introduced by this PR.


h-lame:
Shipped to integration and tested on backend-2 and tunnelling to 3211 instead of 80 worked.  Merging.

PR #6983, opened 2017-12-19
Configure maintenance mode not on load balancers

issyl0:
- In AWS, we don't have loadbalancers. We don't want to rearchitect the entire maintenance mode work right now, so include the loadbalancers::maintenance class for machines with nginx on them, if maintenance mode is turned on.

https://trello.com/c/EkXtb3cN/887-find-out-what-migration-mode-does-and-plan-for-using-it-during-migration

### DIFF
--- a/modules/loadbalancer/manifests/balance.pp
+++ b/modules/loadbalancer/manifests/balance.pp
@@ -80,15 +80,6 @@ define loadbalancer::balance(
     }
   }
 
-  if ! defined(File['/etc/nginx/includes/maintenance.conf']) {
-    file { '/etc/nginx/includes/maintenance.conf':
-      ensure  => present,
-      content => template('loadbalancer/etc/nginx/includes/maintenance.conf.erb'),
-      require => File['/etc/nginx/includes'],
-      notify  => Class['nginx::service'],
-    }
-  }
-
   if $maintenance_mode {
     include loadbalancer::maintenance
   }
--- a/modules/loadbalancer/manifests/maintenance.pp
+++ b/modules/loadbalancer/manifests/maintenance.pp
@@ -4,6 +4,13 @@
 # into maintenance mode.
 #
 class loadbalancer::maintenance {
+  file { '/etc/nginx/includes/maintenance.conf':
+    ensure  => present,
+    content => template('loadbalancer/etc/nginx/includes/maintenance.conf.erb'),
+    require => File['/etc/nginx/includes'],
+    notify  => Class['nginx::service'],
+  }
+
   file { '/usr/share/nginx/www':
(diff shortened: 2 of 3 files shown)


issyl0:
@surminus I moved that stuff into the `loadbalancer::maintenance` class. I don't know if that will break the existing `loadbalancer::balance` stuff - the file definitions existed even when not in migration mode.

modules/nginx/manifests/config.pp
```diff
@@ -17,13 +17,22 @@
...
+# [*maintenance_mode*]
+#   Whether or not the site is in maintenance mode.
+#   Default: undef.
+#
 class nginx::config (
   $server_names_hash_max_size,
   $variables_hash_max_size = 1024,
   $denied_ip_addresses,
   $stack_network_prefix = '10.1.0',
+  $maintenance_mode = undef,
```
surminus: Set to `false` to be explicit that it's a boolean.

modules/loadbalancer/manifests/balance.pp
```diff
@@ -74,21 +74,6 @@
       logstream => present;
   }
 
-  if ! defined(File['/etc/nginx/includes']) {
```
surminus: I wouldn't move this out as it's not specific to maintenance, but would ensure that this file is created in the `nginx::config` class where you're creating it.

modules/loadbalancer/manifests/maintenance.pp
```diff
@@ -4,6 +4,21 @@
 # into maintenance mode.
 #
 class loadbalancer::maintenance {
+  if ! defined(File['/etc/nginx/includes']) {
+    file { '/etc/nginx/includes':
+      ensure => 'directory',
+    }
+  }
+
+  if ! defined(File['/etc/nginx/includes/maintenance.conf']) {
```
surminus: I'm not sure the `defined` function was used here in a conditional, and I'm not sure it's required.

issyl0: So that it doesn't recreate the file if it already exists? Does that mean the condition is no longer required?

surminus:
I tried testing this by deploying and then manually setting the hieradata to `nginx::config::maintenance_mode: true` on the Puppetmaster.

I think this behaves differently on how we configured the NGINX load balancers as turning off an app caused NGINX to throw a 502, which meant that the error page was not displayed.

We should do a complete set of work to properly set maintenance pages on Staging and Production. Integration is less of an issue.

surminus:
We're going to do a full investigation of a maintenance mode so it is AWS compatible. The current method should work for Carrenza when we do the migration.

PR #6984, opened 2017-12-19
Add theme back in

surminus:
I removed the nice blue theme in 4ba6ea4c1d5463a7b5eae654a42bd87e8368a114, but since I did that I've really missed the pleasant colours, which have been replaced with some awful black based theme.

Revert to the nice theme <3

### DIFF
--- a/hieradata_aws/class/jenkins.yaml
+++ b/hieradata_aws/class/jenkins.yaml
@@ -51,6 +51,11 @@ govuk_jenkins::config::user_permissions:
       - 'hudson.model.Item.Read'
 
 govuk_jenkins::config::banner_string: 'AWS Integration'
+govuk_jenkins::config::banner_colour_background: '#005EA5'
+govuk_jenkins::config::banner_colour_text: 'white'
+govuk_jenkins::config::theme_colour: '#005EA5'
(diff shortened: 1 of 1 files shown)


issyl0:
Do we need to think of some other colours?

surminus:
We could [go green](http://www.color-hex.com/color/006400) to keep in keeping with the traffic lights theme

hieradata_aws/class/jenkins.yaml
```diff
@@ -51,6 +51,11 @@ govuk_jenkins::config::user_permissions:
       - 'hudson.model.Item.Read'
 
 govuk_jenkins::config::banner_string: 'AWS Integration'
+govuk_jenkins::config::banner_colour_background: '#005EA5'
+govuk_jenkins::config::banner_colour_text: 'white'
+govuk_jenkins::config::theme_colour: '#005EA5'
+govuk_jenkins::config::theme_text_colour: 'white'
+govuk_jenkins::config::theme_environment_name: 'Integration'
```
deanwilson: Should this be AWS Integration while we're migrating?

surminus: :+1:

surminus:
We're gonna go blue!

PR #6988, opened 2017-12-21
Make the email-alert-check jobs resilient to GitHub being down

tuzz:
https://trello.com/c/0nNVddWb/408-%EF%B8%8F-dont-call-people-when-github-is-down

Previously, the build failed if the clone from GitHub failed. We'd tried
to use the Text Finder plugin to change the job status from 'failed' to
'unstable' so that Failure_Passive_Check wouldn't trigger post-build,
but this approach doesn't work.

The problem is the Text Finder plugin can only downgrade build statuses,
e.g. success -> unstable, it can't go fail -> unstable. See:
https://issues.jenkins-ci.org/browse/JENKINS-19641

This change makes it so we guard against the clone to GitHub failing by
cloning to a `tmp` directory, which only replaces the existing
email-alert-monitoring directory if the clone succeeds. Otherwise, we
echo from this command which causes execution to continue. This
has the effect of re-using the previously cloned directory from the last
time the job ran.

If the clone fails and the email check succeeds, we mark the build as
unstable by matching the 'Cloning failed' text in the build output. This
shouldn't alert in PagerDuty, but will add visibility that something
isn't quite right.

While working on this, I also looked at the Conditional Build Plugin and
Groovy Script Plugin, but I didn't manage to get either of these plugins
to trigger when the clone fails using Jenkins' default SCM settings for
the repository.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/drug_email_alert_check.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/drug_email_alert_check.yaml.erb
@@ -1,12 +1,6 @@
 ---
 - scm:
     name: email-alert-check
-    scm:
-        - git:
-            url: git@github.com:alphagov/email-alert-monitoring.git
-            branches:
-              - master
-
 - job:
     name: email-alert-check
     display-name: EmailAlertCheck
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/templates/jobs/drug_email_alert_check.yaml.erb
```diff
@@ -20,6 +14,10 @@
         - timed: '21 * * * *' # every hour on the 21st minute
     builders:
         - shell: |
+            # If GitHub is down, re-use the directory we cloned from last time the job ran.
+            (git clone https://github.com/alphagov/email-alert-monitoring tmp && rm -rf email-alert-monitoring && mv tmp email-alert-monitoring) || true
```
boffbowsh: If this fails, can we mark the build as Unstable, but not alert through pagerduty? Same for Travel Advice, obvs.

tuzz: Just to clarify, is this what you mean:

If the clone succeeds and the email check succeeds
  - success
If the clone fails and the email check succeeds
  - unstable
If the clone succeeds and the email check fails
  - failure (+ trigger pagerduty)
If the clone fails and the email check fails
  - failure (+ trigger pagerduty)

boffbowsh: Yep 👍

tuzz:
@boffbowsh: ok, updated. I forced the clone to fail and it looks to have the desired effect:
https://deploy.publishing.service.gov.uk/job/travel-advice-email-alert-check/57658/console

PR #6990, opened 2017-12-21
Redirect kibana.*.publishing.service.gov.uk to Logit

tuzz:
This might be more efficient than going through
google apps. The environment is set in the session
from a UUID in the URL.


h-lame:
This is magical!

Nothing is impossible

PR #6991, opened 2017-12-22
[AWS] Set everything as internal in AWS

surminus:
This change comes off the back of a discussion that was had in a [PR that I opened for Plek][1]. It looks to solve an issue where we initially tried to override specific cases of Plek lookups, but kept running into small edge cases.

Set "GOVUK_APP_DOMAIN" as internal by default. Ensure that each application sets a Plek override for itself so that if it ever references itself, it redirects the user to the correct location.

This hopefully solves a bunch of problems we have come across when trying to correctly route traffic internally. We had an issue where applications would misuse Plek to find out what their URL is to redirect users.

[1]: https://github.com/alphagov/plek/pull/54#issuecomment-353390978

### DIFF
--- a/modules/govuk/manifests/deploy/config.pp
+++ b/modules/govuk/manifests/deploy/config.pp
@@ -92,31 +92,25 @@ class govuk::deploy::config(
 
     'ERRBIT_ENVIRONMENT_NAME': value   => $errbit_environment_name;
     'SENTRY_CURRENT_ENV': value        => $errbit_environment_name;
-    'GOVUK_APP_DOMAIN': value          => $app_domain;
     'GOVUK_ASSET_HOST': value          => $asset_root;
     'GOVUK_ASSET_ROOT': value          => $asset_root;
     'GOVUK_WEBSITE_ROOT': value        => $website_root;
   }
 
-  # TODO: Set some internal services specifically before we've figured out
-  # how to entirely use internal domains
   if $::aws_migration {
     $app_domain_internal = hiera('app_domain_internal')
 
     govuk_envvar {
-      'GOVUK_APP_DOMAIN_INTERNAL': value        => $app_domain_internal;
-
-      # These variables are manually set in the draft stack (e.g. s_draft_cache),
-      # make sure they're separated out in those locations otherwise puppet
-      # won't run cleanly.
-      'PLEK_SERVICE_CONTENT_STORE_URI': value   => "https://content-store.${app_domain_internal}";
-      'PLEK_SERVICE_EMAIL_ALERT_API_URI': value => "https://email-alert-api.${app_domain_internal}";
-      'PLEK_SERVICE_LICENSIFY_URI': value       => "https://licensify.${licensify_app_domain}";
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/deploy/config.pp
```diff
@@ -92,31 +92,27 @@
...
-      'PLEK_SERVICE_RUMMAGER_URI': value        => "https://rummager.${app_domain_internal}";
-      'PLEK_SERVICE_SEARCH_URI': value          => "https://search.${app_domain_internal}";
-      'PLEK_SERVICE_STATIC_URI': value          => "https://static.${app_domain_internal}";
+      # By default traffic should route internally
+      'GOVUK_APP_DOMAIN':           value => $app_domain_internal;
+      'GOVUK_APP_DOMAIN_INTERNAL':  value => $app_domain_internal;
+      # We do not host Licensify in AWS, so need to redirect to
+      # it's external domain
+      'PLEK_SERVICE_LICENSIFY_URI': value => "https://licensify.${licensify_app_domain}";
+      # Apps that auth with Signon should do so across the publishing domain
```
boffbowsh: Would be good to have the reason documented here. I can't think of it off the top of my head

modules/govuk/spec/defines/govuk__app__config_spec.rb
```diff
@@ -92,6 +92,45 @@
...
+      it do
+        is_expected.to contain_govuk__app__envvar('giraffe-PLEK_SERVICE_APP_URI').with(
+          :varname => 'PLEK_SERVICE_GIRAFFE_URI',
+          :value   => 'https://giraffe.environment.example.com',
+        )
+      end
+    end
+
+    context 'in AWS with hyphenated app name' do
+      let(:title) { 'bella-the-cat' }
```
boffbowsh: 😸

surminus:
I updated the comment, but I'm still unsure on the specific reason it happened. 

The behaviour was that you would login to Signon (after updating your `/etc/hosts` so that `integration.publishing.service.gov.uk` pointed to the AWS environment), try to click through to an application and be presented with a failed DNS lookup for `integration.govuk-internal.digital`. This change resolved the issue, but I am still unsure exactly where or how the applications or Signon were using Plek to set this URL.

boffbowsh:
Is that not fixed by Signon having an override set for itself? I *think* that's to do with signon having to set a callback URL for itself. Or was it e.g. `publisher.integration.govuk-internal.digital` that it was trying to send you to?

surminus:
That's correct. I will test without this change and see if it still persists.

surminus:
@boffbowsh I've added the commit back in. After testing by browsing to an app, I was redirected to Signon using the incorrect URL which resulted in the following error:

```
This site can’t be reached

signon.integration.govuk-internal.digital’s server DNS address could not be found.
Try running Network Diagnostics.
DNS_PROBE_FINISHED_NXDOMAIN
```

I believe this is because the applications (or gds-sso) use Plek to find the Signon URL to redirect users to, so we need to ensure it is set to the Publishing domain.

I've added it as a separate commit with more detail, and also set comments so they are clearer.

PR #6992, opened 2017-12-22
Add EMAIL_ALERT_API_BEARER_TOKEN for email-alert-service

gpeng:
Adds an env var for email-alert-service.

[Trello](https://trello.com/c/75sVuXhA/455-require-all-apps-to-authenticate-with-email-alert-api)

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_service.pp
+++ b/modules/govuk/manifests/apps/email_alert_service.pp
@@ -25,12 +25,16 @@
 # [*sentry_dsn*]
 #   The URL used by Sentry to report exceptions
 #
+# [*email_alert_api_bearer_token*]
+#   Bearer token for communication with the email-alert-api
+#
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/email_alert_service.pp
```diff
@@ -25,6 +25,9 @@
 # [*sentry_dsn*]
 #   The URL used by Sentry to report exceptions
 #
+# [*email_alert_api_bearer_token*]
+#   Bearer token for communication with the email-alert-api
+#
 class govuk::apps::email_alert_service(
```
thomasleese: You'll need to add `email_alert_api_bearer_token` to the argument list. 😄

gpeng: Just testing to check if you're paying attention.

PR #7003, opened 2017-12-29
Remove files added in error

surminus:


surminus:
I don't need the fixes anymore, but we do need to remove the files committed in error so will repush this PR.

PR #7009, opened 2018-01-03
Update sidekiq paths for grafana

dwhenry:
This moves to the govuk_sidekiq gem naming syntax which includes the server
name in the metric path. We currently only display the metric graph for
rummager, but have made this change global as the assumption is that all apps
should be using our gem.

https://trello.com/c/OBhgB1Lw/531-code-tidy-up

### DIFF
--- a/modules/monitoring/manifests/checks/sidekiq.pp
+++ b/modules/monitoring/manifests/checks/sidekiq.pp
@@ -10,7 +10,7 @@ class monitoring::checks::sidekiq (
   $enable_support_check = true,
 ) {
   icinga::check::graphite { 'check_rummager_queue_latency':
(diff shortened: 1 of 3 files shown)


modules/monitoring/manifests/checks/sidekiq.pp
```diff
@@ -10,7 +10,7 @@
   $enable_support_check = true,
 ) {
   icinga::check::graphite { 'check_rummager_queue_latency':
-    target              => 'keepLastValue(stats.gauges.govuk.app.rummager.workers.queues.default.latency)',
+    target              => 'keepLastValue(maxSeries(stats.gauges.govuk.app.rummager.*.workers.queues.default.latency))',
```
cbaines: 👍 I've done similarly for the Sidekiq Grafana dashboard.

PR #7012, opened 2018-01-03
Display sidekiq graphs & stats on Asset Manager deployment dashboard

floehopper:
I suspect these were not added originally, because Asset Manager used to use Delayed Job rather than Sidekiq.

I think these will be a useful addition, particularly now that we have two different workers - one for virus scanning and the other for uploading to S3 - both of which are integral to the app working correctly.


floehopper:
@brenetic, @JonathanHallam I hope you don't mind me asking you to review this PR, but you were suggested by GitHub.

floehopper:
@sihugh Thanks for reviewing.

floehopper:
This has been deployed to staging and I can see the worker stats and Sidekiq graphs.

floehopper:
This has been deployed to production and I can see the worker stats and Sidekiq graphs. 🎉

PR #7013, opened 2018-01-03
Decrease the xFilesFactor for statsd gauges

cbaines:
Currently, the metrics matching ^stats.* will be stored with 5 second
resolution for 8 days, and then with less resolution over longer time
periods thereafter. Currently if only one value is reported in a 1
minute interval, it won't be aggregated, so that data will only be
stored for 8 days.

I'm interested in the metrics in the stats.gauges.topic-taxonomy
namespace, as I want to use them to track changes over a period like 3
months to see how the Topic Taxonomy is changing over a financial
quarter.

I'm unsure what will happen with an xFilesFactor of 0, as Graphite
doesn't often handle null data as I'd expect, and has a tendency to
treat null as 0. So, I've just set the value low enough such that if
only 1 value is reported in a 1 minute interval, that value will be
aggregated and stored.

With a 1 minute interval being an aggregation of 12 5 second
intervals, if only one value is present, the fraction would be 1/12,
or 0.0833... so use 0.01 as that's a nice round power of 10.

Changing the configuration won't change the value stored in existing
files, that will have to be done manually.

With the current storage schema for statsd metrics, this change in
xFilesFactor should also mean that aggregation for the longer
intervals should also work, as the step up is smaller than this
initial interval change.

### DIFF
--- a/modules/govuk/files/node/s_graphite/storage-aggregation.conf
+++ b/modules/govuk/files/node/s_graphite/storage-aggregation.conf
@@ -44,6 +44,10 @@ xFilesFactor = 0.1
 pattern = ^stats_counts\..*
 aggregationMethod = sum
 
+[statsd-gauges]
+pattern = ^stats.gauges\..*
+xFilesFactor = 0.01
+
 # incremental counter, keep last value
 [mtail-counter-incremental]
 pattern = \.nginx_metrics\.mtail\.http_requests_total.*


rjw1:
As discussed elsewhere it would be good if your why was in the commit message to explain why you don't want the data lost.

rjw1:
Reading http://obfuscurity.com/2012/04/Unhelpful-Graphite-Tip-9 I think you should just set it to '0.0'
You never want to lose the data.

cbaines:
> As discussed elsewhere it would be good if your why was in the commit message to explain why you don't want the data lost.

Ok, I've added a bit more information about the taxonomy to the commit message.

> Reading http://obfuscurity.com/2012/04/Unhelpful-Graphite-Tip-9 I think you should just set it to '0.0'
You never want to lose the data.

The only reason why I didn't do that is that I've had bad experiences with Graphite treating `null` as `0`, so I was worried that it would aggregate regions with no data to 0, but I'm open to trying it.

rjw1:
I think this is a good change and wont break existing stats if it doesn't do what we want.

cbaines:
> I think this is a good change and wont break existing stats if it doesn't do what we want.

Great :)

> My knowledge of Grafana/Graphite is pretty minimal, but let's merge this to see how it performs on integration

Awesome, lets try this 👍

cbaines:
Maybe I should have put this in the commit message, but I've only just worked it out.

I think this is the incantation that needs running to migrate the existing metrics. First the virtualenv needs activating, and then this rather drawn out command needs running:

```
sudo -E -u www-data PYTHONPATH=/opt/graphite/lib find storage/whisper/stats/gauges -type f -not -name "*count.wsp" -exec bin/whisper-resize.py --nobackup --xFilesFactor=0.01 {} 5s:8d 1m:31d 10m:1y 1h:5y \;
```

For some reason, despite using virtualenv, it still doesn't find the python libraries that have been downloaded to /opt/graphite/lib, so set the PYTHONPATH to compensate. Avoid the count files, as they have a different xFilesFactor by default.

PR #7014, opened 2018-01-03
Add publishing-components as a development app

tuzz:
https://trello.com/c/HSre7K10/453-style-the-mvp-email-collection-page-needed-by-jan

The #publishing-frontend team want to be able to work on the gem without
having to mount it into another Rails application first. We can use the
dummy Rails app the test suite uses to achieve this.

Other PR: https://github.com/alphagov/govuk_publishing_components/pull/115


boffbowsh:
What's the benefit of running this in the development VM? Is it so we can access a development version of static?

tuzz:
It hadn't occurred to me to only have it as an app to be run on the host machine. I try not to run anything there if I can help it, but maybe most people don't care?

tuzz:
Merging because Paul would like to use it. I think it's a valuable addition to be able to run this easily in the vm.

PR #7017, opened 2018-01-04
Grouping on deployment databoard

dwhenry:
Fixes for sidekiq data on the deployment dashboard as a result of moving rummager to govuk_sidekiq

### DIFF
--- a/modules/grafana/templates/dashboards/deployment_panels/_sidekiq_errors.json.erb
+++ b/modules/grafana/templates/dashboards/deployment_panels/_sidekiq_errors.json.erb
@@ -9,12 +9,12 @@
   "targets": [
     {
       "refId": "A",
-      "target": "aliasByNode(stats.govuk.app.<%= @app_name %>.workers.*.failure, 5)",
+      "target": "groupByNode(stats.govuk.app.<%= @app_name %>.*.workers.*.*.failure, 6, 'sumSeries')",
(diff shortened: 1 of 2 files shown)


modules/grafana/templates/dashboards/deployment_panels/_sidekiq_errors.json.erb
```diff
@@ -9,12 +9,12 @@
   "targets": [
     {
       "refId": "A",
-      "target": "aliasByNode(stats.govuk.app.<%= @app_name %>.workers.*.failure, 5)",
+      "target": "groupByNode(stats.govuk.app.<%= @app_name %>.*.workers.*.*.failure, 6, 'sumSeries')",
       "textEditor": true
     },
     {
       "refId": "B",
-      "target": "aliasByNode(stats.govuk.app.<%= @app_name %>.workers.*.*.failure, 6)",
```
cbaines: You might want to keep the alias for these two metrics, but the change to the expression looks fine.

dwhenry: the group by node pulls out the alias element and returns it as the key :)

cbaines: Ah, cool, good to know :)

PR #7021, opened 2018-01-04
Remove cron job definitions already marked as absent

floehopper:
These cron job definitions were marked as `ensure => absent` some time ago, so I think it's safe to assume we can remove them and de-clutter the `govuk-puppet` repo very slightly.


floehopper:
@MatMoore Thanks for reviewing!

PR #7026, opened 2018-01-05
Iterating Docker CI pruning

kevindew:
We just had an issue where we ran out of disk space on ci-agent-8 so I've tried to see if there was more that could be done to stop the data usage.

This should remove a cron file that was left hanging around and increase the clearing of volumes/dangling artefacts.

### DIFF
--- a/modules/govuk_ci/manifests/agent/docker.pp
+++ b/modules/govuk_ci/manifests/agent/docker.pp
@@ -15,6 +15,14 @@ class govuk_ci::agent::docker {
     version => '1.17.1',
   }
 
+  # TODO: remove this once the file has been purged
+  cron::crondotdee { 'docker_system_prune' :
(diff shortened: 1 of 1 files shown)


modules/govuk_ci/manifests/agent/docker.pp
```diff
@@ -15,9 +15,16 @@
     version => '1.17.1',
   }
 
-  cron::crondotdee { 'docker_system_prune_dangling' :
+  cron::crondotdee { 'docker_system_prune' :
```
deanwilson: Can you add a TODO please to remind us to remove this code once the file has been purged.

kevindew: @deanwilson Sure done

deanwilson:
:+1:

PR #7027, opened 2018-01-05
Add an envvar for the graphite hostname

issyl0:
- In AWS, it's no longer graphite.cluster.

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -624,6 +624,7 @@ govuk_htpasswd::http_username: "%{hiera('http_username')}"
 govuk_jenkins::packages::gcloud::apt_mirror_hostname: "%{hiera('apt_mirror_hostname')}"
 govuk_jenkins::packages::terraform::apt_mirror_hostname: "%{hiera('apt_mirror_hostname')}"
 
+govuk_jenkins::jobs::deploy_app::graphite_host: 'graphite'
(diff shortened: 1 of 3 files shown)


hieradata/common.yaml
```diff
@@ -610,6 +610,8 @@ govuk_gor::apt_mirror_hostname: "%{hiera('apt_mirror_hostname')}"
 
 govuk_htpasswd::http_username: "%{hiera('http_username')}"
 
+govuk_jenkins::jobs::deploy_app::graphite_host: 'graphite.cluster'
```
surminus: If this is the default we do not need to also have it in hieradata.

hieradata_aws/common.yaml
```diff
@@ -624,6 +624,7 @@ govuk_htpasswd::http_username: "%{hiera('http_username')}"
 govuk_jenkins::packages::gcloud::apt_mirror_hostname: "%{hiera('apt_mirror_hostname')}"
 govuk_jenkins::packages::terraform::apt_mirror_hostname: "%{hiera('apt_mirror_hostname')}"
 
+govuk_jenkins::jobs::deploy_app::graphite_host: "graphite.%{hiera('app_domain_internal')}"
```
surminus: This can just be `graphite`.

hieradata/common.yaml
issyl0: That's true. Removed.

PR #7032, opened 2018-01-08
Add transition DB password to env-sync

afda16:
On AWS the transition db password could be different
to the shared postgresql one. This needs to be reflected
in the environment data sync. There is a PR in the env sync
repository to complete this update.


floehopper:
@afda16 Are you happy for this to be deployed to staging/production?

PR #7033, opened 2018-01-08
Update dashboards for govt-frontend and local links manager

sihugh:
Removing the field prefix means that we look for (for example) the `duration` field in Kibana logs rather than `@fields.duration`.  This changes when we use a version of logstasher that is greater than 1 which is a given if we've upgraded the app to govuk_app_config v1 or higher.

The AWS config for the deployment dashboards appears to have lagged a bit, so I've copy/pasted from the non-aws hieradata file.


sihugh:
👍

PR #7047, opened 2018-01-09
Run publishing-components with `bundle exec`

tuzz:
This is more consistent with the other applications that are run with
foreman.


fofr:
This is consistent with other apps but doesn't address the issue of:

```
vagrant@development:/var/govuk/govuk-puppet/development-vm$ bowl publishing-components
I, [2018-01-09T10:18:06.840030 #19226]  INFO -- : Starting publishing-components...
rbenv: foreman: command not found

The `foreman' command exists in these Ruby versions:
  2.3.1
```

PR #7051, opened 2018-01-09
Display Sidekiq graphs on Whitehall deployment dashboard

floehopper:
c.f. #7012 in which we did the same thing for Asset Manager.

I've found having the graphs accessible from the dashboard very useful.

Note that I've updated both the hieradata & AWS hieradata as per [this commit][1].

[1]:
https://github.com/alphagov/govuk-puppet/commit/418a02bd1bacde312e0fefa287a604ed782a7dc1


floehopper:
@sihugh Thanks for reviewing!

PR #7054, opened 2018-01-10
Update govuk.buildProject to tag docker similar to github tags

kevindew:
This stops docker tags being created for deployed-to-* branches (as we intend to create those automatically at deploy time) and adds a tag for release which is consistent with our github tags.

This is to enable the tags to be consistent with those which are input into deploy jobs as the deploy jobs will copy the tag inputted to them to create deployed-to-* branches.

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -283,10 +283,7 @@ def buildProject(Map options = [:]) {
 
         if (hasDockerfile()) {
           stage("Tag Docker image") {
-            dockerTag = options.newStyleDockerTags ?
-                        getNewStyleReleaseTag() :
-                        "release_${env.BUILD_NUMBER}"
-            pushDockerImage(jobName, env.BRANCH_NAME, dockerTag)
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -347,6 +346,20 @@ def isCurrentCommitOnMaster() {
   ) == 0
 }
 
+/**
+ * Check whether there is a git branch named release
+ * This test is useful for determining whether we should update this branch or
+ * not
+ */
+def releaseBranchExists() {
+  sshagent(["govuk-ci-ssh-key"]) {
```
suzannehamilton: Does this `sshagent` wrapper return the boolean from lines 356-359? Or does it return some value of its own? I can't find it in the docs.

kevindew: That's a good point. I wasn't too sure and in my testing it worked successfully but I may have only testing the truthy route. I'll give it a spin on something where it is falsy to check.

kevindew: @suzannehamilton Yup I've been able to confirm it returns a boolean: https://ci.integration.publishing.service.gov.uk/job/govuk_app_config/job/master/26/console

kevindew:
Thanks @suzannehamilton 👍

PR #7061, opened 2018-01-10
Fix more hardcoded hostnames in Jenkins jobs

issyl0:
- These are the ones I can see are broken in Integration currently since the switchover.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/transition_load_all_data.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/transition_load_all_data.yaml.erb
@@ -10,13 +10,18 @@
         - shell: |
             rm -rf data
             mkdir data
+            <%- if scope.lookupvar('::aws_migration') %>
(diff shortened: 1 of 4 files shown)


surminus:
It may be safer to set the name of the node before each block so we can guarantee we're running tasks on the same instance. 

For example, set the hostnames as an environment variable:
`HOST=$(govuk_node_list -c backend --single-node)`

AFAIK the `--single-node` flag does a random lookup, so if we want all commands in a block of 3 to be run on the same instance, then we cannot guarantee that if we're doing a lookup each time. This may also reduce the amount of repeated code.

If the tasks do not need to be run on a specific instance in order, then this does not apply.

issyl0:
Ah, I figured it would return them in some kind of order consistently. I'll edit this (when I get back) to set an envvar then. Thanks!

issyl0:
@surminus Edited. There's a separate commit, but I can squash it if you like.

PR #7063, opened 2018-01-10
Add a mailmap file.

rjw1:
This standardizes the display of people's names and emails in git blame
and git shortlog. git log can be told to use it as well with `git log
--use-mailmap`


surminus:
:+1: Tested and works well

PR #7066, opened 2018-01-10
Use the Xenial HWE stack on Trusty machines

boffbowsh:
Install the Xenial kernel on Trusty (14.04 LTS) machines in order to
get the latest kernel updates, including for Spectre/Meltdown.

Change the icinga check script to check for this kernel version, and
remove the ability to check our precise machines, for simplicity.

We only run precise on two classes of machines in production -
puppetmaster and licensify-lb. We need to patch these manually for now.

Test this kernel on Carrenza integration only for now. For AWS, we should 
bake an AMI running this kernel, deploy it then enable the check.

See https://wiki.ubuntu.com/Kernel/LTSEnablementStack

### DIFF
--- a/hieradata/staging.yaml
+++ b/hieradata/staging.yaml
@@ -4,7 +4,7 @@ app_domain: 'staging.publishing.service.gov.uk'
 backup::server::backup_hour: 9
 
 base::shell::shell_prompt_string: 'staging'
-base::supported_kernel::enabled: true
+base::supported_kernel::enabled: false
 
 cron::daily_hour: 6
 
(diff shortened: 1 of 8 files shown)


boffbowsh:
Added a flag to selectively enable the check, so we don't spam Icingia

hieradata/staging.yaml
```diff
@@ -4,7 +4,7 @@ app_domain: 'staging.publishing.service.gov.uk'
 backup::server::backup_hour: 9
 
 base::shell::shell_prompt_string: 'staging'
-base::supported_kernel::enabled: true
+base::supported_kernel::enabled: false
```
surminus: In the class this is set to `false` by default, so I think these can just be removed unless we want to keep them for clarify.

boffbowsh: I was keeping them for clarity. We'll need to do a few PRs in succession to change these, so I figured it'd make it easier to see what needs changing

PR #6501, opened 2017-09-19
Sidekiq monitoring for asset manager

chrislo:
This PR enables monitoring for the Asset Manager Sidekiq queues as per [these instructions](https://docs.publishing.service.gov.uk/manual/setting-up-new-sidekiq-monitoring-app.html). 

The [companion PR](https://github.com/alphagov/sidekiq-monitoring/pull/32) against `alphagov/sidekiq-monitoring` adds the monitoring web interface.

PR #7071, opened 2018-01-11
AWS add missing redis parameters to asset-manager

afda16:
On AWS asset-manager is trying to connect with Redis
on localhost, it's missing the host and port parameters
already present in Carrenza Hieradata.


floehopper:
@afda16 I think these lines need to be copied over too, no? As per #6501.

https://github.com/alphagov/govuk-puppet/blob/master/hieradata/common.yaml#L527-L528

afda16:
I'll add them in another PR

floehopper:
👍

PR #7089, opened 2018-01-15
A clone dependent function for govuk_jenkins

kevindew:
It's common that in building a repo we may need to clone another repo, we've solved this problem in different ways: [exhibit A](https://github.com/alphagov/gds-api-adapters/blob/master/Jenkinsfile#L91-L117), [exhibit B](https://github.com/alphagov/publishing-api/blob/master/Jenkinsfile#L72-L99) and this can be quite verbose and lead to inconsistencies.

This PR introduces a new function `cloneDependent` which can be used to perform this operation in a consistent succinct method. Example:

```groovy
  govuk.checkoutDependent("publishing-api", [ branch: PUBLISHING_API_BRANCH ]) {
    stage("Run publishing-api pact") {
      govuk.bundleApp()
      lock("publishing-api-$NODE_NAME-test") {
        govuk.runRakeTask("db:reset")
        govuk.runRakeTask("pact:verify:branch[${env.BRANCH_NAME}]")
      }
    }
  }
```

This uses the GitSCM Jenkins plugin to clone the repositories and uses the same function that we use to clone the main one of a repo. I believe there's a bunch of upsides to using this such as smaller clones, more consistent reset handling. However there a couple of downsides. One is that it will show each checked out revision on the build screen, example: https://ci.integration.publishing.service.gov.uk/job/gds-api-adapters/job/jenkinsfile/29/ and the other that it will set env vars of GIT_URL, etc to the last repo cloned. From my searching we don't use any of those env vars so it doesn't seem like it would have an affect.

An example of this building is https://ci.integration.publishing.service.gov.uk/job/gds-api-adapters/job/jenkinsfile/29/ - there are also some further details in each commit message.

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -322,14 +322,48 @@ def cleanupGit() {
 /**
  * Checkout repo using SSH key
  */
-def checkoutFromGitHubWithSSH(String repository, String org = 'alphagov', String url = 'github.com') {
+def checkoutFromGitHubWithSSH(String repository, Map options = [:]) {
+  def defaultOptions = [
+    branch: null,
+    changelog: true,
+    directory: null,
+    org: "alphagov",
+    poll: true,
+    host: "github.com"
+  ]
+  options = defaultOptions << options
+
+  def branches
+  if (options.branch) {
+    branches = [[ name: options.branch ]]
+  } else {
+    branches = scm.branches
+  }
+
+  def extensions = []
+
+  if(options.directory) {
+    extensions << [
+      $class: "RelativeTargetDirectory",
+      relativeTargetDir: options.directory
+    ]
+  }
+
   withStatsdTiming("github_ssh_checkout") {
-    checkout([$class: 'GitSCM',
-      branches: scm.branches,
-      userRemoteConfigs: [[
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -368,6 +368,35 @@ def checkoutFromGitHubWithSSH(String repository, Map options = [:]) {
...
+    directory: "tmp/${repository}",
+    poll: false
+  ]
+  options = defaultOptions << options
+
+  stage("Cloning ${repository}") {
+    checkoutFromGitHubWithSSH(repository, options)
+  }
+
+  if (closure) {
```
suzannehamilton: It might be nice to wrap this in a stage (maybe using a name configured in `options`) so the pipeline output is clearer. But happy for you to leave it like this for now - we can revisit this if we do add any long processes which use the dependent repo.

kevindew: Yeah I wrestled with that for a little bit as I'd agree it'd be better in a stage. The auto stage naming approach was so vague to be a bit pointless "${repository} tasks" sort of thing. Whereas passing in a stage is probably less intuitive than writing your own within the closure. 

Happy to evaluate after a bit though if it causes problem.

kevindew:
@suzannehamilton Thanks for the review 👍

PR #7092, opened 2018-01-16
Add whitehall govuk_basic_auth_credentials to link checker api for integration & staging

SebAshton:
This PR adds a new env var to set the basic auth credentials for the link checker API to pass the basic auth challenge on integration.

### DIFF
--- a/hieradata/staging.yaml
+++ b/hieradata/staging.yaml
@@ -25,6 +25,7 @@ govuk::apps::email_alert_api::govuk_notify_base_url: 'https://api.staging-notify
 govuk::apps::email_alert_api::govuk_notify_template_id: '76d21ce7-54c3-4fb7-8830-ba3b79287985'
 govuk::apps::hmrc_manuals_api::publish_topics: false
 govuk::apps::kibana::logit_environment: d414187a-2796-4ea7-9b9a-d40c341646d6
+govuk::apps::link_checker_api::govuk_basic_auth_credentials: "%{hiera('http_username')}:%{hiera('http_password')}"
 govuk::apps::publicapi::backdrop_host: 'www.staging.performance.service.gov.uk'
 govuk::apps::publisher::run_fact_check_fetcher: false
 govuk::apps::publisher::fact_check_address_format: 'factcheck+staging-{id}@alphagov.co.uk'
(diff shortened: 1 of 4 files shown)


hieradata/staging.yaml
```diff
@@ -25,6 +25,7 @@ govuk::apps::email_alert_api::govuk_notify_base_url: 'https://api.staging-notify
 govuk::apps::email_alert_api::govuk_notify_template_id: '76d21ce7-54c3-4fb7-8830-ba3b79287985'
 govuk::apps::hmrc_manuals_api::publish_topics: false
 govuk::apps::kibana::logit_environment: d414187a-2796-4ea7-9b9a-d40c341646d6
+govuk::apps::link_checker_api::govuk_basic_auth_credentials: "%{hiera('http_username')}:%{hiera('http_password')}"
```
thomasleese: I think this will need the `Basic ` at the beginning of the string.

dwhenry: Any reason we haven't split these up as per previous use cases: https://github.com/alphagov/govuk-puppet/blob/a2376b163223eef8e2ab859dedd54716eab4c955/hieradata/common.yaml#L815

thomasleese: Actually... since we called the variable `govuk_basic_auth_credentials`, I think we should update https://github.com/alphagov/link-checker-api/pull/120/files#diff-4e80c9e759124f64feef221083d041dbR294 to put in the `Basic` for us.

thomasleese: @dwhenry No real reason except it fits what Whitehall used to do and it works better with how the Link Checker API is configured.

hieradata/staging.yaml
```diff
@@ -25,6 +25,7 @@ govuk::apps::email_alert_api::govuk_notify_base_url: 'https://api.staging-notify
 govuk::apps::email_alert_api::govuk_notify_template_id: '76d21ce7-54c3-4fb7-8830-ba3b79287985'
 govuk::apps::hmrc_manuals_api::publish_topics: false
 govuk::apps::kibana::logit_environment: d414187a-2796-4ea7-9b9a-d40c341646d6
+govuk::apps::link_checker_api::govuk_basic_auth_credentials: "Basic %{hiera('http_username')}:%{hiera('http_password')}"
```
thomasleese: Sorry I've changed my mind... I think we shouldn't include the `Basic` in the string here as the variable has `basic` in the name.

SebAshton: 🙃

PR #7093, opened 2018-01-16
Rename GOV.UK Rate Limit token for Link Checker API

thomasleese:
This better reflects that the token is only used for accessing GOV.UK.


thomasleese:
Depends on https://github.com/alphagov/govuk-secrets/pull/171

PR #7094, opened 2018-01-16
Build docker images in parallel in govuk_jenkins

kevindew:
Building docker images can take quite a substantial amount of time in an
applications build process - regularly 4 or 5 minutes - when there isn't a
recent cache available. This utilises the [parallel][1] command in
jenkins to run the docker parts of the build in a separate thread to the
rest of the build.

A disadvantage to running things in parallel is that the output is less
easy to read as things get mixed together. To try combat this I've set
docker build to run in quiet so it only has minimal output,
unfortunately there is not an option to pass the quiet arg to docker
push tag (we could get around this by running the push tag command via
shell but that seems a bit overkill).

This sets a project that doesn't have a docker file or is running a
schema test to run just the non docker tasks in the main thread.

What was rather tricky was naming for the functions and the labels for
the output. I went with the rather uninspiring `nonDockerBuildTasks` and
`dockerBuildTasks` function names with "build" and "docker" labels for
lack of a better idea.

[1]: https://www.cloudbees.com/blog/parallelism-and-distributed-builds-jenkins

An example of this exact script running is https://ci.integration.publishing.service.gov.uk/job/publishing-api/job/master/1099/ where we probably saved about 2 mins.

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -153,108 +153,13 @@ def buildProject(Map options = [:]) {
       setEnvar("DISPLAY", ":99")
     }
 
-    contentSchemaDependency(params.SCHEMA_BRANCH)
-
-    stage("bundle install") {
-      if (isGem()) {
-        bundleGem()
-      } else {
-        bundleApp()
-      }
-    }
-
-    if (hasLint()) {
-      stage("Lint Ruby") {
-        rubyLinter("app lib spec test")
-      }
+    if (hasDockerfile() && !params.IS_SCHEMA_TEST) {
+      parallel (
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -302,6 +207,106 @@ def buildProject(Map options = [:]) {
   }
 }
 
+def nonDockerBuildTasks(options, jobName, repoName) {
```
cbaines: I don't think the `nonDocker` prefix is adding much here, it might just cause confusion.

kevindew: Yeah I think it's a bad name, I think buildTasks has it conflicting too much with buildProject and it needs something to differentiate it from the docker equivalent. But something better escapes me.

kevindew:
Thanks for the review @cbaines 👍

PR #7097, opened 2018-01-16
AWS set deploy user to manage assets dir

afda16:
On VMware files in the uploads directory are owned by the 'assets'
user. This is required to facilitate SSH key sharing between the
asset-master and asset-slave machines. In the asset-master machine,
all file processing and file sync cron jobs belong to the 'assets'
user; data syncs from other environments are executed with the 'assets'
user, and files written on the directory from application processes
are owned by 'assets' via NFS settings.

On AWS we are not using NFS, we are using EFS instead. This is a problem
because applications are creating files with the app user 'deploy'. Then
file processing and environment sync jobs are failing because the 'assets'
user that executes those processes can't remove files created by 'deploy'.

On AWS we don't have the requirement of data sync between an asset master
and slave machines, so there is not reason for not using the 'deploy'
user to manage the data processes and own the uploads directory. This commit
updates the directories and cron jobs ownership on AWS to use 'deploy'.

### DIFF
--- a/modules/govuk/manifests/apps/publisher.pp
+++ b/modules/govuk/manifests/apps/publisher.pp
@@ -133,11 +133,17 @@ class govuk::apps::publisher(
     content => template('govuk/local_authority_import_check.erb'),
   }
 
+  if $::aws_migration {
+    $data_dir_user = 'deploy'
+  } else {
+    $data_dir_user = 'assets'
+  }
+
   file { ['/data/uploads/publisher', '/data/uploads/publisher/reports']:
     ensure => directory,
     mode   => '0775',
-    owner  => 'assets',
-    group  => 'assets',
+    owner  => $data_dir_user,
+    group  => $data_dir_user,
   }
 
   govuk::procfile::worker { $app_name:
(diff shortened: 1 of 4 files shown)


floehopper:
@afda16:

Thanks for doing this.

The changes look good to me, although I'm definitely not a Puppet expert! 😄 

Can I confirm that when you mention VMWare in the PR description, this relates to the old/non-AWS/Carrenza boxes?

When you've changed the owner/group of a directory, will that change be applied to all directories below that and/or does it need to be? I'm conscious that the directories start off as empty on a newly built machine, but will contain a large number of sub-directories & files on an already-built machine and am unsure how Puppet works in this respect.

Rebuilding the asset master box from scratch as suggested by @surminus sounds appealing in order to ensure the setup is correct & repeatable, although that would then mean re-running some form of sync for the existing files.

floehopper:
@afda16 One other thought: if you end up re-building a new asset-master box, it might be sensible for us to check it's all working OK (e.g. creating new assets and seeing them virus scanned) with empty directories *before* then sync-ing all the existing assets. That first check would confirm whether the new declarations in Puppet are correct in case the sync then affects things, e.g. altering the directory/file permissions.

afda16:
Hi @floehopper , it won't apply changes on subdirectories, only top directories listed in the Puppet file resources. I'll rebuild the master box and let you know for testing, thanks.

PR #7100, opened 2018-01-16
Remove the AWS conditional from the Plek config

cbaines:
For the draft frontend nodes.

As the removed comment says, this was done when the
PLEK_SERVICE_..._URI environment variables were set when using
AWS. This is no longer the case, and these environment variables need
setting for apps on the draft frontend boxes, to make sure they use
the right URLs (e.g. search not draft-search).


cbaines:
I've just realised that this is still wrong, I'll have a go at correcting it...

afda16:
Are these variables going to be used for user redirection or inter-app communications?. With this change those apps will use the publishing domain, so if the variables are used for communications between applications, it probably won't work. We want to use the internal domain

cbaines:
> Are these variables going to be used for user redirection or inter-app communications?. With this change those apps will use the publishing domain, so if the variables are used for communications between applications, it probably won't work. We want to use the internal domain

Yeah, you're right. I've opened another PR here, which will hopefully be more correct.

https://github.com/alphagov/govuk-puppet/pull/7101

PR #7101, opened 2018-01-16
Fix external vs internal routing for draft frontend apps

cbaines:
These environment variables need setting for apps on the draft
frontend boxes, to make sure they use the right URLs (e.g. search not
draft-search). Some of these services need accessing over the internal
domain, so make this the case.


cbaines:
> I think this makes sense.

Awesome, thanks for reviewing.

PR #7102, opened 2018-01-16
Configure Plek for Content Tagger on AWS

cbaines:
This should fix issues where Collections Publisher links to Content
Tagger using the internal domain.


cbaines:
I've been working on a more general approach since I put this up, so closing in favour of fixing this via https://github.com/alphagov/collections-publisher/pull/308

PR #7107, opened 2018-01-17
Set the GOVUK_APP_DOMAIN_EXTERNAL environment variable

cbaines:
This matches up with changes in version 2.1.0 of Plek, which adds
support for explicit external URLs, and can be configured through this
environment variable.


cbaines:
> Assuming nothing is currently relying on the GOVUK_APP_DOMAIN_INTERNAL env var which is being removed, then this looks good to me.

I think it's just Smokey, which I've got a Pull Request open for [1]. Let's integrate in integration! Thanks for reviewing :)

1: https://github.com/alphagov/smokey/pull/324

PR #7109, opened 2018-01-17
BAU: Name Change

suthagarht:
- I have amended my username from suthagarkathirkamathamby to
  suthagarht. The shorther version is already present on the verify
environment. Hence it will be easier to maintain the same format within
gov.uk.

Solo: @suthagarht


afda16:
We need to update users::usernames: in Hieradata

PR #7110, opened 2018-01-17
BAU: Name Change

suthagarht:
- Changing my name from the long version to suthagarht. This format is
  used in verify. Hence the change will help with consistency.

Solo: @suthagarht


issyl0:
You will also need to add yourself to `hieradata_aws/integration.yaml`.

PR #7113, opened 2018-01-18
AWS set UID/GID for deploy user on EFS dep boxes

afda16:
Following the work done in https://github.com/alphagov/govuk-puppet/pull/7097
we need to ensure the UID/GID of the 'deploy' user is the same on all the machines
that share the EFS volume.

Without this setting, files can appear as belonging to different
users, for instance, currently the whitehall-backend machine uses the deploy
user (1058) which == jonathanhallam on asset-master.

At the moment we are forcing this change on asset-master, backend
and whitehall-backend machines, that are the only ones sharing
the volume on Integration AWS. For Staging and Production, we should
probably have the same setting everywhere, and at some point
change it in the rest of Integration too.


afda16:
It should be generated, same as now: https://docs.puppet.com/puppet/3.8/types/user.html#user-attribute-uid

modules/govuk/manifests/deploy/setup.pp
```diff
@@ -24,27 +29,37 @@
     $aws_ses_smtp_password,
     $gemstash_server = 'http://gemstash.cluster',
     $ssh_keys = { 'not set in hiera' => 'NONE_IN_HIERA' },
+    $deploy_uid = undef,
+    $deploy_gui = undef,
```
rubenarakelyan: `deploy_gid`?

chrisroos: Did you see this typo, @afda16? It looks like you need to replace `$deploy_gui` with `$deploy_gid`.

PR #7123, opened 2018-01-22
Update the Publishing API Worker ... grafana dashboard

cbaines:
The Graphite metrics are now recorded per machine, so adjust the
queries accordingly.


cbaines:
Thanks @sihugh :)

PR #7124, opened 2018-01-22
Serve Whitehall's image data attachments from Asset Manager

chrislo:
c.f. https://github.com/alphagov/asset-manager/issues/404

This commit proxies all requests for Whitehall image data assets to
Asset Manager.

We have migrated all existing Whitehall assets under
`system/uploads/image_data/file` on the NFS mount to Asset
Manager[1]. We've been uploading all new assets of this type to Asset
Manager since alphagov/whitehall#3602.

[1] https://github.com/alphagov/asset-manager/issues/404#issuecomment-359395022


chrislo:
Thanks @floehopper.

PR #7134, opened 2018-01-23
Further removes @fields to fix nginx log graphs

elenatanasoiu:
For: https://trello.com/c/Y9Rc6jWR/49-http-stats-not-being-logged-in-graphite

Following a PR to remove "@fields" from the logs nginx template (see:
 #6658), a bunch of graphs in grafana and graphite have stopped reporting data (from 11th of Jan 2018).

screen shot 2018-01-23 at 14 04 52

This PR aims to remove the mentions of "@fields" in order to bring back the graphs.

There is a planned PR from branch `remove-fields-prefix-from-nginx-config` which will rename `@fields` to `access`, as well as renaming some of the fields. You can see this below: 

screen shot 2018-01-23 at 14 03 20

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -957,7 +957,7 @@ grafana::dashboards::deployment_applications:
     # logstasher >1.x
     fields_prefix: ''
   hmrc-manuals-api:
-    # Missing @fields.status
+    # Missing status from logs
     show_controller_errors: false
   imminence:
     # logstasher >1.x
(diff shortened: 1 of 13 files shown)


h-lame:
This looks sensible to me, but the real test is to put it on our infrastructure and see if we start getting logs again.

floehopper:
@elenatanasoiu Thanks for fixing this! 😄  /cc @chrisroos

PR #7137, opened 2018-01-23
Fixes for draft stack in development VM

floehopper:
I've recently been running the draft stack on my development VM and ran into a couple of issues:

* Unlike other apps, the authenticating-proxy app had no OAuth credentials by default. This made it more work to set the credentials on the app in the Signon app. By adding default credentials like the other apps, we can make life a bit simpler.

* The `PLEK_SERVICE_SIGNON_URI` environment variable for the authenticating-proxy app was being set to a URI with the `https` scheme, but none of the apps (including Signon) run with SSL on the development VM and so when attempting to authenticate a user by redirecting to Signon, the request would fail.


floehopper:
@cbaines Thanks for reviewing!

PR #7141, opened 2018-01-24
Split audit tool

chao-xian:
We are splitting out the Content Audit Tool from the Content Performance Manager. We are effectively freezing development on the Audit Tool, but it is in beta with users at the moment and needs to be available.

However we are going to be developing the Content Performance Manager extensively and as the Audit Tool is so coupled it is a risk to keep the two together - changes CPM code may break the Audit Tool and becomes a maintenance headache. This is the best, cleanest option at present.

### DIFF
--- a/hieradata/production.yaml
+++ b/hieradata/production.yaml
@@ -68,6 +68,8 @@ base::supported_kernel::enabled: false
 
 environment_ip_prefix: '10.3'
 
+govuk::apps::content_audit_tool::ensure: 'absent'
+
 govuk::apps::asset_manager::aws_s3_bucket_name: 'govuk-assets-production'
 govuk::apps::asset_manager::aws_region: 'eu-west-1'
 govuk::apps::content_store::performance_platform_big_screen_view_url: 'https://performance-platform-big-screen-view-production.cloudapps.digital'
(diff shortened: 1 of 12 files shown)


hieradata/production.yaml
```diff
@@ -68,6 +68,8 @@ base::supported_kernel::enabled: false
 
 environment_ip_prefix: '10.3'
 
+govuk::apps::content_audit_tool::ensure: 'absent'
```
cbaines: Are you planning to change this to present at some point?

chao-xian: @cbaines yep. We need to coordinate with a dept on when we can do the switchover. I thought it was a good idea to use this flag for now and not worry about the production deployment just yet.

cbaines: Ok, that sounds fine :)

modules/govuk/manifests/apps/content_audit_tool.pp
```diff
@@ -0,0 +1,157 @@
+# == Class: govuk::apps::content_audit_tool
+#
+# App details at: https://github.com/alphagov/content-audit-tool
+#
+# === Parameters
+#
+# [*ensure*]
+# Whether the application should be exist. Can be specified for each
+# environment using deployment hieradata.
```
cbaines: There is some inconsistent indentation here.

modules/govuk/manifests/apps/content_audit_tool/db.pp
```diff
@@ -0,0 +1,23 @@
+# == Class: govuk::apps::content_audit_tool::db
+#
+# === Parameters
+#
+# [*password*]
+#   The DB instance password.
+#
+# [*backend_ip_range*]
+#   Backend IP addresses to allow access to the database.
+#
```
cbaines: The rds parameter is missing from the list in the comment.

chao-xian: Addressed in https://github.com/alphagov/govuk-puppet/pull/7141/commits/9c4dedacec13d8f0f2fbdf3dd1f0620a84e4c920

PR #7142, opened 2018-01-25
Change email-alert-api template ID for Notify

rubenarakelyan:
This commit adds/updates the Notify template IDs used by email-alert-api to reflect the separation of the production/staging/integration services.

Trello: https://trello.com/c/pvDExpMk/551-set-up-notify-callbacks-for-integration-staging

### DIFF
--- a/hieradata/staging.yaml
+++ b/hieradata/staging.yaml
@@ -19,8 +19,7 @@ govuk::apps::email_alert_api::db::backend_ip_range: '10.2.3.0/24'
 govuk::apps::email_alert_api::govdelivery_account_code: 'UKGOVUK'
 govuk::apps::email_alert_api::govdelivery_hostname: 'stage-api.govdelivery.com'
 govuk::apps::email_alert_api::govdelivery_public_hostname: 'stage-public.govdelivery.com'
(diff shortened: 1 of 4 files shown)


hieradata/staging.yaml
```diff
@@ -20,7 +20,7 @@ govuk::apps::email_alert_api::govdelivery_account_code: 'UKGOVUK'
 govuk::apps::email_alert_api::govdelivery_hostname: 'stage-api.govdelivery.com'
 govuk::apps::email_alert_api::govdelivery_public_hostname: 'stage-public.govdelivery.com'
 govuk::apps::email_alert_api::govuk_notify_base_url: 'https://api.staging-notify.works'
```
thomasleese: Should we also stop using Notify's staging environment now?

rubenarakelyan: I thought we might want to run some more load tests there with digests but let me know if not and I can remove it.

thomasleese: I'm just thinking that we can't change the template ID or API key without also changing back to use the production environment because otherwise they won't match.

I suspect we won't need to do any more load testing on staging, so I reckon we can stop using it.

rubenarakelyan: Cool, I've removed it.

PR #7148, opened 2018-01-26
Add draft-assets Nginx virtual host to development VM

floehopper:
We need this host to allow Asset Manager to protect draft assets behind SSO. We're only adding the new virtual host to the development VM for the moment, because we think we need to create SSL certificates, add DNS entries, and create new ELB instances to make use of this in other environments.

The configuration of the new virtual host is based heavily on that of the assets-origin virtual host. One significant difference is that it proxies all requests directly to Asset Manager - rather than doing this via the static virtual host. Note that it is important that the SSO-related paths (i.e. `/auth/gds/*`) as well as the Mainstream asset paths (i.e. `/media/*`) and Whitehall asset paths (i.e. `/government/uploads/*`).

We have intentionally not added a `proxy_set_header` directive to set the `Host` request header inside the `location` block, because this would mean that none of the `X-*` request headers set in the outer `server` block would be set. In particular Asset Manager will need to be able to rely on the value of `X-Forwarded-Host` being set correctly in order to determine whether an asset request came via `assets-origin` or `draft-assets`.

The new virtual host can be enabled in other environments by adding the following snippet to the body of `modules/router/manifests/nginx.pp`.

    class { 'router::draft_assets':
      real_ip_header => $real_ip_header,
    }

We have done some preliminary testing of the above snippet in the integration environment and it seemed to work correctly.

### DIFF
--- a/modules/router/manifests/draft_assets.pp
+++ b/modules/router/manifests/draft_assets.pp
@@ -1,0 +1,32 @@
+# = Class: router::draft_assets
+#
+# Configure vhost for serving draft assets.
+#
+# === Parameters
+#
+# [*real_ip_header*]
+#   Uses the Nginx realip module (http://nginx.org/en/docs/http/ngx_http_realip_module.html)
+#   to change the client IP address to the one in the specified HTTP header.
+#
+# [*vhost_name*]
+#   Primary vhost for draft assets
+#
+
+class router::draft_assets(
+  $real_ip_header = '',
+  $vhost_name = 'draft-assets',
+) {
+  $enable_ssl = hiera('nginx_enable_ssl', true)
+
+  if $::aws_migration {
+    $app_domain = hiera('app_domain_internal')
+    $upstream_ssl = true
+  } else {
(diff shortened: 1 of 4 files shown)


modules/router/manifests/draft_assets.pp
```diff
@@ -0,0 +1,32 @@
...
+) {
+    $enable_ssl = hiera('nginx_enable_ssl', true)
+
+    if $::aws_migration {
+    $app_domain = hiera('app_domain_internal')
+    $upstream_ssl = true
+  } else {
+    $app_domain = hiera('app_domain')
+    $upstream_ssl = $enable_ssl
+  }
```
chrisroos: It looks as though some of the indentation isn't quite right here.

floehopper: @chrisroos You're right. I'll fix that and get this merged. Thanks!

floehopper:
I've amended the commit mentioned by @chrisroos and force-pushed.

PR #7149, opened 2018-01-26
BAU: Adding Hiera Lookup

suthagarht:
- We are adding a hiera lookup for the whitehall mysql DB password.

Pair: @surminus @suthagarht 

https://trello.com/c/licP4V2x/925-build-staging-in-aws-and-review


suthagarht:
The actual job is not needed on staging . Laura is creating a PR to copy the correct jobs.

surminus:
Ref: https://github.com/alphagov/govuk-puppet/pull/7151

PR #7156, opened 2018-01-29
Move default Jenkins Jobs to common.yaml

surminus:
The motivation behind this change is that we need to overwrite the settings for the `jenkins` node. While default configuration lives in the class based `jenkins.yaml`, we would have to create a new hierarchy for `class/staging/jenkins.yaml` to overwrite it. This isn't really required and just adds more places to look to find different data.

If we move the default configuration to `common.yaml`, then we can overwrite it using `staging.yaml`, which is clearer to understand.

This commit also removes some unneeded hieradata related to deploying stuff with vCloud Tools.

https://trello.com/c/licP4V2x/925-build-staging-in-aws-and-review

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -755,6 +755,43 @@ govuk_jenkins::jobs::integration_deploy::applications: *deployable_applications
 govuk_jenkins::jobs::deploy_lambda_app::lambda_apps:
   - 'email_alert_notifications'
 
+govuk_jenkins::config::banner_string: 'AWS Integration'
+govuk_jenkins::config::banner_colour_background: '#005EA5'
+govuk_jenkins::config::banner_colour_text: 'white'
+govuk_jenkins::config::theme_colour: '#005EA5'
+govuk_jenkins::config::theme_text_colour: 'white'
+govuk_jenkins::config::theme_environment_name: 'AWS Integration'
+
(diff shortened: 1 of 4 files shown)


hieradata_aws/common.yaml
```diff
@@ -755,6 +755,43 @@ govuk_jenkins::jobs::integration_deploy::applications: *deployable_applications
 govuk_jenkins::jobs::deploy_lambda_app::lambda_apps:
   - 'email_alert_notifications'
 
+govuk_jenkins::config::banner_string: 'AWS Integration'
+govuk_jenkins::config::banner_colour_background: '#005EA5'
+govuk_jenkins::config::banner_colour_text: 'white'
+govuk_jenkins::config::theme_colour: '#005EA5'
+govuk_jenkins::config::theme_text_colour: 'white'
+govuk_jenkins::config::theme_environment_name: 'AWS Integration'
+
```
issyl0: These names, where do they get changed per environment now?

surminus: The default config shows Integration. The overrides happen in `staging.yaml` and `production.yaml` respectively.

issyl0: Sorry, it helps if I read. :-) `"If we move the default configuration to common.yaml, then we can overwrite it using staging.yaml, which is clearer to understand."`

PR #7158, opened 2018-01-29
Remove private-asset-manager vhost alias

floehopper:
Also removes the associated environment variable which made the vhost name available to the Asset Manager Rails app.

This virtual host alias was never used and the related functionality in Asset Manager was removed in [this PR][1].

Note that we will need to open a separate PR to remove the environment variable definition once this PR has been merged and applied in all environments.

[1]: https://github.com/alphagov/asset-manager/pull/426


floehopper:
Fixing lint error and force-pushing.

PR #7159, opened 2018-01-29
Move db env vars into ensure block

chao-xian:
These should have been placed inside the ensure block in previous
PR https://github.com/alphagov/govuk-puppet/pull/7141

This is to prevent puppet run errors when it tries to create
env vars for an application that is 'absent' on the staging
and production environments.


chao-xian:
[(3) Split Audit Tool](https://trello.com/c/IqfARxlP/21-3-split-audit-tool)

PR #7160, opened 2018-01-29
BAU YAML Syntax Change

suthagarht:
- We identified a syntatctical error graphite hiera lookup. This has
  been corrected via this change.

**Symptom**
We received the following error while performing a puppet run.
`Error: Error from DataBinding 'hiera' while looking up 'govuk::node::s_base::node_apps': (<unknown>): did not find expected key while parsing a block mapping at line 2 column 1 on node i-0a432be76092fd5b4
Error: Error from DataBinding 'hiera' while looking up 'govuk::node::s_base::node_apps': (<unknown>): did not find expected key while parsing a block mapping at line 2 column 1 on node i-0a432be76092fd5b4`

**Prognosis**
This event is very recent hence a change made recently is the most probable cause.

**Diagnosis**
Checked recent changes via the following resources.
[Puppet Manifest](https://validate.puppet.com/)
[YAML Lint](http://www.yamllint.com/)

We noticed that the issue was related to a recent commit #7027.

Solo: @suthagarht


surminus:
Can we post the error this caused Puppet to throw in the commit message? 

It might help someone in the future.

PR #7162, opened 2018-01-29
Separate app specific and user generated asset routes

chrisroos:
We were using the `router::assets_origin::asset_routes` hash to configure nginx `location` blocks for app-specific assets and for assets served by asset-manager. I've split this `asset_routes` hash into the `app_specific_asset_routes` hash and the `asset_manager_routes` array to make the dual purpose clearer.

Extracting the `asset_manager_routes` array allowed me to remove some duplication from the nginx config for the static virtual host.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1638,6 +1638,7 @@ router::assets_origin::app_specific_asset_routes:
   '/smartanswers/': "smartanswers"
 
 router::assets_origin::asset_manager_routes:
+  - '/media/'
   - '/government/uploads/government/uploads/system/uploads/'
   - '/government/uploads/system/uploads/organisation/logo/'
   - '/government/uploads/system/uploads/classification_featuring_image_data/file/'
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1630,6 +1630,14 @@ router::assets_origin::asset_routes:
   '/government/assets/': "whitehall-frontend"
   '/government/placeholder': "whitehall-frontend"
   '/government/uploads/': "whitehall-frontend"
+  '/government-frontend/': "government-frontend"
+  '/info-frontend/': "info-frontend"
+  '/manuals-frontend/': "manuals-frontend"
+  '/licencefinder/': "licencefinder"
+  '/service-manual-frontend/': "service-manual-frontend"
+  '/smartanswers/': "smartanswers"
+
+router::assets_origin::asset_manager_routes:
   '/government/uploads/government/uploads/system/uploads/': "static"
   '/government/uploads/system/uploads/organisation/logo/': "static"
   '/government/uploads/system/uploads/classification_featuring_image_data/file/': "static"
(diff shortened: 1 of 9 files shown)


modules/router/templates/assets_origin.conf.erb
```diff
@@ -61,6 +61,10 @@ server {
     proxy_pass <%= @upstream_ssl ? 'https' : 'http' %>://static.<%= @app_domain %>;
   }
 
+  location /media/ {
+    proxy_pass <%= @upstream_ssl ? 'https' : 'http' %>://static.<%= @app_domain %>;
+  }
+
```
floehopper: I think it would be worth mentioning in the commit note that previously we were relying on the implicit behaviour in the final `location` block for paths matching `/`.

hieradata/common.yaml
```diff
@@ -1638,6 +1638,7 @@ router::assets_origin::app_specific_asset_routes:
   '/smartanswers/': "smartanswers"
 
 router::assets_origin::asset_manager_routes:
+  - '/media/'
```
floehopper: I think it might be worth mentioning in the commit note that you're doing this to DRY up `assets_origin.conf.erb`.

lib/puppet-lint/plugins/check_hiera.rb
```diff
@@ -49,6 +49,10 @@
     # metrics
     'govuk_logging::logstream::metrics_only',
 
+    # Share the array of asset_manager_routes between assets-origin and static
+    # nginx virtual host configuration
+    'router::assets_origin::asset_manager_routes',
```
floehopper: Minor: This comment suggests that it's this line of code which is doing the sharing, but if I understand correctly, it's merely *allowing* the sharing. It would be good to update the comment accordingly.

floehopper: Also did you consider extracting `asset_manager_routes` as a top-level heira variable so that you didn't need this? It feels a bit of a shame to have to add this exception to the lint-ing.

floehopper:
Note that a couple of my comments are on out-dated diffs... there are 4 comments in all.

modules/router/templates/assets_origin.conf.erb
chrisroos: I've added the following to the commit note:

> Note that requests for /media/ would've previously been handled by the catch-all route (`location /`) in this file.

lib/puppet-lint/plugins/check_hiera.rb
chrisroos: I've updated the comment to:

> # Allow sharing of asset_manager_routes array between assets-origin and static
> # nginx virtual host configuration

hieradata/common.yaml
chrisroos: I've added the following to the commit note:

> In preparation for being able to reuse the `router::assets_origin::asset_manager_routes` array in the nginx config file for the static virtual host.

lib/puppet-lint/plugins/check_hiera.rb
chrisroos: > Also did you consider extracting asset_manager_routes as a top-level heira variable so that you didn't need this? It feels a bit of a shame to have to add this exception to the lint-ing.

I don't think that would've avoided the problem, unfortunately. We'd still need to be able to lookup that top-level hiera variable from within assets-origin and static config files, and that lookup would require an exception to be added to this file.

chrisroos:
Thanks for the review, @floehopper. I've addressed your feedback, rebased on master and force-pushed in preparation for merging.

PR #7171, opened 2018-01-29
Add a jenkins job to extract data from graphite

MatMoore:
This job extracts error rate data out of graphite for all the GOV.UK applications.

I'd like to store this data in an accessible place, like google drive or an s3 bucket, but I haven't got that far yet. This measure can be used to understand how GOV.UK-the-programme is doing so we want to make it easy for non-developers to access.

Basically this change takes us from this:
screen shot 2018-01-31 at 13 47 16

To this:
screen shot 2018-01-31 at 13 50 17

See https://github.com/alphagov/app-performance-summary/pull/1

Note: I had to use `app_domain_internal` to get this to work on AWS, but that variable wasn't defined on prod/staging, so I set it to be the same as `app_domain`.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/extract_app_performance.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/extract_app_performance.yaml.erb
@@ -19,5 +19,6 @@
         - shell: |
             PIPENV_VENV_IN_PROJECT=1 pipenv install --three --skip-lock
             PIPENV_VENV_IN_PROJECT=1 GRAPHITE_URL='https://graphite.<%= app_domain_internal %>' pipenv run python run_summary.py
+            PIPENV_VENV_IN_PROJECT=1 pipenv run python lint_data.py output/*.csv
     triggers:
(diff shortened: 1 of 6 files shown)


modules/govuk_jenkins/templates/jobs/extract_app_performance.yaml.erb
```diff
@@ -19,5 +19,6 @@
         - shell: |
             PIPENV_VENV_IN_PROJECT=1 pipenv install --three --skip-lock
             PIPENV_VENV_IN_PROJECT=1 GRAPHITE_URL='https://graphite.<%= app_domain_internal %>' pipenv run python run_summary.py
+            PIPENV_VENV_IN_PROJECT=1 pipenv run python lint_data.py output/*.csv
```
suzannehamilton: Will the output directory get created automatically? I don't see it in the repo.

MatMoore: yeah, the script creates it

PR #7178, opened 2018-01-30
Set GOVUK_APP_DOMAIN_EXTERNAL when precompiling assets

cbaines:
This is new in Plek 2.1.0, so set this along with the
GOVUK_APP_DOMAIN.


cbaines:
> It would be nice if there was a way we could avoid having to set these manually in here and just rely on the variables set in the environment. However, given that a bunch of similar variables are already set in here, this doesn't seem to make things much worse. And this change is blocking another PR, so I'm happy to approve it.

Yeah, I don't think it's necessary for apps to use Plek when performing asset pre-compilation, but apparently they do, maybe because its part of the Rails initialisation process...

PR #7179, opened 2018-01-30
Terraform deployment job

surminus:
This creates a job on ci-deploy that can take some credentials and trigger a build of a `./jenkins.sh` script that lives in the repo.

I'll need to do the stuff for `sops` as well before this will work.

Related: https://github.com/alphagov/govuk-aws/pull/472

https://trello.com/c/qAzTAU4p/952-build-box-spike-on-adding-tf-deploys-to-current-integration-carrenza-deploy-jenkins

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_terraform_govuk_aws.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_terraform_govuk_aws.yaml.erb
@@ -1,0 +1,73 @@
+---
+- scm:
+    name: govuk-aws_Deploy_Terraform_Govuk_AWS
+    scm:
+        - git:
+            url: git@github.com:alphagov/govuk-aws.git
+            branches:
+                - master
+- job:
+    name: Deploy_Terraform_Project
+    display-name: Deploy_Terraform_Project
+    project-type: freestyle
+    description: |
+        Deploy GOV.UK AWS Infrastructure using Terraform.
+
+        For available projects to deploy, please see the <a href=https://github.com/alphagov/govuk-aws/tree/master/terraform/projects>Terraform repository</a>
+
+    properties:
+        - github:
+            url: https://github.com/alphagov/govuk-aws/
+    scm:
+        - govuk-aws_Deploy_Terraform_Govuk_AWS
+    wrappers:
+        - ansicolor:
+            colormap: xterm
+        - build-name:
+          name: '${ENV,var=ENVIRONMENT} ${ENV,var="STACKNAME"} ${ENV,var="PROJECT"} ${ENV,var="COMMAND"}'
+    builders:
+        - shell: |
+            ./jenkins.sh
+
+    parameters:
+        - string:
+            name: AWS_ACCESS_KEY_ID
+            description: Your AWS access key ID
+            default: false
+        - password:
+            name: AWS_SECRET_ACCESS_KEY
+            description: Your AWS secret access key
+            default: false
+        - password:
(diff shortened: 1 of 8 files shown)


modules/govuk_jenkins/templates/jobs/deploy_terraform_govuk_aws.yaml.erb
```diff
@@ -0,0 +1,66 @@
...
+                - production
+        - choice:
+            name: STACKNAME
+            description: Choose which stack to deploy into
+            choices:
+                - blue
+                - green
+        - string:
+            name: PROJECT
+            description: Name of the project you wish to deploy
```
issyl0: Can we make this a `choice` and include all of the `infra-` and `app-` projects programatically somehow, or is that way too much work? Or we could make it a list? But then that's _another_ place to update when we add a new one? Or is people typing still OK?

surminus: I've [been there before](https://github.com/alphagov/govuk-puppet/commit/64a8df8eab)! I'm not opposed to it, but in my experience it was a pain to keep updated and didn't bring a huge amount of benefit. I'm not entirely opposed to it if people would prefer a dropdown for usability.

issyl0: Ah! OK then. It is another place to update which would get in the way a bit, but then so will mis-typing a project and waiting for Jenkins to return a failure. But people can complain abut this when/if they start using it, I suppose. Can we add a "Projects currently available can be found in Github" line like in your linked PR?

surminus: Sure :+1:

surminus:
I've updated this again to include a `TERRAFORM_VERSION` parameter. Context for that change is [here](https://github.com/alphagov/govuk-aws/pull/472).

modules/govuk_jenkins/templates/jobs/deploy_terraform_govuk_aws.yaml.erb
```diff
@@ -0,0 +1,72 @@
...
+            name: AWS_SECRET_ACCESS_KEY
+            description: Your AWS secret access key
+            default: false
+        - password:
+            name: AWS_SESSION_TOKEN
+            description: The AWS session token
+            default: false
+        - choice:
+            name: COMMAND
+            description: Choose whether to plan (default), apply or destroy
```
issyl0: Where do we set `plan` as default? Or is that a Terraform thing?

surminus: It's just how the Jenkins dropdown works (the top is always the default)

surminus:
I've also had a thought - we currently deploy stuff using the [older Terraform code](https://github.com/alphagov/govuk-terraform-provisioning) on the instance we wish to deploy this from, so we won't be able to install the newer version of Terraform until we've moved that job across. I'll investigate this.

modules/govuk_jenkins/templates/jobs/deploy_terraform_govuk_aws.yaml.erb
jljalderman: Do we need govuk as a stackname option too?

surminus: We do, good spot.

PR #7184, opened 2018-01-31
Add a graphite-compatible regex to match either disk name on AWS

issyl0:
- In AWS, disks are named `xvd...` or `nvm...` depending on the instance type.


surminus:
The correct query seems to be `apt-1_blue_integration.disk-{xvd*,nvm*}?.disk_time.*`, so with a comma `,` rather than a pipe `|` within the value list.

issyl0:
D'oh. My brain was half in normal regex and half in Graphite regex.

PR #7189, opened 2018-01-31
Respond with 404 for unrecognised requests to draft-assets

chrisroos:
This avoids us falling through to the default nginx host for requests that we don't recognise; which will mean we avoid rendering the default nginx welcome page at the root of draft-assets.

### DIFF
--- a/modules/router/templates/draft-assets.conf.erb
+++ b/modules/router/templates/draft-assets.conf.erb
@@ -32,6 +32,11 @@ server {
   add_header "Access-Control-Allow-Methods" "GET, OPTIONS";
   add_header "Access-Control-Allow-Headers" "origin, authorization";
 
+  # Avoid falling through to the default host if none of our locations match
+  location / {
+    return 404;
+  }
+
   location /auth/ {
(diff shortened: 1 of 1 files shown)


modules/router/templates/draft-assets.conf.erb
```diff
@@ -32,6 +32,11 @@ server {
   add_header "Access-Control-Allow-Methods" "GET, OPTIONS";
   add_header "Access-Control-Allow-Headers" "origin, authorization";
 
+  # Avoid falling through to the default host if none of our locations match
+  location / {
+    return 404;
+  }
+
```
floehopper: Very minor: I might be tempted to put this location block _below_ the other one, since it's less important. However, I haven't looked at how this is done in other Nginx config files.

chrisroos:
Thanks @floehopper. I've rebased this on master and force pushed in preparation for merging.

modules/router/templates/draft-assets.conf.erb
chrisroos: I don't feel particularly strongly about it so I've left it as it is.

PR #7194, opened 2018-02-01
Make safeDockerTag in the jenkins build script actually safe

h-lame:
The [docker tag documentation](https://docs.docker.com/engine/reference/commandline/tag/)
says that a docker tag must be valid ascii, contain only lowercase, uppercase,
digits, underscore, dash and period characters.  It can't start with a dash or
a period, and it can only be 128 chars long.  Our previous version only replaced
slashes with underscores so didn't entirely guarantee the tag was safe.

Discovered when I used a `+` in the branch name for alphagov/whitehall#3471


adrianclay:
I'm getting some failures on Jenkins related to this change:

https://ci.integration.publishing.service.gov.uk/job/publishing-e2e-tests/job/run-tests-continually/36/console

h-lame:
Oh! It worked in [an online groovy console](https://groovyconsole.appspot.com/).  I'll revert and try again.

PR #7201, opened 2018-02-02
BAU: Add Roch Trinque

suthagarht:
- We have a new team member called Roch Trinque. This commmit adds a new
  user account for Roch.

Pair: @suthagarht @rtrinque


surminus:
We should get Roch to do raise this as a first taster of the process :)

PR #7202, opened 2018-02-02
Enable SSL for draft-assets

chrisroos:
This addresses https://github.com/alphagov/asset-manager/issues/430.

This will allow us to start responding to requests to draft-assets in staging and production. I've copied the necessary bits of the config from assets_origin.pp and assets-origin.conf.erb.

Note that draft-assets doesn't require SSL to be configured on the virtual host in integration because we're terminating SSL requests at the ELB.


chrisroos:
Thanks @floehopper! I'm going to get this merged.

PR #7203, opened 2018-02-02
Define unicorn workers as a govuk::app::config thing and make local-links-manager, whitehall, signon use it

h-lame:
We've upgraded local-links-manager to configure unicorn via
govuk_app_config, and the expected way to change the number of worker
processes used by unicorn is via an env var picked up by govuk_app_config,
rather than hardcoded directly in the apps own config/unicorn.rb.

We set the workers to 4 because local-links-manager has had 4 unicorn
workers since alphagov/local-links-manager#52

Note: this should be merged and deployed before https://github.com/alphagov/local-links-manager/pull/181 otherwise we'll have a blip where local-links-manager only has 2 workers (the default when no env var is present).

### DIFF
--- a/modules/govuk/manifests/apps/local_links_manager.pp
+++ b/modules/govuk/manifests/apps/local_links_manager.pp
@@ -79,6 +79,10 @@
 #   to GA.
 #   Default: false
 #
+# [*unicorn_worker_processe*s]
+#   The number of unicorn worker processes to run
+#   Default: undef
+#
 class govuk::apps::local_links_manager(
   $port = 3121,
   $enabled = true,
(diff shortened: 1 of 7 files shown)


modules/govuk/manifests/apps/local_links_manager.pp
```diff
@@ -104,6 +104,7 @@
   $run_links_ga_export = false,
 ) {
   $app_name = 'local-links-manager'
+  $unicorn_worker_processes = 4
```
dwhenry: would this make more sense in the hiera data instead of being declared inline (I know this is whats done for whitehall)

h-lame: shrug.  I just followed what whitehall does - but no idea really if that's the preferred / best way.

h-lame: Actually, after talking about it on Slack, it seems like using hieradata might be a better option.  I'll change it.

PR #7212, opened 2018-02-05
Set unicorn worker processes for content store

kevindew:
This moves this configuration which is currently hardcoded into content
store into an environment variable.

### DIFF
--- a/modules/govuk/manifests/apps/content_store.pp
+++ b/modules/govuk/manifests/apps/content_store.pp
@@ -45,6 +45,9 @@
 # [*secret_key_base*]
 #   The key for Rails to use when signing/encrypting sessions.
 #
+# [*unicorn_worker_processes*]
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/apps/content_store.pp
```diff
@@ -89,6 +93,13 @@
     }
   }
 
+  if $unicorn_worker_processes {
+    govuk::app::envvar { "${title}-UNICORN_WORKER_PROCESSES":
```
h-lame: In #7203 I added this to govuk::app, so you can replace this stanza with passing `$unicorn_worker_processes` into the govuk::app call on line68 above.

kevindew: Oh excellent. I'll get that sorted out.

kevindew:
@h-lame I've just updated this, does this look right to you?

PR #7224, opened 2018-02-07
Avoid asset requests going via static

chrisroos:
It seems unnecessary to route these requests to asset-manager via static.
    
Routing these requests directly to asset-manager mirrors the configuration in draft-assets.

### DIFF
--- a/lib/puppet-lint/plugins/check_hiera.rb
+++ b/lib/puppet-lint/plugins/check_hiera.rb
@@ -49,7 +49,7 @@ PuppetLint.new_check(:hiera_explicit_lookup) do
     # metrics
     'govuk_logging::logstream::metrics_only',
 
-    # Allow sharing of asset_manager_uploaded_assets_routes array between assets-origin and static
+    # Allow sharing of asset_manager_uploaded_assets_routes array between assets-origin and draft-assets
     # nginx virtual host configuration
     'router::assets_origin::asset_manager_uploaded_assets_routes',
 
--- a/modules/router/templates/assets_origin.conf.erb
+++ b/modules/router/templates/assets_origin.conf.erb
@@ -49,7 +49,16 @@ server {
 
   <%- @asset_manager_uploaded_assets_routes.each do |path| -%>
   location <%= path %> {
-    proxy_pass <%= @upstream_ssl ? 'https' : 'http' %>://static.<%= @app_domain %>;
+    # Explicitly re-include Strict-Transport-Security header, this
+    # forces nginx not to clear Cache-Control headers further up the
+    # stack.
+    include /etc/nginx/add-sts.conf;
+
+    add_header "Access-Control-Allow-Origin" "*";
+    add_header "Access-Control-Allow-Methods" "GET, OPTIONS";
+    add_header "Access-Control-Allow-Headers" "origin, authorization";
+
+    proxy_pass <%= @upstream_ssl ? 'https' : 'http' %>://asset-manager.<%= @app_domain %>;
   }
   <%- end -%>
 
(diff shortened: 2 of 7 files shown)


lib/puppet-lint/plugins/check_hiera.rb
```diff
@@ -49,7 +49,7 @@
     # metrics
     'govuk_logging::logstream::metrics_only',
 
-    # Allow sharing of asset_manager_uploaded_assets_routes array between assets-origin and static
+    # Allow sharing of asset_manager_uploaded_assets_routes array between assets-origin and draft-assets
```
floehopper: Minor: This is more of a correction than a clarification, isn't it?

modules/router/templates/assets_origin.conf.erb
```diff
@@ -49,7 +49,16 @@ server {
...
+    # Explicitly re-include Strict-Transport-Security header, this
+    # forces nginx not to clear Cache-Control headers further up the
+    # stack.
+    include /etc/nginx/add-sts.conf;
+
+    add_header "Access-Control-Allow-Origin" "*";
+    add_header "Access-Control-Allow-Methods" "GET, OPTIONS";
+    add_header "Access-Control-Allow-Headers" "origin, authorization";
+
+    proxy_pass <%= @upstream_ssl ? 'https' : 'http' %>://asset-manager.<%= @app_domain %>;
```
floehopper: Minor: I think it might be worth explaining in the commit note that `@upstream_ssl` incorporates the logic that was in `@enable_ssl || @proxy_pass_asset_manager_host_https`:

* https://github.com/alphagov/govuk-puppet/blob/c68837ae461461319af90df0040fb4c902ccec64/modules/router/manifests/assets_origin.pp#L38-L44
* https://github.com/alphagov/govuk-puppet/blob/c68837ae461461319af90df0040fb4c902ccec64/modules/govuk/manifests/apps/static.pp#L59-L63

modules/router/templates/assets_origin.conf.erb
```diff
@@ -49,7 +49,16 @@ server {
...
   location <%= path %> {
-    proxy_pass <%= @upstream_ssl ? 'https' : 'http' %>://static.<%= @app_domain %>;
+    # Explicitly re-include Strict-Transport-Security header, this
+    # forces nginx not to clear Cache-Control headers further up the
+    # stack.
+    include /etc/nginx/add-sts.conf;
+
+    add_header "Access-Control-Allow-Origin" "*";
+    add_header "Access-Control-Allow-Methods" "GET, OPTIONS";
+    add_header "Access-Control-Allow-Headers" "origin, authorization";
```
floehopper: Minor: Did you consider extracting this into an `include` to avoid the duplication?

lib/puppet-lint/plugins/check_hiera.rb
chrisroos: I've updated the commit note to describe it as a correction instead of a clarification.

modules/router/templates/assets_origin.conf.erb
chrisroos: I've updated the commit note as you've suggested.

modules/router/templates/assets_origin.conf.erb
chrisroos: I've had a quick look at introducing an `include` to reduce the duplication but think it might be worth leaving this for another time. There are a few different `Access-Control-Allow-*` headers/values being set and I think it might be worth seeing whether we can make them consistent before removing the duplication.

chrisroos:
Thanks for the review, @floehopper. While updating the commit note with more information about `@upstream_ssl` versus `@enable_ssl || @proxy_pass_asset_manager_host_https` I noticed another difference in the use of `asset-manager.<%= @app_domain %>` versus `@asset_manager_host`. I've updated the commit and note accordingly.

Would you be able to take another look to make sure it all makes sense, please?

floehopper:
@chrisroos:

> Would you be able to take another look to make sure it all makes sense, please?

Looks good to me! 👍

chrisroos:
Thanks @floehopper. I've rebased this on master in preparation for merging.

PR #7227, opened 2018-02-07
Remove govuk_notify_base_url env var

gpeng:
This was used by email-alert-api when we were testing using the notify staging environment. We no longer use different environments so we no longer need this env var.

This 'ensures' it is absent. It will be fully removed in a subsequent commit.

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api.pp
+++ b/modules/govuk/manifests/apps/email_alert_api.pp
@@ -76,7 +76,7 @@
 #   Template ID for GOV.UK Notify
 #
 # [*govuk_notify_base_url*]
-#   Base URL for GOV.UK Notify API
+#   Deprecated env var to switch notify environments. 
 #
 # [*govuk_notify_rate_per_worker*]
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/email_alert_api.pp
```diff
@@ -218,6 +214,7 @@
           varname => 'GOVUK_NOTIFY_TEMPLATE_ID',
           value   => $govuk_notify_template_id;
       "${title}-GOVUK_NOTIFY_BASE_URL":
+          ensure  => 'absent',
           varname => 'GOVUK_NOTIFY_BASE_URL',
           value   => $govuk_notify_base_url;
```
thomasleese: I reckon we will need to get rid of the `value` if we set it to `absent`.

gpeng: value is a required parameter here so we need to leave it. I did need to change it though as I'd removed the declaration of `$govuk_notify_base_url` so the linter wasn't happy.

PR #7230, opened 2018-02-08
Extract whitehall asset routes

chrisroos:
Splitting these Whitehall asset routes from the `app_specific_asset_routes` should make it clearer that these are the only non-app-specific assets served by Whitehall.


chrisroos:
Thanks, @floehopper!

PR #7231, opened 2018-02-08
Restrict asset requests that end up in Whitehall

chrisroos:
As of the 1 Feb 2018 the only asset requests being served by Whitehall
are for attachments and HMRC. I confirmed this by using the following
query in Kibana:

    request:/GET \/government\/uploads\/.*/
    AND application:/whitehall.*/
    AND status:200
    -attachment_data
    -hmrc

I've also looked at the assets stored in the Whitehall directory on the
NFS mount and confirmed that the only files remaining are attachments
and the special case HMRC files.

Note that the HMRC assets URLs use www.gov.uk instead of
assets.publishing.service.gov.uk. I've added an explicit route for them
here in case anyone was relying on being able to access them via the
assets host.

Moving the '/government/uploads/' catch-all route means that all
non-attachment and non-HMRC asset requests are now sent to asset-manager
(via static). This is preferable to sending them to Whitehall as, since
https://github.com/alphagov/whitehall/pull/3763, sending them to
Whitehall would result in an infinite redirect loop.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1638,23 +1638,6 @@ router::assets_origin::whitehall_uploaded_assets_routes:
   - '/government/uploads/uploaded/hmrc/'
 
 router::assets_origin::asset_manager_uploaded_assets_routes:
-  - '/government/uploads/government/uploads/system/uploads/'
-  - '/government/uploads/system/uploads/attachment/file'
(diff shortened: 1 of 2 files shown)


chrisroos:
Thanks, @chrislo! I've rebased on master in preparation for merging.

PR #7233, opened 2018-02-09
Fix Grafana dashboards

sihugh:
The majority of our apps have been upgraded to logstasher 1.* (mostly through using the govuk_app_config gem).  This removes the `@fields.` prefix from fields that are logged.  Our grafana deploy dashboards expect this to be present unless we specify otherwise (which is what I've done here).  At present, quite a few of the dashboards don't display useful information.

I got this information by looking at each application's logs in Kibana and checking which fields were available.

I also removed some comments like `logstasher > 1` because although they were useful historically, they aren't now.


sihugh:
Fixed, thanks!

PR #7234, opened 2018-02-09
Differentiate between static and uploaded assets

chrisroos:
I've renamed three `assets_origin` hiera variables to differentiate between static and uploaded assets. This mirrors the language we use to talk about assets in [Assets: how they work](https://docs.publishing.service.gov.uk/manual/assets.html).


chrisroos:
Thanks @floehopper!

PR #7243, opened 2018-02-12
Add draft-assets to hosts::development::apps

chrisroos:
This ensures that draft-assets is added to /etc/hosts so that we can
access it from the development VM.


chrisroos:
Thanks @floehopper. I've rebased this branch on master and force-pushed in preparation for merging.

PR #7244, opened 2018-02-13
Disable postgresql collectd plugin on CI

boffbowsh:
The version of collectd we’re using has a bug[1] where it will fail
noisily if a database has no tables. This is quite likely in a CI
environment, and the metrics it collects are not useful. This prevents
27 log lines every 10 seconds across all 5 CI boxes, which accounts for
some logit overuse.

[1]: https://github.com/collectd/collectd/issues/1905


surminus:
Great! This has been filling up logs since the beginning of time.

PR #7245, opened 2018-02-13
Disable notifications in Staging/Prod in AWS

surminus:
When we deploy an application in Staging and Production in AWS, we are also notifying the live Release app and posting to the Slack channel of events that should not be of concern to the 2nd line team while we test these environments.

This adds an environment variable (that is an option and can be overridden for testing in the future) within the job itself. The capistrano job should pick up the env var and act accordingly.

We ensure this is set to true by default, but set to false in the environments we wish to disable this behaviour on.


issyl0:
@surminus You are a ⭐️.

PR #7246, opened 2018-02-13
Record Icinga alert counts in Graphite

boffbowsh:
Adds a collectd plugin that gets the current service alert statuses and
records them as gauges in graphite. Records warnings and criticals
seperately, along with whether they’re acknowledged or not.

The eventual keys would be e.g.
`monitoring-1_blue_integration.icinga.gauge-services_critical_unacknowledged`

### DIFF
--- a/modules/collectd/files/usr/lib/collectd/icinga.sh
+++ b/modules/collectd/files/usr/lib/collectd/icinga.sh
@@ -1,0 +1,31 @@
+#! /bin/bash
+
+HOSTNAME="${COLLECTD_HOSTNAME:-`hostname -f`}"
+INTERVAL="${COLLECTD_INTERVAL:-10}"
+
+# $3 is status, $13 is acked
+
+set -o pipefail
+
+while sleep "$INTERVAL"; do
+  result=$(curl -sH'Host: alert' 'localhost/cgi-bin/icinga/status.cgi?servicestatustypes=20&csvoutput=1' | awk -F ';' -f <(cat - <<-'EOD'
+    BEGIN {
+      statuses["warning_acknowledged"] = 0
+      statuses["warning_unacknowledged"] = 0
+      statuses["critical_acknowledged"] = 0
+      statuses["critical_unacknowledged"] = 0
+    }
+
+    NR>1 {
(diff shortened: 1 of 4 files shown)


modules/collectd/files/usr/lib/collectd/icinga.sh
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    END {
+      for (key in statuses) {
+  print "PUTVAL $HOSTNAME/icinga/gauge-services_"key" interval=$INTERVAL N:"statuses[key]
+      }
+    }
+EOD
+  ) | sed 's/$INTERVAL/'$INTERVAL'/g' | sed 's/$HOSTNAME/'$HOSTNAME'/g') && echo "$result"
+done
```
cbaines: I'd have gone for JSON and Ruby before turning to CSV transforming using Awk and sed in combination, but fair enough.

modules/collectd/templates/etc/collectd/conf.d/icinga.conf.erb
```diff
@@ -0,0 +1,5 @@
+LoadPlugin exec
+
+<Plugin exec>
+  Exec nobody "/usr/lib/collectd/file_handles.sh"
```
cbaines: Not sure how `file_handles.sh` is relevant, should this be `icinga.sh`?

modules/collectd/manifests/plugin/icinga.pp
```diff
@@ -0,0 +1,23 @@
...
+    mode    => '0755',
+    owner   => 'root',
+    group   => 'root',
+    content => file('collectd/usr/lib/collectd/icinga.sh'),
+    tag     => 'collectd::plugin',
+    require => Class['collectd::package'],
+  }
+
+  @collectd::plugin { 'icinga':
+    content => template('collectd/etc/collectd/conf.d/icinga.conf.erb'),
```
cbaines: This template doesn't look like it needs to be a template. It might be simpler to treat it as a file (assuming that is indeed simpler).

boffbowsh:
Good catches, thanks. Updated.

PR #7247, opened 2018-02-14
Add Configuration for Publishing API Consumer in Content Performace Manager

peteglondon:
Creates a queue, user, and Procfile workers

amq.gen* is for randomly generated queue names - I don't think we need it.
We also don't need to deal with the exchange if the queue is configured
and bound in puppet.

this is for [94 - Subscribe to content Changes in Publishing-API](https://trello.com/c/D1ht9Mb5/94-3-subscribe-to-content-changes-in-publishing-api)

There is a related [PR in content performance manager](https://github.com/alphagov/content-performance-manager/pull/491)

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -352,10 +352,17 @@ govuk::apps::content_audit_tool::db::backend_ip_range: "%{hiera('environment_ip_
 govuk::apps::content_audit_tool::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::content_audit_tool::redis_port: "%{hiera('sidekiq_port')}"
 
+govuk::apps::content_performance_manager::rabbitmq::amqp_pass: 'content_performance_manager'
+govuk::apps::content_performance_manager::rabbitmq_password: 'content_performance_manager'
 govuk::apps::content_performance_manager::db_hostname: "postgresql-primary-1.backend"
 govuk::apps::content_performance_manager::db::backend_ip_range: "%{hiera('environment_ip_prefix')}.3.0/24"
+govuk::apps::content_performance_manager::rabbitmq_hosts:
+  - rabbitmq-1.backend
+  - rabbitmq-2.backend
+  - rabbitmq-3.backend
 govuk::apps::content_performance_manager::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::content_performance_manager::redis_port: "%{hiera('sidekiq_port')}"
+
 
 govuk::apps::content_store::nagios_memory_warning: 2300
 govuk::apps::content_store::nagios_memory_critical: 2500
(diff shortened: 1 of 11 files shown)


modules/govuk_rabbitmq/manifests/queue_with_binding.pp
```diff
@@ -13,7 +13,8 @@
 # At present the workaround for changing a binding is to create a new queue for the new binding. Then switch the
 # consumers over to the new queue, and remove the old queue.
 # Another side effect of this problem is you can't define more than one binding for a queue.
-# This may be fixed by https://github.com/voxpupuli/puppet-rabbitmq/pull/504
+# This is possible in version 6.0.0, however that version also drops Pupper 3 support.
```
koetsier: Pupper->t

hieradata/common.yaml
```diff
@@ -352,11 +352,18 @@ govuk::apps::content_audit_tool::db::backend_ip_range: "%{hiera('environment_ip_
 govuk::apps::content_audit_tool::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::content_audit_tool::redis_port: "%{hiera('sidekiq_port')}"
 
+govuk::apps::content_performance_manager::rabbitmq::amqp_pass: 'content_performance_manager'
+govuk::apps::content_performance_manager::rabbitmq_password: 'content_performance_manager'
```
koetsier: Is this ok to be plaintext?

peteglondon: We expect this to be overridden in other environments. However the CI tests only use this file and fail when something is not present in this file.

modules/govuk_rabbitmq/manifests/queue_with_binding.pp
peteglondon: Fixed

hieradata/common.yaml
cbaines: It might be a bit neater to have the password default to undefined then.

modules/govuk/manifests/apps/content_performance_manager.pp
```diff
@@ -46,6 +46,18 @@
...
+#   RabbitMQ hosts to connect to.
+#   Default: localhost
+#
+# [*rabbitmq_user*]
+#   RabbitMQ username.
+#   This is a required parameter
+#
+# [*rabbitmq_password*]
+#   RabbitMQ password.
+#   Default: rummager
```
cbaines: This is a required parameter, there isn't a default.

hieradata/common.yaml
MatMoore: As discussed offline - I don't think we can make it undefined without changing how the tests work, so I'm going to leave this as is for now. Agree it's not ideal but I think we would need a larger refactoring to make this work better for all applications.

PR #7253, opened 2018-02-15
Validate that .ruby-version and Dockerfile have matching version

adrianclay:
It is important that the Dockerfile matches the .ruby-version as we use the Dockerfile as part of the publishing-e2e-tests.  If the .ruby-version gets bumped without the Dockerfile then the publishing-e2e-tests could give you a false negative that the upgrade is non-breaking.

There isn't a way to automatically pull the .ruby-version into the Dockerfile. So in place of that add a CI check which validates they are at parity.

We ignore any information after the - from both sides to allow for minor variations.
```
2.0.0-p648 == 2.0.0-alpine
```

Example mismatching versions build:
https://ci.integration.publishing.service.gov.uk/job/travel-advice-publisher/job/dodgy-ruby-version-patch/34/console

Example matching versions build:
https://ci.integration.publishing.service.gov.uk/job/travel-advice-publisher/job/dodgy-ruby-version-patch/36/console

**TODO:** Check the following apps don't break from this change.

- [x] ~govuk/gemstash-alpine~ Dockerfile comes from inside of this repo, and wouldn't be checked by this change.
- [x] govuk/government-frontend
- [x] ~govuk/logspout-alpine~ Dockerfile comes from inside of this repo, and wouldn't be checked by this change.
- [x] govuk/rummager
- [x] govuk/whitehall
- [x] govuk/static
- [x] govuk/publishing-api
- [x] govuk/asset-manager
- [x] govuk/frontend
- [x] govuk/publisher
- [x] govuk/manuals-publisher
- [x] govuk/collections
- [x] govuk/collections-publisher
- [x] govuk/content-store
- [x] govuk/manuals-frontend
- [x] ~govuk/router~ Does not use the `govuk_jenkinslib.groovy`
- [x] govuk/router-api
- [x] govuk/travel-advice-publisher
- [x] govuk/content-tagger
- [x] govuk/specialist-publisher
- [x] govuk/calendars
- [x] govuk/finder-frontend
- [x] govuk/contacts=
- [x] govuk/content-performance-manager
- [x] govuk/signon
- [x] govuk/release
- [x] govuk/email-alert-frontend
- [x] govuk/calculators
- [x] govuk/info-frontend
- [x] govuk/feedback
- [x] govuk/content-audit-tool
- [x] govuk/licencefinder

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -829,11 +829,28 @@ def hasDatabase() {
   sh(script: "test -e config/database.yml", returnStatus: true) == 0
 }
 
+def validateDockerFileRubyVersion() {
+  if (fileExists(file: ".ruby-version")) {
+    def rubyVersion = readFile(file: ".ruby-version")
+    // Remove any patch information from the ruby version. 2.0.0-p648 -> 2.0.0
+    rubyVersion = rubyVersion.trim().split("-")[0]
+
+    // The Dockerfile base image version can be optionally suffixed with a - followed by a variant
+    // e.g. ruby:2.4.2-slim
+    def hasMatchingVersions = sh(script: "egrep \"FROM ruby:${rubyVersion}(\$|-)\" Dockerfile", returnStatus: true) == 0
+    if (!hasMatchingVersions) {
+      def baseImageDefinition = sh(script: "egrep \"FROM \" Dockerfile", returnStdout: true).trim()
+      error("Dockerfile uses base image \"${baseImageDefinition}\", this mismatches .ruby-version \"${rubyVersion}\"")
+    }
+  }
+}
+
 def hasDockerfile() {
   sh(script: "test -e Dockerfile", returnStatus: true) == 0
 }
 
 def buildDockerImage(imageName, tagName, quiet = false) {
+  validateDockerFileRubyVersion()
   tagName = safeDockerTag(tagName)
   args = quiet ? "--quiet ." : "."
   docker.build("govuk/${imageName}:${tagName}", args)


h-lame:
This looks like it'll do the right thing, but could we make the `Dockerfile` consume the `.ruby-version` file instead so there's no chance of them getting out of sync?  I think we've done similar things when we have a `.ruby-version` file and a `ruby` declaration in the `Gemfile`.

h-lame:
Oh, sorry @adrianclay - I didn't spot that you'd already said there's no way to have the `Dockerfile` consume `.ruby-version`.  You may have already explored and discounted this, but I don't suppose we could read the `.ruby-version` file and pass the contents in as an `--arg` when we run the docker commands?

adrianclay:
Hey @h-lame 

The difficulty with the arg approach is then with the ` docker-compose.yml ` inside publishing-e2e-tests where we define the build.  We'd have to do something like this:

```
travel-advice-publisher:
  image: govuk/travel-advice-publisher:${TRAVEL_ADVICE_PUBLISHER_COMMITISH:-master}
  build:
    context: apps/travel-advice-publisher
    args:
      RUBY_VERSION:${TRAVEL_ADIVCE_PUBLISHER_RUBY_VERSION}
```

From the `Makefile` we could then populate the environment variable with some magic, but it becomes a pain when running `docker-compose` from outside of the `Makefile`.  You have a similar situation if you ever wanted to build the `Dockerfile` locally, having to pass in the `--arg` as you mention.  Ideally, we'd want something similar to the `Gemfile` solution where the version is pulled in instead of having to be pushed in.

I agree this solution isn't ideal, but in my eyes it's the lesser of two evils.

adrianclay:
Hey @kevindew I've put in a couple of links to Jenkins scenarios in the PR description (versions matching and versions not matching), if you're looking for something different let me know.

h-lame:
@adrianclay that does seem painful.  I think you're probably right that if there's no easy way to do it automatically then at least with this PR we have an automatic way to remind us to make the change.

kevindew:
The failed build looks good 👍 
Yeah I don't think there's a particularly practical way to set the ruby version through an arg since it's the base image.

PR #7255, opened 2018-02-15
Set unicorn worker processes in puppet

sihugh:
These apps are set up to read `UNICORN_WORKER_PROCESSES` env vars to set the
number of worker processes to use.

The default is 2 processes, but I know that platform support are likely to change
this as part of https://trello.com/c/5qhHri2F/90-check-if-we-have-enough-server-resource-serving-government-frontend so I've added it in anyway.

If this is merged, then we can remove the default from frontend's https://github.com/alphagov/frontend/blob/41d9f122404216c3c77e77c5933a9a02dc47ef57/config/unicorn.rb#L4

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -398,9 +398,11 @@ govuk::apps::frontend::nagios_memory_warning: 1200
 govuk::apps::frontend::nagios_memory_critical: 1400
 govuk::apps::frontend::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::frontend::redis_port: "%{hiera('sidekiq_port')}"
+govuk::apps::frontend::unicorn_worker_processes: "4"
 
 govuk::apps::government-frontend::nagios_memory_warning: 900
 govuk::apps::government-frontend::nagios_memory_critical: 1000
+govuk::apps::government-frontend::unicorn_worker_processes: "2"
 
 govuk::apps::govuk_cdn_logs_monitor::processed_data_dir: '/mnt/logs_cdn/data'
 
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -401,9 +401,11 @@ govuk::apps::frontend::nagios_memory_warning: 1200
 govuk::apps::frontend::nagios_memory_critical: 1400
 govuk::apps::frontend::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::frontend::redis_port: "%{hiera('sidekiq_port')}"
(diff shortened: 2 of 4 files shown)


h-lame:
You'll need to update https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/government_frontend.pp to pull in the hieradata value and pass it down to `govuk::app`.  Like what https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/whitehall.pp does.

Might be worth setting up the same for frontend while you're at it, then you can drop the worker_process line from it's unicorn.rb and let govuk_app_config deal with it.

h-lame:
Uh, actually, I mean like https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/imminence.pp, not whitehall.  Whitehall does pass a value down to `govuk::app`, but not by pulling it in from hieradata.

sihugh:
Lovely, thanks @h-lame. I had a feeling I hadn't quite finished

sihugh:
Frontend isn't currently using an up to date version of govuk_app_config although Kevin said he'd be looking at it soon for Docker reasons.  When he does bump it, he can remove the unicorn.rb bits :-)

hieradata/common.yaml
```diff
@@ -389,9 +389,11 @@ govuk::apps::frontend::nagios_memory_warning: 1200
 govuk::apps::frontend::nagios_memory_critical: 1400
 govuk::apps::frontend::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::frontend::redis_port: "%{hiera('sidekiq_port')}"
+govuk::apps::government-frontend::unicorn_worker_processes: "3"
```
h-lame: should this one be `govuk::apps::frontend::unicorn_worker_processes` ?

hieradata_aws/common.yaml
```diff
@@ -395,9 +395,11 @@ govuk::apps::frontend::nagios_memory_warning: 1200
 govuk::apps::frontend::nagios_memory_critical: 1400
 govuk::apps::frontend::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::frontend::redis_port: "%{hiera('sidekiq_port')}"
+govuk::apps::government-frontend::unicorn_worker_processes: "3"
```
h-lame: and this one

hieradata/common.yaml
sihugh: I really was feeling funny yesterday :-/

PR #7256, opened 2018-02-15
Install latest version Terraform on ci-deploy

surminus:
We're using the CI Jenkins deploy instance to manage all builds of Terraform infrastructure (a central build box).

This needs to have the latest version of Terraform installed for this to work for our code.

https://trello.com/c/qAzTAU4p/952-build-box-spike-on-adding-tf-deploys-to-current-integration-carrenza-deploy-jenkins


issyl0:
I thought the job downloaded the latest if it wasn't already there?

surminus:
We discussed this in meatspace - the job does download the latest, but it would do this every time if specified.

PR #7262, opened 2018-02-15
[WIP] Update Vagrantfile for AWS

surminus:
In the past it was useful to be able to test Puppet code, but since we moved to AWS we haven't been able to replicate this due to the change in the way we defined nodes.

Ideally a more complete testing framework should be implemented long-term, but for now these changes should hopefully make working with the Puppet codebase easier and safer for people less familiar with all the ins-and-outs.

The changes apply the following changes in mind:

 - Define nodes within the same repo: in previous lives, people had to  have a different repository to gather the nodes; instead, define all  the nodes within the Vagrantfile itself
 - Rely on a default base image: rather than relying on a base image  created by packer and which has to be kept up to date, use the default  Ubuntu base image and install packages on top of that. This relates  more closely to how creation of a new node works in AWS.
 - Making the Vagrantfile simpler with less logic so it's easier for other people to pick up. This is why I have removed some other files.

## This currently does not work, so needs further testing.


issyl0:
We're probably not going to get to fixing this (we probably should have got to it in Firebreak). Thanks for the effort, Laura. :-) One day...

surminus:
If I were coming back to it I'd totally investigate test-kitchen instead ;)

PR #7264, opened 2018-02-16
Prepare staging and integration email-alert-api environments

thomasleese:
This sets up the integration and staging environments to correctly configure the override address to a success Notify address and then also include the whitelist of valid email addresses.

[Trello Card](https://trello.com/c/UwM5NwqV/549-spike-enabling-a-full-test-run-for-integration)

### DIFF
--- a/hieradata/staging.yaml
+++ b/hieradata/staging.yaml
@@ -23,6 +23,7 @@ govuk::apps::email_alert_api::govdelivery_account_code: 'UKGOVUK'
 govuk::apps::email_alert_api::govdelivery_hostname: 'stage-api.govdelivery.com'
 govuk::apps::email_alert_api::govdelivery_public_hostname: 'stage-public.govdelivery.com'
 govuk::apps::email_alert_api::govuk_notify_template_id: '2844a647-6bf1-4b01-a25c-569d2cc00849'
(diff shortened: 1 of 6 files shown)


hieradata/staging.yaml
```diff
@@ -23,6 +23,13 @@ govuk::apps::email_alert_api::govdelivery_account_code: 'UKGOVUK'
 govuk::apps::email_alert_api::govdelivery_hostname: 'stage-api.govdelivery.com'
 govuk::apps::email_alert_api::govdelivery_public_hostname: 'stage-public.govdelivery.com'
 govuk::apps::email_alert_api::govuk_notify_template_id: '2844a647-6bf1-4b01-a25c-569d2cc00849'
+govuk::apps::email_alert_api::email_address_override: 'simulate-delivered@notifications.service.gov.uk'
+govuk::apps::email_alert_api::email_address_override_whitelist:
```
kevindew: Might just want to add this in the common one as it'll only take effect if email_address_override is set.

If so could you also include the emails from: https://trello.com/c/wxCFNWrk/609-recruit-and-set-up-our-little-gang-of-testers

thomasleese: Yep, good point. Fixed that in 2ac79b7.

PR #7272, opened 2018-02-19
Revert "Add Configuration for Publishing API Consumer in Content Performace Manager"

chao-xian:
Reverts alphagov/govuk-puppet#7247

The use of "plaintext" creds in #7247 is causing issues. To avoid the error on production I'll revert this for now.


chao-xian:
We are going to switch off Rabbit MQ for Content Performance Manager using `ensure => 'absent'` instead in https://github.com/alphagov/govuk-puppet/pull/7274

PR #7274, opened 2018-02-19
Allows ensure to be set to `absent` for rabbitmq

elenatanasoiu:
We're trying to allow rabbitmq to be turned off using `ensure => absent`


surminus:
You could also further add to this further and ensure that `govuk_rabbitmq::consumer` has resources set to `absent`.

elenatanasoiu:
Done

PR #7281, opened 2018-02-20
Added user sebastianschmieschek and public key

schmie:


surminus:
Please squash this into a single commit.

PR #7282, opened 2018-02-20
Add global shared library to Jenkins

surminus:
This is related to
https://github.com/alphagov/govuk-jenkinslib/commit/c2e84ed07a17d2afd17d090bf4a87b0d62efeef9.

This sets the global option to enable loading in a shared library from a repository. Pass in a hash of the libraries you wish to load, and then you should be able to load the libraries in using a Jenkinsfile.

I used an embedded hash within the manifest because at the moment it's a single library. In the future it may be easier to create a default set of values, and pass in a hash from hiera.

There are a few advantages over loading in an external shared library, including keeping the library in it's own version control, being able to potentially add unit tests, because able to explictly call a job from a branch of the library to test changes and making the code more portable and visible.

https://trello.com/c/88RFcs4R/994-make-jenkinslibgroovy-its-own-repo-refactor

This requires [this PR](https://github.com/alphagov/govuk-jenkinslib/pull/1) to be merged before merging as otherwise Jenkins will get upset.

### DIFF
--- a/modules/govuk_jenkins/manifests/pipeline.pp
+++ b/modules/govuk_jenkins/manifests/pipeline.pp
@@ -28,4 +28,18 @@ class govuk_jenkins::pipeline (
     source  => 'puppet:///modules/govuk_jenkins/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy',
     require => File['/var/lib/jenkins/groovy_scripts'],
   }
+
+  $jenkins_libraries = {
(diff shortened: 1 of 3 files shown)


surminus:
This has been updated to set implicit load to false.

surminus:
I had to repush because I updated implicit load to `false` and the tests started to fail. 

@issyl0 it's because XML is horrible :)

PR #7284, opened 2018-02-21
Make the dev-vm procfile more consistent

kevindew:
This changes the way specialist-publisher and manuals-publisher start in the dev-vm procfile so they use rails server rather than foreman like every other app. 

Running bundle exec foreman for them is going to break soon when https://github.com/alphagov/specialist-publisher/pull/1167 and https://github.com/alphagov/manuals-publisher/pull/1291 are merged.

### DIFF
--- a/development-vm/Pinfile
+++ b/development-vm/Pinfile
@@ -69,8 +69,8 @@ process :'short-url-manager' => [:'publishing-api']
 process :signon => [:'signon-sidekiq']
 process :'signon-sidekiq'
 process :smartanswers => [:static, :'content-store', :imminence]
-process :'specialist-publisher' => ['asset-manager', :'publishing-api', :'email-alert-api']
-process :'specialist-publisher-worker' => [:'email-alert-api']
+process :'specialist-publisher' => [:'asset-manager', :'publishing-api', :'email-alert-api', :'specialist-publisher-worker']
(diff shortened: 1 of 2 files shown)


kevindew:
Doh! sorry I should have spotted those. I'll add some commits for them too. Cheers @h-lame

kevindew:
@h-lame I've updated this for email-alert-api and HMRC manuals. govuk_publishing_components seems to be a bit more of a special case with how they run things so I've omitted that. Will hopefully return to look at them soon though.

h-lame:
@kevindew loathe to hold this up, but hmrc-manuals-api's Procfile has a sidekiq process, shouldn't we add that in here as `hmrc-manuals-api-worker`?

kevindew:
@h-lame yeah that isn't actually being used. From the commit message:

"This is for consistency. HMRC Manuals API was using foreman start to
start a worker and a web server however they don't use the worker
anymore - the project itself needs updating to remove the worker."

If I get a few mins at the end of the day I'll open a pr to remove sidekiq from it. Or maybe I should put a trello card somewhere in case I forget 🤔

h-lame:
Sorry, didn't read the commits :(

kevindew:
@h-lame that's no bother 👍

PR #7285, opened 2018-02-21
Add new machines for email-alert-api to LB

rubenarakelyan:
This commit adds new machines for email-alert-api to the load balancers and balances between the new and current machines.

We're adding these new machines since email-alert-api is using the majority of resources on the current backend machines and we'd like to give it space as well as reduce any possible impact on other backend apps. This is a result of the work to migrate all email sending to our own systems.

Depends on https://github.com/alphagov/govuk-provisioning/pull/41 and https://github.com/alphagov/govuk-aws/pull/494.

Trello: https://trello.com/c/oqi6Xvnu/629-move-email-alert-api-and-email-alert-service-to-dedicated-vms

### DIFF
--- a/hieradata/production.yaml
+++ b/hieradata/production.yaml
@@ -166,6 +166,13 @@ govuk::node::s_backend_lb::backend_servers:
   - 'backend-1.backend'
   - 'backend-2.backend'
   - 'backend-3.backend'
+govuk::node::s_backend_lb::email_alert_api_backend_servers:
+  - 'backend-1.backend'
+  - 'backend-2.backend'
+  - 'backend-3.backend'
+  - 'email-alert-api-1.backend'
+  - 'email-alert-api-2.backend'
+  - 'email-alert-api-3.backend'
 govuk::node::s_backend_lb::performance_backend_servers:
   - 'performance-backend-1.backend'
   - 'performance-backend-2.backend'
--- a/hieradata/class/backend.yaml
+++ b/hieradata/class/backend.yaml
@@ -5,6 +5,8 @@ govuk::apps::contacts::db_hostname: 'mysql-master-1.backend'
 govuk::apps::contacts::db_username: 'contacts'
 govuk::apps::contacts::db_password: "%{hiera('govuk::apps::contacts::db::mysql_contacts_admin')}"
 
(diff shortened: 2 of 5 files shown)


hieradata/class/email_alert_api.yaml
```diff
@@ -0,0 +1,5 @@
+---
+
+govuk::node::s_base::apps:
```
surminus: We shouldn't add apps in this way anymore. Add it in `common.yaml` in both hieradata and hieradata_aws in [this block](https://github.com/alphagov/govuk-puppet/blob/master/hieradata/common.yaml#L10-L122)

hieradata/production.yaml
```diff
@@ -166,6 +166,13 @@ govuk::node::s_backend_lb::backend_servers:
   - 'backend-1.backend'
   - 'backend-2.backend'
   - 'backend-3.backend'
+govuk::node::s_backend_lb::email_alert_api_backend_servers:
+  - 'backend-1.backend'
```
surminus: Is the idea of this to migrate traffic between both sets of servers for a migration period? Otherwise I don't think the backend servers need to be included.

rubenarakelyan: If we just had the new servers in here, would there be an interruption while the apps were deployed, and would anything happen to requests currently in progress? I took this straight from the way publishing-api was moved onto its own machines.

hieradata/class/email_alert_api.yaml
rubenarakelyan: Fixed

hieradata/production.yaml
rubenarakelyan: Actually, scrap that, since we're removing the apps from the backend machines in `common.yaml`!

surminus: You're correct in that a smoother migration period may be to run them together at the same time, and then filter down the connections. It's up to you ultimately.

rubenarakelyan: Yup, I've set it up to use both the existing backend servers and the new ones, then I'll make up another PR to remove the existing backend servers.

surminus: 👍

PR #7288, opened 2018-02-22
Jenkinslib parameter for selecting publishing-e2e-tests branch

kevindew:
This allows someone triggering a build to select which branch of e2e
tests is used which can be useful when you wish to make a change that
requires a change in the e2e tests.

### DIFF
--- a/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
+++ b/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
@@ -112,9 +112,8 @@ def buildProject(Map options = [:]) {
 
   properties([
     buildDiscarder(
-      logRotator(
-        numToKeepStr: '50')
-      ),
+      logRotator(numToKeepStr: '50')
+    ),
     [$class: 'RebuildSettings', autoRebuild: false, rebuildDisabled: false],
     [$class: 'ParametersDefinitionProperty', parameterDefinitions: parameterDefinitions],
   ])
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy
```diff
@@ -106,6 +106,14 @@ def buildProject(Map options = [:]) {
     )
   ]
 
+  if (options.publishingE2ETests == true && env.PUBLISHING_E2E_TESTS_BRANCH == null) {
+    parameterDefinitions << stringParam(
```
RyanMacG: I hadn't realised the `parameterDefinitions` were in here.  Now it's pointed out it's obvious!

RyanMacG:
How do these changes reconcile with the changes being made with https://github.com/alphagov/govuk-jenkinslib ?

kevindew:
@RyanMacG good point I don't know if there are ideas to keep them both in sync or for that to just have a catch up PR when it's ready.

@surminus is there any process to follow for making changes to jenkinslib while you're working on the switchover?

surminus:
@kevindew if you're happy to also add the same changes that would be appreciated, alternatively I can just do a final sync when I've tested and I think it's ready for use. While I'm testing it's not a problem.

I believe Jenkinsfile's will need to be updated to switch over so the process may be long running unfortunately.

kevindew:
@surminus Sure I've opened it as a PR here: https://github.com/alphagov/govuk-jenkinslib/pull/3

PR #7289, opened 2018-02-23
Enable uuid-ossp Postgres extension on email-alert-api

thomasleese:
We're going to be using this extension to generate UUID values within Postgres.

[Trello Card](https://trello.com/c/B8db764N/637-switch-some-of-our-primary-keys-from-id-to-uuid)

### DIFF
--- a/modules/govuk_postgresql/manifests/db.pp
+++ b/modules/govuk_postgresql/manifests/db.pp
@@ -119,8 +119,9 @@ define govuk_postgresql::db (
         tag      => 'govuk_postgresql::server::not_slave',
     }
 
+    validate_array($extensions)
+
     if ! $rds {
-      validate_array($extensions)
       if (!empty($extensions)) {
           $temp_extensions = prefix($extensions,"${db_name}:")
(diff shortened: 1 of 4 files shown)


modules/govuk_postgresql/manifests/db.pp
```diff
@@ -158,5 +159,12 @@
       collectd::plugin::postgresql_db{$db_name:}
       govuk_postgresql::monitoring::db{$db_name:}
     }
+  } else {
+    if (!empty($extensions)) {
+      postgresql::server::extension { $extensions:
```
rhowe-gds: @deanwilson if `$extensions` is `[]` then this will not do anything so the `if (!empty..` guard doesn't need to exist, amirite?

thomasleese: Hmm, yeah, I would assume so... I included it as the one for non RDS does it for some reason.

https://github.com/alphagov/govuk-puppet/pull/7289/files#diff-127ce7effed9df8b23c79cefcf3195b4R125

PR #7290, opened 2018-02-23
Replication improvements

boffbowsh:
Ask for a MFA token on the command line, and then store the credentials
in a file. The `assume-role` command returns an expiration time, so
check this with a 5 minute margin of error. If the token has expired
or is about to expire, prompt the user for their MFA token again.

This has the added benefit of subsequent runs of this script not
requiring a new MFA token if the STS session is still valid.

The dumps for mysql and postgres can be downloaded per database. This
means we can skip the download of ignored DBs like transition.

### DIFF
--- a/development-vm/replication/common-args.sh
+++ b/development-vm/replication/common-args.sh
@@ -54,60 +54,6 @@ function ignored() {
   return 1
 }
 
-aws_auth() {
-  temporary_aws_credentials=$(dirname $0)/.temporary_aws_credentials
-  test -f ${temporary_aws_credentials} && source ${temporary_aws_credentials}
-
-  if [ -z "${AWS_ACCESS_KEY_ID-}" ] || \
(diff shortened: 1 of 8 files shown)


development-vm/replication/common-args.sh
```diff
@@ -55,14 +54,64 @@ function ignored() {
   return 1
 }
 
-TOKEN_REQUIRED=1
-while getopts "a:hF:u:d:sri:onmpqet" OPTION
+aws_auth() {
```
dwhenry: not sure this is the best file for this method

davbo:
I'm not sure but it could be worth looking at `aws s3 presign` for some of this too. 

That way you'd get a presigned URL using your assumed role and GET the object after that.

boffbowsh:
Nice idea @davbo but most of the work is using `aws s3 sync`, so might not be worth it.

Good call @dwhenry, updated.

PR #7299, opened 2018-02-26
Run Router in a container in Staging

surminus:
We are waiting on some other work to allow deployment of Router as the binary artefact, and this is blocking testing deployment of other apps.

An easy way around this is to deploy Router in a container, which we have prior art for.


surminus:
This is meant to be a temporary work around until the artefact stuff works.

PR #7304, opened 2018-02-26
Adds docs for mongo

elenatanasoiu:
This will add docs for two Icinga alerts:

1. Repairing a mongodb cluster - this already had a doc written up and just need to be hooked up. 
2. Mongo DB backups running out of space - this also required a new [PR in govuk-developer-docs](https://github.com/alphagov/govuk-developer-docs/pull/735)

### DIFF
--- a/modules/mongodb/manifests/backup.pp
+++ b/modules/mongodb/manifests/backup.pp
@@ -89,6 +89,7 @@ class mongodb::backup(
       service_description => $service_desc,
       freshness_threshold => $threshold_secs,
       host_name           => $::fqdn,
+      notes_url           => monitoring_docs_url(low-space-for-automongodbbackup),
     }
   }
 
--- a/modules/mongodb/manifests/monitoring.pp
+++ b/modules/mongodb/manifests/monitoring.pp
@@ -70,5 +70,6 @@ class mongodb::monitoring (
     check_command       => 'check_nrpe!check_mongodb!replset_state',
     service_description => 'mongod replset state',
     host_name           => $::fqdn,
+    notes_url           => monitoring_docs_url(recreate-a-mongo-instance-in-aws),
   }
 }


modules/mongodb/manifests/backup.pp
```diff
@@ -89,6 +89,7 @@
       service_description => $service_desc,
       freshness_threshold => $threshold_secs,
       host_name           => $::fqdn,
+      notes_url           => monitoring_docs_url(mongodb),
```
suzannehamilton: `monitoring_docs_url` generates a link to the alerts docs, so this would link to https://docs.publishing.service.gov.uk/manual/alerts/mongodb.html which doesn't exist.

Is there a doc page for this specific alert that we can link to? That would be more useful than the general docs on MongoDB anyway.

modules/mongodb/manifests/monitoring.pp
```diff
@@ -70,5 +70,6 @@
     check_command       => 'check_nrpe!check_mongodb!replset_state',
     service_description => 'mongod replset state',
     host_name           => $::fqdn,
+    notes_url           => monitoring_docs_url(howto-recreate-a-mongo-instance-in-aws),
```
suzannehamilton: [Ditto](https://github.com/alphagov/govuk-puppet/pull/7304/files#r170865357).

modules/mongodb/manifests/backup.pp
elenatanasoiu: I can create one instead.

elenatanasoiu: @suzannehamilton Done here: https://github.com/alphagov/govuk-developer-docs/pull/735/files

modules/mongodb/manifests/monitoring.pp
elenatanasoiu: Ditto your ditto: https://github.com/alphagov/govuk-developer-docs/pull/735/files

elenatanasoiu: I've moved the file into the alerts folder

suzannehamilton: It might be good to rename that file so that it's clear it's really about an alert. Also, I'd update the metadata to add to the "alerts" section. It's currently under "development VM" even though it's an AWS thing.

elenatanasoiu: Done here: https://github.com/alphagov/govuk-developer-docs/pull/738/files

suzannehamilton: Should this drop the "howto-" now that the file has been renamed?

elenatanasoiu: Indeed, renamed now.

PR #7305, opened 2018-02-26
Use external GOV.UK Jenkins library

surminus:
This uses the library defined in [the external Jenkins library](https://github.com/alphagov/govuk-jenkinslib) hosted in an external repository.

The library is implicitly loaded as defined in the Jenkins configuration.

If the library is not implicitly loaded, then the following would need to be defined in each `Jenkinsfile`:

`library("govuk")`

I am not sure on which approach I prefer - the second approach would allow you specify specific branches if you wanted to test changes in the Groovy library, but the first means you don't have to worry about specifying it all.

https://trello.com/c/88RFcs4R/994-make-jenkinslibgroovy-its-own-repo-refactor

Relates to https://github.com/alphagov/govuk-puppet/pull/7282 which will define the library in Jenkins itself.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,9 +2,9 @@
 
 REPOSITORY = 'govuk-puppet'
 
-node {
-  def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
+library("govuk")
 
+node {
   properties([
     buildDiscarder(logRotator(numToKeepStr: '50')),
   ])


deanwilson:
I'd rather have the explicit `library("govuk")` in each of the jenkinsfiles. It'll make it easier to determine what's using it later. When we move away from Jenkins.

surminus:
👍 It also means you can explicitly load from a branch for testing if required: `library("govuk@my-new-branch")

deanwilson:
I think that'd be much better, and make the testing experience nicer. Thank you for mentioning it as a second option in the PR.

surminus:
Updated with new commit to make it clear that's the decision.

PR #7308, opened 2018-02-27
Mongodb dailybackup storage

suthagarht:
- We have added a 32GB disk called /dev/sdf1 to
  '/var/lib/automongodbbackup'. This was done to accomodate the current
storage requirement.

Solo: @suthagarht

### DIFF
--- a/hieradata/class/mongo.yaml
+++ b/hieradata/class/mongo.yaml
@@ -10,6 +10,7 @@ lv:
     pv:
       - '/dev/sdb1'
       - '/dev/sdd1'
+      - '/dev/sdf1'
     vg: 'backup'
   data:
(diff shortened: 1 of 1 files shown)


hieradata/class/mongo.yaml
```diff
@@ -10,6 +10,7 @@ lv:
     pv:
       - '/dev/sdb1'
       - '/dev/sdd1'
+      - '/dev/sdf1/'
```
dwhenry: Can we remove the extra slash at the end of the string so it is consistent.

suthagarht: thank you @dwhenry . good spot. I have now corrected this.

PR #7309, opened 2018-02-27
[DO NOT MERGE] Remove email-alert apps from backend servers

rubenarakelyan:
This commit removes email-alert-api and email-alert-service from the backend servers since they are now served by their own servers.

Trello: https://trello.com/c/oqi6Xvnu/629-move-email-alert-api-and-email-alert-service-to-dedicated-vms


rubenarakelyan:
Added to https://github.com/alphagov/govuk-puppet/pull/7310

PR #7312, opened 2018-02-28
Create an extension with quotes around the name

thomasleese:
When trying to install the extension 'uuid-ossp' this failed because of the dash in the name, but if we use quotes this should succeed.


thomasleese:
@rubenarakelyan oh yep, fixed!

PR #7315, opened 2018-02-28
[WIP] Add a Jenkins job to sync data between production and AWS staging

issyl0:
- I don't know what to do about the `gor` traffic replay stuff, as that was in the Carrenza sync job but obviously not the AWS integration sync job.

https://trello.com/c/cYTPEGvn/1044-data-sync-from-production-carrenza-to-staging-aws

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/copy_data_from_production_to_aws.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/copy_data_from_production_to_aws.yaml.erb
@@ -1,0 +1,95 @@
+---
+- scm:
+    name: env-sync-and-backup_Copy_Data_to_Production
+    scm:
+        - git:
+            url: git@github.com:alphagov/env-sync-and-backup.git
+            branches:
+              - master
+
+- job:
+    name: Copy_Data_from_Production_to_AWS
+    display-name: Copy_Data_from_Production_to_AWS
+    project-type: freestyle
+    description: |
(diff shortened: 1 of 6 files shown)


modules/govuk_jenkins/templates/jobs/copy_data_from_production_to_aws.yaml.erb
```diff
@@ -0,0 +1,111 @@
...
+            notify-backtonormal: false
+            notify-repeatedfailure: false
+            include-test-summary: false
+            room: <%= @slack_room %>
+    scm:
+      - env-sync-and-backup_Copy_Data_from_Production_to_AWS
+    logrotate:
+        numToKeep: 10
+    triggers:
+        - timed: 'H 1 * * *'
```
surminus: Will this happen at the same time as the current data sync?

issyl0: Good point. The current Integration data sync is set to happen at `H 4 * * 1-5`, and the current Staging one set to happen at `H 1 * * *`, so yes. The Integration AWS sync was set to happen at `H 11 * * 1-5`. I'll make this data sync consistent with the latter.

issyl0:
We should do Carrenza Staging => AWS Staging syncs instead to avoid locking Production database tables. I'll change this.

PR #7316, opened 2018-02-28
Fix the data_sync_complete job to rewrite to blue.staging.

stephenharker:
When data is synced from production, some of the DBs need tweaking to correct the production domain to the staging domain.


stephenharker:
Or should it just be 'staging.govuk.digital' and not 'blue.staging.govuk.digital' and let the local CNAME sort it out?

jljalderman:
If you let the local CNAME sort it out, will you be able to sync data to green before switching over to use it?

PR #7317, opened 2018-03-01
Use external GOV.UK Jenkins library in builds

tijmenb:
We use an external jenkins library now: https://github.com/alphagov/govuk-jenkinslib. This replaces the current library with a reference to that so that we don't have to make changes in both places while we migrate.

@issyl0 @surminus I think this works (I tested it in https://github.com/alphagov/collections/pull/525), thoughts?


surminus:
Oh, interesting approach! We could try devops'ing in the change on `ci-master-1` and seeing if it breaks anything?

PR #7319, opened 2018-03-01
Remove Graham Pengelly

rubenarakelyan:
Trello: https://trello.com/c/GK4ie9lQ/155-leaver-graham-pengelly


rubenarakelyan:
I haven't heard of any issues, but I've found a lot of home directories left over so wanted to do it cleanly :)

kevindew:
@rubenarakelyan sounds good would be good to learn and document too if there are issues rather it being a folklore thing heh.

PR #7320, opened 2018-03-01
Add_mongo_mount

jljalderman:
Add additional mount point for EBS volume for mongo data

### DIFF
--- a/hieradata_aws/class/mongo.yaml
+++ b/hieradata_aws/class/mongo.yaml
@@ -10,3 +10,15 @@ mongodb::server::replicaset_members:
   'mongo-1':
   'mongo-2':
   'mongo-3':
+
+lv:
+  data:
+    pv:
+      - /dev/nvme1n1
+    vg: mongo
+
+mount:
+  /var/lib/mongodb:
+    disk: '/dev/mapper/mongo-data'
+    govuk_lvm: 'data'
+    mountoptions: 'defaults,noatime'


hieradata_aws/class/mongo.yaml
```diff
@@ -10,3 +10,15 @@ mongodb::server::replicaset_members:
   'mongo-1':
   'mongo-2':
   'mongo-3':
+
+lv:
+  data:
+    pv:
+      - /dev/sdf
```
surminus: These machines are `m5.large`, which means they use a different device label. Check other hieradata to see what it should be (as I can't remember off the top of my head).

jljalderman: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html and https://aws.amazon.com/amazon-linux-ami/instance-type-matrix/ suggest that M5.large should be HVM EBS backed 64bit, which it suggests for EBS volumes should be /dev/sd[f-p]

surminus: [Related commit I was talking about](https://github.com/alphagov/govuk-puppet/commit/b1e5add07092de506bee9f0322840fba99acc004)

jljalderman: I did see that, I don't know the answer but happy to change it...

jljalderman: done

PR #7323, opened 2018-03-01
Revert "Use external GOV.UK Jenkins library in builds"

cbaines:
This seems to be causing issues on the CI Jenkins, builds fail with:
`java.lang.ClassNotFoundException: null`

This reverts commit f7fde044fe4a1f64019acfb117affe356e83a737.


cbaines:
An example of a failing build: https://ci.integration.publishing.service.gov.uk/job/govuk-taxonomy-supervised-learning/job/use-ijson-for-measure-average-taxons/2/console

PR #7324, opened 2018-03-02
Revert "Run Router in a container in Staging"

suthagarht:
Reverts alphagov/govuk-puppet#7299

We are making this change to test that the S3 buckets are usable as we would expect during a normal operation.


hieradata_aws/staging.yaml
```diff
@@ -90,7 +90,6 @@ govuk_mysql::server::slow_query_log: true
 govuk::node::s_asset_master::copy_attachments_hour: 7
 
 govuk::node::s_cache::real_ip_header: 'True-Client-Ip'
-govuk::node::s_cache::router_as_container: true
```
surminus: Instead of reverting everything, could we just remove this line? It might be useful in the future.

surminus:
I'd recommend respinning the instances to ensure there are no resources left lying around.

PR #7327, opened 2018-03-02
Add a Grafana dashboard for scheduled publishing

suzannehamilton:
Add a dashboard for monitoring delays to scheduled publishing. This shows data reported by the content-store to help us monitor documents which are published late.

Draft of the dashboard: https://grafana.publishing.service.gov.uk/dashboard/db/publishing-delay-report

screen shot 2018-03-02 at 16 07 01

https://trello.com/c/pMHR10Rv/85-2-scheduled-publication-reporting

### DIFF
--- a/modules/grafana/files/dashboards/scheduled_publishing.json
+++ b/modules/grafana/files/dashboards/scheduled_publishing.json
@@ -1,0 +1,226 @@
+{
+  "annotations": {
+    "list": []
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "hideControls": false,
+  "id": 79,
+  "links": [],
+  "refresh": false,
+  "rows": [
+    {
+      "collapse": false,
+      "height": 250,
+      "panels": [
+        {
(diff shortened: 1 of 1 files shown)


modules/grafana/manifests/dashboards.pp
```diff
@@ -38,6 +38,7 @@
     "${dashboard_directory}/application_health.json": content => template('grafana/dashboards/application_health.json.erb');
     "${dashboard_directory}/edge_health.json": content => template('grafana/dashboards/edge_health.json.erb');
     "${dashboard_directory}/origin_health.json": content => template('grafana/dashboards/origin_health.json.erb');
+    "${dashboard_directory}/scheduled_publishing.json": content => template('grafana/dashboards/scheduled_publishing.json.erb');
```
cbaines: It doesn't look like this needs to be a template, so you could just add it as a file.

suzannehamilton: Thanks, that's much easier. 👍

PR #7328, opened 2018-03-02
Add a jenkins job to clear the frontend caches

tijmenb:
This is intended to be used as a downstream task for the emergency publishing.

Part of improvements during a emergency publishing drill: https://trello.com/c/ktLcZLIN


tijmenb:
Realised that this doesn't actually work because it will only clear the cache for Rails running on a single machine, at least if the app is using file store or memcached. I'm doubly unsure how this works if the application uses memory store (I think it might only clear the cache for the `rails console` process).

tijmenb:
Yeah that could work but I'm still not sure about the in-memory cache in those cases. And we would be recreating fabric :)

kevindew:
true, suppose we could use fabric in this script. I kinda liked that this could be a dependant job of the emergency publishing rake task. Not too much of a biggie though.

PR #7331, opened 2018-03-05
Add extra disks to postgres primary on staging

bevanloon:
Two extra disks - one for data and one for backup to take into account the extra data we expect to generate from Email Alert Api.

[Trello](https://trello.com/c/T2ChiLVo/621-increase-disk-space-on-postgres-machines)


hieradata/class/staging/postgresql_primary.yaml
```diff
@@ -9,7 +9,10 @@ lv:
       - '/dev/sdf1'
       - '/dev/sdg1'
       - '/dev/sdh1'
+      - '/dev/sdj'
```
thomasleese: Should this be `sdj1`?

PR #7334, opened 2018-03-05
Added Puppet ensure for /data/uploads owned by deploy.deploy

schmie:
In order to fix asset-master deployment


suthagarht:
Hi @schmie , I am going to close this PR. I assume that thet the actual problem was caused by puppet not applying correctly when the test was made. This probably meant that the deploy failed. I did some tests to confirm that I am not missing something. Please check the trello card for details.

PR #7338, opened 2018-03-06
Add traffic in metrics to the Email Alert API dashboard

thomasleese:
This adds "Traffic In" metrics to the Email Alert API dashboard based on the "Email Alert API traffic" dashboard that exists in production right now but not in Puppet.

[Trello Card](https://trello.com/c/eNik7kAj/673-include-nginx-monitoring-to-the-dashboard)

### DIFF
--- a/modules/grafana/files/dashboards/email_alert_api.json
+++ b/modules/grafana/files/dashboards/email_alert_api.json
@@ -831,6 +831,193 @@
     },
     {
       "collapse": false,
+      "height": 250,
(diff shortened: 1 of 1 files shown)


kevindew:
Tom and I split this into two graphs and added AWS

Looks like this in Integration:

screen shot 2018-03-06 at 11 40 47

Looks like this in production:

screen shot 2018-03-06 at 12 01 23

PR #7340, opened 2018-03-06
Change namespace for signon background queue icinga check

h-lame:
Since upgrading signon to govuk_sidekiq 3.x and govuk_app_config 1.3.x
it now includes the hostname in the namespace for the queue information
so we need to look in `stats.gauges.govuk.app.signon.*.workers`
instead of `stats.gauges.govuk.app.signon.workers.` or we get no datapoints.


h-lame:
This should stop https://alert.publishing.service.gov.uk/cgi-bin/icinga/extinfo.cgi?type=2&host=monitoring-1.management.publishing.service.gov.uk&service=signon+background+worker+queue+size+unexpectedly+large from reporting as UNKNOWN in icinga.

PR #7344, opened 2018-03-07
Increase the max postgres connections in staging to match production

thomasleese:
There is no reason for them to be different, as both environments should be the same.

This change will restart Postgres in staging.

I decided not to merge the two staging and production settings into a common configuration file, because I'm not sure if that may affect any other production Postgres databases which would lead to them being restarted in office hours.

### DIFF
--- a/hieradata/class/staging/postgresql_primary.yaml
+++ b/hieradata/class/staging/postgresql_primary.yaml
@@ -1,4 +1,4 @@
-govuk_postgresql::server::max_connections: 200
+govuk_postgresql::server::max_connections: 300
 
 lv:
   postgresql:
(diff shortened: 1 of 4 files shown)


kevindew:
Sounds good to me. Do we need to update the files or can they inherit it from a non node specific file? 

Also do you want to update the icinga thresholds too?

thomasleese:
I believe that should be everything, I'm basing it on the last PR I raised to do this: https://github.com/alphagov/govuk-puppet/pull/6953

In terms of the icinga thresholds, they are automatically calculated from this value as 80% and 90%:

https://github.com/alphagov/govuk-puppet/blob/bf597dccef53b63a47a6bd8225537f5eb748d4b3/modules/govuk_postgresql/manifests/server.pp#L98-L99

PR #7348, opened 2018-03-07
Bump the number of Deploy Jenkins executors to 6

cbaines:
The Deploy Jenkins is mostly used to run jobs on other machines, and
the parallelism is controlled individually for each job. As the
resources of the Jenkins machine are not a concern for most jobs (as
the jobs are running on other machines), bump the number of executors
to allow more jobs to run in parallel.


cbaines:
Just tested this on Integration, seems to work fine.

PR #7350, opened 2018-03-07
Increase deploy-jenkins executors on integration

suzannehamilton:
This will help us avoid a long queue of app deployments and other tasks.

Jobs on deploy-jenkins don't tend to use many resources on the Jenkins box itself - they mostly trigger deployments or rake tasks on other hosts. So it's safe to increase the number of parallel jobs run by Jenkins.

cc @boffbowsh

### DIFF
--- a/hieradata/class/integration/jenkins.yaml
+++ b/hieradata/class/integration/jenkins.yaml
@@ -1,5 +1,5 @@
 ---
-govuk_jenkins::config::executors: '4'
+govuk_jenkins::config::executors: '8'
 
 govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::jobs::deploy_licensify


suzannehamilton:
@boffbowsh Fixed. 👍

cbaines:
I went for 6, but we've both ended up doing this...https://github.com/alphagov/govuk-puppet/pull/7348

Our changes are different though, I think the `hieradata_aws` needs editing, as the the Deploy Jenkins is in AWS?

h-lame:
I wonder if we should do it to all of them?  We frequently see some deploys being held up on production because we have a bunch of long lived tasks that run on a schedule.

suzannehamilton:
Closing this PR in favour of #7348.

I agree with @h-lame that it'd be good to bump this on all environments if it works on integration.

cbaines:
I've proposed increasing the executors on staging/production here now https://github.com/alphagov/govuk-puppet/pull/7382

PR #7351, opened 2018-03-07
Increase no of unicorn workers for gov-frontend

elenatanasoiu:
## Background 

Following the work done here: https://github.com/alphagov/govuk-puppet/pull/7255 we are increasing the number of unicorn workers to better handle the number of requests being done on the frontend boxes.

There's now a high amount of requests going to `government-frontend` as it has taken over responsibility from `whitehall`. 

## Current traffic

screen shot 2018-03-07 at 15 54 53

For a period between 9AM and 12PM (3 hours) there were 440K requests (GET & POST) made to `government-frontend`.

That boils down to 440.000 requests / 3 hours / 60 minutes / 60 seconds = **41 requests/s**.

We currently have 2 unicorn workers for each machine so 2 x 3 = 6 workers to consume these requests.

So 41 req / 6 workers = **~7 req / worker**

If we increase the workers from 2 to 4, that would result in 4x3 = 12 workers:

So 41 req / 12 workers = **~3.5 req / worker**

## Conclusion

1. As a first step, we'll increase the number of workers from 2 -> 4. 
2. We'll also increase the cpu warning threshold  for high CPU usage as we've increased `government-frontend` machines from 2 cores to 4 cores.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -411,8 +411,8 @@ govuk::apps::frontend::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::frontend::redis_port: "%{hiera('sidekiq_port')}"
 govuk::apps::frontend::unicorn_worker_processes: "4"
 
-govuk::apps::government-frontend::nagios_memory_warning: 900
-govuk::apps::government-frontend::nagios_memory_critical: 1000
+govuk::apps::government-frontend::nagios_memory_warning: 1800
+govuk::apps::government-frontend::nagios_memory_critical: 2000
 govuk::apps::government-frontend::unicorn_worker_processes: "8"
 
 govuk::apps::govuk_cdn_logs_monitor::processed_data_dir: '/mnt/logs_cdn/data'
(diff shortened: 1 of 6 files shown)


boffbowsh:
I think we'll also need to increase the thresholds for memory usage too. This is especially important as Icinga will restart the processes if they go above the Critical level

elenatanasoiu:
@boffbowsh I've checked how much memory `government-frontend` is using on each box: 
screen shot 2018-03-12 at 17 21 21

Graphite link: https://graphite.publishing.service.gov.uk/render/?width=586&height=308&_salt=1520870375.531&target=movingAverage(frontend-?_frontend.processes-app-government-frontend.ps_data%2C1000)

This can also be seen in grafana: https://grafana.publishing.service.gov.uk/dashboard/db/memory-dashboard?orgId=1&panelId=2&fullscreen&from=now-30d&to=now

It seems the memory usage hovers around 400-700 MB, so I've doubled the warning and critical thresholds from 900 and 1000 to, respectively, 1800 and 2000.

h-lame:
Perhaps the provisioning changes already did this, but if we add another (potential) 1Gb of memory consumption to these boxes do we run out of actual memory on the box?

elenatanasoiu:
The machines have 8GB of RAM each, set here: 
https://github.com/alphagov/govuk-provisioning/blob/master/vcloud-launcher/production_carrenza/frontend.yaml#L106

Looking at how much memory all apps on the frontend boxes, consume, two of them use max 4GB, and one uses 700 MB. 
screen shot 2018-03-12 at 17 47 52

PR #7356, opened 2018-03-08
Modify copy to staging job

suthagarht:
- We noticed an indentation error. This was noticed during a puppet
  apply on the actual jenkins server.

Pair: @schmie @suthagarht


suthagarht:
We have removed an indentation error.

PR #7358, opened 2018-03-08
Fix yaml of template copy_data_from_staging_to_aws.yaml.erb

suthagarht:
- We have edited the layout of the YAML file.

Pair: @schmie @suthagarht


surminus:
We can see _what_ you have done from the code, the commit message should explain _why_ you have done this.

PR #7362, opened 2018-03-09
Change postgres standby critical threshold from 16MB to 512MB

kevindew:
We regularly see this alert come through as a 2nd line critical alert
that quickly resolves itself, which indicates this really isn't a
critical alert.

From reading fbc9f18 these numbers were not picked for any particular
significance and from my (meagre) understanding of 90dc023b we store 4GB of
data to be synced. So this would imply that this isn't really a huge concern
unless we get close to exhausating this.

With the change in postgres usage from changes to Email Alert API and
Content Performance Manager we are seeing Postgres increase in data size
more rapidly and seems to cause this alert to trip. By moving this to
512MB we are much less likely to see this as critical unless something
has actually gone wrong.


boffbowsh:
Yes, in itself, but this is a good proxy for "postgres is doing too much". We should also add some metrics and alerts for CPU usage on postgres I think.

kevindew:
@boffbowsh sounds a good idea - it's had a lot more flowing through it recently.

PR #7363, opened 2018-03-09
Add whitehall 5xx analysis grafana dashboard

h-lame:
These exist on production and staging but were manually provisioned so
they were a) different between staging and prod, b) broken because they
used the old style prefixes for querying kibana.  Moving the defintion
into puppet should avoid those discrepancies in the future.

### DIFF
--- a/modules/grafana/files/dashboards/whitehall_5xx_analysis.json
+++ b/modules/grafana/files/dashboards/whitehall_5xx_analysis.json
@@ -1,0 +1,836 @@
+{
+  "annotations": {
+    "list": []
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "hideControls": false,
+  "id": 8,
+  "links": [],
+  "refresh": false,
+  "rows": [
+    {
+      "collapse": false,
+      "height": "250px",
+      "panels": [
+        {
+          "aliasColors": {},
+          "bars": false,
+          "dashLength": 10,
+          "dashes": false,
+          "datasource": "Elasticsearch",
+          "editable": true,
+          "error": false,
+          "fill": 1,
+          "grid": {},
+          "id": 9,
+          "legend": {
+            "avg": false,
+            "current": false,
+            "max": false,
+            "min": false,
+            "show": true,
+            "total": false,
+            "values": false
+          },
+          "lines": true,
+          "linewidth": 2,
+          "links": [],
+          "nullPointMode": "connected",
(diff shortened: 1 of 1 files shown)


modules/grafana/manifests/dashboards.pp
```diff
@@ -39,6 +39,7 @@
     "${dashboard_directory}/edge_health.json": content => template('grafana/dashboards/edge_health.json.erb');
     "${dashboard_directory}/origin_health.json": content => template('grafana/dashboards/origin_health.json.erb');
     "${dashboard_directory}/whitehall_health.json": content => template('grafana/dashboards/whitehall_health.json.erb');
+    "${dashboard_directory}/whitehall_5xx_analysis.json": content => template('grafana/dashboards/whitehall_5xx_analysis.json.erb');
```
cbaines: This doesn't look to use any of the templating functions, so you might want to just add it as a file.

However, some of the other dashboards that use Elasticsearch do use templating, so it might be good to look at why that is used.

h-lame: Seems the templating used by the other ERB dashboards is to use `fields_prefix` but we know that whitehall doesn't use one so I reckon we're ok without doing that here.  Most of the other ERB dashboards don't actually use templating either, do you know if we're planning to move them all to files rather than templates?

I could argue for leaving them as templates, so it's less noisy if we _do_ need to introduce templating features, but it's not a hill I want to die on.

cbaines: > Seems the templating used by the other ERB dashboards is to use fields_prefix but we know that whitehall doesn't use one so I reckon we're ok without doing that here.

Sounds good.

> Most of the other ERB dashboards don't actually use templating either, do you know if we're planning to move them all to files rather than templates?

I put up a Pull Request for that change a couple of days ago, just because it seemed a bit neater [1].

1: https://github.com/alphagov/govuk-puppet/pull/7349

> I could argue for leaving them as templates, so it's less noisy if we do need to introduce templating features, but it's not a hill I want to die on.

I don't mind particularly either, just though I'd mention it.

h-lame: I hadn't spotted that PR, if this is the direction of travel, I'll follow suit.

h-lame: done in a rebase.

cbaines: Ok :+1:

PR #7371, opened 2018-03-13
Add the right environment variables to run the data pipeline job

MatMoore:
See https://github.com/alphagov/govuk-secrets/pull/264

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/extract_app_performance.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/extract_app_performance.yaml.erb
@@ -18,7 +18,32 @@
     builders:
         - shell: |
             PIPENV_VENV_IN_PROJECT=1 pipenv install --three --skip-lock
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/templates/jobs/extract_app_performance.yaml.erb
```diff
@@ -18,7 +18,32 @@
...
+    wrappers:
+      - ansicolor:
+          colormap: xterm
+      - inject-passwords:
+          global: false
+          mask-password-params: true
+          job-passwords:
+            - name: GOOGLE_PRIVATE_KEY_ID
+              password:
+                '<%= @google_private_key_id %>'
```
cbaines: I'm not sure if you need the `@` here, as it doesn't seem to be used for the `app_domain_internal`.

MatMoore: As far as I can tell the instance variables are the idiomatic style for ERB templates in puppet so I'll change `app_domain_internal`.

PR #7382, opened 2018-03-14
Up executors for the Deploy Jenkins to 6

cbaines:
In staging and production, as sometimes they get quite busy. Most of
the load is on other machines, so this shouldn't cause an issue with
load.


cbaines:
This follows on from increasing the number of executors in Integration https://github.com/alphagov/govuk-puppet/pull/7348

PR #7383, opened 2018-03-14
AWS RDS Postgresql password

suthagarht:
- In our juenkins job to copy data from Carrenza staging to AWS staging,
  we have a neccity to provide 'aws_db_admin' as the destination user.
While we do provide the user name explicity via the script used to
perform the task, we wer not providing the password from secrets.

https://trello.com/c/cYTPEGvn/1044-data-sync-from-staging-carrenza-to-staging-aws

Solo: @suthagarht


suthagarht:
Not necessary

PR #7387, opened 2018-03-15
Increase freshness threshold for duplicity backups

MatMoore:
These backups seem to take a long time. They start at 4:20 in the
morning, and we are being alerted that the backup is still running at
10am.

Our developer docs warn that the task may still be running, so this
seems like a common occurance.
https://docs.publishing.service.gov.uk/manual/alerts/offsite-backups.html

We have no logs for this process, so I'm going to raise a separate
PR to add logs.


MatMoore:
Actually there are some logs, because cron errors get sent to a mailing list. It's just nothing gets logged if the job ran without errors.

PR #7395, opened 2018-03-19
Change app_domain for aws staging to staging.govuk.digital

schmie:
- config in hieradata_aws/staging.yaml still pointed to staging.publishing.service.gov.uk and is causing the
  govuk_crawler_worker to get confused.

- Furthermore, the targets of crawler as they are defined in hieradata_aws/common.yaml.
  Changed this now to prevent the crawler from timing out trying to access Carrenza.

Solo: @schmie

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -473,9 +473,8 @@ govuk::apps::govuk_crawler_worker::blacklist_paths:
 
 govuk::apps::govuk_crawler_worker::enabled: true
 govuk::apps::govuk_crawler_worker::root_urls:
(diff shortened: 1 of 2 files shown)


issyl0:
I don't think it should time out trying to access Carrenza, they're all public-over-the-Internet URLs?

schmie:
Hi @issyl0, both assets urls (see below) do result in a 404 from my workstation. I would assume that is the same from awd->carrenza internal?! 
https://assets.digital.cabinet-office.gov.uk/
https://assets.publishing.service.gov.uk/

PR #7397, opened 2018-03-19
Add redis-tools to db_admin

elenatanasoiu:
Install redis-tools so we can access redis on the AWS integration environment

Related to: https://github.com/alphagov/govuk-aws/pull/522

For: https://trello.com/c/CA3PFmY1/129-install-redis-tools-on-aws-integration

### DIFF
--- a/modules/govuk/manifests/node/s_db_admin.pp
+++ b/modules/govuk/manifests/node/s_db_admin.pp
@@ -49,6 +49,10 @@ class govuk::node::s_db_admin(
     require => Apt::Source['gof3r'],
   }
 
+  package { 'redis-tools':
+    ensure  => $ensure,
+  }
+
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/node/s_db_admin.pp
```diff
@@ -49,6 +49,11 @@
     require => Apt::Source['gof3r'],
   }
 
+  package { 'redis-tools':
+    ensure  => $ensure,
+    require => Apt::Source['redis-tools'],
```
boffbowsh: I don't think this is required as redis-tools is part of the normal apt repo.

elenatanasoiu: Removed that `require`

PR #7401, opened 2018-03-21
Add email alert api to aws backend

bevanloon:
Email alert api is still hanging around on `backend` machines despite now having its own servers. A previous commit set `govuk::apps::email_alert_api::ensure: 'absent'` in `hieradata/class/backend.yaml` but this had no effect because backend machines were not configured to include `email-alert-api`. We still want to stop and remove the various `email-alert-api` services and processes on `backend` machines.

This PR attempts to address that on our integration environment by creating an `else` condition in `email_alert_api` manifest which ensures the app is absent when it is not `$enabled`.

If this successfully removes the recalcitrant services, a subsequent PR will do the same to non-aws environments so that it runs on staging and production.

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api.pp
+++ b/modules/govuk/manifests/apps/email_alert_api.pp
@@ -103,6 +103,8 @@ class govuk::apps::email_alert_api(
 ) {
 
   if $enabled {
+    notify { 'Email alert api enabled': }
+
     govuk::app { 'email-alert-api':
       app_type                 => 'rack',
       port                     => $port,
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/email_alert_api.pp
```diff
@@ -232,4 +232,9 @@
       }
     }
   }
+  else {
+    govuk::app { 'email-alert-api':
+      ensure => 'absent',
```
thomasleese: I think you can use the `absent` keyword here rather than a string.

bevanloon: I thought I would leave it consistent with the other occurrence in the file.

thomasleese: Yeah, I don't mind either way!

bevanloon:
Closing in favour of https://github.com/alphagov/govuk-puppet/pull/7411

PR #7404, opened 2018-03-21
Set the transition authentication variables when syncing staging

thomasleese:
The staging sync was recently [changed to use a new script which has different variable names for the passwords, `PG_TR_SRC_ENV_SYNC_PW` and `PG_TR_DST_ENV_SYNC_PW`](https://github.com/alphagov/env-sync-and-backup/pull/72/files#diff-f99ca43a8a970b1128838bdd6b3d6866R11). Since before we were using the same password for normal postgres and transition postgres, I assume that we can set it to be the same as normal postgres.


issyl0:
Ah, this was only fixed for the AWS copy script in #7392.

PR #7411, opened 2018-03-23
Remove email alert api and procfile worker from backend

bevanloon:
We want to stop email-alert-api and associated profile worker from running on backend machines. 

* Add `email-alert-api` into backend machines in the hieradata
* Set `email_alert_api::enabled` to `false` in backend hieradata
* Remove `email_alert_api::ensure` setting from backend hieradata since we are relying on `enabled` instead.
* Add else condition on `enabled` in `email_alert_api` manifest and when it is `false`, ensure `email-alert-api` and its profile worker are absent

This gets us most of the way to remove all traces of email-alert-api from backend machines. After this there will still be some logship processes hanging around and also a `rackup` command.

[Trello](https://trello.com/c/ida7jtCQ/715-investigate-properly-removing-email-alert-api-from-backend-machines)

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api.pp
+++ b/modules/govuk/manifests/apps/email_alert_api.pp
@@ -232,4 +232,16 @@ class govuk::apps::email_alert_api(
       }
     }
   }
+  else {
+    govuk::app { 'email-alert-api':
+      ensure   => 'absent',
+      app_type => 'rack',
+      port     => $port,
+    }
+
+    govuk::procfile::worker {'email-alert-api':
+      ensure         => 'absent',
+      enable_service => false,
+    }
+  }
 }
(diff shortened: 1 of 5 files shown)


modules/govuk/manifests/apps/email_alert_api.pp
```diff
@@ -232,4 +234,18 @@
       }
     }
   }
+  else {
+    notify { 'Email alert api not enabled': }
```
thomasleese: Do we still need these `notify`s?

bevanloon: probably not....removed them

kevindew:
All looks good to me, you been able to give it a spin on integration?

bevanloon:
@kevindew yep - it looks pretty good....tested on backend and email-alert-api machines and it seems to work.

kevindew:
woop woop 🎉

PR #7415, opened 2018-03-23
Fixing bad var for whisper-backup repo hostname

rtrinque:


modules/govuk/manifests/node/s_graphite.pp
```diff
@@ -10,9 +10,14 @@
...
 #   Hour of the day to run the backup crons.
 #
+#  [*apt_mirror_hostname*]
+#    The hostname of the local aptly mirror.
+#
+
 class govuk::node::s_graphite (
   $graphite_path = '/opt/graphite',
   $graphite_backup_hour = 23,
+  $apt_mirror_hostname = 'apt.production.alphagov.co.uk',
```
issyl0: This URL redirects to `apt.publishing.service.gov.uk` - can we avoid the extra hop and replace `alphagov.co.uk`?

rtrinque: Sure yes, went with alphagov.co.uk as it's what we're using in our code base at the moment and service.gov.uk has a carrenza flavour to me.
We could also go to what domain we want to use when the migration is over, something in .digital maybe?

issyl0: Ah yes, that is the current Carrenza production URL.

For context, Carrenza used to be on the `environment.alphagov.co.uk` domain, which now all redirect to `publishing.service.gov.uk`.

When the migration is over in production, we'll be using `publishing.service.gov.uk` on AWS.

But for now, Carrenza is still the canonical source of packages, I believe?

PR #7418, opened 2018-03-28
Update .ruby-version to 2.4.2

Davidslv:
Can we upgrade this to a more current ruby version please?


cbaines:
Looks like the gem related failure in the CI job might be helped by upgrading the json gem, the changelog entry for 1.8.2 mentions some fixes for Ruby 2 [1].

However, I think there might be more significant issues with this change. If I remember correctly, the current ruby version might be useful when getting govuk-puppet working on a new machine which it hasn't run on yet (aka bootstrapping). So even if the build passes, I'd suggest checking in to reasons why this specific version of Ruby might be in use.

1: https://github.com/flori/json/blob/master/CHANGES.md#2015-01-08-182

h-lame:
I seem to recall that we can't, for some reason.

h-lame:
Oops, didn't mean to close, sorry!

h-lame:
https://github.com/alphagov/govuk-puppet/pull/4896 might have some context

tijmenb:
Closing this because it appears to be a larger piece of work, feel free to reopen!

PR #7420, opened 2018-03-29
Serve Whitehall attachments from Asset Manager

chrisroos:
Removing this more specific path will mean that requests for Whitehall attachments will be caught by the more general '/government/uploads/' path and routed to Asset Manager instead of Whitehall.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1710,8 +1710,12 @@ router::assets_origin::app_specific_static_asset_routes:
 
 router::assets_origin::whitehall_uploaded_assets_routes:
   - '/government/placeholder'
-  - '/government/uploads/system/uploads/attachment_data/file/'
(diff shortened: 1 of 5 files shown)


barrucadu:
I've rebased this and added a couple more routes.  When this is deployed, these routes in whitehall shouldn't be hit at all:

- `get '/government/uploads/system/uploads/consultation_response_form/*path.:extension'`
- `get '/government/uploads/system/uploads/attachment_data/file/:id/*file.:format'`

Furthermore, I don't think the asset-manager redirect one will be hit:

- `get '/government/uploads/*path'`

When we deploy this change to staging we can monitor those routes for a while and confirm these suspicions.

barrucadu:
I am a bit concerned about the overhead of the regex route, but that's why we have so much cache, right?

floehopper:
🎉 Yay! That's great news! 🎉

PR #7427, opened 2018-04-05
Setup sidekiq monitoring for manuals publisher and support api

cbaines:


cbaines:
Thanks for reviewing Kevin :)

PR #7432, opened 2018-04-09
Don't set the Signon instance name in production

cbaines:
Even on AWS. This differs from the current production environment, in
which this is unset, I believe intentionally.

The Signon service assumes that this will be unset in the production
environment, e.g. it won't enforce the use of 2 step verification for
admin users if this is set.


cbaines:
Thanks for taking a look @1pretz1 :)

PR #7454, opened 2018-04-12
Fix Sidekiq monitoring typo related to Manuals Publisher

cbaines:
Remvoe _api from the hieradata. This was causing the redis envvar
configuration to use the default values, e.g. host 127.0.0.1 that
don't work.

I think this was a copy/paste error when I added the configuration.


cbaines:
Awesome, thanks for taking a look @1pretz1

PR #7457, opened 2018-04-12
Add Strict-Transport-Security headers in integration

rubenarakelyan:
This commit includes the `add-sts.conf` file in nginx config for integration, which will set the `Strict-Transport-Security` header. AWS ELBs should pass this through to the requester.

This is done separately in integration since we disable SSL completely, and a side-effect of this is
to stop setting the `Strict-Transport-Security` header. SSL is disabled because SSL in integration is
 terminated at the AWS ELBs.

### DIFF
--- a/modules/nginx/templates/default-vhost.conf
+++ b/modules/nginx/templates/default-vhost.conf
@@ -1,6 +1,8 @@
 server {
   listen 80 default_server;
   <%- if scope.lookupvar('::aws_migration') %>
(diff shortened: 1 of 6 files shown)


issyl0:
It would be good to include why we're doing this, eg why the usual approach doesn't work - in the commit message.

rubenarakelyan:
Yup I've added some more info to the commit about the reasons behind this.

PR #7458, opened 2018-04-13
Update to java8 and install graphviz on dev VM

MatMoore:
We already use graphviz to generate some diagrams, and I want to generate some more.

We've already updated Java on CI in https://github.com/alphagov/govuk-puppet/pull/7449

cc @kevindew


kevindew:
@MatMoore This will be good to merge now

PR #7459, opened 2018-04-13
Fix openjdk running each install

kevindew:
I noticed that every time puppet runs for openjdk it would come up with `Notice: /Stage[main]/Govuk_java::Set_defaults/Exec[java-set-default-jdk]/returns: executed successfully` for openjdk 8 which would cause Jenkins restarts each time puppet runs.

This was occuring because `update-java-alternatives` was failing to update javac due to a change in the jinfo files for openjdk8.

/cc @MatMoore

### DIFF
--- a/modules/govuk_java/files/update-java-alternatives
+++ b/modules/govuk_java/files/update-java-alternatives
@@ -1,0 +1,155 @@
+#! /bin/bash
+
+shopt -s dotglob
+
+prog=$(basename $0)
+
+usage()
+{
+    rv=$1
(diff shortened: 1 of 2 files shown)


kevindew:
@thomasleese it's just the standard ubuntu one from a later version. I'll update the commit with the source of the file.

thomasleese:
@kevindew aha, I see. Makes sense!

kevindew:
@thomasleese ha yeah I wouldn't trust my bash skills to produce that.

PR #7469, opened 2018-04-17
Add all jobs provided by sync scripts to jenkins job description

schmie:
- The Jenkins job has an incomplete list and updating of the joblist via the
  web interface is currently broken due to error:
  SlackNotifier$SlackJobProperty is missing its descriptor

- Will open another card for the error, but this change is helping consistency
  in any case

Solo: @schmie

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/copy_data_from_staging_to_aws.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/copy_data_from_staging_to_aws.yaml.erb
@@ -96,6 +96,8 @@
             description: 'Choose the thing to sync. All is the default, but some jobs may run but not do anything due to your config for the destination environment.'
             choices:
                 - all
+                - attachments
+                - assets
                 - elasticsearch-rummager
                 - mongo-api
                 - mongo-normal


modules/govuk_jenkins/templates/jobs/copy_data_from_staging_to_aws.yaml.erb
```diff
@@ -96,10 +96,13 @@
             description: 'Choose the thing to sync. All is the default, but some jobs may run but not do anything due to your config for the destination environment.'
             choices:
                 - all
+                - attachments
+                - assets
                 - elasticsearch-rummager
                 - mongo-api
                 - mongo-normal
                 - mongo-router
+                - mongo-pp
```
issyl0: We don’t need to sync mongo-pp because Performance Platform (pp) is staying on Carrenza/PaaS.

issyl0: Maybe this is another story to remove it from AWS if it is in staging. The database shouldn’t have anywhere to sync to as we shouldn’t have any `performance-mongo` machines.

schmie:
Removed mongo-pp from the list of selectable jobs as deprecated following comment of @issyl0

PR #7474, opened 2018-04-18
Schedule daily jenkins job to send Whitehall notifications

elliotcm:
Part of https://trello.com/c/QVPUgQem/68-alert-the-author-of-a-consultation-document-via-email
Triggers https://github.com/alphagov/whitehall/pull/3972


elliotcm:
@bevanloon This should be OK to go out as-is because the jenkins job won't be created in production, and should make for a much smaller PR to enable it once we're happy with the content.

PR #7475, opened 2018-04-18
Invert the default for the default field prefix on deployment dashboards

cbaines:


cbaines:
> For apps that didn't have a field_prefix value set do they now need a field_prefix: '@fields.' value?

If we wanted to keep the same behaviour, then yes. But I checked in Logit, and assuming the query I used was correct, I couldn't find any data using the prefix of `@fields.`, so I think by changing to not using a prefix, this will fix the missing data on the dashboards for some apps (e.g. collections).

PR #7480, opened 2018-04-19
Run Jenkins Copy_Data_to_Integration at 3am UTC instead of 4.

schmie:
- At the moment this jobs runs well into the day (finished at 11am).

- An aspect seems to be Jenkins avoid job collisions.

- We like to test the impact of scheduling an hour early.

Pair: @schmie @pmanrubia


kevindew:
@schmie Ooh I was just looking into this one too. Glad I found this!

You might want to make this even earlier as it finishing at 10am is also pretty limiting.

schmie:
"Merged" with time zone update #7484

PR #7483, opened 2018-04-19
Add an AWS_SESSION_TOKEN parameter to the Deploy DNS job

issyl0:
- People deploying DNS these days don't have old style credentials per
  environment, they have to `aws assume-role`, for which AWS also
  requires a `SESSION_TOKEN`.


stephenharker:
LGTM

issyl0:
I can't find anything else that uses these values in the Puppet manifests - they're probably just envvars that the underlying Terraform command uses.

PR #7503, opened 2018-04-25
Remove Chris Patuzzo who is no longer at GDS

tuzz:


tijmenb:
Superseded by https://github.com/alphagov/govuk-puppet/pull/7519. Sad to see you go @tuzz!

PR #7504, opened 2018-04-25
Add the `terraform-docs` package to the CI machines

issyl0:
- This is needed for https://github.com/alphagov/govuk-aws/pull/553 to
  be able to run CI checks for up to date Terraform documentation.


issyl0:
This is ready to go as per https://docs.publishing.service.gov.uk/manual/debian-packaging.html#initial-setup as we need to get the repo into aptly via Puppet in order to add the new package (which I've generated [here](https://deploy.publishing.service.gov.uk/job/build_fpm_package/47/) for CI to be able to use it.

PR #7517, opened 2018-05-02
Try publishing overdue Whitehall publications before calling the on-call person

thomasleese:
This sets up an Icinga event handler to try and publish overdue Whitehall publications before calling the on-call person.

[Trello Card](https://trello.com/c/2fww91ri/118-make-icinga-try-publishing-overdue-whitehall-publications-before-calling-the-on-call-person-2)

### DIFF
--- a/modules/monitoring/manifests/checks.pp
+++ b/modules/monitoring/manifests/checks.pp
@@ -39,13 +39,15 @@ class monitoring::checks (
   }
 
   icinga::check { "check_whitehall_overdue_from_${::hostname}":
-    check_command       => 'check_whitehall_overdue',
-    service_description => 'overdue publications in Whitehall',
-    use                 => 'govuk_urgent_priority',
-    host_name           => $::fqdn,
-    notes_url           => monitoring_docs_url(whitehall-scheduled-publishing),
-    action_url          => "https://${whitehall_hostname}${whitehall_overdue_url}",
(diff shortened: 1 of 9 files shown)


modules/monitoring/manifests/checks.pp
```diff
@@ -39,12 +39,15 @@
...
-    notes_url           => monitoring_docs_url(whitehall-scheduled-publishing),
-    action_url          => "https://${whitehall_hostname}${whitehall_overdue_url}",
+    check_command              => 'check_whitehall_overdue',
+    service_description        => 'overdue publications in Whitehall',
+    use                        => 'govuk_urgent_priority',
+    host_name                  => $::fqdn,
+    notes_url                  => monitoring_docs_url(whitehall-scheduled-publishing),
+    action_url                 => "https://${whitehall_hostname}${whitehall_overdue_url}",
+    event_handler              => 'publish_overdue_whitehall',
+    attempts_before_hard_state => 2,
```
thomasleese: @tijmenb I've just added a line to make this more explicit. Here it will check for overdue publications twice (before I made it explicit, it would check four times) with an interval of 10 seconds, triggering a `SOFT` alert if there are overdue publications before changing to the `HARD` state if necessary, which is when someone will be called.

tijmenb: 👍 awesome thanks, clear now!

PR #7525, opened 2018-05-03
Add more unicorn workers to asset manager

thomasleese:
We're seeing timeouts while large uploads are happening so this should help the situation.

As asset manager continues to serve and receive more assets this will become more necessary.

[Trello Card](https://trello.com/c/ktNXJSsb/220-improve-resources-for-asset-manager)


thomasleese:
@barrucadu It's potentially related, but I don't think it'll completely solve that problem.

boffbowsh:
This is ok as a stopgap, but perhaps having a pool of unicorns for doing read requests only would be a good idea too. They're always* pretty quick, it's the writes that take the time.

thomasleese:
@boffbowsh Agreed, I'll make it clear in the card that this is the next thing to do.

PR #7528, opened 2018-05-03
PuppetDB Java Heap Size.

suthagarht:
- We have noticed that the AWS Staging pupet master is crashing
frequently.
- The symptom is always `java.lang.OutOfMemoryError: Java heap space`.
- We can see that Carrenza staging is still using 192MB. We also know
that we have 4CPU cores in AWS compared to 8 in Carrenza. However,
reducing the heap size might/should trigger frequent garbage collection
and ultimately less probability of getting full.
- This might have an adverse effect. however, this is a simpler change
to test in the new environment before choosing an expensive instance
type.

Note: Please check the trello card for more details.

https://trello.com/c/nr8kuCOv/1170-aws-staging-puppet-is-broken

Solo: @suthagarht

### DIFF
--- a/modules/govuk_puppetdb/manifests/config.pp
+++ b/modules/govuk_puppetdb/manifests/config.pp
@@ -7,7 +7,7 @@ class govuk_puppetdb::config (
   # installing puppetdb-2.x on the new Trusty Puppetmasters on AWS
   # Use aws_migration fact
   if $::aws_migration {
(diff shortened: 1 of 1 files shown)


issyl0:
Closing this as it's been superseded by #7550 as we're upping the max size rather than reducing it, as there is commonly an out of memory error.

PR #7533, opened 2018-05-08
Remove Go Free Range from Integration

bevanloon:
Remove Chris Lowis, Chris Roos and James Mead from Integration.


vanitabarrett:
Completed as part of https://github.com/alphagov/govuk-puppet/pull/7534

PR #7536, opened 2018-05-09
Automate cache clearing in emergency banner publishing

elliotcm:
This is currently done as another step via fabric scripts.

Incorporates the steps from here:
https://docs.publishing.service.gov.uk/manual/emergency-publishing.html#adding-emergency-publishing-banners

https://trello.com/c/yFrK9Hue/119-update-emergency-publishing-process-to-not-require-fabric-2


elliotcm:
There's lots of repetition here but I'm not super comfortable with bash scripting, especially with an SSH layer involved.

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
```diff
@@ -25,7 +25,59 @@
           LINK=$(escapeRakeArg $LINK)
```
kevindew: I wonder if you want to add `set -ex` to this file so it can output everything and break on errors.

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
```diff
@@ -25,7 +25,59 @@
...
 
+          echo "Deploying emergency banner"
           ssh deploy@$(govuk_node_list -c frontend --single-node) "cd /var/apps/static && govuk_setenv static bundle exec rake \"emergency_banner:deploy[$CAMPAIGN_CLASS,$HEADING,$SHORT_DESCRIPTION,$LINK,$LINK_TEXT]\""
+
+          echo "Clearing template cache.."
+          for node in `govuk_node_list frontend`; do
+            echo "-- $node"
+            ssh deploy@$node "sudo rm /var/apps/static/public/templates/{wrapper,homepage,header_footer_only,core_layout}.html.erb"
+            ssh deploy@$node "sudo rm -rf /var/apps/{,government-}frontend/tmp/cache/*"
+          done
```
kevindew: nitpick - but think you missed a new line

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
```diff
@@ -25,7 +25,59 @@
...
+            ssh deploy@$node "sudo service memcached restart"
+          done
+
+          echo "Clearing varnish cache.."
+          for node in `govuk_node_list cache`; do
+            echo "-- $node"
+            ssh deploy@$node "sudo varnishadm 'ban req.url ~ .'"
+          done
+
+          <% if @clear_cdn_cache %>
```
kevindew: Something that makes me a bit nervous here - but not something where I know a particularly good solution - is that I don't think we'll ever get to test this code and would only find out if it's good at the point of doing this for real. 

Admittedly this is more a problem with our infrastructure than the script. 

Best I can think of for this is to have some sort of help message alert before it of a link for if it fails. 

Sidenote: I wonder how this small set of paths got defined - they seem quite abstract.

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
```diff
@@ -25,7 +25,59 @@
           LINK=$(escapeRakeArg $LINK)
           LINK_TEXT=$(escapeRakeArg $LINK_TEXT)
 
+          echo "Deploying emergency banner"
           ssh deploy@$(govuk_node_list -c frontend --single-node) "cd /var/apps/static && govuk_setenv static bundle exec rake \"emergency_banner:deploy[$CAMPAIGN_CLASS,$HEADING,$SHORT_DESCRIPTION,$LINK,$LINK_TEXT]\""
+
+          echo "Clearing template cache.."
```
thomasleese: I wonder if we should extract this part and the part below about clearing the CDN caches into separate Jenkins jobs which get triggered from this job. If we don't then it's likely people will still need to use the Fabric scripts for just clearing some caches unrelated to the emergency publishing drill and it would be very easy for the cache clearing code in the Fabric scripts to end up different to the code here. My suggestion would be to move it into separate Jenkins tasks and then get rid of the Fabric scripts related to caching completely (or at least deprecate them for the moment).

As it is, people seem to prefer using Jenkins over the Fabric scripts so this might be nice anyway, and it allows us to more easily see when the tasks have been run and why.

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
elliotcm: What does this do?

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
elliotcm: That was kinda deliberate as both steps are part of clearing the template cache.  Maybe I can clarify that with comments (or better, functions) instead.

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
kevindew: -e will make it stop executing if one of the commands returns a non 0 exit code
-x will echo out the commands that are being run (which should make it easier to follow progress)

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
elliotcm: Perhaps if we take @thomasleese's suggestion to the extreme we could extract CDN clearing into its own job which we could test during a quiet period.

As for the paths, those are (were?) the top 10 most visited paths according to the fabric scripts.  I'll put a comment in to that effect and see if they're still current.

elliotcm:
Also I tricked you, there's a bug which none of you spotted which I definitely put in deliberately and didn't just think of in the shower this morning.  This is why we deffo need to be able to test these things thoroughly.

I've not run it on integration yet, it's a bit of a process so I'll do it once we're happy with the overall structure of this.

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
elliotcm: Sounds good.

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
```diff
@@ -26,6 +28,14 @@
           LINK_TEXT=$(escapeRakeArg $LINK_TEXT)
 
           ssh deploy@$(govuk_node_list -c frontend --single-node) "cd /var/apps/static && govuk_setenv static bundle exec rake \"emergency_banner:deploy[$CAMPAIGN_CLASS,$HEADING,$SHORT_DESCRIPTION,$LINK,$LINK_TEXT]\""
+    publishers:
+      - trigger:
+        project: clear_template_cache
```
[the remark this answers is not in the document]
elliotcm: I was basing it on this: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/templates/jobs/benchmark_search.yaml.erb#L35-L45

Which I guess the dashes turn it into an array?  I don't read YAML well.

modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
```diff
@@ -0,0 +1,45 @@
+---
+- job:
+    name: clear-cdn-cache
+    display-name: Clear production CDN cache
+    project-type: freestyle
+    description: "Clears the production CDN cache for the 10 most popular pages on GOV.UK."
```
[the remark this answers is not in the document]
elliotcm: Likely true, but I'm inclined to leave it because I've found YAML string handling to be finicky and the description is the one most likely to get special characters added to it later.

What do you think?

modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
```diff
@@ -0,0 +1,45 @@
...
+    properties:
+      - build-discarder:
+          days-to-keep: 30
+          artifact-num-to-keep: 5
```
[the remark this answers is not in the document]
elliotcm: Do you have some docs for that property?  Can't find anything on it.

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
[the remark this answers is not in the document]
thomasleese: Yeah, the dashes mean it's an array, and then the indentation after `trigger:` turns it into a hash. So actually, it should be like this:

```yaml
publishers:
  - trigger-parameterized-builds:
      - project: clear_template_cache
      - project: clear_frontend_memcache
      ...
```

modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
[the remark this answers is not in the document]
thomasleese: https://docs.openstack.org/infra/jenkins-job-builder/properties.html#properties.build-discarder

modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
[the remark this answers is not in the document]
thomasleese: I don't really mind, I just prefer to not have the quotes where it's not required. It might be better do to the following which should avoid any issues with special characters:

```yaml
description: >
  Clears the production CDN cache for the 10 most popular pages on GOV.UK.
```

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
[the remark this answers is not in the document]
elliotcm: Except without the parameterized part, right?  because there's no params here

modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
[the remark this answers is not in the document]
elliotcm: Looks good.

modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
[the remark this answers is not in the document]
elliotcm: Sounds fine.

modules/govuk_jenkins/templates/jobs/deploy_emergency_banner.yaml.erb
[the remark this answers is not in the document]
thomasleese: Oh yeah, good point.

elliotcm:
@thomasleese OK I've updated again with the above input.  I've added the jobs to any environment which already had the deploy banner job listed.

I've also added the cache clear triggers to the `remove_emergency_banner` job and updated that and the deploy banner job in the same way as the others regarding keeping builds.

modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
```diff
@@ -0,0 +1,46 @@
+---
+- job:
+    name: clear-cdn-cache
+    display-name: Clear production CDN cache
+    project-type: freestyle
+    description: >
+      Clears the production CDN cache for the 10 most popular pages on GOV.UK.
+    properties:
+      - build-discarder:
+          days-to-keep: 30
```
thomasleese: I think these should also use `num-to-keep`, and I would be tempted to make it slightly higher than 10, perhaps 30.

elliotcm: Yeah I just noticed that myself.  Updated, and changed to 30.

I've also removed the CDN job entirely from non-prod environments since it's hard coded to production.

elliotcm:
Tested on staging with @thomasleese.  Merging.

Docs update PR incoming.

PR #7537, opened 2018-05-09
Directing gor replay traffic from carrenza staging cache machines to …

rtrinque:
…AWS staging

  We set a global conf for the whole cache class and delete the machine specific configs

### DIFF
--- a/hieradata/class/staging/cache.yaml
+++ b/hieradata/class/staging/cache.yaml
@@ -4,5 +4,5 @@
 #
 # If changed, the old host record must be cleaned up;
 router::gor::replay_targets:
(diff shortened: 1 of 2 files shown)


cbaines:
Could you please put the commit message in the present tense, and watch the line length (GOV.UK uses 50 characters for the first line, which is quite short https://github.com/alphagov/styleguides/blob/master/git.md#structure ).

Also, please add the reason for the change in.

issyl0:
Closing this as we need to talk to GOV.UK and do some testing. Also, Gor does a DNS lookup on every request if it’s not pointed to an IP, which seems suboptimal.

PR #7542, opened 2018-05-10
Allow --sandbox option on govuk_app_console

Davidslv:
Rails provides a `--sandbox` option, this means that anything done during the session is done inside a transaction which is then rollbacked upon end of the session.

This commit is to enable that functionality when using `govuk_app_console` script.

Suggestions are highly accepted, my bash skills are not the best.

Thank you :)

### DIFF
--- a/modules/govuk_scripts/files/usr/local/bin/govuk_app_console
+++ b/modules/govuk_scripts/files/usr/local/bin/govuk_app_console
@@ -9,19 +9,19 @@
 
 set -eu
 
-if [ "$#" -ne "1" ] || [ "$1" = "-h" ]; then
-  echo "Usage: $(basename "$0") app_name" >&2
+if [ "$#" -gt 0 ] || [ "$1" = "-h" ]; then
+  echo "Usage: $(basename "$0") app_name [--sandbox]" >&2
   exit 1
 fi
 
 cd /var/apps/$1
 
 if [ -f bin/rails ]; then
-  sudo -u deploy govuk_setenv $1 bundle exec ./bin/rails console
+  sudo -u deploy govuk_setenv $1 bundle exec ./bin/rails console $2
 elif [ -f script/rails ]; then
-  sudo -u deploy govuk_setenv $1 bundle exec ./script/rails console
+  sudo -u deploy govuk_setenv $1 bundle exec ./script/rails console $2
 elif [ -f console ]; then
-  sudo -u deploy govuk_setenv $1 bundle exec ./console
+  sudo -u deploy govuk_setenv $1 bundle exec ./console $2
 else
   echo "I don't know how to run a console for $1"
   exit 1
--- a/modules/govuk_scripts/files/usr/local/bin/govuk_app_console
+++ b/modules/govuk_scripts/files/usr/local/bin/govuk_app_console
@@ -9,7 +9,7 @@
 
 set -eu
 
-if [ "$#" -gt 0 ] || [ "$1" = "-h" ]; then
+if [[ ($# -ne 1 && $# -ne 2) || "$1" == "-h" ]]; then
   echo "Usage: $(basename "$0") app_name [--sandbox]" >&2
   exit 1
 fi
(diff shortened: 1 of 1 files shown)


modules/govuk_scripts/files/usr/local/bin/govuk_app_console
```diff
@@ -9,19 +9,19 @@
...
 
 if [ -f bin/rails ]; then
-  sudo -u deploy govuk_setenv $1 bundle exec ./bin/rails console
+  sudo -u deploy govuk_setenv $1 bundle exec ./bin/rails console $2
 elif [ -f script/rails ]; then
-  sudo -u deploy govuk_setenv $1 bundle exec ./script/rails console
+  sudo -u deploy govuk_setenv $1 bundle exec ./script/rails console $2
 elif [ -f console ]; then
-  sudo -u deploy govuk_setenv $1 bundle exec ./console
+  sudo -u deploy govuk_setenv $1 bundle exec ./console $2
```
cbaines: I didn't know of this before, but it looks to be a Rails thing. This part of the script seems to be used for Rummager (I didn't see any other apps when I looked).

I'm guessing it won't work for Rummager, so maybe don't pass in the option, or show a message, or both?

modules/govuk_scripts/files/usr/local/bin/govuk_app_console
```diff
@@ -9,19 +9,19 @@
 
 set -eu
 
-if [ "$#" -ne "1" ] || [ "$1" = "-h" ]; then
-  echo "Usage: $(basename "$0") app_name" >&2
+if [ "$#" -gt 0 ] || [ "$1" = "-h" ]; then
```
cbaines: This says if the number of provided arguments is greater than 0, then show the help message right? If so, won't that happen always?

elliotcm: Agreed.  It also won't work because the LHS is a string but the RHS is an integer.

After [some research](https://stackoverflow.com/questions/18568706/check-number-of-arguments-passed-to-a-bash-script) I'd recommend `[[ ($# -ne 1 || $# -ne 2) || "$1" == "-h" ]]`, bearing in mind I wrote the not-amazing original clause.

modules/govuk_scripts/files/usr/local/bin/govuk_app_console
elliotcm: Correct, rummager is a Sinatra app so it's not likely to support this option (which I'd not heard of either).

I don't *think* we have any more Sinatra apps?  So yeah, should be OK to just error out if this option is provided in a Sinatra context.  We probably don't want to warn and continue as someone might do something they later regret, thinking it'll be rolled back.

Davidslv:
I will try to address your comments as soon as I can, haven't been forgotten :)

modules/govuk_scripts/files/usr/local/bin/govuk_app_console
```diff
@@ -9,19 +9,25 @@
...
 
 if [ -f bin/rails ]; then
-  sudo -u deploy govuk_setenv $1 bundle exec ./bin/rails console
+  sudo -u deploy govuk_setenv $1 bundle exec ./bin/rails console $2
 elif [ -f script/rails ]; then
-  sudo -u deploy govuk_setenv $1 bundle exec ./script/rails console
+  sudo -u deploy govuk_setenv $1 bundle exec ./script/rails console $2
 elif [ -f console ]; then
-  sudo -u deploy govuk_setenv $1 bundle exec ./console
+  if [[ -z $2 ]] ; then
```
Davidslv: I took this magic from here: https://askubuntu.com/questions/444082/how-to-check-if-1-and-2-are-null

Davidslv:
@elliotcm @cbaines Thank you for your comments, could you double check if this is a better approach please?

cbaines:
> @elliotcm @cbaines Thank you for your comments, could you double check if this is a better approach please?

Sure, I think it looks good now.

PR #7548, opened 2018-05-14
Update static dependent apps

sihugh:
We had an incident recently where static removed a thing that a frontend app was reliant on.  

We should keep this list of apps up to date as it will help us detect when we break things, as it will show up on the [deployment dashboard](https://grafana.publishing.service.gov.uk/dashboard/file/deployment_static.json?refresh=5s&orgId=1)


sihugh:
Thanks @bevanloon 👍

PR #7550, opened 2018-05-16
In AWS, bump the Puppet JVM heap size to 2G

issyl0:
- We're seeing the Puppetmaster run out of memory a lot and crash. Along with the other mitigations, like making the instance more powerful in staging and production, increase the maximum Java heap size as this is always the error we see.

https://trello.com/c/357jwiXx/1130-puppetmaster-in-need-of-bigger-heap-size

### DIFF
--- a/modules/govuk_puppetdb/manifests/config.pp
+++ b/modules/govuk_puppetdb/manifests/config.pp
@@ -7,7 +7,7 @@ class govuk_puppetdb::config (
   # installing puppetdb-2.x on the new Trusty Puppetmasters on AWS
   # Use aws_migration fact
   if $::aws_migration {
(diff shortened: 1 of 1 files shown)


issyl0:
Integration’s instance is [still an m5.xlarge](https://github.com/alphagov/govuk-aws-data/pull/137), but that does come with 16G RAM so we’d be fine to bump it. I’ll do that too.

About the integration heap size, that’s currently 1G, not 192M, as far as I can see from the file?

schmie:
Ah, OK... I reckon the AWS machines provided more RAM to begin with at all times... The 192M popped up in SK's card - that will have been Carrenza staging, then...

issyl0:
@schmie I bumped it up to 2G in AWS.

PR #7558, opened 2018-05-17
Add organisations-publisher to dev machine config

sihugh:
Initially we just want to depend on publishing-api.

Instructions taken from https://docs.publishing.service.gov.uk/manual/setting-up-new-rails-app.html#defining-the-app-in-the-development-repository


Davidslv:
🚢 🍏

PR #7574, opened 2018-05-22
Monitor the data warehouse ETL job

MatMoore:
This adds a passive icinga check to the jenkins job.
If it runs, it will notify icinga of success/fail. If it doesn't
run for 29 hours, the alert will hit its freshness threshold and
fail.

The job failing is a critical alert in icinga, but it shouldn't trigger pagerduty, because the data is not being consumed on a daily basis (yet).

Trello: https://trello.com/c/HzWvnSCy/345-2-alert-2nd-line-when-no-data-is-retrieved

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/content_performance_manager.pp
+++ b/modules/govuk_jenkins/manifests/jobs/content_performance_manager.pp
@@ -11,10 +11,23 @@
 #
 class govuk_jenkins::jobs::content_performance_manager (
   $rake_etl_master_process_cron_schedule = undef,
+  $app_domain = hiera('app_domain'),
 ) {
   file { '/etc/jenkins_jobs/jobs/content_performance_manager.yaml':
     ensure  => present,
     content => template('govuk_jenkins/jobs/content_performance_manager.yaml.erb'),
     notify  => Exec['jenkins_jobs_update'],
+  }
+
+  $check_name = 'etl-data-warehouse'
+  $service_description = 'Data warehouse ETL'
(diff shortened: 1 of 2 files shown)


MatMoore:
Since the job doesn't run on any other environment, I'm going to deploy it and test the monitoring by commenting out the actual job. I'll roll back/fail forward if it doesn't work.

thomasleese:
I think this is causing a problem in staging where `@rake_etl_master_process_cron_schedule` isn't `true` meaning that the job never runs, and then the freshness alert is triggered.

modules/govuk_jenkins/manifests/jobs/content_performance_manager.pp
```diff
@@ -11,10 +11,23 @@
...
+  $check_name = 'etl-data-warehouse'
+  $service_description = 'Data warehouse ETL'
+  $job_url = "https://deploy.${app_domain}/job/content_performance_manager_import_etl_master_process/"
+
+  @@icinga::passive_check { "${check_name}_${::hostname}":
+    service_description => $service_description,
+    host_name           => $::fqdn,
+    freshness_threshold => 104400,
+    action_url          => $job_url,
+    notes_url           => monitoring_docs_url(data-warehouse-etl-failed),
```
thomasleese: This docs URL doesn't exist at the moment, is there a plan to create one?

MatMoore:
Whoops, sorry about this.

Is it a good idea to have a manual page for every alert? If so I wonder if we could store them all in this repo and load them into the tech docs at build time (like we do with some app documentation). Then we could review the alerts and the docs about them at the same time.

PR #7577, opened 2018-05-23
Add additional message queue consumer for CPM

chao-xian:
The CPM/Data Warehouse will need to preload its database with a bulk
import of data. We will be adding a rake task to the publishing-api
to send out all events with a corresponding routing key to this
queue.

Note that we will have to go into the RabbitMQ manager and manually delete the binding for the `content_performance_manager` queue to pick up the new routing key pattern we've set for it.

Paired with @MatMoore


chao-xian:
[[Tech debt] Use only publishing api to preload items](https://trello.com/c/m3vzLnR5/372-tech-debt-use-only-publishing-api-to-preload-items)

PR #7578, opened 2018-05-24
Fix queue length stats for Publishing API.

benthorner:
https://trello.com/c/sSFxdCHU/86-fix-publishing-api-worker-speed-and-throughput-to-not-show-sum-of-sidekiq-workers

The queue length reported by each instance is for the entire queue,
which means the use of 'sum' was counting the entire queue length
multiple times. Using avg should give us a more reasonable number!


cbaines:
Good spot Ben, no idea why I thought that was a good idea...

PR #7583, opened 2018-05-25
Add missing procfile worker

chao-xian:
New consumer and RabbitMQ queue for CPM were added, but the
corresponding Procfile worker was missing.


chao-xian:

PR #7585, opened 2018-05-25
Allow rabbit consumer to read both CPM queues

MatMoore:
Content performance manager now has 2 queues set up,
so the user needs to read from both of them.


chao-xian:

PR #7586, opened 2018-05-25
Fix process name for procfile worker

chao-xian:
It should match what's defined in the Procfile in the CPM


chao-xian:

PR #7588, opened 2018-05-25
Remove logspout in favour of AWS implementation

benthorner:
https://trello.com/c/Vc2ZF7R5/94-stop-docker-logspout-being-installed-on-ci-machines

Logspout was being incorrectly installed on some machines and not
installed on others. This is because CI/management machines match the
conditional, while actual integration machines have AWS domains and do
not match. The comment suggests this code is out of date, as logit is
actually available in staging and production.

For now, it seems best to remove this feature until we finish moving to
AWS and can make it work in all environments.


kevindew:
This looks good to me, but I think there are a few specs and files that might need to be cleaned out too. 

I get: 

- modules/govuk_docker/files/logspout/build.sh
- modules/govuk_docker/files/logspout/Dockerfile
- modules/govuk_docker/files/logspout/modules.go
- modules/govuk_docker/files/logspout/README.md
- modules/govuk_docker/spec/classes/govuk_docker__logspout_spec.rb
- modules/govuk_docker/spec/classes/govuk_docker_spec.rb

PR #7609, opened 2018-06-01
Make backend_lb class machines safe to reboot over night

thomasleese:
The traffic should be balanced between both backend boxes so there shouldn't be a problem with rebooting these over night.


boffbowsh:
I'm not convinced this is correct - looking at the nginx logs on backend-lb-2, it does always get traffic. So the NAT rule does actually balance between the two boxes.

thomasleese:
Ah, well that's even better than, I'll remove the previous comment.

PR #7613, opened 2018-06-01
James Alderman has left, so `ensure => absent` his user account

issyl0:
- As per the [new
  process](https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html),
  removing a user is a two-step process so their home directory doesn't
  hang around.


boffbowsh:
Merging this to expedite things

PR #7614, opened 2018-06-01
Switch to 1.1.1.1 as primary DNS server

thomasleese:
It's faster than 8.8.8.8 and we're seeing some problems resolving domains which we think could be related.

I've also changed the Icinga DNS check on www.gov.uk to use 1.1.1.1 instead of 8.8.8.8.

### DIFF
--- a/modules/govuk/templates/asset_manager_extra_nginx_config.conf.erb
+++ b/modules/govuk/templates/asset_manager_extra_nginx_config.conf.erb
@@ -70,9 +70,9 @@ location ~ /cloud-storage-proxy/(.*) {
   proxy_hide_header x-amz-replication-status;
   proxy_hide_header x-amz-meta-md5-hexdigest;
 
-  # Add Google DNS server to avoid "no resolver defined to resolve"
-  # errors when trying to connect to S3
-  resolver 8.8.8.8;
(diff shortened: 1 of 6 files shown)


modules/govuk/templates/asset_manager_extra_nginx_config.conf.erb
```diff
@@ -72,7 +72,7 @@ location ~ /cloud-storage-proxy/(.*) {
 
   # Add Google DNS server to avoid "no resolver defined to resolve"
   # errors when trying to connect to S3
-  resolver 8.8.8.8;
+  resolver 1.1.1.1;
```
boffbowsh: [This directive](http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver) supports multiple resolvers.

modules/govuk/templates/asset_manager_extra_nginx_config.conf.erb
```diff
@@ -72,7 +72,7 @@ location ~ /cloud-storage-proxy/(.*) {
 
   # Add Google DNS server to avoid "no resolver defined to resolve"
```
boffbowsh: This comment will need updating

thomasleese:
Thanks @boffbowsh, have amended and pushed!

PR #7619, opened 2018-06-04
Run the autopostgresql backup earlier

thomasleese:
The backup takes longer to run now and usually isn't ready by 9am when the backups get collected together.


boffbowsh:
Might want to add something to the docs for this alert to say why it might be failing

PR #7623, opened 2018-06-05
Give content-tagger an email-alert-api token

barrucadu:
I think I'll need someone else to actually create the token and add it to the secrets(?) repo.

---

[Trello card](https://trello.com/c/3fxFsB93/201-add-subscriber-list-count-stats-to-taxons-in-content-tagger)


cbaines:
> I think I'll need someone else to actually create the token and add it to the secrets(?) repo.

Yep, let me know if you're stuck with this.

PR #7626, opened 2018-06-06
Reduce robots.txt crawl-delay to 0.1s

barrucadu:
#7620 went out yesterday morning and nothing is on fire, so let's reduce it again.

---

[Trello card](https://trello.com/c/CdL0aCEb/197-reduce-robotstxt-crawl-delay-and-evaluate-the-impact)


barrucadu:
Going to ask Tim Blair how long we want to monitor.

thomasleese:
We've left it at 0.3 seconds over the weekend which didn't seem to cause any problems, so this is now fine to merge.

PR #7627, opened 2018-06-06
Fix Pinfile for CPM bulk import

chao-xian:
The bulk import api consumer was not being started on the Dev VM, because we were calling `publishing-api-read` as a dependency for the CPM instead of `publishing-api`. The former does not have `publishing-api-worker` as a dependency itself but the latter does.


chao-xian:
[(3) [Tech debt] Use only publishing api to preload items (basepath)](https://trello.com/c/m3vzLnR5/372-3-tech-debt-use-only-publishing-api-to-preload-items)

PR #7631, opened 2018-06-06
Change the collectd naming scheme when on AWS

rtrinque:
We need to change the way we name machines on collectd when
on AWS since the machines have different hostnames now.


cbaines:
@rtrinque @stephenharker This change is causing a large number of Unknown alerts in Icinga, as the collectd namespace is now inconsistent with the alerts that are also configured through govuk-puppet (e.g. https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/app/config.pp#L254 ).

It might be best to revert this change, as it's not immediately obvious to me how to rectify this?

rtrinque:
Yep, we just noticed the problem. I'm in the process of fixing it. We'll revert if I can't find a quick enough fix

PR #7635, opened 2018-06-07
Modify fqdn_metrics at the facter module level

rtrinque:
By changing the variable at this level we keep the codebase coherent


cbaines:
Has this been replaced by https://github.com/alphagov/govuk-puppet/pull/7647 now?

rtrinque:
Yes it adds a script to do some cleanup in the whisper data

cbaines:
> Yes it adds a script to do some cleanup in the whisper data

Ok, that sounds good.

PR #7646, opened 2018-06-12
Configure machine in GOV.UK infrastructure for CKAN deployment

boffbowsh:
Adds classes to configure CKAN and its underlying dependencies, including Solr.


modules/govuk/manifests/apps/ckan.pp
```diff
@@ -0,0 +1,84 @@
...
+class govuk::apps::ckan (
+  $enabled = false,
+  $port    = '3108',
+  $db_hostname = undef,
+  $db_username = 'ckan',
+  $db_password = 'foo',
+  $db_name = 'ckan_production',
+  $redis_host = undef,
+  $redis_port = undef,
+  $secret     = undef,
```
andrewgarner: Should these be documented in the parameters docs above? It's not obvious to me what `secret` is for.

modules/govuk/manifests/node/s_postgresql_base.pp
```diff
@@ -5,6 +5,7 @@
 class govuk::node::s_postgresql_base inherits govuk::node::s_base {
   include govuk::apps::content_audit_tool::db
   include govuk::apps::content_tagger::db
+  include govuk::apps::ckan::db
```
andrewgarner: This should be the first item in the list if we're to maintain the alphabetical ordering.

modules/govuk/templates/ckan/ckan.ini.erb
```diff
@@ -0,0 +1,217 @@
...
+#ckan.datastore.write_url = postgresql://ckan_default:pass@localhost/datastore_default
+#ckan.datastore.read_url = postgresql://datastore_default:pass@localhost/datastore_default
+
+# PostgreSQL' full-text search parameters
+ckan.datastore.default_fts_lang = english
+ckan.datastore.default_fts_index_method = gist
+
+## Site Settings
+
+ckan.site_url = https://data.gov.uk/
```
andrewgarner: Should we make this configurable? Our staging equivalent is `https://test.data.gov.uk/`

boffbowsh: So would this be `https://ckan.publishing.service.gov.uk/` or still `https://data.gov.uk/` I wonder? I bet it uses this for internal redirects and things like that.

boffbowsh: But yes will put it in hiera, good shout.

modules/govuk/manifests/apps/ckan.pp
```diff
@@ -0,0 +1,84 @@
...
+) {
+  $ckan_home = '/var/ckan'
+  $ckan_ini  = "${ckan_home}/ckan.ini"
+
+  if $enabled {
+    govuk::app { 'ckan':
+      app_type           => 'bare',
+      command            => "unicornherder --gunicorn-bin ./venv/bin/gunicorn -p /var/run/ckan/unicornherder.pid -- --pythonpath ${ckan_home} wsgi_app:application --bind 127.0.0.1:3108 --workers 8",
```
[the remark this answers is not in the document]
andrewgarner: The `3108` port is hardcoded here but available as a configurable option.

modules/govuk/manifests/apps/ckan.pp
```diff
@@ -0,0 +1,103 @@
...
+      port               => $port,
+      vhost_ssl_only     => true,
+      health_check_path  => '/',
```
andrewgarner: We've created a `/healthcheck` path for this. See https://github.com/alphagov/ckanext-datagovuk/pull/54.

andrewgarner:
There's a lint failure:

```
modules/govuk/manifests/apps/ckan.pp - WARNING: double quoted string containing no variables on line 43
```

PR #7647, opened 2018-06-12
Change fqdn_metrics and install cron job to clean up whisper data

rtrinque:
We change fqdn_metrics when on AWS to be in the format class-ip so that things are better organised in graphite, making the change at the variable definition level prevents problems we had in other places.
We also add a cron job to do a daily clean up of the whisper data, as the ever changing hostnames of the aws machines means we're left with useless leftover data

### DIFF
--- a/modules/collectd/manifests/package.pp
+++ b/modules/collectd/manifests/package.pp
@@ -10,12 +10,24 @@
 # [*yajl_package*]
 #   The libyajl (Yet Another JSON Library) package to install.
 #
+# [*apt_mirror_hostname*]
+#   The hostname of the local aptly mirror.
+#
 class collectd::package (
   $collectd_version = '5.4.0-3ubuntu2',
   $yajl_package = 'libyajl2',
+  $apt_mirror_hostname = 'apt.production.alphagov.co.uk',
+
 ) {
   include govuk_ppa
 
+  apt::source { 'collectd':
+    ensure       => present,
+    location     => "http://${apt_mirror_hostname}/collectd",
+    release      => $::lsbdistcodename,
+    architecture => $::architecture,
+    key          => '3803E444EB0235822AA36A66EC5FE1A937E3ACBB',
+  }
   # collectd contains only configuration files, which we're overriding anyway
   package { 'collectd':
     ensure => purged,
@@ -23,7 +35,7 @@ class collectd::package (
   # collectd-core contains collectd itself, and all the compiled plugins
   package { 'collectd-core':
     ensure  => $collectd_version,
-    require => Package['collectd'],
+    require => [ Apt::Source['collectd'], Package['collectd'] ],
   }
   package { $yajl_package:
     ensure => present,
--- a/modules/govuk/lib/facter/fqdn_metrics.rb
+++ b/modules/govuk/lib/facter/fqdn_metrics.rb
@@ -7,7 +7,7 @@ Facter.add("fqdn_metrics") do
     if Facter.value("aws_migration").nil?
       fqdn_metrics = Facter.value("fqdn_short").dup
     else
-      fqdn_metrics = "#{Facter.value(:aws_hostname)}.#{Facter.value(:aws_stackname)}.#{Facter.value(:aws_environment)}"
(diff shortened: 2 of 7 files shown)


modules/govuk/lib/facter/fqdn_metrics.rb
```diff
@@ -7,7 +7,7 @@
     if Facter.value("aws_migration").nil?
       fqdn_metrics = Facter.value("fqdn_short").dup
     else
-      fqdn_metrics = "#{Facter.value(:aws_hostname)}.#{Facter.value(:aws_stackname)}.#{Facter.value(:aws_environment)}"
+      fqdn_metrics = "#{Facter.value(:aws_migration)}-#{Facter.value('fqdn')}"
```
cbaines: Given the warning at the top of the file about the 63 byte limit, I am unsure if this is going to be ok?

As far as I'm aware, the current collectd namespaces use this pattern, and one long one I pulled out comes to exactly 63 characters by my count (`rummager_elasticsearch-ip-10-1-4-149_eu-west-1_compute_internal`).

modules/govuk/templates/node/s_graphite/remove_old_whisper_data.erb
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+WHISPER_DIR='<%= @graphite_path %>/storage/whisper'
+
+for dir in $(ls $WHISPER_DIR); do
+  if [[ -z "$(find $WHISPER_DIR/$dir -mtime 0 )" && $? -eq 0 ]] ; then
```
cbaines: If I'm reading this bit correctly, I'd prefer a retention period of may a week or two (e.g. `-mtime 14`), so that we still have some opportunity to look at recent metrics for all machines operating recently.

cbaines: Just realised Stephen's made a similar comment :+1:

modules/govuk/lib/facter/fqdn_metrics.rb
rtrinque: ah yeah didn't think of that, on AWS staging we have rummager_elasticsearch-ip-10-2-4-246_eu-west-1_compute_internal which is 64 characters long, and it seems to be collecting normally though

cbaines: Looks from the 5.6 changelog that this is now configurable (look for DATA_MAX_NAME_LEN), and might be longer by default, but we are using 5.4 as far as I can see...

it would be good to confirm that this is no longer a problem if it really isn't, and remove the comment if that's the case.

cbaines:
In the last commit, the message says:

>  Upgrade collectd to version 5.6.0
>
>  This version of collectd removes the 63 chars limit.

I'm not quite sure that's correct, it looks to me like the limit is still there, but now 128 characters by default, this is looking at the current master branch of the repository on GitHub though.
https://github.com/collectd/collectd/blob/master/src/daemon/globals.h#L29-L31

modules/govuk/templates/node/s_graphite/remove_old_whisper_data.erb
cbaines: I see this has changed now.

modules/govuk/lib/facter/fqdn_metrics.rb
cbaines: I see this has changed on Friday, I think it's fine to try using the more recent version of collectd.

modules/collectd/manifests/package.pp
```diff
@@ -10,20 +10,32 @@
...
 #   The libyajl (Yet Another JSON Library) package to install.
 #
+# [*apt_mirror_hostname*]
+#   The hostname of the local aptly mirror.
+#
 class collectd::package (
   $collectd_version = '5.4.0-3ubuntu2',
   $yajl_package = 'libyajl2',
+  $apt_mirror_hostname = 'apt.production.alphagov.co.uk',
+
```
cbaines: This line is probably redundant?

PR #7650, opened 2018-06-13
Remove leaver Ross Jones

Davidslv:
Trello: https://trello.com/c/a05bucQW/107-leaver-ross-jones


cbaines:
There is some updated guidance on this process now https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html

PR #7678, opened 2018-06-14
Correctly strip Authorization header when proxying to s3

barrucadu:
I think I've got it right this time: https://stackoverflow.com/questions/44536548/how-to-remove-client-headers-in-nginx-before-passing-request-to-upstream-server

This reverts #7677

---

[Trello card](https://trello.com/c/1shBFAQX/232-strip-authorization-header-when-forwarding-requests-to-s3)

### DIFF
--- a/modules/govuk/templates/asset_manager_extra_nginx_config.conf.erb
+++ b/modules/govuk/templates/asset_manager_extra_nginx_config.conf.erb
@@ -70,6 +70,12 @@ location ~ /cloud-storage-proxy/(.*) {
   proxy_hide_header x-amz-replication-status;
   proxy_hide_header x-amz-meta-md5-hexdigest;
 
+  # Strip any Authorization header from the client, as this will cause
+  # S3 to return an error.  We do not use proxy_hide_header here, as
+  # that hides response headers sent FROM S3 to the client; whereas we
+  # want to unset a request header sent TO S3 by nginx.
+  proxy_set_header Authorization "";
+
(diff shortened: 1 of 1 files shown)


modules/govuk/templates/asset_manager_extra_nginx_config.conf.erb
```diff
@@ -71,6 +70,10 @@ location ~ /cloud-storage-proxy/(.*) {
   proxy_hide_header x-amz-replication-status;
   proxy_hide_header x-amz-meta-md5-hexdigest;
 
+  # Strip any Authorization header from the client, as this will cause
+  # S3 to return an error.
+  proxy_set_heasder Authorization "";
```
thomasleese: I'm guessing this should actually be `proxy_set_header` :)

barrucadu: Oh man

barrucadu:
To do: deploy this branch to integration tomorrow morning to check.

kevindew:
Any chance you could explain in the commit why we need to set it to empty string rather than use proxy_hide_header? feels like something where someone could accidentally set it back thinking it was a mistake

barrucadu:
`proxy_hide_header` hides response headers *from* the proxy, whereas `proxy_set_header` sets request headers sent *to* the proxy.  Setting a header to the empty string removes it.

barrucadu:
Deployed to integration, it works.

PR #7679, opened 2018-06-14
[DO NOT MERGE] Environment variable for Design System reverse proxy

kevindew:
Trello: https://trello.com/c/j3ZOI1fQ/227-add-route-to-the-design-system-at-design-system

This environment variable is to be used when content store registers a
backend with router. Rather than using the plek default of
design-system.publishing... this will instead use the URI in the ENV var
and reverse proxy to the PAAS.

### DIFF
--- a/modules/govuk/manifests/apps/content_store.pp
+++ b/modules/govuk/manifests/apps/content_store.pp
@@ -22,6 +22,10 @@
 #   Performance platform spotlight url
 #   Default: undef
 #
+# [*design_system_uri*]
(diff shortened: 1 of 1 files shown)


kevindew:
@elliotcm yeah it really smells at env var level :(

Marking as DNM while design system decide whether to proceed with this.

bevanloon:
https://trello.com/c/j3ZOI1fQ/227-add-route-to-the-design-system-at-design-system

PR #7682, opened 2018-06-18
Serve staging whitehall assets from asset-manager

barrucadu:
I'm assuming here that the staging hieradata will override the common hieradata, as both define `router::assets_origin::whitehall_uploaded_assets_routes` - is that right?

---

#7648, but for staging.  Part of #7420.


barrucadu:
I don't know if you'll get a notification that I force-pushed a commit, so: fixed.

PR #7683, opened 2018-06-18
Deploy_Node_Apps - Release Tag [WIP]

suthagarht:
- The Deploy_Node_Apps triggers the Deploy_Apps jenkins job. The
deployment happens from a branch for the related application (for
example deployed-to-{environment}).

- The above mentioned process means that the "signon" web interface will
show a release name like "deployed-to-{environment}". The is a generic
name that can't be matched to a release tag.

- The change attempts to indicate the release version/tag.

https://trello.com/c/7TuoLXtJ/1193-l-replace-deployed-to-x-branch-with-release-tag-in-deploy-apps

Solo: @suthagarht

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
@@ -32,6 +32,13 @@
               export GIT_ORIGIN_PREFIX="git@gitlab.com:govuk"
             fi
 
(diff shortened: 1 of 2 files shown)


kevindew:
@schmie do you think this should be closed?

deanwilson:
This is documented and linked to from the trello card linked in the PR so we're going to close this until it resurfaces as an issue.

PR #7689, opened 2018-06-20
Set max heap size for puppetserver to 8G

schmie:
- We encounter OOM kills of the puppetserver

- There is no direct comparison to the ruby version in staging, but
  between threads that instance seem to use 6G+ RAM

- Changing the default of 2G manually to 8G has increased stability

@schmie

### DIFF
--- a/modules/puppet/manifests/puppetserver/config.pp
+++ b/modules/puppet/manifests/puppetserver/config.pp
@@ -1,6 +1,8 @@
 # FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
 class puppet::puppetserver::config {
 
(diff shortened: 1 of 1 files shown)


modules/puppet/manifests/puppetserver/config.pp
```diff
@@ -1,6 +1,8 @@
 # FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
 class puppet::puppetserver::config {
 
+  $java_args = '-Xms8g -Xmx8g -XX:MaxPermSize=1g -Dhttps.protocols=TLSv1.2'
```
dougneal: The `-Dhttps.protocols=TLSv1.2` was added to the config on AWS Staging temporarily by me earlier and didn't have the desired effect, so this should be taken back out.

PR #7697, opened 2018-06-21
Send asset requests on www/www-origin/draft-origin to assets hosts

barrucadu:
This is to [replace a redirect in whitehall](https://github.com/alphagov/whitehall/blob/master/app/controllers/attachments_controller.rb#L5-L11), as we eventually want to drop that controller entirely.

---

[Trello card](https://trello.com/c/e1I8Cjhf/263-redirect-whitehall-assets-path-to-the-assets-host-in-nginx)

### DIFF
--- a/modules/router/templates/base.conf.erb
+++ b/modules/router/templates/base.conf.erb
@@ -48,5 +48,21 @@ server {
   include             /etc/nginx/ssl.conf;
   <%- end %>
 
+  # Redirect asset requests to the assets host
+  location ~ ^/government/uploads/system/uploads/attachment_data/file/[0-9]+/[^/]*$ {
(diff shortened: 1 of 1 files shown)


barrucadu:
Having slept on this, I don't think this is necessary, given: https://github.com/alphagov/whitehall/blob/820deaaa18178497ba4d9e06d25ccd4ea2522644/config/routes.rb#L429

So we can drop the `attachments` controller, and this other route will catch and handle the redirect.  We're also dropping the `LongLifeRedirect` route, so that only leaves the `csv_preview#show` and `hmrc_assets#show` routes.  If we move those to asset-manager at some point (I'll add them to the board as stretch goals), then we can just drop all routing of asset paths to whitehall.

No need for this redirect.

PR #7701, opened 2018-06-22
Serve draft whitehall assets from whitehall-admin

barrucadu:
Requests to draft-origin for draft whitehall assets get routed to
draft-whitehall-frontend, which serves these files directly from NFS.

This is unlike the behaviour for www-origin, which redirects to
assets-origin.  We want to redirect draft-origin too, so we send the
request on to whitehall-admin.

This should be deployed *after* https://github.com/alphagov/whitehall/pull/4152

---

[Trello card](https://trello.com/c/sMwEMvsv/265-redirect-draft-origin-assets-to-assets-origin)

### DIFF
--- a/modules/govuk/manifests/apps/whitehall.pp
+++ b/modules/govuk/manifests/apps/whitehall.pp
@@ -252,8 +252,15 @@ class govuk::apps::whitehall(
       $whitehall_admin_vhost_ = "whitehall-admin.${app_domain}"
     }
 
+    if $::aws_migration {
+      $whitehall_admin_aliases = ['draft-whitehall-frontend']
+    } else {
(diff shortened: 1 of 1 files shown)


barrucadu:
The `draft-assets` host is being served by asset-manager, draft whitehall assets served through `draft-origin` are not.

barrucadu:
`govuk::app::nginx_vhost` sticks the `.*` after each alias automatically, whereas `nginx::config::vhost::static` doesn't.  Special cases! :(

https://github.com/alphagov/govuk-puppet/blob/2d2677f4f5cba73b2a29bb63ef1fd57bf2d63613/modules/nginx/manifests/config/vhost/proxy.pp#L122-L124

PR #7704, opened 2018-06-22
docker-collectd-plugin

suthagarht:
- We have noticed the following error logged in /var/log/syslog.

  collectd[11566]: Failed to open types database `/usr/share/collectd/docker-collectd-plugin/dockerplugin.db': No such file or directory

- This error appears on instances that have collectd  and
Docker installed. This is visible on instances that belong to docker_management.

- The initial change was introduced by https://github.com/alphagov/govuk-puppet/commit/ae7468e0a0d5caad56363e316eb6bb972b923e7b#diff-8f1f4c9a523ffbaf50a9977a99bfe92d

- pip doesn't install docker-collectd-plugin. Hence, we have taken the
content from https://github.com/lebauce/docker-collectd-plugin.

- This issue was noticed when a new graphite instance was created in the
integration environment and null data was recorded.

Pair: @kevindew @suthagarht


thomasleese:
Instead of copying the `docker-collectd-plugin` into `govuk-puppet` which we now have to maintain and keep up to date, could we not just clone the repo into `/usr/share/collectd/docker-collectd-plugin`? We could pin to a specific commit to make sure it doesn't break underneath us?

suthagarht:
Hi @thomasleese , 

Thank you. That sounds great. I will try this.

Thank You
SK

modules/collectd/manifests/plugin/docker.pp
```diff
@@ -2,7 +2,10 @@
 #
 # Collectd plugin for docker
 #
```
kevindew: You should put some documentation of the parameters here

modules/collectd/manifests/plugin/docker.pp
```diff
@@ -2,7 +2,10 @@
...
-class collectd::plugin::docker {
+class collectd::plugin::docker(
+  $repo = undef,
+  $commit = undef,
```
kevindew: I think you should set these to the values from hieradata and then delete them from hieradata - we shouldn't need it there since this will be very static.

modules/collectd/manifests/plugin/docker.pp
suthagarht: done

modules/collectd/manifests/plugin/docker.pp
suthagarht: The test actually worked with hieradata. I am going to leave it there.

PR #7706, opened 2018-06-25
Force removal of cdn logs archive

kevindew:
This couldn't be removed before as it's a directory and you need to pass the force parameter to delete one.


kevindew:
Thanks @thomasleese !

PR #7707, opened 2018-06-25
Refine jenkins Email Check approach

kevindew:
Trello: https://trello.com/c/xeq75b8T/128-update-email-alert-checks-to-not-call-people-on-http-issues

This changes the Email Check jobs that run in Jenkins so that they are flagged as unstable if they exit with a code of 1 and are fails if they exit with 2 (which is the code used in https://github.com/alphagov/email-alert-monitoring).

This then sets the freshness threshold to return a critical when these have not had a passive check for 90 minutes.

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/email_alert_check.pp
+++ b/modules/govuk_jenkins/manifests/jobs/email_alert_check.pp
@@ -52,18 +52,22 @@ class govuk_jenkins::jobs::email_alert_check (
 
   @@icinga::passive_check {
     "${drug_check_name}_${::hostname}":
-      service_description => $drug_service_description,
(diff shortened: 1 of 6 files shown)


elliotcm:
I can see where the process exits with 2 for a failure, where do you see unstable builds exiting as 1?

kevindew:
@elliotcm Sure, they're here: https://github.com/alphagov/email-alert-monitoring/blob/master/Rakefile#L68-L87

elliotcm:
That's the exit 2s, I'm looking for the exit 1s

kevindew:
Oh sorry Elliot I misread. The exit 1s are if the rake task exits on a Ruby exception.

PR #7713, opened 2018-06-26
Mongodb - Maximum open files

suthagarht:
- We have experienced two recorded occasions where the mongo cluster
failed.

- During the investigation , we found-out that the number of openfiles
exceeded 3000. However, the current limit is set at 2000.

- This change will increase the acceptable number of openfiles to 4096.

https://govuk.zendesk.com/agent/tickets/2782624

Solo: @suthagarht

### DIFF
--- a/modules/govuk/manifests/node/s_api_mongo.pp
+++ b/modules/govuk/manifests/node/s_api_mongo.pp
@@ -10,6 +10,13 @@ class govuk::node::s_api_mongo inherits govuk::node::s_base {
     outgoing => 27017,
   }
 
+  limits::limits { 'api_mongo_nofile':
+    ensure     => present,
+    user       => 'mongodb',
+    limit_type => 'nofile',
+    both       => 4096,
+  }
+
   if ! $::aws_migration {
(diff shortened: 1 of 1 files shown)


suthagarht:
hi @barrucadu , That is a very good question. This particular incident times relate to the time when we do a mongo backup to AWS S3.

In theory, this limit should apply to the designated user. This also means that each process forked by the user will have the limit inherited. 

For example, if **mongodb** started two processes, each of these processes can open 2000 files. In that occasion, we should see 4000 open files. 

We can see a high rate of forking around the time in question.

screen shot 2018-06-26 at 14 41 55

However, this is purely based on informed assumptions on the actual reason why the number of open files went up-to 3000 while the limit is 2000. I do understand that, if we had forks, they them selves would have been able to open 2000 files of their own without triggering a problem. 

This change is tries to test the hypothesis that if we increase the acceptable number then we won't see this issue again. If we do see the issue again and had 3000 open files, then we clearly know that the hypothesis is wrong and will revert the change.

PR #7717, opened 2018-06-26
Use more recent upstream Puppet module for Apparmor

dougneal:
The `apparmor` module was forked from https://github.com/simondeziel/puppet-apparmor to https://github.com/alphagov/puppet-apparmor in 2013. It has received a few fixes since then, including one reported today, regarding ordering constraints missing from the resources. This was manifesting as an error when Puppet looked for Apparmor's upstart config before the package had been installed.

Switch to the newer upstream instead of continuing to maintain a fork.


.librarian/puppet/config
```diff
@@ -3,4 +3,3 @@ LIBRARIAN_PUPPET_DESTRUCTIVE: 'true'
 LIBRARIAN_PUPPET_PATH: vendor/modules
 LIBRARIAN_PUPPET_INSTALL__STRIP_DOT_GIT: '1'
 LIBRARIAN_PUPPET_MODE: local
-LIBRARIAN_PUPPET_USE_V1_API: '1'
```
dougneal: Not sure how this got in there

dougneal:
```
 Puppet::Error:
       Duplicate declaration: Package[apparmor] is already declared in file /var/lib/jenkins/workspace/ppet_apparmor-module-update-F3XC2DFNNW54CQVT2RD6R5F37HCYCE4O3XS453RDZV5RKWAO2ESA/vendor/modules/apparmor/manifests/init.pp:39; cannot redeclare at /var/lib/jenkins/workspace/ppet_apparmor-module-update-F3XC2DFNNW54CQVT2RD6R5F37HCYCE4O3XS453RDZV5RKWAO2ESA/vendor/modules/docker/manifests/repos.pp:9 on node apt-1.management.somethingsomething
```

I don't get why this is failing, as this is what `ensure_packages()` is supposed to prevent. The resources declared are the same so it's not clear why this is failing now but was passing previously.

dougneal:
Setting `package_ensure => present` did the trick (thanks @deanwilson)

PR #7728, opened 2018-06-28
Env vars for Email Alert API AWS integration

kevindew:
Trello: https://trello.com/c/FM3MCuAl/255-investigate-email-alert-api-database-size-and-push-email-archives-to-s3

We're planning to archive data stored in Email Alert API databases in
S3. This sets up the environment variables so it can integrate with it.


kevindew:
@MatMoore Thanks 👍

PR #7731, opened 2018-07-02
Fix the Puppetmaster build and the Vagrant dev environment for Puppet

dougneal:
Fix the Puppetmaster build:
 * Drop the `govuk_puppetdb` module and merge its functionality into the `puppet` module
 * Drop the legacy Puppetmaster (Unicorn) entirely - all new builds to be on Ubuntu Trusty with PuppetServer

In order to test the build, I also needed to get Vagrant working when the `$::aws_migration` top-scope variable is set.

This caused tests to fail, which on investigation turned out to be because tests were running against the legacy `hieradata` directory. Added some logic to point them at `hieradata_aws` except in the case of legacy node classes that are not included in the AWS migration.


issyl0:
Please can you test this on AWS staging?

dougneal:
@issyl0 this is currently underway...

README.md
```diff
@@ -185,51 +185,46 @@ runs the single test at line 7 of `collectd__package_spec.rb` with colour mode e
...
 Vagrant will run the Puppet provisioner against the node when it boots up.
 Nodes should look almost identical to that of our real
-production/staging/preview environments, including network addresses. To
-access a node's services like HTTP/HTTPS you can point your `hosts` file to
-the host-only IP address (eth1).
-
-Physical attributes like `memory` and `num_cores` will be ignored because
-they don't scale appropriately to local VMs, but can still be customised as
-described below.
+production/staging/preview environments.
```
issyl0: "Preview" was renamed to "Integration".

modules/puppet/manifests/init.pp
```diff
@@ -12,18 +12,31 @@
...
   validate_bool($future_parser)
   validate_bool($use_puppetmaster)
 
-  include puppet::cronjob
-  include puppet::repository
-  include puppet::package
-  include puppet::monitoring
+  include '::puppet::cronjob'
+  contain '::puppet::cronjob'
+
```
issyl0: I'm not sure I understand the difference between `include` and `contain` - [this](https://puppet.com/docs/puppet/5.5/lang_containment.html#containing-classes) doesn't help, yet includes both examples. This is a more fundamental question than probably appropriate for this code review, though.

modules/puppet/spec/classes/puppet_spec.rb
```diff
@@ -18,7 +18,7 @@
     }}
 
     it do
-      is_expected.to contain_file('/etc/puppet/puppet.conf').with_content(/environmentpath = \/usr\/share\/puppet\/production\/current\/environments/)
+      is_expected.to contain_file('/etc/puppet/puppet.conf').with_content(/environmentpath = \/usr\/share\/puppet\/rp_env\/current\/environments/)
```
issyl0: This looks like it's a variable - where does it come from? The `puppet.conf.erb` file contains `<%= scope.lookupvar('::environment') %>`.

vagrant.nodes.yaml
```diff
@@ -0,0 +1,37 @@
+---
+defaults:
+  box: ubuntu/trusty64
+  memory: 3072
+
+nodes:
```
issyl0: These are AWS node names (eg `db-admin`), not Carrenza names. Are we going the AWS way with the vagrant setup? 🚀

modules/puppet/manifests/init.pp
dougneal: Containment means that if a resource forms an ordering relationship with the containing class, it also forms the same ordering relationship with the contained class.

From another look at the docs it looks like the `include` might be redundant in this case.. I'll try taking that out.

modules/puppet/spec/classes/puppet_spec.rb
dougneal: `rp_env` is what rspec-puppet sets `$::environment` to. I'll add a comment to clarify that.

vagrant.nodes.yaml
dougneal: That was my thinking, yes - make the Vagrant dev environment representative of the future...

dougneal:
I'm going to close this PR and split it up into branches of a more manageable size, that will allow us to fix the Puppet Master build on AWS before tackling the Carrenza legacy.

PR #7733, opened 2018-07-03
Don't send Postgres backups offsite

boffbowsh:
We've replaced autopostgresql backups with WAL-E, which by design backs up to S3. This job backs up almost 300GB of data to S3 daily, and ends up taking longer than 24 hours, meaning the safety of our MySQL and Mongo backups is reduced.

TODO:
- [x] Finish testing the WAL-E restore in staging


issyl0:
So, WAL-E replaces this job in this case, for PostgreSQL backups?

boffbowsh:
Yep. I think we need a bit more discussion before turning off autopostgresqlbackup completely, but WAL-E definitely* provides our offsite needs at the moment.

PR #7752, opened 2018-07-11
Rewrite the doc for adding a new app

benthorner:

The original doc was quite verbose and missing a few changes. It seems
better to give a few rich example files, together with a list of things
to check, instead of having a section for every individual change to
make. The doc is now somewhat more Rails/PG oriented as a consequence.

### DIFF
--- a/docs/adding-a-new-app.md
+++ b/docs/adding-a-new-app.md
@@ -1,248 +1,178 @@
-# Adding a new app to GOV.UK
+[rds-database-management]: https://github.com/alphagov/govuk-aws/blob/master/doc/guides/rds-database-management.md
+[govuk-secrets]: https://github.com/alphagov/govuk-secrets
 
-This Puppet repository defines all the apps which run on GOV.UK.
+# Adding a new app to GOV.UK (Rails+PostgreSQL)
 
 Create a new file in `modules/govuk/manifests/apps` named `my_app.pp`:
 
 ```
-# Be sure to include documentation of the class at the top of the file, as
-# per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc.
+# == Class: govuk::apps::myapp
 #
-# You may wish to copy an existing app, eg. calculators:
-# https://github.com/alphagov/govuk-puppet/blob/3e926f0d/modules/govuk/manifests/apps/calculators.pp
+# Read more: https://github.com/alphagov/myapp
 #
-class govuk::apps::my_app (
-  $port = 3456,
+# === Parameters
+# [*port*]
+#   What port should the app run on? Find the next free one in development-vm/Procfile
+#
+# [*enabled*]
+#   Whether to install the app. Don't change the default (false) until production-ready
+#
+# [*secret_key_base*]
+#   The key for Rails to use when signing/encrypting sessions (in govuk-secrets)
+#
+# [*sentry_dsn*]
+#   The app-specific URL used by Sentry to report exceptions (in govuk-secrets)
+#
+# [*oauth_id*]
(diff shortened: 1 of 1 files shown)


docs/adding-a-new-app.md
```diff
@@ -1,248 +1,178 @@
...
+    deny_framing      => true,
+    asset_pipeline    => true,
   }
-}
-```
 
-The `health_check_path` must be an endpoint on the application returning a
-HTTP 200 status for an unauthenticated request. Monitoring checks for the application are
-configured using this path and fail if the request isn't successful.
+  govuk::app::envvar {
```
kevindew: If I remember right doesn't this need to have capital letter prefixes to set the defaults

e.g: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/email_alert_api.pp#L155-L159

based on this wonderfully intuitive syntax: https://puppet.com/docs/puppet/5.3/lang_defaults.html

docs/adding-a-new-app.md
```diff
@@ -1,248 +1,178 @@
...
-The `secret_key_base` is used to encrypt user sessions. If it's not defined the application will not start. The value of
-`secret_key_base` should be encrypted in the `govuk-secrets` repo.
-
-Additional arguments can be specified to configure basic auth, monitoring checks and logging.
-These are defined and explained in `modules/govuk/manifests/app.pp`.
-
-## Database setup
-
-Create a file in `modules/govuk/manifests/apps/my_app` named `db.pp`:
+Also add a file in `modules/govuk/manifests/apps/my_app` named `db.pp`:
```
kevindew: I think since we've dropped the section on `Database setup` it's probably clearer if we say "If your application uses a Postgres/MySQL add a file to" (as I don't think we do this for MongoDB)

docs/adding-a-new-app.md
```diff
@@ -1,248 +1,178 @@
...
-```
-
-## Environment variables
+govuk::apps::myapp::db_hostname: "postgresql-primary-1.backend"
+govuk::apps::myapp::db::backend_ip_range: "%{hiera('environment_ip_prefix')}.3.0/24"
 
-If your application needs other environment variables, they will need to be added to
-the application manifest too.
+# hieradata_aws/common.yaml
+govuk::apps::myapp::db_hostname: "postgresql-primary-1.backend"
```
kevindew: Small nitpick but that hostname is carrenza specific if I remember correctly and we have a different one for AWS.

docs/adding-a-new-app.md
```diff
@@ -1,248 +1,178 @@
...
+  * hieradata_aws/common.yaml (grafana::dashboards::deployment_applications)
+  * hieradata/development.yaml (hosts::development::apps)
+  * modules/govuk/manifests/node/s_db_admin.pp (see docs for [RDS](rds-database-management))
+  * modules/govuk/manifests/node/s_postgresql_base.pp
+  * modules/grafana/files/dashboards/processes.json
+  * modules/govuk/manifests/node/s_backend_lb.pp (if a backend app)
+  * modules/govuk/manifests/node/s_frontend_lb.pp (if a frontend app)
+  * development-vm/Procfile
+  * development-vm/Pinfile
+  * development-vm/alphagov_repos
```
kevindew: Holy cow that's a long list 😱

docs/adding-a-new-app.md
benthorner: Ah, I assumed there wasn't a difference :-|

kevindew:
@benthorner yup, so long as it's clear 👍

PR #7754, opened 2018-07-12
Add JWS_AUTH_SECRET to collections-publisher

leenagupte:
Trello: https://trello.com/c/9ZMVHe0u

## Motivation
We want to allow people with an access_limiting token to be able to view content published
by collections-publisher on the draft stack without the need of a signon account.

The first step is to create JWS_AUTH_SECRET environment variable that will be used to generate
the token.


cbaines:
I can see from the linked Trello card that the motivation for this is previewing Step by steps, but it might be useful to have this information in the commit message as well.

PR #7755, opened 2018-07-12
Add content publisher app

benthorner:

This PR adds a new Rails app (content-publisher) with additional configuration for Postgres and Grafana. We've only enabled the app for integration to avoid spurious staging/production alerts.


kevindew:
Ooh could you fix the WIP commit please?

PR #7758, opened 2018-07-12
Remove signon from the backend app list on integration

rtrinque:
The only machines on integration that still use the hosts file
to resolve hostnames are the licensify machines
We want the licensify backend to resolve the signon service
hostname through DNS. So we need to remove it from the hosts file


cbaines:
> The only machines on integration that still use the hosts file
> to resolve hostnames are the licensify machines
> We want the licensify backend to resolve the signon service
> hostname through DNS. So we need to remove it from the hosts file

Ok, although if most of this list is actually redundant, it would be good to remove more things eventually as well.

rtrinque:
Yes I suppose the only hosts that the licensify machines really need are signon and the others licensify hosts

PR #7760, opened 2018-07-13
Try removing cyclic dependency for db_admin

benthorner:

The puppet logs are showing a cyclic dependency involving
content_audit_tool and content_publisher when attempting to modify the
pg_hba config for the latter. It's unclear why there's a dependency
chain for these databases, but since content_publisher is new it's
possible the previous dependencies were satisfied 'in bulk' at some
point, while new dependencies should be at the top of the list.

### DIFF
--- a/modules/govuk/manifests/node/s_db_admin.pp
+++ b/modules/govuk/manifests/node/s_db_admin.pp
@@ -141,7 +141,6 @@ class govuk::node::s_db_admin(
 
   # include all PostgreSQL classes that create databases and users
   class { '::govuk::apps::content_audit_tool::db': } ->
-  class { '::govuk::apps::content_publisher::db': } ->
   class { '::govuk::apps::content_tagger::db': } ->
   class { '::govuk::apps::email_alert_api::db': } ->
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/node/s_db_admin.pp
```diff
@@ -150,7 +149,8 @@
   class { '::govuk::apps::policy_publisher::db': } ->
   class { '::govuk::apps::publishing_api::db': } ->
   class { '::govuk::apps::service_manual_publisher::db': } ->
-  class { '::govuk::apps::support_api::db': }
+  class { '::govuk::apps::support_api::db': } ->
+  class { '::govuk::apps::content_publisher::db': }
```
rhowe-gds: Is the ordering necessary? Can also do

```
class { [
    '::govuk::apps::support_api::db',
    '::govuk::...',
  ]: }
```

benthorner: Yeah, we don't think it is. We thought we'd avoid breaking anything else, for now, though ;-)

PR #7774, opened 2018-07-16
Alert on the Publishing API downstream_high latency

cbaines:
The downstream_high queue in the Publishing API is a Sidekiq queue
used to process new and changed content which is a direct result of a
publisher action, for example, someone pressing a publish button.

It's expected that jobs in this queue are processed very quickly, so
the warning threshold has been set to 20 seconds, and the critical
threshold to 60 seconds.

As this is a new alert, just have it notify people in hours, but I'd
consider also alerting on this out of hours eventually.

### DIFF
--- a/modules/monitoring/manifests/checks/sidekiq.pp
+++ b/modules/monitoring/manifests/checks/sidekiq.pp
@@ -45,6 +45,21 @@ class monitoring::checks::sidekiq (
     args      => '--dropfirst -60',
   }
 
+  @@icinga::check::graphite { 'check_publishing_api_downstream_high_queue_latency':
+    target              => 'transformNull(keepLastValue(maxSeries(stats.gauges.govuk.app.publishing-api.*.workers.queues.downstream_high.latency)), 0)',
+    warning             => 20, # seconds
+    critical            => 60, # seconds
+    # Use data from the last 24 hours, to avoid not getting any
+    # datapoints back.
+    from                => '24hours',
+    # Take an average over the most recent 36 datapoints, which at 5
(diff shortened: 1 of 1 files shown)


modules/monitoring/manifests/checks/sidekiq.pp
```diff
@@ -45,6 +45,21 @@
     args      => '--dropfirst -60',
   }
 
+  @@icinga::check::graphite { 'check_publishing_api_downstream_high_queue_latency':
```
thomasleese: I'll approve this, but I think it would be better as a healthcheck.

Using [`SidekiqQueueLatencyCheck`](https://github.com/alphagov/govuk_app_config/blob/master/docs/healthchecks.md#sidekiqqueuelatencycheck) it's in the app itself rather than in Puppet config, you get more customisation and a clearer message in Icinga which includes the actual latency and the thresholds.

cbaines: Thanks for taking a look Tom. I guess using the healthcheck wouldn't immediately get to the point where Pagerduty can be used for notifications (unless I'm missing something), but maybe that's something that can be sorted as well...

I'll take a look at adding the healthcheck, and close this if it works out...

thomasleese: Hmm true, you would have to assign an alert to the critical health check which might call on other things.

cbaines: The healthcheck is deployed now, so I'll have a look at notifications in the context of healthchecks... at some point.

PR #7778, opened 2018-07-17
Add a rake task to count the number of superfluous taggings.

koetsier:
This runs every day at 3 AM and counts all content that is tagged
to a taxon and its ancestor.

Trello: https://trello.com/c/VOLclckO/18-start-tracking-the-number-of-superfluous-tags-to-the-topic-taxonomy


1pretz1:
LGTM 👍

PR #7781, opened 2018-07-17
Vagrant for testing Puppet in AWS

dougneal:
Decouple this repo from the legacy `govuk-provisioning` repo (which contains the vCloud server definitions) and make the Vagrant node manifest more representative of what we're moving towards in AWS.

Instead of scraping an external source for a list of servers, simply list the classes of node in `vagrant.nodes.yaml` and assume we only ever want to run one of each locally.

Move some files around into `tools/vagrant/`

Make the mock EC2 metadata service persistent across Vagrant VM reboots.


kevindew:
@dougneal this seems to have broken govuk_puppet in dev vm. 

We're getting errors of:

```
    default: /var/govuk/govuk-puppet/tools/puppet-apply-dev: 82: exec:
    default: /var/govuk/govuk-puppet/tools/puppet-apply: not found
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

Which seems to be because govuk_puppet uses the contents of: https://github.com/alphagov/govuk-puppet/blob/master/modules/puppet/templates/govuk_puppet_development which runs https://github.com/alphagov/govuk-puppet/blob/master/tools/puppet-apply-dev#L82 which is pointing to the file that no longer exists.

PR #7795, opened 2018-07-20
Handle draft-whitehall-frontend through the whitehall-frontend

cbaines:
Rather than the whitehall-admin load balancer and service. Currently,
some of the resources on the draft-whitehall-frontend rendered pages
don't load, because they are loaded from the Whitehall Admin, and the
browsers Same Origin Policy doesn't allow loading these.

Serving the requests from the whitehall-frontend uses different
configuration for the assets, which use relative URLs which work.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1360,7 +1360,6 @@ hosts::production::backend::app_hostnames:
   - 'content-publisher'
   - 'content-tagger'
   - 'docs'
-  - 'draft-whitehall-frontend'
   - 'email-alert-api'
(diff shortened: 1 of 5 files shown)


cbaines:
Seems to work in Staging :tada: , I've noticed that there is some stuff to update in `govuk-aws-data` as well, so I'll do that soon.

cbaines:
> Seems to work in Staging tada , I've noticed that there is some stuff to update in govuk-aws-data as well, so I'll do that soon.

https://github.com/alphagov/govuk-aws-data/pull/182

PR #7797, opened 2018-07-20
Configure memcached to listen publicly

cbaines:
This will enable using memcached in a distributed way. This makes more
memory available, but the main advantage of this is that it allows
explicit cache invalidation to be done.

At the moment, you could remove entries from the cache when they
become invalid, but this would only take effect on the current
machine. When using memcached in a distributed way, the cache
invalidation would affect all machines.


cbaines:
> This looks fine to me, but are there any security implications?

I believe because of the network isolation in both Carrenza and AWS, this should be secure. It's likely that this needs tweaking even to make communication between the Publishing API machines possible.

PR #7799, opened 2018-07-20
WIP: Share memcached between the Publishing API machines

cbaines:
This makes it possible to do distributed cache invalidation, as well
as increasing the available space in the cache.

The AWS configuration is a bit tricky, as I'm not even sure there is a
way to address the machines...

It also depends on https://github.com/alphagov/govuk-puppet/pull/7797 and possibly some more changes related to the actual networking...


theseanything:
@cbaines is this still WIP?

cbaines:
> @cbaines is this still WIP?

Well, the "In Progress" bit is definitely inaccurate :) I'll close it for now, I was a bit too hopeful when I created it.

PR #7801, opened 2018-07-20
Add mongo sync scripts

schmie:
Testing/Linting puppet code


issyl0:
How WIP is this? It looks to be green in CI now?

Could we possibly merge this so we're not branching off branches to add new sync scripts to `modules/govuk_env_sync`?

schmie:
Apologies for the delay. Yes, this can be reviewed/closed.

PR #7802, opened 2018-07-23
Better support testing changes while locked

cbaines:
Currently, you can't use govuk_puppet to run puppet if a lock is in
place. This makes it difficult to test what running govuk_puppet would
do.

This change makes it possible to run commands like the following to
test a what actually running govuk_puppet would do.:

  govuk_puppet --noop --test --agent_disabled_lockfile=/tmp/puppet.noop

### DIFF
--- a/modules/puppet/templates/govuk_puppet
+++ b/modules/puppet/templates/govuk_puppet
@@ -191,10 +191,12 @@ if [[ " ${puppet_args} " =~ "--enable " ]]; then
   remove_lock
 fi
 
-locked=$(has_locks)
-if [ "$locked" == 1 ];then
-  list_locks
-  exit 1
+if [[ ! " ${puppet_args} " =~ "--noop " ]]; then
+  locked=$(has_locks)
+  if [ "$locked" == 1 ];then
+      list_locks
+      exit 1
+  fi
 fi
 
 set +e


modules/puppet/templates/govuk_puppet
```diff
@@ -191,10 +191,12 @@ if [[ " ${puppet_args} " =~ "--enable " ]]; then
   remove_lock
 fi
 
-locked=$(has_locks)
-if [ "$locked" == 1 ];then
-  list_locks
-  exit 1
+if [[ ! " ${puppet_args} " =~ "--noop " ]]; then
```
thomasleese: Why are there spaces around the argument and the `puppet_args` variable?

cbaines: I just copied the format of the `--enable` and others, (for example: `if [[ " ${puppet_args} " =~ "--enable " ]]; then`).

`=~` is a regular expression based match (I think), so I can see why you want a space at the end, as there is a space at the end of `--noop ` to guard against matching things like `--noop-foo` (made up example). I don't know why there is a space at the start, I assume it would work without it.

thomasleese: That makes sense, thanks!

PR #7807, opened 2018-07-24
Update transition_logs.pp to handle suspicious comments

akondasif:
Greetings,

I am a security researcher, who is looking for security smells in Puppet scripts. I found instances 
of *suspicious comments* i.e. comments with content that are indicative of bad practices. If there are keywords such as TODO, HACK, FIXME and bug/issue information is exposed through comments, then such comments should be avoided. According to the Common Weakness Enumeration (CWE) organization `Many suspicious comments, such as BUG, HACK, FIXME, LATER, LATER2, TODO, in the code indicate missing security functionality and checking. Others indicate code problems that programmers should fix, such as hard-coded variables, error handling, not using stored procedures, and performance issues.` Reff: https://cwe.mitre.org/data/definitions/546.html 

I followed the recommendations from CWE and removed such comments. Any feedback is appreciated.

### DIFF
--- a/modules/govuk_cdnlogs/manifests/transition_logs.pp
+++ b/modules/govuk_cdnlogs/manifests/transition_logs.pp
@@ -99,7 +99,6 @@ class govuk_cdnlogs::transition_logs (
     group  => $user,
   }
 
-  # @TODO remove these once this has been run in production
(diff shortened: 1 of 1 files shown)


akondasif:
Thanks so much for your consideration @issyl0

PR #7815, opened 2018-07-25
Run Smokey loop with its own user

rubenarakelyan:
This commit changes the Smokey loop to run as a new `smokey` user rather than `root`. This is because:

* Running as `root` is not a good idea
* Google Chrome does not like running as `root`

Trello: https://trello.com/c/80OiCEtH/308-more-reliable-smokey-tests-on-aws-environments

### DIFF
--- a/modules/monitoring/manifests/checks/smokey.pp
+++ b/modules/monitoring/manifests/checks/smokey.pp
@@ -21,7 +21,10 @@ class monitoring::checks::smokey (
 
   include govuk::apps::smokey
 
-  # TODO: Should this really run as root?
+  govuk_user { 'smokey':
+    fullname    => 'Smokey',
+  }
+
   file { $service_file:
     ensure  => present,
     content => template('monitoring/smokey-loop.conf'),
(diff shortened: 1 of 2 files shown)


modules/monitoring/manifests/checks/smokey.pp
```diff
@@ -21,7 +21,10 @@
 
   include govuk::apps::smokey
 
-  # TODO: Should this really run as root?
+  govuk_user { 'smokey':
+    fullname    => 'Smokey',
+  }
```
cbaines: I think this class is more intended for a person, not a service (in the admin, deploy and adm groups by default). Something like this would be more appropriate (and you might be able to omit the shell):
```
  user { 'smokey':
    ensure => present,
    shell  => '/bin/bash',
  }
```

cbaines: When I did this in staging, I also made it a `system` user, as that seemed sortof sensible.

rubenarakelyan: https://github.com/alphagov/govuk-puppet/pull/7817

PR #7827, opened 2018-07-25
Monitoring AWS RDS & Elastic Cache

suthagarht:
- In the initial version, we have referenced the AWS Region as a
variable. However, this variable was called from inside the resource.

- This meant that declaring the variable as ```
monitoring::checks::rds_config::region: ``` will not be useful.

- To correct this, we have declared the variable inside the actual
class (For example: monitoring::checks::rds::region). Then we pass the
variable by calling the resource within the class.

- We have introduced this change and corrected comments within files.

https://trello.com/c/C7SY8edZ/1123-monitor-rds-and-elasticache-in-aws-and-alert-icinga

Solo: @suthagarht

### DIFF
--- a/modules/monitoring/manifests/checks/cache.pp
+++ b/modules/monitoring/manifests/checks/cache.pp
@@ -4,8 +4,14 @@
 #
 # === Parameters
 #
+# [*enabled*]
+#   This variable enables to define whether to perform Icinga Elastic-Cache checks. [This is set to 'true' by default.]
+#
 # [*servers*]
 #   List of RDS instances.
+#
+# [*region*]
+#   Which AWS region should be checked ['us-east-1','eu-west-1']
 #
 class monitoring::checks::cache (
       $enabled = true,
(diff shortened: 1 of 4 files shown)


modules/monitoring/manifests/checks/cache.pp
```diff
@@ -4,9 +4,15 @@
 #
 # === Parameters
 #
+# [*enabled*]
+#   This variable enables to define whether to perform Icinga Elastic-Cache checks.
```
issyl0: It's probably worth specifying here that `true` is the default (according to line 17).

modules/monitoring/manifests/checks/cache.pp
```diff
@@ -4,9 +4,15 @@
...
+#
 # [*servers*]
 #   List of RDS instances.
 #
+# [*region*]
+#   Which AWS region should be checked ['us-east-1','eu-west-1']
```
issyl0: Should we set the default `region` to `eu-west-1` (Ireland)?

modules/monitoring/manifests/checks/cache.pp
suthagarht: Done

modules/monitoring/manifests/checks/cache.pp
suthagarht: Hi @issyl0 , We can. However, I don't think that there is some value on adding here. Also, this might be an additional step for someone searching for the first time to understand.

PR #7831, opened 2018-07-26
Change database load balancer IP range to 10.1.0.0/16

barrucadu:
Load balancers are distributed between availability zones, which places them on different subnets.  `10.1.5.0/24` is just one such subnet, but there are others.

This is fine as we use security groups for actual security.

---


barrucadu:
Tested this on integration now, seems good.

PR #7834, opened 2018-07-26
Upgrade Jenkins plugins

rubenarakelyan:
Trello: https://trello.com/c/Z9OlZ9Y2/341-upgrade-all-of-the-jenkins-05-day


barrucadu:
These all look pretty small version bumps.  Is our jenkins fairly up to date, are jenkins plugins just really stable, or is this not upgrading them to the latest versions?

rubenarakelyan:
Yeah, it's a precursor to upgrading Jenkins itself and we did this not too long ago so I wouldn't expect a lot of large version bumps.

PR #7836, opened 2018-07-26
Add port and prepared statements options to database url env var

barrucadu:
When switching an app to use pgbouncer, we will have to change the port and disable prepared statements.

---

### DIFF
--- a/modules/govuk/manifests/app/envvar/database_url.pp
+++ b/modules/govuk/manifests/app/envvar/database_url.pp
@@ -21,18 +21,39 @@
 # [*database*]
 #   The database name.
 #
+# [*port*]
+#   The database port.  If not given, no port is included in the URL.
+#
+# [*allow_prepared_statements*]
+#    Whether to allow prepared statements.  If not given, no
+#    prepared_statements parameter is included in the URL.
+#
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/app/envvar/database_url.pp
```diff
@@ -21,18 +21,39 @@
...
+#    Whether to allow prepared statements.  If not given, no
+#    prepared_statements parameter is included in the URL.
+#
 define govuk::app::envvar::database_url (
   $type,
   $username,
   $password,
   $host,
   $database,
+  $port = undef,
```
boffbowsh: I'd default this to 5432, and always include it. Cleaner code further down.

barrucadu: The problem is that `database_url` is used for non-postgres things too.  For example, mysql: https://github.com/alphagov/govuk-puppet/blob/6e29bbd109db5ff3165f5826c97672c9629c7105/modules/govuk/manifests/apps/whitehall.pp#L239-L244

boffbowsh: Ah, of course, sorry

PR #7840, opened 2018-07-26
Make content-audit-tool use pgbouncer in AWS

barrucadu:


barrucadu:
And I can see in the pgbouncer logs that content-audit-tool is connecting to it.

PR #7842, opened 2018-07-26
Bump module puppetlabs-stdlib to 4.25.1

dougneal:
This is a dependency for a lot of other modules, so should unblock a few other upgrades.

Removed the version pin.


issyl0:
Are we planning on “a lot of other upgrades” soon? Or is this a precursor to The Future when we get around to it?

dougneal:
The second one. Opportunistic house keeping!

PR #7852, opened 2018-07-27
Add content-data-admin for integration

chao-xian:
New frontend app for the Content Data Admin, which will be the interface for the Content Data API.

Enabled for Integration only.

These are following the steps from
https://github.com/alphagov/govuk-puppet/blob/master/docs/adding-a-new-app.md

Related PR:

* https://github.com/alphagov/govuk-secrets/pull/400


chao-xian:
[(8) Provision and deploy a new barebones Admin app](https://trello.com/c/kOuisVYY/547-5-provision-and-deploy-a-new-barebones-admin-app)

modules/govuk/manifests/node/s_backend_lb.pp
```diff
@@ -53,6 +53,7 @@
     'collections-publisher',
     'contacts-admin',
     'content-audit-tool',
+    'content-data-admin',
```
KyleMacPherson: I thought this was a frontend app (but I definitely could be wrong!). If so I think this should be in `s_frontend_lb.pp`.

hieradata/vagrant_credentials.yaml
```diff
@@ -36,6 +36,7 @@ govuk::apps::contacts::db::mysql_contacts_frontend: 'DfM7oUW8Hsn9g23WZW3Hvv4wmr6
 govuk::apps::contacts::db::mysql_contacts_admin: 'DO7910Lz5ohJukmnvC6GrN3e4jtUP82l'
 govuk::apps::content_audit_tool::db::password: 'MDgTmQKvEAax4wgefQAZRNDTajkdtEVf'
 govuk::apps::content_audit_tool::db_password: "%{hiera('govuk::apps::content_audit_tool::db::password')}"
+govuk::apps::myapp::db_password: 'FrkwKJGjLAjbznLtPhFjVy7YEonLKJTr'
```
KyleMacPherson: Should not be `myapp`

KyleMacPherson:
[(8) Provision and deploy a new barebones Admin app](https://trello.com/c/kOuisVYY/547-8-provision-and-deploy-a-new-barebones-admin-app)

modules/govuk/manifests/apps/content_data_admin.pp
```diff
@@ -0,0 +1,103 @@
...
+
+  $ensure = $enabled ? {
+    true  => 'present',
+    false => 'absent',
+  }
+
+  # see modules/govuk/manifests/app.pp for more options
+  if $enabled {
+    govuk::app { $app_name:
+      ensure            => $ensure,
```
erino: As `$ensure` is only set to `present` if `$enabled` is `true` then do we need the `if $enabled {` block on line 62?

chao-xian: Doesn't look like it according to this: https://github.com/alphagov/govuk-puppet/pull/7762/files Cc @KyleMacPherson @peteglondon

peteglondon: This was the initial advise from 2nd line early on Wednesday. We've since gained a bit more understanding about puppet so you are correct, this is not required.

We'll get rid of it now

PR #7863, opened 2018-07-30
Add support for elasticsearch in the sync/backup script

rtrinque:


modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -53,19 +53,44 @@ function set_filename {
...
 function is_writable_mongo {
   mongo --quiet --eval "print(db.isMaster()[\"ismaster\"]);" "localhost/$database"
 }
 
+function is_writable_elasticsearch {
+# We need a bit of black magic here unfortunately
+# We're not looking for a writable node, but rather a unique one to avoid restoring to every nodes
+# We arbitrary pick node one, but we cannot get this info from the cluster itself as it assigns random id to nodes
+# So we use the metadata from the AWS instance to get this
+  NODE_ID=$(curl -s http://169.254.169.254/latest/meta-data/iam/security-credentials | awk -F "-" '{print $NF}')
```
dougneal: I was curious as to what this was doing so tried it out... the URL needs a trailing slash :)

rtrinque: ah crap, thanks!

modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -57,6 +57,21 @@ function is_writable_mongo {
   mongo --quiet --eval "print(db.isMaster()[\"ismaster\"]);" "localhost/$database"
 }
 
+function is_writable_elasticsearch {
```
suthagarht: Hi @rtrinque ,

Could you please tell us where ** is_writable_elasticsearch** is called/used.

modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -84,6 +99,16 @@ function restore_mongo {
     "${tempdir}/${database}"
 }
 
+function dump_elasticsearch {
+  es_dump_restore dump http://localhost:9200/ ${database} ${tempdir}/${filename}
```
suthagarht: Hi @rtrinque , 

Sorry, could you please tell us where is **es_dump_restore** defined.

modules/govuk_env_sync/files/govuk_env_sync.sh
suthagarht: Also, this function is using the hostname from the AWS meta data to decide whether the node is the master. This might cause a problem if node 1 is not the master.

modules/govuk_env_sync/files/govuk_env_sync.sh
issyl0: `es_dump_restore` is [here](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_elasticsearch/manifests/dump.pp#L11-L15) in Puppet, it's a package.

modules/govuk_env_sync/files/govuk_env_sync.sh
issyl0: This is called [here](https://github.com/alphagov/govuk-puppet/pull/7863/files#diff-4bba70a7d7214e66f3873d258dfa41bcR169).

rtrinque: There is no master/slave, we arbitrary choose one node (number 1) just to avoid restoring on several nodes at the same time

PR #7866, opened 2018-07-30
Install pgbouncer to Carrenza postgresql-primary machines

barrucadu:

### DIFF
--- a/modules/govuk/manifests/node/s_postgresql_primary.pp
+++ b/modules/govuk/manifests/node/s_postgresql_primary.pp
@@ -21,6 +21,8 @@ class govuk::node::s_postgresql_primary (
     slave_password => $standby_password,
   }
 
+  include govuk_pgbouncer
+
   limits::limits { 'postgres_nproc':
     ensure     => present,
     user       => 'postgres',


modules/govuk/manifests/node/s_postgresql_primary.pp
```diff
@@ -21,6 +21,8 @@
     slave_password => $standby_password,
   }
 
+  include govuk::pgbouncer
```
boffbowsh: Think this needs to be `::govuk::pgbouncer`

barrucadu: `s_postgresql_base` doesn't have the leading `::`: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/node/s_postgresql_base.pp

barrucadu:
Are there any load balancers that connections will go through, or is pgbouncer allowing the `backend_ip_range` (`"%{hiera('environment_ip_prefix')}.3.0/24"`) enough?

boffbowsh:
That should be enough.

barrucadu:
How about firewall rules to open ports?

boffbowsh:
I *think* we only have firewalls between networks, and this is all within the same network for now.

PR #7890, opened 2018-07-31
Use the same hieradata value everywhere.

deanwilson:
This will make migrating the service easier as we only have to change
one value in hiera.

### DIFF
--- a/modules/collectd/manifests/package.pp
+++ b/modules/collectd/manifests/package.pp
@@ -16,8 +16,7 @@
 class collectd::package (
   $collectd_version = '5.4.0-3ubuntu2',
   $yajl_package = 'libyajl2',
-  $apt_mirror_hostname = 'apt.production.alphagov.co.uk',
-
+  $apt_mirror_hostname = undef,
 ) {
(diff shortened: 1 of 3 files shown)


modules/collectd/manifests/package.pp
```diff
@@ -16,8 +16,7 @@
 class collectd::package (
   $collectd_version = '5.4.0-3ubuntu2',
   $yajl_package = 'libyajl2',
-  $apt_mirror_hostname = 'apt.production.alphagov.co.uk',
-
+  $apt_mirror_hostname = undef,
```
dougneal: Isn't this invalid as a default value?

deanwilson: It's valid and consistent with the existing code but it's not ideal as a missing hieradata entry will result in broken URLs rather than an aborted run. We'll get this in to make all the apt_mirror lines the same and make the default better across the code base in a further PR.

deanwilson:
There is a whole doc about this as part of the migration moves. There are many moving parts.

PR #7892, opened 2018-07-31
Add Grafana datasources automatically

schmie:
- Grafana datasources have been added manually in the past.

- If AWS restarts an instance this introduces potential monitoring
  outages

- The web API of Grafana can be used to automatically add datasources

- This relies on https://github.com/alphagov/govuk-secrets/pull/395 to work

solo @schmie

### DIFF
--- a/modules/grafana/manifests/datasources.pp
+++ b/modules/grafana/manifests/datasources.pp
@@ -1,0 +1,74 @@
+# == Class: grafana::datasources
+#
+# Adds datasources to Grafana via its web API
+#
+# [*webapi_user*]
+#   The user used to authenticate to the web API
+#
+# [*webapi_password*]
+#   The password used to authenticate to the web API
+#
+# [*webapi_user*]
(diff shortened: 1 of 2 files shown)


modules/grafana/manifests/datasources.pp
```diff
@@ -0,0 +1,79 @@
...
+#
+class grafana::datasources(
+  $webapi_user = undef,
+  $webapi_password = undef,
+  $elasticsearch_endpoint = undef,
+  $elasticsearch_user = undef,
+  $elasticsearch_password = undef,
+) {
+
+  $graphitesourcejson = '{\
```
issyl0: Is there a nicer way than a load of escaped new lines?

schmie:
Yes, I had hope too wrap this in a puts, but Puppet lacks the serialisation facilities out of the box (Puppet 3.8 that is...)

deanwilson:
You could put the JSON in a template and load it in that way.

PR #7895, opened 2018-07-31
Shellcheck scripts that have changed in pull requests

dougneal:
Rather than going back through all shell scripts in the repo and making them Shellcheck-compliant, just fail CI if any script that has been changed in the PR has not passed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,6 +18,10 @@ node {
       govuk.bundleApp()
     }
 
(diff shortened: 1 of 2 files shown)


dougneal:
That was the intention! Investigating...

dougneal:
test this please

dougneal:
I've made a few changes and things are now working as expected.

PR #7898, opened 2018-08-01
Add rudimentary file transfer capability

schmie:
- We may want to include a number of ordinary files in the data sync.

- The dbms type "files" allows to compress/push/pull/decompress a
  directory provided as "database" argument.

solo @schmie

### DIFF
--- a/modules/govuk_env_sync/files/govuk_env_sync.sh
+++ b/modules/govuk_env_sync/files/govuk_env_sync.sh
@@ -101,13 +101,13 @@ function restore_mongo {
 }
 
 function dump_files {
-  tar --create --gzip --file "${tempdir}/$filename" "${database}"
+  tar --create --gzip --force-local --file "${tempdir}/$filename" "${database}"
 }
 
 function restore_files {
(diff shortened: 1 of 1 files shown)


modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -92,13 +93,23 @@ function dump_mongo {
 
 function restore_mongo {
   cd "${tempdir}"
-  tar --extract --gzip --file "${filename}"
+  tar --extract --gzip --force-local --file "${filename}"
```
issyl0: The [reason behind `--force-local`](https://steve-parker.org/linux/tar-colon-filename/) sounds like _fun_.

issyl0: I'd prefer if the `--force-local` argument addition in each of these cases was split out into a separate commit, but that might just be me being picky.

schmie: it is done :)

PR #7916, opened 2018-08-07
Jenkins Job - Mongo data-sync [WIP]

suthagarht:
- We have a requirement to create a jenkins job that could used to
trigger the `data-sync`.

- This change addresses the following.

a) `govuk-backup` is the group that has the ownership of the backup
location. Hence, we are adding the `deploy` user to this group to
inherit the permissions.

b) The backup dumps were stored with complete permissions to the owner.
Here we are issuing write permissions to the group.

Solo: @suthagarht


suthagarht:
There has been some changes.  Hence, not required.

PR #7927, opened 2018-08-08
Jenkins Job - Mongo data-sync [WIP]

suthagarht:
- We have a requirement to create a jenkins job that could used to
trigger the `data-sync`.

- This change addresses the following.

a) `govuk-backup` is the group that has the ownership of the backup
location. Hence, we are adding the `deploy` user to this group to
inherit the permissions.

b) The backup dumps were stored with complete permissions to the owner.
Here we are issuing write permissions to the group.

Solo: @suthagarht


modules/govuk_env_sync/manifests/task.pp
```diff
@@ -58,6 +58,20 @@
     content => template('govuk_env_sync/govuk_env_sync_job.conf.erb'),
   }
 
+  file { $temppath:
```
issyl0: The indentation is off in these two `file` blocks.

modules/govuk_jenkins/templates/jobs/mongo_data_sync.yaml.erb
```diff
@@ -0,0 +1,30 @@
...
+    builders:
+      - shell: |
+          #!/usr/bin/env bash
+          set -x
+
+          # Create an array of mongo cluster members.
+          members=$(govuk_node_list -c mongo)
+
+          # Iterate over the array.
+          # If the member is a master, then iitiate data-sync.
```
issyl0: Typo: "iniiate" => "initiate".

modules/govuk_jenkins/templates/jobs/mongo_data_sync.yaml.erb
```diff
@@ -0,0 +1,30 @@
...
+          for member in ${members[@]};do
```
issyl0: There's a missing space between `;` and `do`.

modules/govuk_jenkins/templates/jobs/mongo_data_sync.yaml.erb
```diff
@@ -0,0 +1,30 @@
...
+            if $(ssh deploy@${member} "mongo --quiet --host "$member" --eval 'rs.isMaster().ismaster'"); then
```
issyl0: Can we make this exit once it runs the command on the master? At the moment, if the master is first, the sync happens there and then exits, it loops to check if the other `members` are masters too, returning `false`, then exits successfully. Although a tiny extra couple of steps, this feels like a waste as there should only ever be one master.

See [the job output](https://deploy.blue.staging.govuk.digital/job/mongo_data_sync/1/console) for an example.

modules/govuk_env_sync/manifests/task.pp
suthagarht: good spot done.

modules/govuk_jenkins/templates/jobs/mongo_data_sync.yaml.erb
suthagarht: SK! :)

modules/govuk_jenkins/templates/jobs/mongo_data_sync.yaml.erb
suthagarht: corrected

modules/govuk_env_sync/manifests/task.pp
schmie: I do not think the lock file should be "ensured" - it is rather created by the setlock command to indicate that no auto-update/reboot should happen during the operation

suthagarht: removed ensure

kevindew:
@suthagarht Hey, we've hit a problem with this on the dev vm - the govuk-puppet group doesn't exist by default. I think we need something similar to: https://github.com/alphagov/govuk-puppet/blob/2778edd41113b2dd1a0d24eaab4fc367e27e98b0/modules/govuk/manifests/deploy/setup.pp#L42 and https://github.com/alphagov/govuk-puppet/blob/2778edd41113b2dd1a0d24eaab4fc367e27e98b0/modules/govuk/manifests/deploy/setup.pp#L45

PR #7932, opened 2018-08-09
Configure & test env_sync mysql push

szd55gds:
Configure & test env_sync mysql push

- I needed to configure heiradata and other areas in order to test the 'push' side of env_sync (mysql).
- Source instance is: whitehall-mysql-backup-1.backend.staging.publishing.service.gov.uk.yaml
- DB: whitehall_production
- I had to edit a file in 'backup' module in order to retain correct ownership of /var/lib/automysqlbackup/s3backup because it is reset to root:root by daily cron automysqlbackup.

### DIFF
--- a/hieradata_aws/class/staging/db_admin.yaml
+++ b/hieradata_aws/class/staging/db_admin.yaml
@@ -9,7 +9,6 @@ govuk_env_sync::tasks:
     temppath: "/tmp/content_publisher_production"
     url: "govuk-staging-database-backups"
     path: "postgresql-backend"
-govuk_env_sync::tasks:
   "pull_contacts_production_daily":
     hour: "1"
     minute: "12"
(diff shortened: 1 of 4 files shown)


issyl0:
Thanks for this, @szd55gds! ⭐️

Could you squash 1b6f886, c5d8323, 5a5bce3 and 084e52a together into 8a2b543 please? They look to all have the same end result - creating a (YAML) file for mysql env-sync, but with some trial and error to get it setup correctly. Of course, trial and error is perfectly fine, but for conciseness...

Does 39b3cb5 work as the temp path and supersede 4dfea1d?

hieradata_aws/class/staging/db_admin.yaml
```diff
@@ -9,7 +9,6 @@ govuk_env_sync::tasks:
     temppath: "/tmp/content_publisher_production"
     url: "govuk-staging-database-backups"
     path: "postgresql-backend"
-govuk_env_sync::tasks:
```
issyl0: I just noticed that this was removed. Why?

issyl0: Ah, it was a duplicate.

PR #7934, opened 2018-08-09
Add Yarn to GOV.UK infrastructure

kevindew:
This is a semi-popular alternative to NPM that Rails has a preference
for. Installing this so that it can be used to install node packages as
part of Rails asset pipeline.

The reason for installing this in addition to NPM already existing on
our boxes (as part of Node) is because Rails will utilise this package
manager automatically as part of the asset-pipeline process. Rails also
has a preference for Yarn over NPM which means that all Node package
related tasks become more difficult when using NPM.


kevindew:
@benthorner I've updated the commit documentation

benthorner:
Good stuff 👍 .

PR #7938, opened 2018-08-10
Mongo Data-Sync

suthagarht:
- There were some errors noticed during puppet runs.

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Duplicate declaration: File[/etc/unattended-reboot/no-reboot/govuk_env_sync] is already declared in file /usr/share/puppet/production/current/modules/govuk_env_sync/manifests/task.pp:72; cannot redeclare at /usr/share/puppet/production/current/modules/govuk_env_sync/manifests/task.pp:72 on node postgresql-primary-1.backend.staging.publishing.service.gov.uk
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run```

- This change is intended to correct the duplicate declaration.

Note: Original change https://github.com/alphagov/govuk-puppet/pull/7927

Solo: @suthagarht

### DIFF
--- a/modules/govuk_env_sync/manifests/lock_file.pp
+++ b/modules/govuk_env_sync/manifests/lock_file.pp
@@ -1,0 +1,15 @@
+# == Class: govuk_env_sync::lock_file
+#
+# Set permissions on the lock file
+#
+class govuk_env_sync::lock_file
+{
+
+  # Lock file
+  #
+  file { '/etc/unattended-reboot/no-reboot/govuk_env_sync':
+    mode  => '0660',
+    owner => $govuk_env_sync::user,
+    group => $govuk_env_sync::user,
+  }
+}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: modules/govuk_env_sync/manifests/lock_file.pp [hunk 1]
```diff
@@ -0,0 +1,15 @@
+# == Class: govuk_env_sync::lock_file
+#
+# Set permissions on the lock file
+#
+class govuk_env_sync::lock_file
```
deanwilson: Formatting is incorrect.

modules/govuk_env_sync/manifests/lock_file.pp
```diff
@@ -0,0 +1,15 @@
...
+{
+
+  # Lock file
+  #
+  file { '/etc/unattended-reboot/no-reboot/govuk_env_sync':
+    mode  => '0660',
+    owner => $govuk_env_sync::user,
```
deanwilson: You are using a variable that doesn't exist in this scope. It should be a param that is hiera'd or  default.

deanwilson:
There should be a spec for the new class. That will help pick up things like the variables being  used outside of their scope.

PR #7943, opened 2018-08-10
Add hieradata for pulling the MySQL databases for the data sync

issyl0:
- Alphabetize the existing hieradata.
- Arbitrarily set timings for the cron jobs. Several hours later than
  their `push` counterparts.


hieradata_aws/class/staging/db_admin.yaml
```diff
@@ -29,3 +29,33 @@ govuk_env_sync::tasks:
     temppath: "/tmp"
     url: "govuk-staging-database-backups"
     path: "mysql"
+  "pull_pg_content_publisher_production_daily":
```
issyl0: Not sure why this is `pull_pg_...` as we specify the database in the YAML below...

hieradata_aws/class/staging/db_admin.yaml
```diff
@@ -29,3 +29,33 @@ govuk_env_sync::tasks:
...
+    hour: "0"
+    minute: "12"
+    action: "pull"
+    dbms: "postgresql"
+    storagebackend: "s3"
+    database: "content_publisher_production"
+    temppath: "/tmp/content_publisher_production"
```
issyl0: Should we be consistent about `temppath` locations? They seem to vary between `/home/govuk-backup/...` and `/tmp`.

hieradata_aws/class/staging/db_admin.yaml
issyl0: We can optimize this inconsistency when we've tested that all these jobs work.

hieradata_aws/class/staging/db_admin.yaml
issyl0: We can optimize this inconsistency when we've tested that all these jobs work.

PR #7945, opened 2018-08-10
Increase number of processes deploy user can run

leenagupte:
All of the processes, e.g. running a rails console, deploying applications) that the `deploy` user tries to invoke in Integration is failing with the following error:

`sudo: unable to fork: Resource temporarily unavailable`

This was due to the deploy user using up all of it’s available processes.

We devops a fix by upping the process limit to 2048 in /etc/security/limits.d/deploy_nproc.conf on both machines.

This change replicates that fix.


leenagupte:
@deanwilson Could you tick the approve box? Thanks.

PR #7950, opened 2018-08-10
Fix the data sync pull "path" value for transition_production

issyl0:
- If this doesn't match its [`push` counterpart](https://github.com/alphagov/govuk-puppet/blob/master/hieradata/node/transition-postgresql-master-1.staging.publishing.service.gov.uk.yaml#L11), then the script won't
  find the data in the S3 bucket.

https://trello.com/c/l7dNI7Th/1411-configure-the-postgres-data-sync-policy-via-hieradata-and-ensure-it-is-functioning


issyl0:
Thanks Dean. I added a link in the PR description.

PR #7956, opened 2018-08-13
Allow smartanswers to run in draft preview

leenagupte:
Trello: https://trello.com/c/W86Z7ABG/214-get-smart-answers-working-on-draft-stack

At the moment a 500 error is returned when you try to navigate the landing page of a smartanswer in draft preview.
I think this is because the smartanswers app is not running in draft.

These changes are based on the set-up for `draft-government-frontend`
Rather than starting up a new `calculators` machine, draft-smartanswers will be run on the `frontend` machine on port 3231

### DIFF
--- a/modules/govuk/manifests/apps/smartanswers/enable_running_in_draft_mode.pp
+++ b/modules/govuk/manifests/apps/smartanswers/enable_running_in_draft_mode.pp
@@ -1,0 +1,73 @@
+# == Class govuk::apps::smartanswers::enable_running_in_draft_mode
+#
+# Enables running smartanswers for serving draft content.
+#
+# Intended to be used only in development.
+# In draft mode smartanswers:
+#
+# 1. runs on a different port: 3231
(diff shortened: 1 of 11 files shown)


modules/govuk/manifests/apps/smartanswers/enable_running_in_draft_mode.pp
```diff
@@ -0,0 +1,73 @@
+# == Class govuk::apps::smartanswers::enable_running_in_draft_mode
+#
+# Enables running smartanswers for serving draft content.
+#
+# Intended to be used only in development.
```
cbaines: I'm not sure this is quite the case? This will improve the publisher experience with the draft site right?

leenagupte: Classic copy and paste error.

leenagupte: @cbaines It turns out this comment is correct. This manifest does only run in the dev vm.

PR #7961, opened 2018-08-14
Only ip_forward on Docker machines

kevindew:
Trello: https://trello.com/c/obJqITrJ/213-docker-is-causing-ci-builds-to-fail

This firstly changes sysctl to use systctl.d rather than a single sysctl.conf file

Secondly it then creates an extra sysctl.d file for docker which applies the ip_forward setting so that this only applies to boxes with docker running. It does this via creating a define for building the sysctl files. This is done so that we have a loose dependency between modules that define sysctl files and govuk_harden::sysctl itself and thus only create those files at the point of running govuk_harden::sysctl (and modules will compile without govuk_harden::sysctl)

As this changes how sysctl is configured I've compared the differences of running this on a backend box on integration to check nothing has changed dramatically:

```
➜  /tmp diff backend-before-sysctl backend-after-sysctl
41c41
< fs.dentry-state = 211066	187595	45	0	0	0
---
> fs.dentry-state = 220702	196447	45	0	0	0
45,47c45,47
< fs.file-nr = 13536	0	1584982
< fs.inode-nr = 229212	177819
< fs.inode-state = 229212	177819	0	0	0	0	0
---
> fs.file-nr = 13568	0	1584982
> fs.inode-nr = 235836	177819
> fs.inode-state = 235836	177819	0	0	0	0	0
149c149
< kernel.ns_last_pid = 17653
---
> kernel.ns_last_pid = 30681
177c177
< kernel.pty.nr = 106
---
> kernel.pty.nr = 107
180c180
< kernel.random.entropy_avail = 840
---
> kernel.random.entropy_avail = 779
184c184
< kernel.random.uuid = 89860ca8-7ef6-4c36-93b2-d31fe32352c1
---
> kernel.random.uuid = f1462bef-10f5-4b8c-b61c-f02c6f5ca6e7
352c352
< net.ipv4.conf.all.accept_redirects = 0
---
> net.ipv4.conf.all.accept_redirects = 1
363c363
< net.ipv4.conf.all.forwarding = 1
---
> net.ipv4.conf.all.forwarding = 0
391c391
< net.ipv4.conf.default.forwarding = 1
---
> net.ipv4.conf.default.forwarding = 0
419c419
< net.ipv4.conf.eth0.forwarding = 1
---
> net.ipv4.conf.eth0.forwarding = 0
447c447
< net.ipv4.conf.lo.forwarding = 1
---
> net.ipv4.conf.lo.forwarding = 0
477c477
< net.ipv4.ip_forward = 1
---
> net.ipv4.ip_forward = 0
797c797
< net.netfilter.nf_conntrack_count = 2425
---
> net.netfilter.nf_conntrack_count = 2040
```

Only things that seem to have changed are the forwarding which is expected since this is applied to just docker boxes.


kevindew:
@dougneal argh of course! I've fixed that in a4864821e. Thanks for reviewing.

PR #7963, opened 2018-08-14
Change the logging level

deanwilson:
Setting it to VERBOSE logs the users key fingerprint on login


cbaines:
I guess this could be useful, but is there a specific motivation here?

deanwilson:
Without this there is no way to see which key was used to login. We will eventually need this.

It's also pentest related but that's a side point.

cbaines:
> Without this there is no way to see which key was used to login. We will eventually need this.
> 
> It's also pentest related but that's a side point.

Ok.

PR #7970, opened 2018-08-15
Fix up the config, includes and typos for transition-postgresql data sync

issyl0:


issyl0:
Thanks for adding in the things I missed, David! 😅

Because I raised this PR, I can't interact with it further, so someone else will have to review.

PR #7976, opened 2018-08-15
Increase db::lb_ip_range for Content Data Admin's PG Bouncer

chao-xian:
So that our app can finally connect to the database from specific
boxes.


chao-xian:
[(8) Provision and deploy a new barebones Admin app](https://trello.com/c/kOuisVYY/547-8-provision-and-deploy-a-new-barebones-admin-app)

PR #7982, opened 2018-08-16
Declaring temppath resource in task breaks sync

schmie:
- We want to be able to declare multiple tasks using the same temppath.

- Puppet requires resources to be unique.

- Will need ensure directory existence/access rights by other means.

@schmie


issyl0:
So we didn't need to do the Mongo portion of #7980 after all!

PR #7984, opened 2018-08-16
Set temppath to /tmp in place of /var/lib/autopostgresqlbackup

szd55gds:
Set temppath to /tmp in place of /var/lib/autopostgresqlbackup

- govuk-backup user unable to write to /var/lib
- Discussed this short term measure with @schmie 
- Longer team plan is to provide separate storage for the backup app.

@szd55gds

### DIFF
--- a/hieradata/node/warehouse-postgresql-1.backend.staging.publishing.service.gov.uk.yaml
+++ b/hieradata/node/warehouse-postgresql-1.backend.staging.publishing.service.gov.uk.yaml
@@ -6,6 +6,6 @@ govuk_env_sync::tasks:
     dbms: "postgresql"
     storagebackend: "s3"
     database: "content_performance_manager_production"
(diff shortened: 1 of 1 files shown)


issyl0:
Sorry to be picky, but could you set the first line of your commit message to the thing you _did_, and the follow on description to be _why_?

I thought the `/var/lib/autopostgresqlbackup` path was OK for `push` jobs? This'll therefore need fixing across the board if this isn't the case, as a lot of the `push` jobs have `/var/lib/autopostgresqlbackup` as a temppath.

schmie:
I think the missing autopostgresqlbackup directory is specific to warehouse / it was never added to the whole autobackup mechanism...

szd55gds:
@schmie is correct. There is no autopostgresqlbackup dir and the user is unable to create one due to permissions. I will edit the commit message, will this be acceptable?

PR #7986, opened 2018-08-16
Adding this DB to the env_backup process

szd55gds:
Tested end to end - successful.


schmie:
The GDS style guide suggests you include a short header describing the change as well as a longer message describing the "why". See:
https://github.com/alphagov/styleguides/blob/master/git.md

issyl0:
I rebased this on `master` to get rid of the shellcheck error, and edited the commit message while I was there.

PR #7987, opened 2018-08-16
Turn off GDS API adapter caching

elliotcm:
This can be turned back on on a per-app basis by adding `cache_api_calls => true` to `govuk::app` invocations.

https://trello.com/c/OOMMRcLJ/397-disable-gds-api-adapters-cache-in-any-frontend-app-that-still-has-it-enabled

### DIFF
--- a/modules/govuk/manifests/app/config.pp
+++ b/modules/govuk/manifests/app/config.pp
@@ -32,6 +32,11 @@
 # [*cpu_critical*]
 #   CPU usage percentage that alerts are sounded at
 #
+# [*cache_api_calls*]
+#   Control whether the app uses the in-memory cache that comes with
+#   GDS API adapters.
+#   Default: undef
+#
 define govuk::app::config (
   $app_type,
   $domain,
@@ -63,6 +68,7 @@ define govuk::app::config (
   $unicorn_worker_processes = undef,
   $cpu_warning = 150,
   $cpu_critical = 200,
+  $cache_api_calls = undef,
 ) {
   $ensure_directory = $ensure ? {
     'present' => 'directory',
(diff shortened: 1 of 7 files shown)


modules/govuk/manifests/app/config.pp
```diff
@@ -139,6 +145,9 @@
       "${title}-SENTRY_DSN":
         varname => 'SENTRY_DSN',
         value   => $sentry_dsn;
+      "${title}-GDS_API_DISABLE_CACHE":
+        varname => 'GDS_API_DISABLE_CACHE',
+        value   => "${!cache_api_calls}";
```
thomasleese: If I understand [how it decides whether to disable the cache](https://github.com/alphagov/gds-api-adapters/blob/6d1576dd0340720fb7bdedabec08b8fea64125b3/lib/gds_api/json_client.rb#L42) correctly, a string `"false"` will be truthy, should it be that the environment variable is either `""` or `"true"` rather than `"false"` or `"true"`?

elliotcm: `""` is truthy too..

elliotcm: Perhaps that env var override never worked.

thomasleese: Ah that's where Ruby and Python differ about the truthness of strings 😉

elliotcm: So what shall we do?  If we want to take this approach we'd have to release a new version of the gem and bump it in every app.

thomasleese: Can we do something like:

```puppet
if $cache_api_calls {
  # ensure present the environment variable with the value
} else {
  # ensure absent the environment variable
}
```

elliotcm: (as a bit of an FYI then, in Ruby only `nil` and `false` are falsey)

elliotcm: We can, yes, though that wouldn't override any where it's explicitly turned on in the app code, though I don't *think* there are any cases of that.

elliotcm:
@rhowe-gds Could you give this a final once-over?

modules/govuk/manifests/app/config.pp
```diff
@@ -141,6 +147,20 @@
         value   => $sentry_dsn;
     }
 
+    $cache_api_calls_final = pick($cache_api_calls, hiera('govuk::app::cache_api_calls', true))
```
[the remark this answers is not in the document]
elliotcm: Done

rhowe-gds:
Looks fine to me, but I don't any approval powers over GOV.UK... yet :)

PR #7993, opened 2018-08-17
Change the way the temp path are created to avoid duplicate

rtrinque:
The way tasks cfg files are created at the moment raises a
duplicate resources error in puppet if we use the same temp dir
more than once. This forces us to use different dirs and is
extremely error prone. This fixes this problem.


issyl0:
Didn't #7982 fix this?

PR #8007, opened 2018-08-20
Display thread count on the deployment dashboard

barrucadu:
I considered replacing the process count graph with a thread count graph, but decided against it.  It's useful to see both, because this allows distinguishing between process leaks and thread leaks.  A process leak will have both graphs going up, whereas a thread leak will only have the thread graph going up.

To make space, I moved the "useful links" panel to the bottom.

---

[Trello card](https://trello.com/c/mPzlp9D2/413-make-the-deployment-dashboard-show-the-number-of-threads)

### DIFF
--- a/modules/grafana/manifests/dashboards/deployment_dashboard.pp
+++ b/modules/grafana/manifests/dashboards/deployment_dashboard.pp
@@ -107,8 +107,8 @@ define grafana::dashboards::deployment_dashboard (
 
   $panel_partials = concat(
     [
-      ['processor_count', '5xx_rate'],
-      ['error_counts_table', 'links']
+      ['processor_count', 'thread_count'],
+      ['error_counts_table', '5xx_rate']
     ],
     $worker_row,
     $errors_by_controller_row,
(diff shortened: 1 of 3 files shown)


cbaines:
What's expected to happen to the thread count when deploying apps? I guess like the process count, when using unicorn it'll go up, but I don't really see what it shows beyond the thread count (when deplyoing apps).

barrucadu:
It'll go up, and then go back down - just like the process count.

But the panel is useful because it'll let us spot thread leaks, as recently happened after a deploy of Content Performance Manager, which ultimately caused other apps to crash.

cbaines:
> But the panel is useful because it'll let us spot thread leaks, as recently happened after a deploy of Content Performance Manager, which ultimately caused other apps to crash.

This reminded me of the recent monitoring added for the procfile worker thread counts, I've now put up a Pull Request to extend that to the apps [1].
1: https://github.com/alphagov/govuk-puppet/pull/8010

In general, I think the alerts are a better way of watching for this, but I'm not totally against adding the threads to the deployment dashboard.

barrucadu:
Closing in favour of #8010

PR #8021, opened 2018-08-22
Add configuration for separate publishing-api rds and db_admin

schmie:
- To reduce load on individual instances, we are splitting
publishing-api and email-alert-api into seperate rds instances

- This does not "flip the switch" for use by publishing-api, yet, but will configure the instances and allow us to test and sync data before switching the app over to use it.

- Files created from warehouse equivalents and s/warehouse/publishing-api/g

Update/Do not merge: This now includes publishing_api::db, where I like to double check it does not
interact with the app in it's current state

@schmie

### DIFF
--- a/modules/govuk/manifests/node/s_publishing_api_db_admin.pp
+++ b/modules/govuk/manifests/node/s_publishing_api_db_admin.pp
@@ -1,0 +1,102 @@
+# == Class: Govuk_Node::S_publishing_api_db_admin
+#
+# This machine class is used to administer publishing_api PostgreSQL RDS instances.
+#
+# === Parameters
+#
+class govuk::node::s_publishing_api_db_admin(
+  $backup_s3_bucket     = undef,
+  $postgres_host        = undef,
+  $postgres_user        = undef,
+  $postgres_password    = undef,
+  $postgres_port        = '5432',
+  $postgres_backup_hour = 7,
+  $postgres_backup_min  = 30,
+  $apt_mirror_hostname,
+) {
+  include govuk_env_sync
+  include ::govuk::node::s_base
+
+  $alert_hostname = 'alert'
+
+  if $backup_s3_bucket {
+    $ensure = 'present'
+  } else {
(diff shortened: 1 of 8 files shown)


modules/govuk/manifests/node/s_publishing_api_db_admin.pp
```diff
@@ -0,0 +1,102 @@
...
+
+  # This class collects the resources that are exported by the
+  # govuk_postgresql::server::db defined type
+  include ::govuk_postgresql::server::not_slave
+
+  # Ensure the client class is installed
+  class { '::govuk_postgresql::client': } ->
+
+  # include all PostgreSQL classes that create databases and users
+  class { '::govuk::apps::content_performance_manager::db': }
```
issyl0: Wrong database.

modules/govuk/manifests/node/s_publishing_api_db_admin.pp
```diff
@@ -0,0 +1,102 @@
...
+    architecture => $::architecture,
+    key          => '3803E444EB0235822AA36A66EC5FE1A937E3ACBB',
+  }
+
+  package { 'gof3r':
+    ensure  => $ensure,
+    require => Apt::Source['gof3r'],
+  }
+
+  $alert_hostname = 'alert'
```
issyl0: Could you move this up to where the other variables are defined?

schmie: This is done

modules/govuk/manifests/node/s_publishing_api_db_admin.pp
schmie: I have changed this to publishing_api::db, it got me thinking, there may be interactions, despite not touching the app config, yet, thus added do not merge for now

schmie:
Tested on AWS staging today and I am happy it did not break publishing API - next up await sync and the redirect pgbouncer (again testing in AWS staging)

PR #8029, opened 2018-08-23
Permit varnish cache purge from 10.1.0.0/16

elliotcm:
This is so our backend machines can issue specific URL cache purges when a document is (re)published.

https://trello.com/c/f8U7EfUh/427-make-the-new-cache-clearing-app-invalidate-cache-in-varnish

### DIFF
--- a/modules/varnish/templates/default.vcl.erb
+++ b/modules/varnish/templates/default.vcl.erb
@@ -13,6 +13,7 @@
 
 acl purge_acl {
   "localhost";
+  "10.1.0.0"/16;
 }
 
 sub vcl_recv {


modules/varnish/templates/default.vcl.erb
```diff
@@ -13,6 +13,7 @@
 
 acl purge_acl {
   "localhost";
+  "10.1.0.0"/16;
```
thomasleese: I'm not too familiar with the syntax, but the `/16` is fine outside the quotes?

elliotcm: I wouldn't normally write it like that, but I was basing it off the varnish docs:
https://varnish-cache.org/docs/4.1/users-guide/purging.html#http-purging

thomasleese: Weird!

PR #8033, opened 2018-08-23
Create a Puppet Class for email-alert-api db-admin

suthagarht:
- This change creates a puppet class to manage email-alert-api db-admin
instances.

https://trello.com/c/ljsMBGQ4/1419-split-email-alert-api-into-its-own-rds-instance

Solo: @suthagarht

### DIFF
--- a/modules/govuk/manifests/node/s_email_alert_api_db_admin.pp
+++ b/modules/govuk/manifests/node/s_email_alert_api_db_admin.pp
@@ -1,0 +1,102 @@
+# == Class: Govuk_Node::s_email_alert_api_db_admin
+#
+# This machine class is used to administer email-alert-api PostgreSQL RDS instances.
+#
+# === Parameters
+#
+class govuk::node::s_email_alert_api_db_admin(
+  $backup_s3_bucket     = undef,
+  $postgres_host        = undef,
+  $postgres_user        = undef,
+  $postgres_password    = undef,
+  $postgres_port        = '5432',
+  $postgres_backup_hour = 7,
+  $postgres_backup_min  = 30,
+  $apt_mirror_hostname,
+) {
+  include govuk_env_sync
+  include ::govuk::node::s_base
+
+  if $backup_s3_bucket {
+    $ensure = 'present'
+  } else {
+    $ensure = 'absent'
+  }
+
+  apt::source { 'gof3r':
+    ensure       => $ensure,
+    location     => "http://${apt_mirror_hostname}/gof3r",
+    release      => $::lsbdistcodename,
+    architecture => $::architecture,
+    key          => '3803E444EB0235822AA36A66EC5FE1A937E3ACBB',
+  }
+
+  package { 'gof3r':
+    ensure  => $ensure,
+    require => Apt::Source['gof3r'],
+  }
+
+  $alert_hostname = 'alert'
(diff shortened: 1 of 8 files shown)


modules/govuk/manifests/node/s_email_alert_api_db_admin.pp
```diff
@@ -0,0 +1,102 @@
...
+
+  # This class collects the resources that are exported by the
+  # govuk_postgresql::server::db defined type
+  include ::govuk_postgresql::server::not_slave
+
+  # Ensure the client class is installed
+  class { '::govuk_postgresql::client': } ->
+
+  # include all PostgreSQL classes that create databases and users
+  class { '::govuk::apps::content_performance_manager::db': }
```
issyl0: This is Email Alert API, not Warehouse.

modules/govuk/manifests/node/s_email_alert_api_db_admin.pp
```diff
@@ -0,0 +1,102 @@
...
+
+  $postgres_backup_desc = 'RDS Warehouse PostgreSQL backup to S3'
```
issyl0: Same.

modules/govuk/manifests/node/s_email_alert_api_postgresql.pp
```diff
@@ -0,0 +1,28 @@
...
+  $wale_private_gpg_key,
+  $wale_private_gpg_key_fingerprint,
+  $alert_hostname = 'alert.cluster',
+) inherits govuk::node::s_base  {
+  include govuk_env_sync
+  Govuk_mount['/var/lib/postgresql'] -> Class['govuk_postgresql::server::standalone']
+
+  include govuk_postgresql::server::standalone
+  include govuk_postgresql::tuning
+  include govuk::apps::content_performance_manager::db
```
issyl0: Wrong database.

modules/govuk/manifests/node/s_email_alert_api_db_admin.pp
suthagarht: thank you, corrected

modules/govuk/manifests/node/s_email_alert_api_db_admin.pp
suthagarht: thank you done

modules/govuk/manifests/node/s_email_alert_api_postgresql.pp
suthagarht: thank you , done

deanwilson:
We should either do a small refactor or add a dedup card to the backlog board to refactor the duplicate code out into a class/define when this work is done and deployed. the `gof3r` code for example should live in one place.

issyl0:
@deanwilson Yep, this is duplicated across all the db-admin manifests so far.

schmie:
Laura's solution to this was: https://github.com/alphagov/govuk-puppet/blob/b588e4ade996e97b8975e69cb00800521fff4a48/modules/hosts/manifests/purge.pp..
I am not sure, if there are a differences(?!) Should we unify this?

issyl0:
Yeah, I pointed that solution out, but it still failed the tests.

PR #8034, opened 2018-08-23
Fix typo in hieradata for government-frontend

benthorner:
https://trello.com/c/vticNyhn/233-fix-government-frontend-unicorn

The variables were incorrectly using a hyphen in 'government-frontend',
instead of an underscore, which is what the module is expecting.


cbaines:
@benthorner good find :+1: It would be great if you could make the same change in the AWS specific hieradata [1], so that the behaviour doesn't revert in the event of production being migrated to AWS.

1: https://github.com/alphagov/govuk-puppet/blob/master/hieradata_aws/common.yaml#L495-L497

PR #8037, opened 2018-08-23
Remove gds api adapters cache

thomasleese:
This has no effect now as the cache has been disabled.

This depends on https://github.com/alphagov/gds-api-adapters/pull/848.

[Trello Card](https://trello.com/c/0HSDBjNk/398-remove-gds-api-adapters-cache)

### DIFF
--- a/modules/govuk/manifests/app/config.pp
+++ b/modules/govuk/manifests/app/config.pp
@@ -32,11 +32,6 @@
 # [*cpu_critical*]
 #   CPU usage percentage that alerts are sounded at
 #
-# [*cache_api_calls*]
-#   Control whether the app uses the in-memory cache that comes with
-#   GDS API adapters.
-#   Default: false
-#
 define govuk::app::config (
   $app_type,
   $domain,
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/app/config.pp
```diff
@@ -148,18 +142,6 @@
...
-    if $cache_api_calls {
-      $ensure_disable_cache_var = 'absent'
-    } else {
-      $ensure_disable_cache_var = 'present'
-    }
-
-    govuk::app::envvar { "${title}-GDS_API_DISABLE_CACHE":
-      ensure  => $ensure_disable_cache_var,
-      varname => 'GDS_API_DISABLE_CACHE',
-      value   => "${!cache_api_calls_final}",
```
elliotcm: I'm a little confused, do you know where the code went that was setting this variable from hiera?

thomasleese: Yeah, it went away when we deployed to production: https://github.com/alphagov/govuk-puppet/pull/8036

PR #8047, opened 2018-08-29
Revert "Disable gds api adapters cache everywhere"

thomasleese:
Reverts alphagov/govuk-puppet#8036

Just in case... this will make the PR ready to merge quickly by already running the tests on CI.


thomasleese:
We probably don't need this in the end.

PR #8050, opened 2018-08-29
Add spurious errors encountered in pg databases

schmie:
- email-alert-api add different COMMENT lines / uses different
extensions than other databases, these create warnings we have to handle
manually

- Also changed the array separator used to avoid clashes with use of
hypens in the data set/pg errors.

solo @schmie


issyl0:
Are we* ever going to want to actually fix these errors, or will we just suppress them until the end of time?

* Is someone

schmie:
These are all related to incompatibilities between RDS and our "normal" PG servers: 
- Different super user "postgres" vs "aws_db_admin" 
- "COMMENT ON" commands for extensions are invalid, because the AWS account is not the canonical PG root/does not own extensions.
They will go away once this operates RDS->RDS, I will add a card to the post-move backlog.

PR #8057, opened 2018-08-30
Add post processing functionality to data sync

schmie:
- signon and router require replacement of environment specific urls
post sync

- This might also be used for crude data sanitation

solo @schmie

### DIFF
--- a/modules/govuk_env_sync/files/govuk_env_sync.sh
+++ b/modules/govuk_env_sync/files/govuk_env_sync.sh
@@ -44,7 +44,13 @@ set -o errtrace
 #
 args=("$@")
 
+# Get local ip addr, avoiding using puppet templating of this script.
 ip_address=$(ip addr show dev eth0 | grep -Eo 'inet ?([0-9]*\.){3}[0-9]*' | grep -Eo '([0-9]*\.){3}[0-9]*')
+
+# LOCAL_DOMAIN as defined in /etc/govuk_env_sync/env.d
+# No, it is not a typo
+# shellcheck disable=SC2153
+local_domain="${LOCAL_DOMAIN}"
 
 function log {
   echo -ne "$(basename "$0"): $1\\n"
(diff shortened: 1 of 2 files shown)


modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -270,6 +274,57 @@ function get_timestamp_rsync {
...
+
+  mongo --quiet --eval \
+    "db = db.getSiblingDB(\"${database}\"); \
+    db.backends.find( { \"backend_id\": { \$ne: \"licensify\" } } ).forEach( \
+      function(b) { b.backend_url = b.backend_url.replace(\".${source_domain}\", \".${local_domain}\"); \
+    db.backends.save(b); } ); \
+    db.backends.find( { \"backend_id\": \"licensify\" } ).forEach( \
+      function(b) { b.backend_url = b.backend_url.replace(\".${source_domain}\", \".${licensify_domain}\"); \
+    db.backends.save(b); } );"
+}
```
cbaines: One possible alternative to this would be to run the `register_backends` Rake task [1] in the Content Store. That would update the backends in the Router API, from the Content Store, meaning that you wouldn't need to touch the database directly.

1: https://github.com/alphagov/content-store/blob/master/lib/tasks/register_backends.rake

schmie: Hi @cbaines, if you can live with it, I rather update the database directly, as otherwise I have to enable the sync to talk to the content_store somehow.
Having had to add in this post processing stuff is really not ideal, but I would like to move on from the data sync soon (and keep things in one place - We have penciled in a day for a final refactor, though).

cbaines: I'm not too fussed. Keep in mind though that other parts of the data sync currently rely on running rake tasks, e.g. https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/templates/jobs/data_sync_complete_integration.yaml.erb . In my mind, this Content Store action could be done similarly.

schmie: I will keep this in mind - We were talking to @benthorner earlier about potentially triggering downstream tasks via Jenkins (to avoid giving govuk-backup any direct access and execution privileges..) will continue to think about what to do..

PR #8068, opened 2018-08-31
Add local DNS caching to Carrenza staging cache machines

rtrinque:
We want to cache DNS request during trafic replay so as not to overwhelm the resolver on which cache machines depend.


modules/nscd/manifests/service.pp
```diff
@@ -0,0 +1,11 @@
+# == Class: nscd::service
+#
+# Set up and run the SSH service.
```
deanwilson: SSH?

PR #8075, opened 2018-09-04
Remove asset-master and asset-slave boxes

barrucadu:
The NFS asset share is on asset-master, so that share is also gone.

I'm pretty confident that these can be safely removed:

- `sudo lsof /data/uploads` on a production whitehall-admin and backend machine: no files.
- `ps aux | grep 2900` (the UID of the "assets" user) on the same: no processes.

Other things which will need to be touched after this change:

- The env sync script for assets can just be deleted now.
- The actual boxes in Carrenza can be removed.
- Terraform will need a change somewhere too I guess.
- I think once this is deployed we can just delete the entire `assets` module.

---

[Trello card](https://trello.com/c/M3Dfrj8S/443-remove-attachments-from-nfs)


barrucadu:
Yep, we'll have to follow this process once we're confident it's all worked: https://docs.publishing.service.gov.uk/manual/remove-machines.html

barrucadu:
Just pushed a change to mark the mountpoint as `absent`, not just `unmounted`, which unmounts *and* removes from fstab: https://puppet.com/docs/puppet/5.3/types/mount.html#mount-attribute-ensure

barrucadu:
So, we can't actually do this, because there is nothing to ensure that (eg) the virus scan worker runs on the same box that the file was uploaded to.  Fixing this in the Asset Manager (and possibly whitehall?  I experienced some weirdness around PDF thumbnails) would be too much work.

So instead we're going to delete everything we can from the share and shrink the box.

PR #8082, opened 2018-09-04
Move integration Jenkins jobs

benthorner:
The list of jobs available on Jenkins was previously in the shared
jenkins config, which is then overridden in staging and production.

It makes more sense to move this into the integration-specific config
file for Jenkins, as that's the only place where it's used.

This also applies to a group of config settings, such as the
'banner_string' and the 'banner_colour_text'.


benthorner:
I have now and remarkable everything is fine ;-).

PR #8083, opened 2018-09-04
Add YAML for non-sanitized database syncs into Integration

issyl0:
- Configure the data sync pull for ElasticSearch, MongoDB, PostgreSQL and MySQL.
- This `pull`s from S3 at an arbitrary time of times within the 4am hour.
- In the PostgreSQL and MySQL cases, this changeset does not include Email Alert API, Publishing API or Whitehall databases as they need to be sanitized first so we can't pull them straight from staging.

https://trello.com/c/9Oa4A8ji/1439-make-the-new-data-sync-work-for-integration-s


issyl0:
This was missing Transition, Content Performance Manager and the databases on router_backend Mongo instances. I've added those now.

PR #8089, opened 2018-09-05
Execute AWS Jenkins job for data sync complete

schmie:
- The data_sync_complete Jenkins job will have never been triggered in
  integration since it has moved to AWS.

- If this is working, the logic around the
trigger_data_sync_complete(_aws) jobs should be revisited as it is quite
messy (at least added to backlog/techdebt

solo: @schmie


issyl0:
This is the job that repoints URLs to environment specific ones, right, as well as some other stuff? :disappointed: :scream:

benthorner:
Not exactly - this trigger the job you're thinking of, which is defined in integration. Some of the repointing stuff also happens as part of the data sync process; but triggering the Data_Sync_Complete job on integration is necessary to do other post-processing, such as republishing tasks.

schmie:
This has been finished on another branch by Ben. Closing.

PR #8104, opened 2018-09-10
Add AWS X-Ray daemon to app machines

rubenarakelyan:
This commit adds the AWS X-Ray daemon to all node types that host Ruby applications.

Depends on https://github.com/alphagov/govuk-puppet/pull/8102 being deployed.

Trello: https://trello.com/c/YMZBPMGF/450-provision-x-ray-and-set-up-permissions-%F0%9F%8D%90-portunity

### DIFF
--- a/modules/govuk/manifests/node/s_backend.pp
+++ b/modules/govuk/manifests/node/s_backend.pp
@@ -28,6 +28,8 @@ class govuk::node::s_backend inherits govuk::node::s_base {
     ensure => installed,
   }
 
+  include govuk_aws_xray_daemon
+
   include imagemagick
 
   include nginx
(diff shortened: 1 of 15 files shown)


modules/govuk/manifests/node/s_backend.pp
```diff
@@ -28,6 +28,8 @@
     ensure => installed,
   }
 
+  include govuk_aws_xray_daemon
```
barrucadu: If we're going to have config like this which applies to all app nodes, does it make sense to have an `s_app_base` class?  There are a few other classes like this.

rubenarakelyan: Yeah, we have https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/node/s_app_server.pp but it has the comment "This class is deprecated; nothing new should be added to it and things should be removed over time."

PR #8105, opened 2018-09-10
Remove asset_master_and_slave_disk_space_similar alert + clean-up an unused mountpoint

barrucadu:
The NFS share is only used as temporary scratch space to coordinate
the asset manager instances, and to store a few files by publisher and
support-api.

---

[Trello card](https://trello.com/c/M3Dfrj8S/443-remove-attachments-from-nfs)

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -100,8 +100,6 @@ govuk::node::s_api_lb::search_servers:
 
 govuk::node::s_asset_master::copy_attachments_hour: 11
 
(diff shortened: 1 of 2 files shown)


thomasleese:
Is there any concern if the files by publisher or support-api are missing on the asset slaves because we no longer have this alert?

barrucadu:
That is a concern, but I think it'll be covered by the copy_attachments_to_slaves alert.  Actually, I'm not sure what the similar disk space alert gets us over that.

PR #8112, opened 2018-09-11
Remove a bunch of ensured-absent stuff

barrucadu:
Also removed the govuk_java::oracle8 class as nothing uses it any more.


issyl0:
We need to remove the `davidpye` and `suthagahrt` users as 2ndline go through the leaver's tickets for them. Indeed, there's [currently a PR](https://github.com/alphagov/govuk-puppet/pull/8107) for one of those. I don't know what's best, to remove them in this and then carry on with the leaver's process, or do the leavers process in its entirety?

barrucadu:
Probably to do the leavers process in its entirety.  I'll not remove their manifests in this PR.

PR #8116, opened 2018-09-11
Don't clear Fastly in integration and staging

thomasleese:
We currently don't have a suitable Fastly set up in integration and staging and this causes exceptions in those environments.

[Trello Card](https://trello.com/c/9k1peA1F/434-make-the-new-cache-clearing-app-invalidate-cache-in-fastly)


thomasleese:
I thought we would need this but actually we don't, you only get exceptions for 404 pages, which is fine.

PR #8117, opened 2018-09-11
Don't clear the cache on republish

thomasleese:
The content of the page won't have changed if it's been republished, so we don't need to clear the cache.

[Trello Card](https://trello.com/c/9k1peA1F/434-make-the-new-cache-clearing-app-invalidate-cache-in-fastly)

### DIFF
--- a/modules/govuk/manifests/apps/cache_clearing_service/rabbitmq.pp
+++ b/modules/govuk/manifests/apps/cache_clearing_service/rabbitmq.pp
@@ -56,7 +56,7 @@ class govuk::apps::cache_clearing_service::rabbitmq (
   }
 
   rabbitmq_binding { "binding_republish_${amqp_exchange}@${amqp_queue}@/":
-    ensure           => 'present',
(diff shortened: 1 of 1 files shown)


cbaines:
I'm not sure the reasoning here matches at least our current use of that update type. I think often when people "republish" things, it's because something isn't showing up correctly.

thomasleese:
Hmm, basically I wanted to reduce the number of `PURGE` requests we're sending to Fastly: we're currently sending on average 9 every second. I figured ignoring `republish` could be a quick win here, but I guess you're right.

The real reason we're sending this many requests is because of the increased Publishing API `downstream_low` activity, so really we should be solving that.

cbaines:
> The real reason we're sending this many requests is because of the increased Publishing API downstream_low activity, so really we should be solving that.

:+1: 

In case you don't know, there is a visualisation in Kibana that breaks down the dependency resolution activity by source, which helps identifying where the `downstream_low` activity is coming from.

PR #8131, opened 2018-09-19
Add content publisher to data_sync_complete_integration

kevindew:
This allows the app to have it's migrations run each time the data sync
complete task runs.

It does make me a little nervous that this task may form some sort of
race condition with the whitehall import for content into content
publisher but I think we'll get around this by the order of the builds.
If this becomes more of a problem we may want to run a db:migrate at the
start of the whitehall import.


benthorner:
Yay 👍. As part of this, could you also add this file to the list at the bottom of `docs/adding-a-new-app.md`.

kevindew:
@benthorner Sure, added

PR #8133, opened 2018-09-20
Changing IP ranges for in AWS in line with the new migration plan

rtrinque:
We need different IP ranges in AWS and carrenza so that we can run part of gov.uk on one platform and the rest on the other during the migration

### DIFF
--- a/hieradata_aws/production.yaml
+++ b/hieradata_aws/production.yaml
@@ -9,7 +9,7 @@ base::supported_kernel::enabled: false
 
 cron::daily_hour: 6
 
-environment_ip_prefix: '10.3'
+environment_ip_prefix: '10.13'
 
 govuk::apps::asset_manager::aws_s3_bucket_name: 'govuk-assets-production'
 govuk::apps::asset_manager::aws_region: 'eu-west-1'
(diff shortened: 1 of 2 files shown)


hieradata_aws/production.yaml
```diff
@@ -19,7 +19,7 @@ govuk::apps::content_publisher::google_tag_manager_id: "GTM-NQXC4TG"
 govuk::apps::content_publisher::google_tag_manager_preview: "env-7"
 govuk::apps::content_store::performance_platform_big_screen_view_url: 'https://performance-platform-big-screen-view-production.cloudapps.digital'
 govuk::apps::content_store::performance_platform_spotlight_url: 'https://performance-platform-spotlight-production.cloudapps.digital'
-govuk::apps::email_alert_api::db::backend_ip_range: '10.2.3.0/24'
+govuk::apps::email_alert_api::db::backend_ip_range: '10.12.3.0/24'
```
issyl0: Is this supposed to be `10.13` instead?

rtrinque: Yeah I wasn't sure. I followed the original schema as I thought there might be a good reason for it? But if nobody can think of any we might as well put it on the prod range.

issyl0: I think the production AWS hieradata was copied from staging verbatim, hence `10.2` in places that should be different. Let's put it on the production range, as it's production.

PR #8136, opened 2018-09-20
Disable processing transition logs

thomasleese:
This is now handled by a Lambda in AWS which reads from Athena, exports to an S3 bucket which is then read by the Transition tool directly.

Depends on #8137.

[Trello Card](https://trello.com/c/dNtEfmJM/395-move-transition-hit-stats-processing-to-s3-and-athena-%F0%9F%8D%90-portunity)

### DIFF
--- a/hieradata/production.yaml
+++ b/hieradata/production.yaml
@@ -107,8 +107,6 @@ govuk::deploy::setup::ssh_keys:
 govuk_bouncer::gor::enabled: true
 govuk_bouncer::gor::target: '195.225.216.149'
 
-govuk_cdnlogs::transition_logs::enable_cron: true
-
 govuk_jenkins::config::banner_colour_background: '#df3034'
 govuk_jenkins::config::banner_colour_text: 'white'
 govuk_jenkins::config::banner_string: 'Carrenza PRODUCTION'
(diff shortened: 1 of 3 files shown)


bevanloon:
My puppet knowledge isn't really good enough to know the answer to these, so I was wondering how / if the actual cron job will be deleted once `$enable_cron` is `false` - it looks like the code to create the cron is inside an `if` https://github.com/alphagov/govuk-puppet/pull/8136/files#diff-8def73b3037908d4605b2496346bb283R115 which won't get triggered so `ensure => $ensure` won't fire? 

I've a similar question about the passive check https://github.com/alphagov/govuk-puppet/pull/8136/files#diff-8def73b3037908d4605b2496346bb283R125 i.e. will it be left hanging around without explicitly deleting it.

and last question - will we still need https://github.com/alphagov/govuk-puppet/pull/8136/files#diff-8def73b3037908d4605b2496346bb283R134, because its installation not variable controlled

rubenarakelyan:
Yup agreed, `enabled` needs to be left true for now but `ensure` set to `absent`, then the whole lot can be delete. With the Icinga check, you should also be able to add an `ensure => absent` to it. Same with `mawk`.

thomasleese:
@bevanloon @rubenarakelyan Good point! I've updated the class to correctly remove resources when `enabled` or `enable_cron` is set to `false`.

PR #8137, opened 2018-09-20
Use new transition rake task to fetch stats from s3

barrucadu:
This depends on https://github.com/alphagov/transition/pull/610

---

[Trello card](https://trello.com/c/dNtEfmJM/395-move-transition-hit-stats-processing-to-s3-and-athena-%F0%9F%8D%90-portunity)


rubenarakelyan:
This should first be set to `ensure => absent` then it can be deleted otherwise it may leave stuff lying around.

PR #8143, opened 2018-09-21
Add Terraboard to the monitoring machines

rubenarakelyan:
This commit installs Terraboard to the monitoring machines along with dependencies.

* Installs Docker
* Creates a new custom bridge network called terranet
* Creates a configuration file for OAuth2 Proxy
* Downloads and runs Docker images for Terraboard, PostgreSQL and OAuth2 Proxy
* Creates an Icinga check to ensure Terraboard is running
* Adds non-secret hieradata
* Adds an nginx proxy for the external Terraboard subdomain to the OAuth2 Proxy Docker container

Depends on https://github.com/alphagov/govuk-secrets/pull/446

Trello: https://trello.com/c/6UMlTRpa/490-get-terraboard-running-on-the-monitoring-machines-in-each-environment

### DIFF
--- a/modules/govuk_containers/manifests/terraboard.pp
+++ b/modules/govuk_containers/manifests/terraboard.pp
@@ -1,0 +1,130 @@
+# == Class: govuk_containers::terraboard
+#
+# Install and run Terraboard in a Docker container with PostgreSQL backend
+# and OAuth2 Proxy for authentication, also in Docker containers. All containers
+# are connected to the `terranet` bridge network.
+#
+# The PostgreSQL database is used to cache Terraform states to reduce the amount
+# of traffic.
+#
+# === Parameters
+#
+# [*aws_region*]
+#   The AWS region to fetch Terraform state from.
+#
+# [*aws_access_key_id*]
+#   The AWS access key ID to use when fetching Terraform state.
+#
+# [*aws_secret_access_key*]
+#   The AWS secret access key to use when fetching Terraform state.
+#
+# [*aws_bucket*]
+#   The AWS bucket to fetch the Terraform state from.
+#
+# [*db_password*]
+#   The database password for the PostgreSQL Docker instance used by Terraboard.
+#
+# [*github_oauth_client_id*]
+#   The OAuth client ID provided by GitHub for authentication.
+#
+# [*github_oauth_client_secret*]
+#   The OAuth client secret provided by GitHub for authentication.
+#
+# [*oauth2_proxy_base_url*]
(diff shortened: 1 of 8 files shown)


modules/govuk_containers/manifests/terraboard.pp
```diff
@@ -0,0 +1,127 @@
+# == Class: govuk_containers::terraboard
+#
+# Install and run Terraboard in a Docker container with PostgreSQL backend
```
barrucadu: Is it worth adding a note here about why we're using a dockerised postgres (it's just storing transient state which can be recovered from s3)?  Otherwise someone will see this in the future and wonder why this is different to everything else.

rubenarakelyan: Yup, done!

modules/govuk_containers/manifests/terraboard.pp
```diff
@@ -0,0 +1,130 @@
...
+  $aws_secret_access_key = undef,
+  $aws_bucket = undef,
+  $db_password = undef,
+  $github_oauth_client_id = undef,
+  $github_oauth_client_secret = undef,
+  $oauth2_proxy_base_url = undef,
+  $oauth2_proxy_cookie_secret = undef,
+) {
+
+  file { '/opt/terraboard':
```
thomasleese: Might be nice to combine this and the one below with something like:

```puppet
file { ['/opt/terraboard', '/opt/terraboard/conf']:
  ensure => directory,
  ...
}
```

modules/govuk_containers/manifests/terraboard.pp
```diff
@@ -0,0 +1,130 @@
...
+    require   => Class['govuk_docker'],
+    image_tag => '10.5',
+  }
+
+  ::docker::run { 'terraboard-db':
+    net     => 'terranet',
+    image   => 'postgres',
+    require => Docker::Image['postgres'],
+    env     => [
+      'POSTGRES_USER=gorm',
```
thomasleese: Might be worth pulling `gorm` out into a variable with a comment to explain why it's hardcoded as `gorm`. 😄

rubenarakelyan: Done, yes it is a weird name but I've got used to it!

modules/govuk_containers/manifests/terraboard.pp
rubenarakelyan: TIL!

PR #8145, opened 2018-09-24
Upgrade postgresql to 9.6 on all db_admin classes

rtrinque:
We need version 9.6 of postgresql on email_alert_api_db_admin,
 publishing_api_db_admin, transition_db_admin, and
warehouse_db_admin


schmie:
I'll test with transition in AWS staging

PR #8148, opened 2018-09-25
Remove Terraboard Icinga check

rubenarakelyan:
This check doesn’t work since Terraboard is run inside Docker so we can’t detect it.

Trello: https://trello.com/c/6UMlTRpa/490-get-terraboard-running-on-the-monitoring-machines-in-each-environment

### DIFF
--- a/modules/govuk_containers/manifests/terraboard.pp
+++ b/modules/govuk_containers/manifests/terraboard.pp
@@ -122,6 +122,7 @@ class govuk_containers::terraboard(
   }
 
   @@icinga::check { "check_terraboard_running_${::hostname}":
+    ensure              => absent,
     check_command       => 'check_nrpe!check_proc_running!terraboard',
     service_description => 'terraboard running',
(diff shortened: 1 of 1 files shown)


modules/govuk_containers/manifests/terraboard.pp
```diff
@@ -122,6 +122,7 @@
   }
 
   @@icinga::check { "check_terraboard_running_${::hostname}":
+    ensure              => absent,
     check_command       => 'check_nrpe!check_proc_running!terraboard',
```
thomasleese: I think this should work, there's one for `gemstash` which seems to work: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_containers/manifests/gemstash.pp#L38

https://alert.publishing.service.gov.uk/cgi-bin/icinga/extinfo.cgi?type=2&host=apt-1.management.publishing.service.gov.uk&service=gemstash+running

PR #8153, opened 2018-09-26
Disable the use of insecure arcfour SSH ciphers

issyl0:
- This was a recommendation from our pentest.
- From sshd_config(5):
    If the specified value begins with a `+' character,
    then the specified ciphers will be appended to the default set
    instead of replacing them.  If the specified value begins with a
    `-' character, then the specified ciphers (including wildcards)
    will be removed from the default set instead of replacing them.


issyl0:
For the avoidance of doubt, I got this syntax from [this StackOverflow answer](https://unix.stackexchange.com/questions/333728/ssh-how-to-disable-weak-ciphers/410252#410252) and felt it better to be explicit (as these were the specific ones the pentest report mentioned). 😳

PR #8157, opened 2018-09-26
Add Jenkins job for govuk-alert-tracker

deborahchua:
We have a new app `govuk-alert-tracker` which runs a script that scrapes Icinga for alerts and stores it in a Google sheet. Our [Platform Health Dashboard](https://datastudio.google.com/u/0/reporting/1bXgS9j2mgMJtifuQrVHaZ5rYwxKCns0k/page/gKOG) then extracts the data from the sheet to populate the dashboard metrics.

This commit adds a Jenkins job that runs the rake task on the first day of each month at 7am.

Trello card: [Improve alerting information on the Platform Health dashboard](https://trello.com/c/B1dtJ48q/420-improve-alerting-information-on-the-platform-health-dashboard)

Related PR setting up the Google credentials: https://github.com/alphagov/govuk-secrets/pull/451

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/scrape_icinga_alerts_for_dashboard_metrics.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/scrape_icinga_alerts_for_dashboard_metrics.yaml.erb
@@ -1,0 +1,49 @@
+---
+- scm:
+    name: scrape_icinga_alerts_for_dashboard_metrics
+    scm:
+        - git:
+            url: git@github.com:alphagov/govuk-alert-tracker.git
+            branches:
+              - master
+
+---
+- job:
+    name: scrape_icinga_alerts_for_dashboard_metrics
+    display-name: Scrape Icinga alerts for dashboard metrics
+    project-type: freestyle
+    scm:
+      - scrape_icinga_alerts_for_dashboard_metrics
+    description: "Extracts information about alerts into a Google sheet which the dashboard extracts data from"
+    logrotate:
+        numToKeep: 10
+    triggers:
+        - timed: |
+            TZ=Europe/London
+            0 7 1 * *
+    properties:
+      - build-discarder:
+          days-to-keep: 30
+          artifact-num-to-keep: 5
+    builders:
+       - shell: |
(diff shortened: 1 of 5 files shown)


deborahchua:
I'm not 100% sure I've set this up correctly so any feedback is welcome!

modules/govuk_jenkins/templates/jobs/scrape_icinga_alerts_for_dashboard_metrics.yaml.erb
```diff
@@ -0,0 +1,47 @@
...
+        - timed: |
+            TZ=Europe/London
+            0 7 1 * *
+    properties:
+      - build-discarder:
+          days-to-keep: 30
+          artifact-num-to-keep: 5
+    builders:
+       - shell: |
+          ssh deploy@$(govuk_node_list -c backend --single-node) 'cd /data/apps/govuk-alert-tracker/current; govuk_setenv govuk-alert-tracker bundle exec rake run_monthly_report'
```
thomasleese: I don't think you need to `ssh` onto a backend machine here as the application gets `git clone`'d onto the monitoring machine by Jenkins (from the `scm` section above).

Instead, I reckon you can do something like this:

```bash
cd govuk-alert-tracker
bundle install
bundle exec rake run_monthly_report
```

I'm also wondering if we need to get this Jenkins job to create a `credentials.json` file for it to use? Perhaps we can add another variable stored in `govuk-secrets` and then do something like:

```bash
echo "<%= @google_credentials %>" > credentials.json
```

deborahchua: Thanks @thomasleese! Do you mean something similar to https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/templates/google_api/credentials.json.erb

deborahchua: I think I've made all the necessary changes now @thomasleese 🤞

PR #8161, opened 2018-09-27
Remove Policy Publisher

cbaines:
Now that the policy pages are being replaced by topic pages, there
isn't any need for Policy Publisher any more.

These changes will have govuk-puppet remove the app from the machines,
and then once this has been deployed, the govuk-puppet configuration
can be removed.


tijmenb:
image

PR #8165, opened 2018-09-28
Fix aws monitoring boto3.client calls

schmie:
- The boto client seems to require an explicit stating of the AWS region in order to be happy

solo: @schmie


schmie:
Had the desired effect in integration :)

PR #8169, opened 2018-10-01
Update Ruby version in Jenkins Job

deborahchua:
The current Ruby version doesn't match the one set in `[govuk-alert-tracker](https://github.com/alphagov/govuk-alert-tracker/blob/master/Gemfile)` (completely my fault!)

Trello card: https://trello.com/c/B1dtJ48q/420-improve-alerting-information-on-the-platform-health-dashboard


deborahchua:
Ah I see, that was what I was missing! Thanks @cbaines!

PR #8175, opened 2018-10-03
Ensure the home directory for smokey exists

cbaines:
Somehow, this exists on Staging and Production, but not
Integration. Smokey doesn't really require it's home directory, but
bundler ends up logging stuff [1] that breaks reading the output file
if the directory doesn't exist. So, just create the directory to try
to get this working.

1:
`/home/smokey` is not a directory.
Bundler will use `/tmp/user/999/bundler/home/smokey' as your home directory temporarily.

### DIFF
--- a/modules/monitoring/manifests/checks/smokey.pp
+++ b/modules/monitoring/manifests/checks/smokey.pp
@@ -22,10 +22,11 @@ class monitoring::checks::smokey (
   include govuk::apps::smokey
 
   user { 'smokey':
-    ensure => present,
-    name   => 'smokey',
-    shell  => '/bin/false',
-    system => true,
+    ensure     => present,
+    name       => 'smokey',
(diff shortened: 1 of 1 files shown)


deanwilson:
Adding a `    managehome => true,`  to the smokey user resource might be a more inline way to fix this unless those permissions need to be exactly that.

cbaines:
> Adding a `managehome => true,` to the smokey user resource might be a more inline way to fix this unless those permissions need to be exactly that.

Ah, I didn't see that. However, now reading that bit of the documentation, I'm not sure this would help in this case, as the documentation suggests that the home directory will only be created if the user is also being created. At least on Integration, we've got ourselves in to the state of having the user, but no home directory, so if I'm reading the documentation correctly, I don't think `managehome => true,` would have Puppet create it.

PR #8181, opened 2018-10-08
Adds user murilodalri to integration and CI

MuriloDalRi:


MuriloDalRi:
@benthorner cool, I still don't have access to that repo. Good to know, I just used that title because it's on the getting started doc.

PR #8182, opened 2018-10-08
Disable retries for non-idempotent requests

rubenarakelyan:
This commit disables the nginx proxy setting which retries failed requests, when the HTTP request is non-idempotent. This replicates a fix made to nginx in later versions.

Trello: https://trello.com/c/l0zAThUM/521-set-nginx-proxynextupstream-off

### DIFF
--- a/modules/loadbalancer/templates/nginx_balance.conf.erb
+++ b/modules/loadbalancer/templates/nginx_balance.conf.erb
@@ -52,6 +52,12 @@ server {
   # Try next upstream if one returns an error
   proxy_next_upstream error http_500;
 
+  # Disable trying next upstream for non-idempotent requests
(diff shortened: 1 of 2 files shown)


modules/loadbalancer/templates/nginx_balance.conf.erb
```diff
@@ -52,6 +52,12 @@ server {
   # Try next upstream if one returns an error
   proxy_next_upstream error http_500;
```
thomasleese: I wonder if we should move this into an `else` block (if that's even possible in Nginx?) to avoid someone seeing that line and missing the bit below it.

rubenarakelyan: Unfortunately you can't do `else` in nginx, you just need loads of consecutive `if` statements.

PR #8186, opened 2018-10-09
[DO NOT MERGE] Correct Jenkins job credentials path

binaryberry:
[DO NOT MERGE] Failing tests.

Because that's where it is: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/templates/google_api/credentials.json.erb

https://trello.com/c/B1dtJ48q/420-improve-alerting-information-on-the-platform-health-dashboard


binaryberry:
The issue was not coming from the path being wrong, but a JSON parsing issue on the private key. I'll open another PR to correct that.

PR #8196, opened 2018-10-10
Allow restoring WAL-E backups on replica boxes

barrucadu:
Currently restoring a WAL-E backup leaves the replicas in an unhealthy state, unable to catch up with the primary, as you can see from this somewhat cryptic graph:

download

I think that we can just run the WAL-E restore script on replicas to bring them up to speed, at which point normal replication would take over: https://groups.google.com/forum/#!topic/wal-e/usseZZYxE3s

However, our puppet is currently not set up to do this.  For a node to get the restore script, it also must be doing backups.  This is an attempt to split the two responsibilities up.

Rather than splitting the class, another approach would be to pass a `do_backups` param, which is `true` by default, and use that to enable/disable all the cron and icinga stuff.

---

[Trello card](https://trello.com/c/50tQsfZE/538-postgres-replicas-dont-recover-after-restoring-wal-e-backup)

### DIFF
--- a/modules/govuk/manifests/node/s_postgresql_standby.pp
+++ b/modules/govuk/manifests/node/s_postgresql_standby.pp
@@ -10,8 +10,21 @@
 #
 class govuk::node::s_postgresql_standby (
   $primary_password,
+  $aws_access_key_id,
+  $aws_secret_access_key,
+  $s3_bucket_url,
+  $wale_private_gpg_key,
+  $wale_private_gpg_key_fingerprint,
 ) inherits govuk::node::s_postgresql_base {
   class { 'govuk_postgresql::server::standby':
     master_password => $primary_password,
+  }
+
+  govuk_postgresql::wal_e::restore { $title:
+    aws_access_key_id                => $aws_access_key_id,
+    aws_secret_access_key            => $aws_secret_access_key,
+    s3_bucket_url                    => $s3_bucket_url,
+    wale_private_gpg_key             => $wale_private_gpg_key,
+    wale_private_gpg_key_fingerprint => $wale_private_gpg_key_fingerprint,
   }
 }
--- a/modules/govuk_postgresql/manifests/wal_e/restore.pp
+++ b/modules/govuk_postgresql/manifests/wal_e/restore.pp
@@ -45,8 +45,6 @@ define govuk_postgresql::wal_e::restore (
 ) {
     include govuk_postgresql::wal_e::package
 
-    validate_re($wale_private_gpg_key_fingerprint, '^[[:alnum:]]{40}$', 'Must supply full GPG fingerprint')
-
     file { '/etc/wal-e/env.d':
       ensure => directory,
       owner  => 'postgres',
--- a/modules/govuk_postgresql/manifests/wal_e/restore.pp
+++ b/modules/govuk_postgresql/manifests/wal_e/restore.pp
@@ -1,0 +1,135 @@
+# == Define: Govuk_postgresql::Wal_e::Restore
+#
+# Allows restoring from an offsite backup of a PostgreSQL database
+# from an S3 bucket using WAL-E: https://github.com/wal-e/wal-e
+#
+# === Parameters:
+#
+# [*aws_access_key_id*]
+#   The AWS access ID for the user that is allowed to
+#   access the S3 bucket.
+#
+# [*aws_secret_access_key*]
+#   The AWS secret access key for the user that is allowed
+#   to access the S3 bucket.
+#
+# [*s3_bucket_url*]
+#   The unique address of the S3 bucket, in the format of:
+#   s3://bucketaddress/directory/foo
+#
+# [*aws_region*]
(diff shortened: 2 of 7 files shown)


modules/govuk_postgresql/manifests/wal_e/restore.pp
```diff
@@ -0,0 +1,135 @@
...
+  $aws_secret_access_key,
+  $s3_bucket_url,
+  $aws_region = 'eu-west-1',
+  $wale_private_gpg_key = undef,
+  $wale_private_gpg_key_fingerprint = undef,
+  $db_dir = $postgresql::params::datadir,
+) {
+    include govuk_postgresql::wal_e::package
+
+    validate_re($wale_private_gpg_key_fingerprint, '^[[:alnum:]]{40}$', 'Must supply full GPG fingerprint')
```
barrucadu: I think this should go away, as the same check happens later too.

modules/govuk/manifests/node/s_postgresql_standby.pp
```diff
@@ -14,4 +14,12 @@
   class { 'govuk_postgresql::server::standby':
     master_password => $primary_password,
   }
+
+  govuk_postgresql::wal_e::restore { $title:
```
barrucadu: The common hieradata should probably be moved to the `s_postgresql_base` class

hieradata/vagrant_credentials.yaml
```diff
@@ -157,11 +157,11 @@ govuk::node::s_transition_postgresql_slave::master_password: "%{hiera('govuk::no
 govuk::node::s_transition_postgresql_primary::standby_password: 'eiGhuSaiNgai0OhZai1aeyahrei8Thiikie7amieci'
 govuk::node::s_transition_postgresql_standby::primary_password: "%{hiera('govuk::node::s_transition_postgresql_primary::standby_password')}"
 
-govuk::node::s_postgresql_primary::aws_access_key_id: 'foo'
-govuk::node::s_postgresql_primary::aws_secret_access_key: 'bar'
-govuk::node::s_postgresql_primary::s3_bucket_url: 'bucket'
-govuk::node::s_postgresql_primary::wale_private_gpg_key: 'key'
-govuk::node::s_postgresql_primary::wale_private_gpg_key_fingerprint: 'AIFAED1POOX9MOA3THI9CIE0IEQU7ULAIWEI4SOF'
+govuk::node::s_postgresql_base::aws_access_key_id: 'foo'
```
barrucadu: Moving these hiera keys means secrets also need changing.

modules/govuk/manifests/node/s_postgresql_standby.pp
```diff
@@ -10,8 +10,21 @@
 #
 class govuk::node::s_postgresql_standby (
   $primary_password,
+  $aws_access_key_id,
+  $aws_secret_access_key,
+  $s3_bucket_url,
+  $wale_private_gpg_key,
```
thomasleese: I think we should name these `wal_e` for consistency with the Puppet class.

modules/govuk_postgresql/manifests/wal_e/restore.pp
```diff
@@ -0,0 +1,133 @@
...
+#   so there should be no impact, and the default backups up
+#   everything.
+#   Default: /var/lib/postgresql/9.3/main
+#
+define govuk_postgresql::wal_e::restore (
+  $aws_access_key_id,
+  $aws_secret_access_key,
+  $s3_bucket_url,
+  $aws_region = 'eu-west-1',
+  $wale_private_gpg_key = undef,
```
thomasleese: I suspect we don't need the `wale` prefix on these two variables since we're already in the `wal_e` class.

modules/govuk/manifests/node/s_postgresql_standby.pp
barrucadu: They'd be consistent with the Puppet class, but inconsistent with the `s_postgresql_primary` class.  Unless that were to change too, which would require changing secrets.

modules/govuk_postgresql/manifests/wal_e/restore.pp
barrucadu: Same reasoning as above.  I took the current names to be consistent with the classes which exist, and to avoid needing to also change secrets to make everything consistent.

modules/govuk/manifests/node/s_postgresql_standby.pp
thomasleese: Hmm yeah, that's unfortunate. Probably better in this case to be consistent with the `s_postgresql_primary` class.

modules/govuk_postgresql/manifests/wal_e/restore.pp
thomasleese: Sure, that sounds reasonable!

barrucadu:
Tested on staging, this worked quite nicely.

PR #8198, opened 2018-10-12
[WIP] Add cadvisor running code

deanwilson:
As we add more docker containers to our stack we need ways to monitor
and observe their running. Cadvisor provides a quick to deploy,
low resource overhead initial pass at this.

### DIFF
--- a/modules/govuk_containers/manifests/cadvisor.pp
+++ b/modules/govuk_containers/manifests/cadvisor.pp
@@ -1,0 +1,45 @@
+# == Class: govuk_containers::cadvisor
+#
+# Install and run a dockerised cadvisor container.
+# This provides metrics and graphs for the docker
+# daemon itself and other running containers.
+#
+# === Parameters
+#
+# [*image_name*]
+#   Docker image name to use for the container.
+#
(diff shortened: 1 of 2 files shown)


th31nitiate:
@deanwilson Is this really what we are looking to for monitoring our containers ?

I believe prometheus will be a more suitable robust solution for this.

deanwilson:
cAdvisor is one of the datasources prometheus will need to scrape. In the meantime, as we don't actually have prometheus, it will provide visibility somewhere we have none and require no rework once prometheus is available to us.

th31nitiate:
Ok, I have reviewed the following: https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md

This means sounds interesting and should be workable solutions. I have just never been a big fan of it based on the way it stores its data in Kubernetes. I think for ad-hoc solutions is cool where we deploy as daemon on node running docker.

I am not sure how or where we may deploy the export when moving to Kubernetes. 

How are we going to centralise these metrics, are they going to graphite ?

deanwilson:
I'm closing this as it's not part of this quarters work.

PR #8212, opened 2018-10-19
cast password parameter to string

fredericfran-gds:
need to cast password parameter to string because puppet interprets the password parameter as a non-string type for certain values

Co-Authored-By: Conor Glynn


ronocg:
Closing this, its a duplicate

PR #8218, opened 2018-10-19
Add HTTP/2 preconnect header for assets

rubenarakelyan:
This commit adds a `Link` HTTP header pointing at the relevant assets subdomain for the environment.

This header is used by browsers that support HTTP/2 to indicate that they should immediately initiate a TCP connection to this subdomain in anticipation of downloading assets from it.

This allows the connection initiation overheads to happen while the HTML is being downloaded, so that referenced assets can then be downloaded immediately rather than waiting for the connection to be initiated.

### DIFF
--- a/modules/router/templates/router_include.conf.erb
+++ b/modules/router/templates/router_include.conf.erb
@@ -63,6 +63,8 @@ location ~ ^/contact/govuk/(|service-feedback|problem_reports|foi|page_improveme
 location / {
   <%= scope.function_template(['router/fragments/_basic_auth.erb']) -%>
 
+  add_header Link "<https://assets.<%= @app_domain %>>; rel=preconnect";
+
   # HTML verification for DWP YouTube channel
   location = /dla-ending/google6db9c061ce178960.html {
     add_header Content-Type text/html;


modules/router/templates/router_include.conf.erb
```diff
@@ -63,6 +63,8 @@ location ~ ^/contact/govuk/(|service-feedback|problem_reports|foi|page_improveme
 location / {
   <%= scope.function_template(['router/fragments/_basic_auth.erb']) -%>
 
+  add_header Link "<https://assets.<%= @app_domain %>>; rel=preconnect";
```
thomasleese: I wonder if this makes more sense to go in the meta tags in the HTML of pages because it might get more visibility there, people might not think to look here for that?

thomasleese: I wonder if this makes more sense to go in the meta tags in the HTML of pages because it might get more visibility there, people might not think to look here for that?

Nooshu:
Adding some context for this change here. Switching to HTTP/2 has revealed an issue with the domain sharding we are doing when using the assets domain. Where HTTP/1.1 had 6 connections in parallel, HTTP/2 now only has 1. So when the DNS / connection / SSL for assets hasn't resolved on this single connection, it becomes a real bottleneck. This can be seen in red in the waterfall below:

dns-lookup

This delay can also be seen in visual progression graphs like the ones below:

screen shot 2018-10-22 at 13 33 12
screen shot 2018-10-22 at 13 33 37
screen shot 2018-10-22 at 13 33 49

In theory the use of the [preconnect resource hint](https://www.w3.org/TR/resource-hints/#preconnect) should remove some if not all of this initial lag.

PR #8220, opened 2018-10-22
Fix CKAN cron jobs

brucebolt:
These cron jobs run regularly to: start automatic scheduled harvesters, mark finished harvesters as finished, and clear up the harvesting logs.

Previously they were not running on integration, with an error `/bin/sh: 1: govuk_setenv: not found`.  This fixes that error and also makes the start/stop job execute more frequently (resulting in jobs getting marked as 'Finished' sooner).

Trello card: https://trello.com/c/MTf9rBKK/3-get-harvesting-working-on-integration


elliotcm:
How long do the harvesters take to run?  Are we likely to see similar overlap issues to the ones we saw with sidekiq?

brucebolt:
They seem to take a few minutes, for a 90 dataset source.  New jobs will not get started if a harvest source has a running job (determined using the database), so there won't be any overlap issues.

PR #8233, opened 2018-10-24
Remove configuration for Policy Publisher

cbaines:
We're no longer running this app anymore, as the policy pages on
GOV.UK have been replaced by the new Topic pages.

Following on from [1], when Policy Publisher was removed as an app,
this commit removes the remaining configuration in govuk-puppet.

1: 0fb32e690ca6057de87e4812bbcf7d365f1c5e96
   https://github.com/alphagov/govuk-puppet/pull/8161


th31nitiate:
cool, this looks clean.

PR #8237, opened 2018-10-25
Add on/off switch for govuk_env_sync

schmie:
- So far our hiera config only allowed to add things, rather than to
  remove them

- We want to be able to turn off data sync to preserve the state
  of integration and/or in some cases, e.g. autumn budget prep

solo: @schmie

### DIFF
--- a/modules/govuk_env_sync/manifests/task.pp
+++ b/modules/govuk_env_sync/manifests/task.pp
@@ -45,13 +45,14 @@ define govuk_env_sync::task(
   $database,
   $url,
   $path,
+  $ensure = 'absent',
 ) {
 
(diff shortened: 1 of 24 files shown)


modules/govuk_env_sync/manifests/task.pp
```diff
@@ -45,13 +45,14 @@
   $database,
   $url,
   $path,
+  $ensure = 'absent',
```
barrucadu: Why did you go for a default of `absent`?

schmie: I want it to break in case it gets ill configured/ I miss the configuration update. Mainly with focus on having to turn off integration sync tonight..

PR #8238, opened 2018-10-26
Use Postgresql module devel for mapit dependencies 

th31nitiate:
This add's the postgres package to mapit nodes. I have added this to the mapit deployment puppet manifest. This seemed to be the simplest method since it will only run on nodes with the mapit role enabled.

### DIFF
--- a/modules/govuk/manifests/apps/mapit.pp
+++ b/modules/govuk/manifests/apps/mapit.pp
@@ -56,6 +56,9 @@ class govuk::apps::mapit (
       ensure => present,
     }
 
+    # Install postgrest developments dependencies
+    include postgresql::lib::devel
+
     include gdal::repo
     package { 'gdal':
       ensure  => $gdal_version,


modules/govuk/manifests/apps/mapit.pp
```diff
@@ -52,6 +52,7 @@
         # https://docs.djangoproject.com/en/1.8/ref/contrib/gis/install/geolibs/
         'binutils',
         'libproj-dev',
+        'postgresql-server-dev-9.3',
```
cbaines: Have you considered doing this through the postgresql module?

cbaines:
It would be good if you could put your motivation for making this change in the commit message. That way others can better understand the "why", not just the what from the diff.

modules/govuk/manifests/apps/mapit.pp
th31nitiate: This is in essence just installing the package. There is no need to use the module since we want to keep dependencies at a minimum.

th31nitiate:
@cbaines This is related to migration work on Gov.UK. This PR is mainly about deploying the mapit app. There was package missing which is the one that has been added. The application will not run without this dependency.

cbaines:
> There was package missing which is the one that has been added. The application will not run without this dependency.

Given this relates to the backend of PostgreSQL, I'd like to know more about why this is. If I look at the current mapit machines in production, they don't have this installed.

modules/govuk/manifests/apps/mapit.pp
cbaines: One of the reasons I'm mentioning the module is that I'm currently looking at changing the version of PostgreSQL in the development-vm [1]. If PostgreSQL packages are installed in more places, without using a common version, this will potentially make managing it more complex.

1: https://github.com/alphagov/govuk-puppet/pull/8240

cbaines:
> As you suggested we switched the initial package to just include the lib. We have now also change the way we install to use the Postgres repository.

Looking good :) I'd suggest squashing the commits together, as I don't think there's any value in having two commits rather than one.

modules/govuk/manifests/apps/mapit.pp
```diff
@@ -56,6 +56,9 @@
       ensure => present,
     }
 
+    # Install postgrest developments dependencies
```
cbaines: At least for me, I don't see this adding any information over what the line of code below says. Maybe put something like "Ensure the libpq-dev package is installed", as that gives some more specific information on what the code should do?

th31nitiate:
Fair enough. It makes sense. I am not too sure why I left that on there :). 

Commit squashed.

PR #8245, opened 2018-10-29
Remove mapit from the /etc/hosts on Staging Carrenza

rtrinque:
We're migrating the mapit service to AWS, we have a new DNS
record to point to the AWS ALB and we need to disable the local
resolving on Staging Carrenza machines for it to take effect.


fredericfran-gds:
looks good!

PR #8247, opened 2018-10-29
Point to draft content store when request is draft

AlanGabbianelli:
This is to enable viewing content store on draft stack.
- Add a draft public api class and pointed it to the draft content store
- Add that class to the draft cache machines in all environments
- Revert changes from https://github.com/alphagov/govuk-puppet/pull/8234

Trello card: https://trello.com/c/JDiLJ8h1/570-enable-viewing-content-store-on-draft-stack

### DIFF
--- a/modules/govuk/manifests/apps/draft_publicapi.pp
+++ b/modules/govuk/manifests/apps/draft_publicapi.pp
@@ -1,12 +1,12 @@
 # FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
-class govuk::apps::publicapi (
+class govuk::apps::draft_publicapi (
   $backdrop_protocol = 'https',
   $backdrop_host = 'www.performance.service.gov.uk',
   $privateapi_ssl = true,
 ) {
 
   $app_domain = hiera('app_domain')
-  $app_name = 'publicapi'
+  $app_name = 'draft-publicapi'
 
   if $::aws_migration {
     # In AWS the upstream should use the internal domain. The nginx server_name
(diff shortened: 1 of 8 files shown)


modules/govuk/manifests/apps/draft_publicapi.pp
```diff
@@ -0,0 +1,50 @@
+# FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
+class govuk::apps::draft_publicapi (
```
thomasleese: I wonder if there is a way we can avoid the duplication between this and `publicapi`.

Could we maybe have a common `govuk::apps::publicapi` class and do something like:

```
govuk::apps::publicapi { 'draft-publicapi':
  backdrop_protocol => $backdrop_protocol,
  backdrop_host => $backdrop_host,
  ...
}
```

Similar to how we've defined the parts of `govuk::app`, for example [`govuk::app::package`](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/app/package.pp).

rubenarakelyan: How would that work with including the apps in the YAML files?

thomasleese: I think we could have two classes `draft_publicapi` and `live_publicapi` which then just use the `publicapi` definition.

```puppet
class govuk::apps::draft_publicapi (...) {
  govuk::app::publicapi { 'draft-publicapi':
    backdrop_protocol => $backdrop_protocol,
    backdrop_host => $backdrop_host,
    ...
  }
}

class govuk::apps::live_publicapi (...) {
  govuk::app::publicapi { 'live-publicapi':
    backdrop_protocol => $backdrop_protocol,
    backdrop_host => $backdrop_host,
    ...
  }
}
```

Maybe it's too much refactoring for this PR, I'm just thinking that if we need to change something in the live one it might get missed in the draft one.

PR #8254, opened 2018-10-31
Remove gtm_auth and gtm_preview for production env

KyleMacPherson:
### What?
This PR removes the gtm auth and gtm preview env variables from content data admin's production environment

### Why?
The auth and preview env variables are not needed in production. . The component will default to using a nil value which seems to be what is required in the production environment if it is not passed these values.


KyleMacPherson:
[(1) Embed Google Tag Manager code in header and body of each page of the content-data app](https://trello.com/c/ldax4jWq/631-1-embed-google-tag-manager-code-in-header-and-body-of-each-page-of-the-content-data-app)

PR #8256, opened 2018-10-31
This adds the node_class list to the staging yaml file used in Carren…

jstandring-gds:
…za, with "transition" removed. This is done in order to migrate the backend app to AWS.
With this code in place transition will not be installed in Staging (Carrenza),
so when it's stopped manually it will not return unexpectedly.

The idea being to stop access to transition in Carrenza then switch DNS so the
new transition in AWS is used.


jstandring-gds:
No longer relevant.

PR #8258, opened 2018-10-31
Revert "Disable retries for non-idempotent requests"

rubenarakelyan:
Reverts alphagov/govuk-puppet#8182

Looks like nginx doesn't like this configuration so reverting until we can take a better look at it. This will prevent issues if nginx tries to restart and fails because of this config.

### DIFF
--- a/modules/loadbalancer/templates/nginx_balance.conf.erb
+++ b/modules/loadbalancer/templates/nginx_balance.conf.erb
@@ -52,12 +52,6 @@ server {
   # Try next upstream if one returns an error
   proxy_next_upstream error http_500;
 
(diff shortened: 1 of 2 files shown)


barrucadu:
So has the problem of retrying requests still existed all this time?

rubenarakelyan:
Yes apparently so. If the nginx config test fails then it just doesn't reload the config. Seems like this is what has happened, so as far as Puppet is concerned, it all worked fine.

barrucadu:
Any idea why we've not had a bunch of zendesk tickets about lock version states?

rubenarakelyan:
No idea at all. I'm going to take a deeper look at this once we get the immediate issue out of the way.

PR #8284, opened 2018-11-05
change the app_domain_external also for imminence in AWS staging

fredericfran-gds:
# CONTEXT

In the govuk migration, there is a need to set the `GOVUK_APP_DOMAIN_EXTERNAL` variable only for the imminence app on the backend machines in AWS staging, not the other APPs. 

# DECISIONS

1. modify the imminence app for if `$app_domain` is defined, then then the `GOVUK_APP_DOMAIN_EXTERNAL` is set to the `$app_domain`


cbaines:
I'd be interested to know what the need is for this, and it would have been good to have this information in the commit message as well?

PR #8289, opened 2018-11-05
pip depends in monitoring

th31nitiate:
This commit adds pip dependencies monitoring instances. This is to prevent errors in Nagios checks.


schmie:
Please test your branch in AWS and make sure all dependencies needed by the monitoring get installed by Puppet as part of this PR as well

PR #8296, opened 2018-11-06
Revert "Proxy CSS and JS requests from main domain to assets domain"

rubenarakelyan:
Reverts alphagov/govuk-puppet#8250

This didn't work as expected, so we're removing it.


rubenarakelyan:
`assets-origin` itself proxies requests to various apps, so this "proxy to a proxy to an app" approach didn't work correctly for all apps and it would have been complicated to make it work. Instead, we're going with using HTTP/2 server push to proactively push CSS and JS assets to browsers in the hope this will give us a similar speed-up. This will work now we've upgraded nginx to support HTTP/2.

PR #8297, opened 2018-11-06
Enable HTTP/2 and server push

rubenarakelyan:
This commit enables HTTP/2 and server push for www.gov.uk.

This will allow server push link headers set by apps to be picked up by nginx and the relevant files proactively pushed to clients. These will in turn be cached by Fastly.

In AWS, since we terminate HTTPS at the load balancers, we enable H2C (HTTP/2 cleartext) in nginx so it can continue to pick up the server push headers. Clients will still connect using HTTPS to the load balancers.

Trello: https://trello.com/c/0Ek526Jn/602-enable-http-2-and-server-push

### DIFF
--- a/modules/router/templates/router_include.conf.erb
+++ b/modules/router/templates/router_include.conf.erb
@@ -63,7 +63,11 @@ location ~ ^/contact/govuk/(|service-feedback|problem_reports|foi|page_improveme
 location / {
   <%= scope.function_template(['router/fragments/_basic_auth.erb']) -%>
 
+  # Enable asset domain preconnect for HTTP/2-capable clients
   add_header Link "<https://assets.<%= @app_domain %>>; rel=preconnect; crossorigin";
+
+  # Enable server push for HTTP/2-capable clients
+  http2_push_preload on;
 
   # HTML verification for DWP YouTube channel
   location = /dla-ending/google6db9c061ce178960.html {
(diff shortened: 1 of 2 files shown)


modules/router/templates/router_include.conf.erb
```diff
@@ -63,7 +63,9 @@ location ~ ^/contact/govuk/(|service-feedback|problem_reports|foi|page_improveme
 location / {
   <%= scope.function_template(['router/fragments/_basic_auth.erb']) -%>
 
+  # Enable asset domain preconnect and server push for HTTP/2-capable clients
   add_header Link "<https://assets.<%= @app_domain %>>; rel=preconnect; crossorigin";
+  http2_push_preload on;
```
barrucadu: The nginx docs say:
> Enables automatic conversion of preload links specified in the “Link” response header fields into push requests.
So does the line above need `rel=preconnect` changing to `rel=preload`?

rubenarakelyan: They're actually separate things - the first line hints to the browser to set up the connection to assets even before it has worked out it needs to request anything from there. The second line proactively pushes assets to the browser that have been marked by individual apps.

barrucadu: Maybe add a comment, or even just a blank line, to make it clearer that these lines deal with different things.

rubenarakelyan: Separated with comments

PR #8312, opened 2018-11-07
Remove imminence from backend load balancers

rtrinque:
Imminence is now running on AWS


ronocg:
I guess this is ok to do as these load balancers do Not exist in AWS where imminence now lives for both staging and production?

PR #8315, opened 2018-11-08
Edit Icinga checks 

th31nitiate:
- Remove unmonitored servers from checks in prod and staging


th31nitiate:
Thanks and I will take note of that

PR #8317, opened 2018-11-08
Adding the staging entry for db pull for local-links-manager

jstandring-gds:
This entry creates the cfg file used for completing data_sync for the
local-links-manager. This is required for the migration of this app.


rtrinque:
Why do we need that since we already have pull_local-links-manager_production_daily defined?

th31nitiate:
@rtrinque We need this in order to pull the data into the service. The other one I believe does a push while this will only be pulling.

rtrinque:
"pull_link_checker_api_production_daily":   is a pull job that is defined just above this new job, and that does the same thing.

PR #8322, opened 2018-11-08
Migrate local links manager to aws

jstandring-gds:
During the migration we need to test local-links-manager in AWS by having it
talk to the AWS signon service. This uses a different app_domain to Carrenza so
not only needs a new value but a conditional statement in order to determine
what VAR and VALUE is used. This adds the puppet code to accomodate for this


jstandring-gds:
Test this please.

cbaines:
> Test this please.

Is that directed at Jenkins?

PR #8329, opened 2018-11-09
Schedule for Etl:Master process now 07:00

peteglondon:
We have been having discrepancies in the pageviews and unique pageviews
between GA and the data warehouse. We think this is because GA has not
finished processing the data by 1am when we run the Etl:Master process.

This commit re-schedules it to run at 7am.


peteglondon:
[(5) Data quality bug - Pageviews and unique pageviews don't match Google Analytics](https://trello.com/c/eVYqQ5Rz/818-5-data-quality-bug-pageviews-and-unique-pageviews-dont-match-google-analytics)

PR #8330, opened 2018-11-09
Restore the smokey_deploy job in AWS Staging and AWS Production

cbaines:
This job was removed in [1], but the smokey-loop serivce is still
running on the monitoring boxes in both environments.

While there may be some problems with smokey in AWS environments, it's
still useful to be able to run the deploy Jenkins job, so that at
least an up to date version is running.

1: ce9f0051dcfd8f43d6b7324cdd72f0b353ad7d7f


cbaines:
Is this fine with you @fredericfran-gds , @schmie ?

fredericfran-gds:
I'm fine with that. May I know what is the `smokey-loop` service?

fredericfran-gds:
Can we find when the fix smokey tests will work for AWS?

cbaines:
> I'm fine with that. May I know what is the `smokey-loop` service?

It's the name of the upstart service that runs the Smokey tests, in a tight loop, on the monitoring machine [1].

1: https://github.com/alphagov/govuk-puppet/blob/master/modules/monitoring/manifests/checks/smokey.pp

> Can we find when the fix smokey tests will work for AWS?

I don't know much about Smokey in AWS, but I think the first step forward is to restore the deploy job, and get AWS running the latest version of Smokey.

PR #8338, opened 2018-11-12
Reduce monitoring disk warning threshold to 12%

barrucadu:
This was reduced to 15% in #8314, to handle the increased disk usage
due to the CDN logs.  However, we're still frequently going past that
threshold towards the end of each log rotation period (hourly).  I
haven't seen it drop below 13% free, so set 12% as the threshold.

This does mean that there's only 2 percentage points between the
warning and the critical thresholds, which seems not great though.


barrucadu:
Note: this is also pushed to the `msw/12pc` branch, as the github jenkins plugin we're using [doesn't urlencode branch names](https://ci.integration.publishing.service.gov.uk/job/govuk-puppet/job/msw%252F12%2525/1/console).  So that branch will also need deleting when we're happy with this.

PR #8347, opened 2018-11-13
Add AWS production and staging ssh config

barrucadu:
This is based on the integration ssh config, and lets you do things
like `ssh 10.12.4.142.staging-aws` to get to (at the moment) a
draft-content-store machine.


barrucadu:
@cbaines I started with `aws-staging` / `aws-production`, but decided to check `govukcli list-contexts` and see what that did.

PR #8348, opened 2018-11-13
Add Whitehall to Rummager dependencies

bilbof:
This is a temporary dependency. Rummager now needs Whitehall, as Whitehall is the source
of world locations. Once these have been extracted from Whitehall, we can remove this
depedency.

Related pull request in Rummager:
https://github.com/alphagov/rummager/pull/1318

### DIFF
--- a/development-vm/Pinfile
+++ b/development-vm/Pinfile
@@ -62,7 +62,7 @@ process :'publishing-components'
 process :release
 process :router
 process :'router-api'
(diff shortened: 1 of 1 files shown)


sihugh:
I was wondering if it would be possible to point rummager at the live api in development by default?  I know that people tend to have trouble spinning  up whitehall on the VM so it might be worth trying to avoid this unless absolutely necessary?

bilbof:
Yes that would be possible. I thought it would be nicer to default to the local Whitehall, as the alternative would make offline development a bit harder. This way one can use `--live` locally if Whitehall isn't up and/or run `-w whitehall` to not start Whitehall.

PR #8350, opened 2018-11-13
Remove Tatiana from mailmap

barrucadu:
She has left GDS.


cbaines:
@barrucadu we're not removing Tatiana from the git repository history, so I think the mailmap entries are still useful?

barrucadu:
That's a good point.  I'll revert this.

PR #8354, opened 2018-11-14
 Enable alerts for unprocessed messages in email-alert-service queue

barrucadu:
Looking back over the 2018 data, the queues sizes are always below 3,
except when the service failed, and there was a sudden large build up.
I'm confident that these very low thresholds won't be too noisy, and
will usefully indicate when the service is up but failing to process
messages.

---

**Production:**
screen shot 2018-11-14 at 15 08 13

**Staging:**
screen shot 2018-11-14 at 15 08 54

---

[Trello card](https://trello.com/c/d3ouXvUN/626-add-alerting-around-high-rabbitmq-queues-for-email-alert-service)

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -449,6 +449,8 @@ govuk::apps::email_alert_service::rabbitmq_hosts:
   - rabbitmq-2.backend
   - rabbitmq-3.backend
 govuk::apps::email_alert_service::redis_host: "%{hiera('sidekiq_host')}"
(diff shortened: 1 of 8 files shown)


barrucadu:
Trello card says to have icinga first try restarting email-alert-service.  That might be tricky, as this monitoring lives on the rabbitmq boxes.

barrucadu:
Just had another look, I think this fixup is an improvement as I don't think the previous way would let you monitor different queues with different thresholds. cc @rubenarakelyan

PR #8356, opened 2018-11-15
Remove Tatiana from mailchimp

th31nitiate:
- This PR removes Tatiana entries from the mail chimp file


th31nitiate:
@cbaines Thank you for clarifying this. I am working on a ticket which requested this change be carried. I believe it may already have been actioned.

PR #8357, opened 2018-11-15
Remove hieradata for ckan in Carrenza

barrucadu:
This app has been migrated to AWS and the VMs deleted.


barrucadu:
The tests try to build `govuk::node::s_ckan` with the normal hieradata, which is now failing.  I'll leave this for when the big deletion of Carrenza hieradata happens.

PR #8359, opened 2018-11-16
add publishing_api entries to /etc/hosts to AWS Staging Backends

fredericfran-gds:
# Context

During the AWS migration, the `link-checker-api` is being migrated to AWS without migrating the `publishing-api` which still lives in Carrenza. For the AWS `link-checker-api` migration, the AWS backend machines (which hosts the `link-checker-api` application) need to retrieve the IP of the Carrenza `publishing-api` machines. Hence, the {ip, fqdn} entry needs to be added in '/etc/hosts' file of the AWS backend machines.

# Decision

For AWS staging environment:
1. Add the `hosts::migration` class to the `backend` node class
2. Configure the AWS hiera data to include {ip, fqdn} entry

### DIFF
--- a/hieradata_aws/class/staging/backend.yaml
+++ b/hieradata_aws/class/staging/backend.yaml
@@ -1,0 +1,7 @@
+---
+
+hosts::migration::hosts:
+  publishing-api-1:
+    ip: '10.2.3.45'
+  publishing-api-2:
+    ip: '10.2.3.46'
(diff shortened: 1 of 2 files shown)


hieradata_aws/class/staging/backend.yaml
```diff
@@ -0,0 +1,7 @@
+---
+
+hosts::migration::hosts:
+  publishing-api-1:
+    ip: '10.2.3.45'
```
jstandring-gds: I think this needs to be 10.3.3.x. Should it also include `publishing-api-3`? If so, that's 10.3.3.47.

fredericfran-gds: yes, we are checking now, maybe the list needs to be expanded

fredericfran-gds:
closing this PR for now since we need to investigate further

Issue #1725, opened 2015-12-16
[for review] Migrate http credentials extdata


alexmuller:
This pull request was [originally opened on GitHub Enterprise](https://github.gds/gds/puppet/pull/1725) on Thursday 01 May 2014 at 16:25 GMT before GOV.UK's Puppet code was migrated to a public GitHub repository.

PR #8361, opened 2018-11-16
Trello-1728: mongo-api cluster

th31nitiate:
### What ###
We need to create a mongo cluster (mongo_api) that will be used for testing addition and removal on mongo cluster nodes which will be carried out in card #1725

The mongo cluster will be built using the same module as existing mongo
module: mongodb::server

### Why ###
If our tests are successful in #1725 we will be able to have a no-downtime migration of the mongo databases.
It is also envisaged that the mongo cluster will be split up, in other
words the content-store will use mongo-api cluster (that we are
currently building here) and immenence will continue to use the existing
mongo cluster.

There is hieradata being added to the carrenza side as part of this
commit, that is to satisfy the spec tests, once migrations are fully
completed this hieradata will be collapsed into one, so it is
temporary.

Pair: @th31nitiate @ronocg

### DIFF
--- a/hieradata/class/mongo_api.yaml
+++ b/hieradata/class/mongo_api.yaml
@@ -1,0 +1,10 @@
+---
+
+# Temporarily placed to in carrenza hieradata to satisfy the spec tests
+# The monog-api-* VMs do not exist in Carrenza.
+# This will be collapsed down once the migrations to AWS are complete.
+
+mongodb::server::replicaset_members:
+  'mongo-api-1':
+  'mongo-api-2':
+  'mongo-api-3':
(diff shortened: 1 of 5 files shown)


fredericfran-gds:
The reason why this code is failing in Jenkins originally is:
1. the node definition`s_mongo_api` was added to `modules/govuk/manifests/node/s_mongo_api.pp`
2. in Jenkins, the nodes are compiled to check if they can: [here](https://github.com/alphagov/govuk-puppet/blob/362cd2610acb36d3b8d9cc83740916a7121d93fd/modules/govuk/spec/classes/govuk_nodes_spec_optional.rb#L30)
3. My hypothesis is that the test in 2) is taking the hieradata from the Carrenza side which do not contain the `mongodb::server::replicaset_members:` parameter and therefore the node compilation test fails.
4. 2 options:
     1. add the hieradata in the carrenza side, might be awkward because this node should be aws only
     2. modify the node definition to run only when `aws_migration` fact is true but then the code for node compilation test should include aws-only node compilation handling.
     3. create a aws-only node list in the testing and then branch out the tests in the node compilation test

PR #8363, opened 2018-11-19
Fix crawlers to crawl their own environment

benthorner:
https://trello.com/c/IDBl5Xna/349-fix-govuk-crawler-on-staging

Previously the crawlers for each environment were all crawling
www.gov.uk, which has recently (for some reason) generated rate limit
alerts, as the limit bypass tokens aren't shared between environments.

This changes the crawler config for integration and staging (and
similarly for AWS) to crawl origin for each environment. We've chosen to
crawl the origin instead of the CDN in these environments because:

   * The Carenza machines map www -> www-origin anyway
   * The AWS mirror machines can't access the CDN
   * The load impact of the crawler isn't such an issue

Ideally we should crawl the CDN on all environments, using a common URL
structure (compare www.gov.uk vs www.staging.publishing...), but this
isn't essential, although we should still aim to do this for AWS.

### DIFF
--- a/hieradata/production.yaml
+++ b/hieradata/production.yaml
@@ -203,6 +203,7 @@ govuk_postgresql::backup::auto_postgresql_backup_minute: 0
 
 govuk_crawler::seed_enable: true
 govuk_crawler::sync_enable: true
+govuk_crawler::site_root: 'https://www.gov.uk'
 
 govuk_crawler::targets:
   - 'mirror-rsync@mirror0.mirror.provider1.production.govuk.service.gov.uk'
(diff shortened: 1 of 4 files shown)


hieradata/production.yaml
```diff
@@ -216,6 +217,11 @@ govuk_crawler::ssh_keys:
     key: 'AAAAB3NzaC1yc2EAAAADAQABAAABAQDHi69quH4i1jY4uVLuIYOoUi7yc+6/fJjDKJciWIIgyH1rCSGtTXrfcJQHQKxKEyndQ+LoUs8krImD9CtzQTvoqdXYKu+XgpjqiZ5HfQhx55ZoMCXnx9vqCwGtx2LwH4PaBZRvsNJbbY+sG3W13eXkgwQnIcI2FgGpDDJQ9lfxSNzt6jMJQuvAhhDrKtZEsA56AznPIRd1mIHmuZ6gILBzJmDw5EIpqQhSb4+NrNTNnGxHyFkZISTMG0mx9lRtlyS1SMZVoPJQbShSEScKMXRlacrpkiNOTvrGDrhsLjTUtT5g1bo4XMFRubReAwFCeTZI0Qv1Bvpzj8V9c8B4+p6b'
     type: 'ssh-rsa'
 
+govuk::apps::govuk_crawler_worker::root_urls:
+  - 'https://assets.digital.cabinet-office.gov.uk/'
+  - 'https://assets.publishing.service.gov.uk/'
```
andrewgarner: How about using the app domain here to remove duplication?

```suggestion
  - "https://assets.%{hiera('app_domain')}"
```

benthorner: Could do, but I think I'd rather emphasise it's hard-coded. Also, we can't do it for all of them consistently :-|.

andrewgarner: `%{hiera('app_domain')}` is referenced consistency throughout `production.yaml`. If we were to change the configure app domain we would want that change to apply here too.

benthorner: Done.

PR #8366, opened 2018-11-20
add carrenza hosts details on AWS staging backend machines

fredericfran-gds:
# Context

During the AWS migration, the `link-checker-api` is being migrated to AWS without migrating the other services which still live in Carrenza. For the AWS `link-checker-api` migration, the AWS backend machines (which hosts the link-checker-api application) need to retrieve the IP of the Carrenza services machines. Hence, the {ip, fqdn} entry needs to be added in '/etc/hosts' file of the AWS backend machines.

# Decision
For AWS staging environment:
1. Add the hosts::migration class to the backend node class
2. Configure the AWS hiera data to include {ip, fqdn} entry

Co-Authored-By: rtrinque <rtrinque@users.noreply.github.com>


th31nitiate:
I am pretty much happy with this. I cannot be sure as to the validity of the hosts considering, the fact, I fear that if some hosts are required within AWS and the FQDN is added to the hosts then we may get an outage. I base this on the fact that Linux will check /etc/hosts first. 

I would recommend pinging the nodes from AWS to Carenza randomly just to ensure there is a path to all nodes.

PR #8377, opened 2018-11-27
Add alert app domain management subdomain to /etc/hosts

thomasleese:
Trying to access Icinga alerts from Jenkins leads to connection errors because there are no appropriate hosts defined which connect to the right server. Using the external domain right now doesn't map to an internal IP address.

We can't use alert.cluster here because the HTTPS server is checking that the domain name matches the server name in the certificate.

[Trello Card](https://trello.com/c/serdtvuZ/635-improve-alerting-information-on-the-platform-health-dashboard)

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1561,6 +1561,7 @@ hosts::production::management::hosts:
     legacy_aliases:
       - 'monitoring'
       - "grafana.%{hiera('app_domain')}"
+      - "alert.%{hiera('app_domain')}"
     service_aliases:
       - 'alert'
       - 'monitoring'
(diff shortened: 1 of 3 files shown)


hieradata/common.yaml
```diff
@@ -1561,6 +1561,7 @@ hosts::production::management::hosts:
     legacy_aliases:
       - 'monitoring'
       - "grafana.%{hiera('app_domain')}"
+      - "alert.%{hiera('app_domain')}"
```
barrucadu: Adding a new "legacy" alias seems a bit not-nice, but the other solution is disabling certificate veriication... this is probably the lesser evil.

thomasleese: Yeah, I couldn't think of a better solution... makes me sad too :cry:

PR #8383, opened 2018-11-28
Add the correct memory threshold

th31nitiate:
Fix the issue with the threshold alerting at the wrong times. This I believe was caused by the decimal point within the code. Please review trello card for more infromation: https://trello.com/c/GJDx8cFx/1755-adjust-mysql-rds-instance-memory-warning-threshold

### DIFF
--- a/modules/monitoring/files/usr/lib/nagios/plugins/check_aws_rds_memory
+++ b/modules/monitoring/files/usr/lib/nagios/plugins/check_aws_rds_memory
@@ -65,8 +65,8 @@ import boto3
 #################################
 parser = argparse.ArgumentParser()
 parser.add_argument("-r","--region", type=str, choices=['us-east-1','eu-west-1'], help="AWS region to check")
-parser.add_argument("-w","--warning", type=float, help="Value(GB) at which to raise a warning alert (5.0)", default=5.0)
-parser.add_argument("-c","--critical", type=float, help="Value(GB) at which to raise a critical alert (2.0)", default=2.0)
+parser.add_argument("-w","--warning", type=float, help="Value(GB) at which to raise a warning alert (2.0)", default=2.0)
+parser.add_argument("-c","--critical", type=float, help="Value(GB) at which to raise a critical alert (1.0)", default=1.0)
 parser.add_argument("-i","--instanceid", type=str, help="Instanceid", default=0)
 args = parser.parse_args()
 
--- a/modules/monitoring/manifests/checks/rds_config.pp
+++ b/modules/monitoring/manifests/checks/rds_config.pp
@@ -23,8 +23,8 @@ define monitoring::checks::rds_config (
   $region = undef,
   $cpu_warning = 80,
   $cpu_critical = 90,
-  $memory_warning = 20,
-  $memory_critical = 10,
+  $memory_warning = 2,
+  $memory_critical = 1,
   $storage_warning = 20,
   $storage_critical = 10,
 ){


modules/monitoring/files/usr/lib/nagios/plugins/check_aws_rds_memory
```diff
@@ -65,8 +65,8 @@ import boto3
 #################################
 parser = argparse.ArgumentParser()
 parser.add_argument("-r","--region", type=str, choices=['us-east-1','eu-west-1'], help="AWS region to check")
-parser.add_argument("-w","--warning", type=float, help="Value(GB) at which to raise a warning alert (5.0)", default=5.0)
-parser.add_argument("-c","--critical", type=float, help="Value(GB) at which to raise a critical alert (2.0)", default=2.0)
+parser.add_argument("-w","--warning", type=float, help="Value(GB) at which to raise a warning alert (50.0)", default=50.0)
```
schmie: These values need to be lower (We are trying to avoid a current spurious error in MySQL.. and you had a look at PG from what I gather from the card).. Can you please change this to 2.0GB for warning and 1.0GB for critical?

modules/monitoring/files/usr/lib/nagios/plugins/check_aws_rds_memory
```diff
@@ -65,8 +65,8 @@ import boto3
...
+parser.add_argument("-c","--critical", type=float, help="Value(GB) at which to raise a critical alert (20.0)", default=20.0)
```
schmie: Change to 1.0GB here, please (see above)

modules/monitoring/files/usr/lib/nagios/plugins/check_aws_rds_memory
th31nitiate: I understand but my only problem with this is that the command line check seems to the through strange errors when I use this value, so, for example, it does not trigger an alert :( 

It only seems to work to a specific decimal point. I have done the change to fix staging alerts.

modules/monitoring/files/usr/lib/nagios/plugins/check_aws_rds_memory
th31nitiate: Done

modules/monitoring/files/usr/lib/nagios/plugins/check_aws_rds_memory
```diff
@@ -65,8 +65,8 @@ import boto3
 #################################
 parser = argparse.ArgumentParser()
 parser.add_argument("-r","--region", type=str, choices=['us-east-1','eu-west-1'], help="AWS region to check")
-parser.add_argument("-w","--warning", type=float, help="Value(GB) at which to raise a warning alert (5.0)", default=5.0)
-parser.add_argument("-c","--critical", type=float, help="Value(GB) at which to raise a critical alert (2.0)", default=2.0)
+parser.add_argument("-w","--warning", type=float, help="Value(GB) at which to raise a warning alert (50.0)", default=2.0)
+parser.add_argument("-c","--critical", type=float, help="Value(GB) at which to raise a critical alert (20.0)", default=1.0)
```
schmie: Could you please adjust the help strings as well?

modules/monitoring/manifests/checks/rds_config.pp
```diff
@@ -23,8 +23,8 @@
   $region = undef,
   $cpu_warning = 80,
   $cpu_critical = 90,
-  $memory_warning = 20,
-  $memory_critical = 10,
+  $memory_warning = 2.0,
+  $memory_critical = 1.0,
```
schmie: For consistency with the other values provided in this file, I think these should integers, so 2 and 1 here...

schmie:
Thanks for adding the Trello link. You will need to `git rebase master` and `git push origin change-default-memeory-threshold --force-with-lease` once more to make the CI checks pass again...

PR #8391, opened 2018-12-03
Trello-1731: Gor Replay ste header

ronocg:
The content-store application only supports the "Forwarded:" header
therefore the "X-Forwarded-Host:" header can be changed for the content
store.
NOTE: defaulting to the "X-Forwarded-Host:" header as the cache router
that also uses the gor replay class needs to support this.

Pair: @ronocg <conor.glynn@digital.cabinet-office.gov.uk>
& @fredericfran-gds <frederic.francois@digital.cabinet-office.gov.uk>


fredericfran-gds:
fine my side. need someone to look at it.

PR #8395, opened 2018-12-03
Remove stale database checks 

th31nitiate:
This removes Nagios checks that are created by alongside data-sync jobs.


modules/icinga/manifests/passive_check.pp
```diff
@@ -40,7 +40,7 @@
 define icinga::passive_check (
   $service_description,
   $host_name,
-  $ensure                  = 'present',
+  $ensure                  = 'absent',
```
schmie: Please revert this to "present" - it is the safer default ; ) (We may have entries which have no "ensure" defined...)

th31nitiate: Ok, good catch on that. I will make sure that we revert it in case of config files created without the explicit value.

schmie:
I am merging this as David is on leave and it is a minor change

PR #8419, opened 2018-12-07
Always point /api/content to AWS

rubenarakelyan:
Content Store has been migrated to AWS, so always point `/api/content` to AWS even when accessed via Carrenza.


schmie:
This would not have worked as we do not resolve govuk-internal via DNS. Anyway the original should work

PR #8430, opened 2018-12-10
Ensure data upload dir present

fredericfran-gds:
# Context

As part of the AWS migration of support-api, there is a need to deploy the `support-api` app on the Backend boxes but unfortunately the puppet run fails because `/data/uploads` directory is not present.

# Decisions

1. Ensure the directory is created in puppet code


fredericfran-gds:
Closed because asset-manager is now deployed via puppet after terraforming the project.

PR #8431, opened 2018-12-10
Increase number of unicorn workers on finder-frontend from 2 to 4

JonathanHallam:
This commit increases the number of unicorn workers from 2 to 6 per instance. This allows more requests to be handled simultaneously. This affects proper finders and advanced search and should make them able to handle larger numbers of unique requests at the same time 

### Testing
This was tested using gatling on staging. In general it has helped to lower the upper limits of response time. The following screenshots were from data following tests involving ~5k requests using about 500 unique URL's

Before
screen shot 2018-12-10 at 13 57 46

After
screen shot 2018-12-10 at 13 57 57

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -456,6 +456,7 @@ govuk::apps::email_alert_service::rabbitmq::queue_size_warning_threshold: 5
 govuk::apps::finder_frontend::enabled: true
 govuk::apps::finder_frontend::nagios_memory_warning: 2000
 govuk::apps::finder_frontend::nagios_memory_critical: 2500
+govuk::apps::finder_frontend::unicorn_worker_processes: "4"
 
 govuk::apps::frontend::nagios_memory_warning: 1200
 govuk::apps::frontend::nagios_memory_critical: 1400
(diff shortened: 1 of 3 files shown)


thomasleese:
I can see that finder-frontend doesn't support that option yet, you'll need [to update the file for it](https://github.com/alphagov/govuk-puppet/blob/874c9aaf7669a3e3572e64779defba667aff49be/modules/govuk/manifests/apps/finder_frontend.pp#L30-L40) to have something similar to government-frontend.

- https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/government_frontend.pp#L44
- https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/government_frontend.pp#L67

Once that's done, it would be good to add a similar line to the `hieradata_aws/common.yml` file so we have the same in AWS (GOV.UK is being migrated to AWS at the moment).

JonathanHallam:
@thomasleese i think this now has everything?

thomasleese:
@JonathanHallam Just one more thing, you'll need to add a line in the finder-frontend configuration similar to this: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/government_frontend.pp#L67

JonathanHallam:
@thomasleese am hoping I've now cracked it!

PR #8440, opened 2018-12-13
Switch DB connection from Carrenza to AWS for support_api

rtrinque:
As part of the migration we redirect support_api to the db on
AWS instead of Carrenza.


ronocg:
This will need a rebase from master when back in the new year.It looks like it this should work for getting support-api (carrenza) speaking to the DB in AWS but I think this security group may need updating also: govuk_db-admin_access to allow 6432 inbound?

schmie:
This is no longer required. We had hoped to migrate zero-downtime, but loss of db connection does not write cleanly to queue / keeps app functionality up. We therefore opted to migrate with downtime and not switch out DBs

PR #8453, opened 2018-12-18
Increase number of unicorn workers for Static

Rosa-Fox:
[Trello](https://trello.com/c/vN1bhcxG/680-from-load-testing-make-static-more-resilient)

- Response to outcome of load test in which `frontend` had some difficulty talking to `static`, resulting in 500s

Co-authored-by: Jonathan Kerr <jonathan.kerr@digital.cabinet-office.gov.uk>

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -687,6 +687,7 @@ govuk::apps::smokey::rate_limit_token: "%{hiera('smokey_rate_limit_token')}"
 
 govuk::apps::static::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::static::redis_port: "%{hiera('sidekiq_port')}"
+govuk::apps::static::unicorn_worker_processes: "8"
 
 govuk::apps::support::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::support::redis_port: "%{hiera('sidekiq_port')}"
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/static.pp
```diff
@@ -51,20 +51,22 @@
   $ga_universal_id = undef,
   $redis_host = undef,
   $redis_port = undef,
+  $unicorn_worker_processes = undef
```
thomasleese: It would be good to add some documentation for this field similar to the others above.

hieradata/common.yaml
```diff
@@ -683,6 +683,7 @@ govuk::apps::smokey::rate_limit_token: "%{hiera('smokey_rate_limit_token')}"
 
 govuk::apps::static::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::static::redis_port: "%{hiera('sidekiq_port')}"
+govuk::apps::static::unicorn_worker_processes: "8"
```
thomasleese: This will also need to be added to the `hieradata_aws/common.yaml` file for integration (and for when our apps are all in AWS in the future).

modules/govuk/manifests/apps/static.pp
Rosa-Fox: Thanks @thomasleese! Have amended.

hieradata/common.yaml
Rosa-Fox: Thanks @thomasleese! Have amended.

thomasleese:
We'll test this in a load test before merging. https://trello.com/c/kxP64gJo/700-redo-the-load-test-to-see-what-impact-weve-had-with-our-changes

PR #8454, opened 2018-12-18
Load common_credentials hieradata

tijmenb:
This is a new hieradata file that contains secrets shared between integration, staging and production.

They're created in https://github.com/alphagov/govuk-secrets/pull/523. Do not merge this PR until that PR is merged (I'm not sure what happens if `common_credentials` doesn't exist yet).

### DIFF
--- a/hiera.yml
+++ b/hiera.yml
@@ -7,6 +7,7 @@
   - '%{::environment}_credentials'
   - '%{::environment}'
   - 'common.%{::lsbdistcodename}'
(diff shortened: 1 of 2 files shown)


kevindew:
This should be ok to test and merge before govuk_secrets as you don't have access to govuk_secrets when building the dev VM.

One thing I was wondering related to this is if you're considering a way to manage the auth credentials that are duplicated in staging and prod but distinct in integration?

PR #8456, opened 2018-12-19
Change python version of govuk_node_list_aws to python3

th31nitiate:
This is a PR to add python3 to GOVUK node list AWS within AWS. This will fix a few bugs and ensure proper node information.


schmie:
Fixed some whitespace errors

modules/govuk_scripts/manifests/init.pp
```diff
@@ -33,6 +33,14 @@
       provider => 'pip',
     }
 
+    # Verify the boto installation and all script dependencies are installed
+    exec { 'check_boto':
+          path    => ['/usr/bin', '/usr/sbin'],
```
deanwilson: The indentation here doesn't match the rest of the puppet code and so needs to be fixed.

modules/govuk_scripts/manifests/init.pp
```diff
@@ -33,6 +33,14 @@
...
+          command => ['/usr/bin/pip3 install boto3'],
+          require => Class['base::packages'],
```
deanwilson: This `exec` does not have an `unless` and so will run on every puppet run. This shouldn't happen.

schmie:
Cheers!

PR #8460, opened 2018-12-19
Alert on the Publishing API downstream_high latency

cbaines:
The downstream_high queue in the Publishing API is a Sidekiq queue
used to process new and changed content which is a direct result of a
publisher action, for example, someone pressing a publish button.

It's expected that jobs in this queue are processed very quickly, so
the warning threshold has been set to 20 seconds, and the critical
threshold to 60 seconds.

As this is a new alert, just have it notify people in hours, but I'd
consider also alerting on this out of hours eventually.


cbaines:
I've just seen I opened this before... https://github.com/alphagov/govuk-puppet/pull/7774

PR #8462, opened 2018-12-20
remove backup on the transition postgres

fredericfran-gds:
# Context

Since the transition app has been migrated to AWS, there is no need to backup the transition app postgres

# Decisions

1. disable backup for the transition postgres database.

### DIFF
--- a/modules/govuk_postgresql/manifests/server/primary.pp
+++ b/modules/govuk_postgresql/manifests/server/primary.pp
@@ -30,8 +30,13 @@ class govuk_postgresql::server::primary (
   $slave_password,
   $type = 'host',
   $user = 'replication',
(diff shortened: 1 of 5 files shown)


modules/govuk_postgresql/manifests/server/primary.pp
```diff
@@ -30,8 +30,13 @@
   $slave_password,
   $type = 'host',
   $user = 'replication',
+  $backup_enable = true,
```
deanwilson: Can you add this to the docs at the top of the file please?

fredericfran-gds: yep, will do now. thanks

PR #8470, opened 2018-12-21
Work towards supporting notifications for app healthchecks

cbaines:
This relates to the addition of monitoring for the high priority downstream queue in to the Publishing API healthcheck ( https://github.com/alphagov/publishing-api/pull/1308 ), and then the change in govuk-puppet to actually monitor the details of the healthcheck in Icinga ( https://github.com/alphagov/govuk-puppet/pull/8463 ).

We've had a couple of incidents now where the growing length of the `downstream_high` queue could have been caught sooner, and these changes work towards that.


cbaines:
Great, thanks Ben :)

PR #8479, opened 2019-01-02
Enable authentication in Support API

leenagupte:
Trello: https://trello.com/c/WR2MNxo6

Related to: https://github.com/alphagov/support-api/pull/266

Adds the environment variables needed to allow applications to authenticate with support-api using signon.

Do not merge until govuk-secrets has been updated.


kevindew:
You're also safe to merge this if I remember correctly.

PR #8482, opened 2019-01-03
Disable X-Ray entirely for bouncer

barrucadu:
**Updated to use the mechanism introduced in https://github.com/alphagov/govuk_app_config/pull/63**

---

Without this, we'll begin to record a lot of uninteresting samples
when https://github.com/alphagov/bouncer/pull/171 gets merged.

### DIFF
--- a/modules/govuk/manifests/apps/bouncer.pp
+++ b/modules/govuk/manifests/apps/bouncer.pp
@@ -173,6 +173,15 @@ class govuk::apps::bouncer(
     document_root => '/var/apps/bouncer/assets-directgov/directgov_campaigns',
   }
 
+  # disable X-Ray entirely, as bouncer gets a lot of traffic but
+  # doesn't really do anything interesting (from a tracing
+  # point-of-view).
+  govuk::app::envvar { 'bouncer-GOVUK_APP_CONFIG_DISABLE_XRAY':
+    app     => 'bouncer',
+    varname => 'GOVUK_APP_CONFIG_DISABLE_XRAY',
+    value   => 'this just needs to be set',
+  }
+
   nginx::config::site { 'www.mhra.gov.uk':
     content => template('govuk/www.mhra.gov.uk_nginx.conf.erb'),
   }


modules/govuk/manifests/apps/bouncer.pp
```diff
@@ -173,6 +173,15 @@
     document_root => '/var/apps/bouncer/assets-directgov/directgov_campaigns',
   }
 
+  # disable X-Ray sampling, as bouncer gets a lot of traffic but
```
cbaines: I haven't looked in to X-Ray, but does this mean that we're still using X-Ray for Bouncer (it's just unsampled, rather than sampled data), or that it's being disabled?

barrucadu: It'll be enabled but unsampled, as there's nothing to actually remove it in https://github.com/alphagov/govuk_app_config/blob/master/lib/govuk_app_config/govuk_xray.rb

cbaines: So I'm still uncertain around the intent of this change. Does "unsampled" mean in the context of XRay that we won't be getting any data? I'd usually interpret "unsampled" as you get all the data, rather than a sample of it.

barrucadu: Ah, that was me misusing terminology.  Nothing will be sampled with a sampling rate of 0:

https://github.com/aws/aws-xray-sdk-ruby/blob/1869ca5b72dfb206e42ef92d40a7fd2f4b53f465/lib/aws-xray-sdk/sampling/local/sampler.rb#L91

thomasleese: We might still get some results if I read that correctly since `Random.rand` could return `0`. Not sure how important that is though.

thomasleese: I wonder if a better solution would be to update `govuk_app_config` to allow complete disabling of X-Ray (if we're setting the sample rate to 0, we may as well not bother doing any X-Ray related processing at all).

barrucadu: Yeah, that is better, as with a rate of 0 there would still be a bit of overhead on each request as it decides whether to sample it or not.  I'll add a new env var to govuk_app_config to switch off -Ray entirely.

cbaines: > I'll add a new env var to govuk_app_config to switch off X-Ray entirely.

That sounds good :+1:

PR #8491, opened 2019-01-04
Add Router API authentication

bilbof:
This adds the environment variables required by router-api
and content-store for use in GDS-SSO authentication.

Prior art that I want to follow: https://github.com/alphagov/govuk-puppet/pull/8426.


kevindew:
@bilbof @koetsier heads up that this'll need to be set up for the draft router api too

PR #8505, opened 2019-01-08
Add Draft Router API keys and access token

bilbof:
Following 25c0efb, its also necessary to provide a draft set
of keys and tokens for the router-api, as we run an instance
of the router-api in the draft stack.

### DIFF
--- a/hieradata/class/draft_cache.yaml
+++ b/hieradata/class/draft_cache.yaml
@@ -5,6 +5,9 @@ govuk::apps::authenticating_proxy::mongodb_nodes:
   - 'router-backend-2.router'
   - 'router-backend-3.router'
 
+govuk::apps::router_api::oauth_id: "%{hiera('govuk::apps::router_api::draft_oauth_id')}"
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/apps/content_store.pp
```diff
@@ -57,6 +57,9 @@
 # [*router_api_bearer_token*]
 #   The bearer token that will be used to authenticate with the router api
 #
+# [*draft_router_api_bearer_token*]
+#   The bearer token that will be used to authenticate with the draft router api
+#
```
kevindew: You don't need to add in this draft_router_api_bearer_token, we just need to override the hieradata of the router_api_bearer_token to use the draft one like we've done with oauth_id and oauth_secret: https://github.com/alphagov/govuk-puppet/blob/81cb1eadd1c3457df22c030b8fb2b01cdc186437/hieradata/class/draft_content_store.yaml#L11 (and hieradata_aws)

bilbof: Thank you for spotting that! I've pushed an amended commit.

PR #8506, opened 2019-01-08
Adds oauth environment variables for Rummager

oscarwyatt:
For oauth authentication for Rummager write/delete endpoints

Trello card: https://trello.com/c/M9Xek3W4

### DIFF
--- a/hieradata/class/search.yaml
+++ b/hieradata/class/search.yaml
@@ -5,6 +5,9 @@ govuk_elasticsearch::local_proxy::servers:
   - 'rummager-elasticsearch-2.api'
   - 'rummager-elasticsearch-3.api'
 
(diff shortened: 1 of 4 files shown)


hieradata/class/search.yaml
```diff
@@ -5,6 +5,9 @@ govuk_elasticsearch::local_proxy::servers:
   - 'rummager-elasticsearch-2.api'
   - 'rummager-elasticsearch-3.api'
 
+govuk::apps::rummager::oauth_id: "%{hiera('govuk::apps::rummager::oauth_id')}"
+govuk::apps::rummager::oauth_secret: "%{hiera('govuk::apps::rummager::oauth_secret')}"
```
kevindew: We don't need this - it'll actually cause a circular dependency as it references the hiera it's setting. If you got this from content store, it was done to set the draft hieradata there.

PR #8509, opened 2019-01-08
Remove backups for "email-alert-monitor_production"

tijmenb:
This database doesn't exist (the email-alert-monitor isn't an app and doesn't use a database). The resulting file that we download in data sync is an empty SQL dump.

Was added in https://github.com/alphagov/govuk-puppet/pull/7960, but I'm not sure why.

### DIFF
--- a/hieradata_aws/class/staging/db_admin.yaml
+++ b/hieradata_aws/class/staging/db_admin.yaml
@@ -65,17 +65,6 @@ govuk_env_sync::tasks:
     temppath: "/tmp/contacts_production"
     url: "govuk-staging-database-backups"
     path: "mysql"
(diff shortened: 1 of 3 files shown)


hieradata_aws/class/staging/db_admin.yaml
```diff
@@ -65,17 +65,6 @@ govuk_env_sync::tasks:
     temppath: "/tmp/contacts_production"
     url: "govuk-staging-database-backups"
     path: "mysql"
-  "pull_email-alert-monitor_production_daily":
-    ensure: "present"
```
cbaines: I'm guessing you'll need to remove this manually, as it's not already absent.

PR #8515, opened 2019-01-09
add the govuk-splunk-configurator as a local repo

fredericfran-gds:


ronocg:
The certificate that will sit within a deb package within this apt repo, will be a secret no? As it can be used to put anything into the GDS splunk cloud ?

fredericfran-gds:
a new configurator will be used which will not have any certificates or privary keys etc.

PR #8519, opened 2019-01-09
Trigger reload of xray deamon

erino:


rubenarakelyan:
Yes but you'd need to do that across all the machines running the X-Ray daemon manually which is quite a few...

erino:
@benthorner I started doing that but then realised that as @rubenarakelyan says that's quite a few

benthorner:
@erino but I thought the point was just to prove the code works on one node (and should therefore work on the rest of them)?

Issue #8518, opened 2019-01-09
Lack of blank lines in robots.txt


nacnudus:
[`robots.txt`](https://gov.uk/robots.txt) doesn't separate records by blank lines

```
User-agent: *
Disallow: /*/print$
# Don't allow indexing of user needs pages
Disallow: /info/*
Sitemap: https://www.gov.uk/sitemap.xml
# https://ahrefs.com/robot/ crawls the site frequently
User-agent: AhrefsBot
Crawl-delay: 10
# https://www.deepcrawl.com/bot/ makes lots of requests. Ideally
# we'd slow it down rather than blocking it but it doesn't mention
# whether or not it supports crawl-delay.
User-agent: deepcrawl
Disallow: /
# Complaints of 429 'Too many requests' seem to be coming from SharePoint servers
# (https://social.msdn.microsoft.com/Forums/en-US/3ea268ed-58a6-4166-ab40-d3f4fc55fef4)
# The robot doesn't recognise its User-Agent string, see the MS support article:
# https://support.microsoft.com/en-us/help/3019711/the-sharepoint-server-crawler-ignores-directives-in-robots-txt
User-agent: MS Search 6.0 Robot
Disallow: /
```

This broke a robots.txt parser, and the discussion in the issue pointed out that lack-of-blank-lines is non-standard.  But it depends what standard you read.  Google says blank lines are optional, W3C doesn't say anything definite, and robotstxt.org says they're required.  Here's a [summary](https://github.com/ropensci/robotstxt/issues/48#issuecomment-452719231) with links to sources.

I don't have a strong opinion, but since blank lines are always allowed and are sometimes required, perhaps they should be used.

PR #8520, opened 2019-01-09
Add blank lines to robots.txt

tijmenb:
This adds some blank lines to robots.txt, which some parsers need. Fixes https://github.com/alphagov/govuk-puppet/issues/8518.


elliotcm:
@tijmenb Could you squash those commits down?

PR #8521, opened 2019-01-09
Fixups to support Puppet future parser

rhowe-gds:
The Puppet 4.x parser is stricter than the 3.x one. Here are a bunch of rather random changes to improve our compatibility with Puppet 4. There are still a few more things to fix before the tests all pass, but these are the ones which I think are nice and safe.

The fix to `govuk_jenkins` is rather ugly and I'm not proud of it. Maybe we should find a better way to do that, maybe not?

### DIFF
--- a/modules/govuk/manifests/app/config.pp
+++ b/modules/govuk/manifests/app/config.pp
@@ -138,7 +138,7 @@ define govuk::app::config (
         value   => $app_type;
       "${title}-PORT":
         varname => 'PORT',
-        value   => $port;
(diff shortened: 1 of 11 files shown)


issyl0:
This needs testing in the dev VM or equivalent.

issyl0:
[Here's the output](https://gist.github.com/issyl0/75460cae33a11581d663dcb4a174b313) of my (new) dev VM running `govuk_puppet` on master and this branch. Looks fine.

PR #8539, opened 2019-01-15
Check that AWS and Carrenza hieradata are in sync

rubenarakelyan:
We currently have to duplicate the AWS and Carrenza hieradata. This adds a test to check that they have the same keys defined. To account for legitimate differences, we also implement two whitelists.

We currently have a *lot* of differences, so I've whitelisted everything for now. This will at least stop us from drifting further apart. We'll need to manually whitelist & fix any existing discrepancies.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,6 +17,10 @@ node {
     stage("Bundle install") {
       govuk.bundleApp()
     }
+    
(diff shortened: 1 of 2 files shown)


rubenarakelyan:
Yeah, similarly to govuk-secrets, a lot of them are actually legit differences but there are some that will either be fixed once the app is migrated, or need to be fixed separately.

PR #8541, opened 2019-01-15
Run nginx configtest before reloading or restarting

rubenarakelyan:
This commit amends the Puppet classes that reload or restart nginx after various changes to always run a configtest beforehand. If the configtest fails, nginx will not be reloaded or restarted. This will prevent nginx either:

* Not restarting since the restart command was only run on a successful configcheck
* Reloading and failing


rubenarakelyan:
PS. This also means that if nginx configtest fails, the puppet run will fail and Icinga will show an error. This is currently not the case since nginx restart is silently blocked by a failed configtest.

PR #8545, opened 2019-01-16
Remove nginx::restart

rhowe-gds:
We already define a restart command on the 'nginx' service, so we can replace
this class with a notification against the service.

Use a resource collector in nginx::package, because that doesn't include the
service (so it may not be defined)

Sprinkle some code and test tidyups around whilst we're at it.

### DIFF
--- a/modules/nginx/manifests/init.pp
+++ b/modules/nginx/manifests/init.pp
@@ -58,11 +58,6 @@ class nginx (
     ],
   }
 
-  # Include ability to do a full restart of nginx. This does not explicitly
-  # trigger a restart, but simply makes the class available to any manifest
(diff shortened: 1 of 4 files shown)


rhowe-gds:
This is indeed a fair point. There are subtleties around all of this..

`nginx` postinst script either starts nginx if it's not running or does `/etc/init.d/nginx upgrade` which sends a `USR2` signal to nginx which makes it re-exec its binary as per http://nginx.org/en/docs/control.html

`nginx-module-perl` does not touch the nginx service in its postinst script, so will require the service to be restarted via some other mechanism as we're doing here.

The nginx init script's `restart` command does a `configtest` before restarting the service, but `reload` does not.

PR #8552, opened 2019-01-18
Remove people who no longer need GOV.UK access

rubenarakelyan:
They have either left GDS or no longer work on GOV.UK.


modules/users/manifests/isabelllong.pp
```diff
@@ -1,6 +1,7 @@
 # Creates the isabelllong user
 class users::isabelllong {
   govuk_user { 'isabelllong':
+    ensure   => absent,
```
issyl0: This feels like the end of a long era. 😭

PR #8555, opened 2019-01-18
set aws staging and production not to have search upstream

fredericfran-gds:


fredericfran-gds:
close because no longer needed due to bug with ci tool

PR #8564, opened 2019-01-21
Increase thread count threshold for cache clearing service

steventux:
The increase in workers for cache clearing service has increased
the number of processes spawned so alert on a higher thread count.


sihugh:
Are we doing anything with this?

steventux:
We held off doing this while we investigated the original problem. Assuming it's not still a problem I'll close this.

sihugh:
noice

PR #8568, opened 2019-01-21
Clean up the postfix module; future parser compatibility

rhowe-gds:
The postfix module has some issues under future parser related to variable scoping. Refactor it to be cleaner and avoid these issues.
See individual commits for details.

### DIFF
--- a/modules/postfix/manifests/postmapfile.pp
+++ b/modules/postfix/manifests/postmapfile.pp
@@ -1,19 +1,14 @@
 # FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
 define postfix::postmapfile() {
-  exec { "postmap_${title}":
-    command     => "/usr/sbin/postmap /etc/postfix/${title}",
-    refreshonly => true,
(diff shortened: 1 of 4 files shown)


modules/postfix/manifests/postmapfile.pp
```diff
@@ -1,17 +1,18 @@
...
-                mode    => '0644',
-                content => template("postfix/etc/postfix/${named}.erb"),
-                notify  => Exec["postmap_${named}"],
-        }
+define postfix::postmapfile(
+  $content,
+) {
+  $filename = "/etc/postfix/${title}"
+
+  Package['postfix']
```
deanwilson: This should be a class. You shouldn't depend on things from inside other classes.

rhowe-gds: It should, but it did before

afda16: I think the Package here is not needed, we already have https://github.com/alphagov/govuk-puppet/blob/master/modules/postfix/manifests/init.pp#L46, maybe it was residual

PR #8574, opened 2019-01-23
Enable CKAN basic auth only for integration

brucebolt:


brucebolt:
I've documented `vhost_protected`.  The rest should be addressed in a separate PR.

PR #8583, opened 2019-01-24
Increase the number of Graphite unicorn workers

schmie:
- The Graphite node in production is struggling intermittently with load

- We hope allowing for more graphite workers does help with the issue

solo: @schmie


cbaines:
I think the workers change is a good one to make, but it looks to me like it's hardcoded in a dependency: https://github.com/gds-operations/puppet-graphite/pull/66

cbaines:
I've rebased this on master so that it includes the graphite module update.

cbaines:
I'm on 2ndline at the moment, so I'm going to take the opportunity to merge and test this...

PR #8585, opened 2019-01-25
Allow selectively changing where apps look for rummager

barrucadu:
This is going to be useful as we migrate from rummager to search-api.  When we're happy the data in rummager and search-api is consistent, and that updates are being sent to both apps, we can gradually change which one is used for queries.

---

[Trello card](https://trello.com/c/YTAtoA21/22-add-an-option-to-each-app-which-uses-rummager-to-override-plek)

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -237,6 +237,9 @@
 # [*collectd_process_regex*]
 #   Regex to use to identify the process.
 #
+# [*override_search_location*]
+#   Alternative hostname to use for Plek("search") and Plek("rummager")
+#
 define govuk::app (
   $app_type,
   $port = 0,
@@ -273,6 +276,7 @@ define govuk::app (
   $cpu_warning = 150,
   $cpu_critical = 200,
   $collectd_process_regex = undef,
+  $override_search_location = undef,
 ) {
 
   if ! ($app_type in ['procfile', 'rack', 'bare']) {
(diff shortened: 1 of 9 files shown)


barrucadu:
I *think* these are the only apps which use rummager.

The idea is that we'll get the data in rummager and search-api in sync, with updates going to both apps, and then gradually switch the apps over as we gain confidence that it's all working.

Once these apps have all been switched to use search-api, this special configuration can be removed and we'll just update the apps to use `Plek("search-api")`.

cbaines:
I don't quite follow...

What I'm suggesting is applying the configuration in govuk-puppet here for example https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/deploy/config.pp#L87-L116

Providing you are going to have all the apps using the same value, that should work similarly, but not require as many changes in govuk-puppet.

barrucadu:
We're planning to roll out the switch to search-api like this:

1. For each app:
    1. Set `override_search_location` for that app
    2. Deploy puppet to staging and run it on the affected nodes
    3. Test app in staging
    4. Deploy puppet to production and run it on the affected nodes
    5. Test app in production
2. Check X-Ray logs for Rummager (dial up the reporting rate if necessary) to make sure nothing is still talking to it
3. Update each app to reference `search-api` directly, instead of `search` or `rummager`
4. Remove this special configuration

So the ability to do this on a per-app basis is important.

The reason we're doing it this way is that it seems less risky.  If we discover a problem only when we hit production, only one app will be busted rather than everything which uses rummager.

cbaines:
> So the ability to do this on a per-app basis is important.

Ok.

PR #8593, opened 2019-01-28
Trello-1921: nginx loadbalancer config

ronocg:
As part of the migration of the frontends to aws, there needs to be a
way for the aws finder-frontend to access the search-api (search-1/2/3
VMs) in carrenza. While in carrenza this was accessed using an
internal-lb via a hosts entry but aws will dns resolve
search.publishing.service.gov.uk which will be cnamed into the
backend-lb.

Therefore the backend-lb will require the nginx configuration for search
which is being laid down by this PR.

ALso limiting the nginx for search to only allow from the 3 nat gateway
egress ips from the corresponding AWS environment.


hieradata_aws/production.yaml
```diff
@@ -153,6 +153,9 @@ govuk::deploy::config::licensify_app_domain: 'publishing.service.gov.uk'
 govuk::deploy::setup::ssh_keys:
     jenkins_production_carrenza: 'AAAAB3NzaC1yc2EAAAADAQABAAACAQCfPjubgzCkZo1aTPlkgeXb1eh3IonRBRptx0qLMCjOV+e+M8uRAT/Xx3ydJYPd7sOgZDyx2xjSGb7Eefau0jSUAcMD1Xd01SXWBQPJRDfPmQLrdbM0xxOFH8nft39uo4Mz6ccZc34xrudL6q/urp732HZHYwltnNnbk9h58n1QIhemRtN3u9RrSSOILqw/F42S6Aj8lZ1v/DGgfc6F5pKyJ7TByHL1RlqwpZHbEjYYuvK0ZJJsKPlyVPbNDsX7UEYWwbpPsFs9LPvCC6epmj+7Lv25bTU8rKK8J3rNWa1FybpWS0VXbF/+mrLjtT0/vwvbwUzsjK6dSUsbEsBEn+cOqomxCYkLjMzUy1+ReYAh6+CjmzutPs1g4OjQRel2ONprhPTEsNUu+oNObnGDOUpzHK10ntAZxguA4QEUmOBBWfxuQhmJO60/b1zedCcc7MR8e9S0y4jtpXa8GBCe40+napArZTW9QXlHLWz+khkYQfO107Q+z1QaLFojdcrHlUfpqAc6DtVJQu7tsBt2vXTn0qq6mU5Eg6UY+X1l/3gWdFS3ZEvCUoGK6bLU3i50jZ1xsFogFFfvSux46S1DYW2Fk8a/2IBBdcQcL1YoM73jiAQgpU8Vs50wtk4mWhK1yBaMYmMAeL7mKFbJla7SjTAwaDdo5uezyrJlbZxqTb/Y3w=='
 
+govuk::node::s_backend_lb::aws_egress_nat_ips: []
```
fredericfran-gds: Minor comment: there is no need for this if the default value of `aws_egress_nat_ips` is already an empty list in the module: `s_backend_lb`

hieradata_aws/staging.yaml
```diff
@@ -155,6 +155,9 @@ govuk::deploy::setup::ssh_keys:
...
     jenkins_production_carrenza: 'AAAAB3NzaC1yc2EAAAADAQABAAACAQCfPjubgzCkZo1aTPlkgeXb1eh3IonRBRptx0qLMCjOV+e+M8uRAT/Xx3ydJYPd7sOgZDyx2xjSGb7Eefau0jSUAcMD1Xd01SXWBQPJRDfPmQLrdbM0xxOFH8nft39uo4Mz6ccZc34xrudL6q/urp732HZHYwltnNnbk9h58n1QIhemRtN3u9RrSSOILqw/F42S6Aj8lZ1v/DGgfc6F5pKyJ7TByHL1RlqwpZHbEjYYuvK0ZJJsKPlyVPbNDsX7UEYWwbpPsFs9LPvCC6epmj+7Lv25bTU8rKK8J3rNWa1FybpWS0VXbF/+mrLjtT0/vwvbwUzsjK6dSUsbEsBEn+cOqomxCYkLjMzUy1+ReYAh6+CjmzutPs1g4OjQRel2ONprhPTEsNUu+oNObnGDOUpzHK10ntAZxguA4QEUmOBBWfxuQhmJO60/b1zedCcc7MR8e9S0y4jtpXa8GBCe40+napArZTW9QXlHLWz+khkYQfO107Q+z1QaLFojdcrHlUfpqAc6DtVJQu7tsBt2vXTn0qq6mU5Eg6UY+X1l/3gWdFS3ZEvCUoGK6bLU3i50jZ1xsFogFFfvSux46S1DYW2Fk8a/2IBBdcQcL1YoM73jiAQgpU8Vs50wtk4mWhK1yBaMYmMAeL7mKFbJla7SjTAwaDdo5uezyrJlbZxqTb/Y3w=='
 
+govuk::node::s_backend_lb::aws_egress_nat_ips: []
```
fredericfran-gds: Minor comment: there is no need for this if the default value of aws_egress_nat_ips is already an empty list in the module: s_backend_lb

modules/loadbalancer/manifests/balance.pp
```diff
@@ -38,12 +38,17 @@
...
+#   NAT Gateway Egress IPs from AWS Environment
+#
+
 define loadbalancer::balance(
     $servers,
     $deny_crawlers = false,
     $error_on_http = false,
     $https_port = 443,
     $internal_only = false,
+    $aws_egress_nat_ips = false,
```
fredericfran-gds: was this not supposed to be a list? Might be a bit confusing if you are using a variable as both a list and a boolean.

modules/loadbalancer/templates/nginx_balance.conf.erb
```diff
@@ -89,6 +89,12 @@ server {
     allow 10.0.0.0/8;
     allow 127.0.0.1;
     deny all;
+    <%- elsif @aws_egress_nat_ips -%>
+    # Only accept connnections from AWS Environment
+    <%- [@aws_egress_nat_ips].flatten.each do |ip| -%>
+    allow <%= ip %>;
```
fredericfran-gds: I will put `deny all` statement in line 91 and 97 after line 98. Should be there in all cases

modules/loadbalancer/templates/nginx_balance_no_ssl.conf.erb
```diff
@@ -72,7 +72,13 @@ server {
     allow 10.0.0.0/8;
     allow 127.0.0.1;
     deny all;
```
fredericfran-gds: I will put deny all statement in line 74 and 80 after line 81. Should be there in all cases

modules/loadbalancer/manifests/balance.pp
fredericfran-gds: maybe you want to use this code in your template `@aws_egress_nat_ips.length > 0`

hieradata_aws/staging.yaml
ronocg: this was to satisfy a spec test that checks if the parameters are defined in both aws and cr hiera configs...

modules/loadbalancer/templates/nginx_balance.conf.erb
ronocg: not sure if I follow this... if there is a "deny all" after line 98 then nothing can use the that nginx load balancer unless it has @internal_only or @aws_egress_nat_ips defined... there are nginx configs that allow from everywhere by not having any allow or deny in the config

fredericfran-gds: cool, I get it.

hieradata_aws/staging.yaml
fredericfran-gds: cool, got it.

ronocg: yep I think I can make this equal to undef here and that will clear it up...

modules/loadbalancer/manifests/balance.pp
ronocg: I think I can make that equal to undef here and that will clear it up..

modules/loadbalancer/templates/nginx_balance.conf.erb
fredericfran-gds: cool, got it.

PR #8595, opened 2019-01-28
Increase timeout for Content Data Admin

chao-xian:
We are seeing timeout issues when the Content Performance Manager
is running slow queries. The timeout does not occur at the API end
and we are not able to reproduce this error on other environments.
This problem only occurs on production. For now we are upping the
timeout on the client end to address the timeouts, and allow
opportunities for performance optimisation later.

### DIFF
--- a/modules/govuk/manifests/apps/content_data_admin.pp
+++ b/modules/govuk/manifests/apps/content_data_admin.pp
@@ -50,6 +50,11 @@
 # [*google_tag_manager_auth*]
 #   The identifier of an environment for Google Tag Manager
 #
+# [*read_timeout*]
+# Configure the amount of time the nginx proxy vhost will wait for the
+# backing app before it sends the client a 504. We override the default 15 seconds
+# to 60.
+#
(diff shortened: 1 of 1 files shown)


chao-xian:
[(3) Bug - timeout when querying all content filtering by search_term](https://trello.com/c/dL9n9KJb/1081-3-bug-timeout-when-querying-all-content-filtering-by-searchterm)

modules/govuk/manifests/apps/content_data_admin.pp
```diff
@@ -50,6 +50,11 @@
 # [*google_tag_manager_auth*]
 #   The identifier of an environment for Google Tag Manager
 #
+# [*read_timeout*]
+# Configure the amount of time the nginx proxy vhost will wait for the
+# backing app before it sends the client a 504. We override the default 15 seconds
+# to 60.
+#
```
cbaines: This comment is a bit misleading, as this class doesn't have a `read_timeout` option. It might be good to remove this to keep the commends and code consistent.

PR #8601, opened 2019-01-29
Increase timeout for Content Performance Manager

chao-xian:
We are seeing timeout issues when the Content Performance Manager
is running slow queries. We are not able to reproduce this error on
other environments other than on production.

For now we are upping the timeout limit to address the timeouts,
and allow opportunities for performance optimisation later.

[Related PR](https://github.com/alphagov/govuk-puppet/pull/8595).

### DIFF
--- a/modules/govuk/manifests/apps/content_performance_manager.pp
+++ b/modules/govuk/manifests/apps/content_performance_manager.pp
@@ -66,6 +66,11 @@
 #   Redis port for Sidekiq.
 #   Default: undef
 #
+# [*read_timeout*]
+# Configure the amount of time the nginx proxy vhost will wait for
+# before it sends the client a 504. We override the default 15 seconds
+# to 60.
+#
 # [*secret_key_base*]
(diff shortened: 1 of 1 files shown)


chao-xian:
[(3) Bug - timeout when querying all content filtering by search_term](https://trello.com/c/dL9n9KJb/1081-3-bug-timeout-when-querying-all-content-filtering-by-searchterm)

modules/govuk/manifests/apps/content_performance_manager.pp
```diff
@@ -66,6 +66,11 @@
 #   Redis port for Sidekiq.
 #   Default: undef
 #
+# [*read_timeout*]
+# Configure the amount of time the nginx proxy vhost will wait for
+# before it sends the client a 504. We override the default 15 seconds
+# to 60.
+#
```
cbaines: Same comment as here https://github.com/alphagov/govuk-puppet/pull/8595/files#r255893256

PR #8608, opened 2019-01-30
fix nginx restart

fredericfran-gds:
# Context

Some AWS machines (e.g. content-store) fails on nginx start phase during the second run of Puppet when a new machine comes up and it is being configured by Puppet. The main complain of Puppet is that the address `0.0.0.0:80` is already taken and the new instance of nginx to be started can't have such a port.

Upon further investigation, it was found nginx was already running on port 80 and there was no PID number in the `/var/run/nginx.pid`

# Decisions
1. when nginx is to be started/restarted, ensure that no nginx is running on the machine.

### DIFF
--- a/modules/nginx/manifests/restart.pp
+++ b/modules/nginx/manifests/restart.pp
@@ -1,7 +1,7 @@
 # FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
 class nginx::restart {
 
-  exec { '/etc/init.d/nginx configtest && /etc/init.d/nginx restart':
+  exec { '/etc/init.d/nginx configtest && (/etc/init.d/nginx restart || { nginx -s quit; /etc/init.d/nginx start;})':
     refreshonly => true,
   }
 
--- a/modules/nginx/manifests/service.pp
+++ b/modules/nginx/manifests/service.pp
@@ -3,6 +3,7 @@ class nginx::service {
 
   service { 'nginx':
     ensure    => running,
+    start     => '/etc/init.d/nginx start || { nginx -s quit; /etc/init.d/nginx start;}',
     hasstatus => true,
     restart   => '/etc/init.d/nginx configtest && /etc/init.d/nginx reload',
   }


.mailmap
```diff
@@ -7,7 +7,6 @@ Alex Torrance <alex.torrance@digital.cabinet-office.gov.uk> <hello@alextorrance.
 ambrozic <ambrozic@gmail.com> <ambrozic@users.noreply.github.com>
 Ana Fernandez <ana.fernandez@digital.cabinet-office.gov.uk> <afda16@users.noreply.github.com>
 Andrea Grandi <a.grandi@gmail.com> <andreagrandi@gds5060.local>
-Andrew Garner <andrew.garner@digital.cabinet-office.gov.uk> <andrew@andrewgarner.com>
```
barrucadu: Users shouldn't removed from the mailmap, it's used by git to attribute commits where someone has used multiple email addresses

fredericfran-gds: cool, sorry. I will revert that. Was trying to solve another error with that user on my branch but someone told me the reason.

modules/nginx/manifests/restart.pp
```diff
@@ -1,7 +1,7 @@
 # FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
 class nginx::restart {
 
-  exec { '/etc/init.d/nginx configtest && /etc/init.d/nginx restart':
+  exec { '/etc/init.d/nginx configtest && killall nginx && /etc/init.d/nginx restart':
```
schmie: :) Can we please make this sth a bit more graceful, like: 
```
 '/etc/init.d/nginx configtest && { /etc/init.d/nginx restart || { pkill nginx; /etc/init.d/nginx start; } }'
```
I do not think the above will work directly - maybe we can use shellquote to create a $command variable or something similar..

modules/nginx/manifests/service.pp
```diff
@@ -3,6 +3,7 @@
 
   service { 'nginx':
     ensure    => running,
+    start     => 'killall nginx && /etc/init.d/nginx start',
```
schmie: Same here..

modules/nginx/manifests/restart.pp
```diff
@@ -1,7 +1,7 @@
 # FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
 class nginx::restart {
 
-  exec { '/etc/init.d/nginx configtest && /etc/init.d/nginx restart':
+  exec { '/etc/init.d/nginx configtest && /etc/init.d/nginx restart || {nginx -s quit; /etc/init.d/nginx start}':
```
schmie: This will not work as expected/ run the nginx stop and start on failure of the configtest - maybe we can move the configtest into an onlyif?
```    
exec { '/etc/init.d/nginx restart || {nginx -s quit; /etc/init.d/nginx start}':
    onlyif     => '/etc/init.d/nginx configtest'
```
Or add more brackets ;)

fredericfran-gds: how I read it is:
1. run the configtest,
2. if successful, run the restart command
3. if restart command fails, run quit command and the nginx start.

fredericfran-gds: If I use the `onlyif`, this will revert what Reuben did here: https://github.com/alphagov/govuk-puppet/pull/8541/files

fredericfran-gds: I added more brackets and testes all scenarios on my bash

schmie: Sry to be a pain, but this is still not working in bash for me - let's sit down together for a bit tomorrow to get this out of the door

schmie: Works in situ - I only checked on my Mac's bash... cheers!

PR #8610, opened 2019-01-30
Increase warning threshold for nginx high active connections

peteglondon:
The number of active connection is regularly above the 500 warning
threshold. Increasing to 650 as this is still 350 below the critical
threshold.

this covers the levels we are seeing regularly

screen shot 2019-01-30 at 16 46 53


barrucadu:
It looks like it regularly goes over 600 too, should it be even higher?

peteglondon:
I've made it 650, which covers what we are seeing all the time now.

PR #8611, opened 2019-01-30
Enable GTM for Content Data in VM

chao-xian:
We have Google Tag Manager enabled for Integration, Staging and
Production but not in local development which makes it hard to
work on. There is no real reason not to enable it.


chao-xian:
[(3) Adapt GTM custom script to track multiple autocomplete input fields](https://trello.com/c/R5syptvD/1100-3-adapt-gtm-custom-script-to-track-multiple-autocomplete-input-fields)

PR #8612, opened 2019-01-31
tie the default nginx config creation either in the app or node class

fredericfran-gds:
# Context

The default nginx config can be created in 2 places: 1. the govuk::content_store and 2. govuk::apps::content_store. At `least one` class needs to do it. Hence we introduce a hiera data which links these 2 classes so that they select which one to create the default nginx config.

# Decisions
1. Introduce new hieradata to link the 2 classes.


fredericfran-gds:
close, may not longer be needed.

PR #8623, opened 2019-02-01
Schedule tag_metadata Rummager rake task twice daily

steventux:
https://trello.com/c/spSQHcZ9/34-bug-some-items-of-content-are-disappearing-from-finder

There's a bug in how updates to the mainstream index write metadata
via the [`GovukIndex::PublishingEventWorker` class in Rummager](https://github.com/alphagov/rummager/blob/master/lib/govuk_index/publishing_event_worker.rb) which surfaces when Mainstream Publisher content updates are made to content present in the 'Find EU Exit guidance for your business' finder.
This PR adds a job to schedule running the retagging rake task which successfully re-applies the relevant metadata, this is scheduled in line with 2ndline timing for applying updates for convenience.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/run_metadata_tagger.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/run_metadata_tagger.yaml.erb
@@ -1,7 +1,7 @@
 ---
 - job:
-    name: Run_Metadata_Tagger
-    display-name: Run_Metadata_Tagger
+    name: Run_Rummager_Metadata_Tagger
+    display-name: Run_Rummager_Metadata_Tagger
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/templates/jobs/run_metadata_tagger.yaml.erb
```diff
@@ -0,0 +1,24 @@
+---
+- job:
+    name: Run_Metadata_Tagger
```
thomasleese: I wonder if should call this `Run Rummager/Search Metadata Tagger` just in case it's unclear what it's for.

PR #8636, opened 2019-02-05
Add a default SRID to CKAN config

elliotcm:
The CKAN database contains a constraint which requires all SRIDs
in geometry columns to be set to 4258.  This constraint was copied over
with the DB dump from CKAN in Bytemark, where it was presumably set manually.

Without this setting, the CKAN spatial extension will default new geometric
packages (INSPIRE/GEMINI) to SRID 4326, which will then conflict with the
constraint requiring SRID 4258.

This setting overrides the default and will allow spatial packages to be
harvested.

https://trello.com/c/qOelMdaw/743-pull-in-the-expected-data-when-we-harvest-geospatial-data-strongly-recommend-pairing


elliotcm:
Tested on integration.

PR #8651, opened 2019-02-07
Increase frontend app 5xx thresholds

thomasleese:
Currently we're experiencing higher than usual 5xx rate for our frontend apps due to the AWS migration of content-store.

It makes it harder to understand which alerts need action and which don't because there are constant alerts for high 5xx rates in our frontend apps, when in reality there isn't much we can do about it until the apps are migrated to AWS. Fastly is still able to serve requests from the mirrors, so our overall 5xx rate hasn't increased by much.

This is very much designed as a temporary measure and should be reverted immediately after the migration has completed. I picked the numbers by looking at the graphs and seeing what would make the alerts less noisy.


bevanloon:
@edwardkerry 👍 good idea. I've captured it on our [trello](https://trello.com/c/1UC0sMaz/386-revert-https-githubcom-alphagov-govuk-puppet-pull-8651)

PR #8673, opened 2019-02-12
Trello-2020: Plek envvar for whitehall on content-store and draft-content-store

ronocg:
When router-api and its database migrate to AWS, whitehall is staying in
carrenza, therefore the whitehall backends in the router-api DB will need
to have the associated publishing.service.gov.uk domain.
This has been completed already by manually setting the whitehall backend
in the router api database with the correct domain.
BUT these could possibly be updated at any time from content-store, so
content-store and draft-content-store need the PLEK envvars in place for
the whitehall-frontend backend URI.

The end state for the variables should be;

staging:
https://whitehall-frontend.staging.publishing.service.gov.uk

production:
https://whitehall-frontend.publishing.service.gov.uk

Also because this parameter will be undef for other environments other than
aws staging and aws production they will not be laid down for those
other envs.

@jstandring-gds <julian.standring@digital.cabinet-office.gov.uk>

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -184,6 +184,7 @@ task :check_consistency_between_aws_and_carrenza do
     govuk::apps::content_publisher::db::allow_auth_from_lb
     govuk::apps::content_publisher::db::lb_ip_range
     govuk::apps::content_publisher::db::rds
+    govuk::apps::content_store::plek_service_whitehall_frontend_uri
     govuk::apps::content_tagger::db::allow_auth_from_lb
     govuk::apps::content_tagger::db::lb_ip_range
     govuk::apps::content_tagger::db::rds
--- a/hieradata_aws/production.yaml
+++ b/hieradata_aws/production.yaml
@@ -122,6 +122,7 @@ govuk::apps::content_publisher::google_tag_manager_preview: "env-7"
 govuk::apps::content_store::app_domain: publishing.service.gov.uk
 govuk::apps::content_store::performance_platform_big_screen_view_url: 'https://performance-platform-big-screen-view-production.cloudapps.digital'
 govuk::apps::content_store::performance_platform_spotlight_url: 'https://performance-platform-spotlight-production.cloudapps.digital'
+govuk::apps::content_store::plek_service_whitehall_frontend_uri: 'https://whitehall-frontend.publishing.service.gov.uk'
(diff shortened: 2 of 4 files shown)


modules/govuk/manifests/node/s_content_store.pp
```diff
@@ -17,5 +17,12 @@
     include ::hosts::default
     include ::hosts::backend_migration
     include icinga::client::check_pings
+
+    $app_domain = hiera('app_domain')
+
+    govuk_envvar {
+    'PLEK_SERVICE_WHITEHALL_FRONTEND_URI': value  => "https://whitehall-frontend.${app_domain}";
+    'PLEK_SERVICE_WHITEHALL_URI': value  => "https://whitehall.${app_domain}";
+    }
```
cbaines: I'm not quite following this. What value are you expecting without this change for the Whitehall Frontend backend, and what's the desired value?

ronocg: Without this change the value would be <whitehall-frontend|whitehall>.<production|staging>.govuk-internal.digital  after this change the value should be <whitehall-frontend|whitehall>.<production|staging>.publising.service.gov.uk... 
from speaking with Seb, this should be enough to get the backend URL correct if content-store was to update the router-api db

cbaines: Right, thanks, that makes sense.

In my mind, it would be preferable to keep this configuration with the app, rather than the node (so in this file https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/content_store.pp ) as that's a bit more specific. At the moment, this change will only affect the "live" content store, not the "draft" content store, as it runs on a different node, whereas I think the Content Store configuration overall should be changed.

One other thing, I believe the Content Store will change the Router/Router API backend information everytime it does something with a relevant bit of content, so if someone publishes something in Whitehall, the routing information will be updated. So in my mind, it's maybe more common than you're considering.

ronocg: right thanks Chris, understood... we forgot completely about the draft-content-store.
Also, the app_domain for both the content_store and draft_content_store will also need to change from (staging).publishing.service.gov.uk to <staging|production>.govuk-internal.digital  ... but that will need to be done at migration time.

PR #8679, opened 2019-02-13
 Production content store app domain update

jstandring-gds:
# Context

As part of the migration we need to update the app domain for the content store
so that the content store is aware of the new app (frontend) locations.

# Decision

Remove the hiera data so that we fall back to the default defined in the govuk::deploy::config class.

### DIFF
--- a/hieradata_aws/production.yaml
+++ b/hieradata_aws/production.yaml
@@ -119,7 +119,6 @@ govuk::apps::content_publisher::aws_s3_bucket: "govuk-production-content-publish
 govuk::apps::content_publisher::google_tag_manager_auth: "O0DrItJxJeQ5Q2W6YCZzvw"
 govuk::apps::content_publisher::google_tag_manager_id: "GTM-NQXC4TG"
 govuk::apps::content_publisher::google_tag_manager_preview: "env-7"
-govuk::apps::content_store::app_domain: publishing.service.gov.uk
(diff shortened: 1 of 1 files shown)


ronocg:
This will be merged as part of the migration of the frontends to AWS

hieradata_aws/production.yaml
```diff
@@ -119,7 +119,7 @@ govuk::apps::content_publisher::aws_s3_bucket: "govuk-production-content-publish
 govuk::apps::content_publisher::google_tag_manager_auth: "O0DrItJxJeQ5Q2W6YCZzvw"
 govuk::apps::content_publisher::google_tag_manager_id: "GTM-NQXC4TG"
 govuk::apps::content_publisher::google_tag_manager_preview: "env-7"
-govuk::apps::content_store::app_domain: publishing.service.gov.uk
+govuk::apps::content_store::app_domain: production.govuk-internal.digital
```
cbaines: As the GOVUK_APP_DOMAIN is set for all apps (through `/etc/govuk/env.d`), removing this value from the hieradata should have the same effect. Perhaps do that? Or maybe keep this, and add a comment saying that it can be removed in the future?

jstandring-gds: I like the idea of this being removed... so I'll follow up and do the same in staging and force a change on this PR to remove the line. Thanks for checking and commenting!

PR #8689, opened 2019-02-14
License Finder Frontend does not use cookies

erino:
Session cookies are [disabled in License Finder](https://github.com/alphagov/licence-finder/blob/master/config/initializers/session_store.rb#L3), so there is no need for this rule to not strip cookies.

### DIFF
--- a/modules/varnish/templates/default.vcl.erb
+++ b/modules/varnish/templates/default.vcl.erb
@@ -82,9 +82,8 @@ sub vcl_recv {
   # Strip cookies from inbound requests. Corresponding rule in vcl_fetch{}
   # For simplicity and security most applications should not use cookies.
   # With the exception of:
-  #   - Licensing
   #   - email-alert-frontend (for subscription management)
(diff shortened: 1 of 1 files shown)


erino:
Ah, although when I browse [/apply-for-a-licence](https://www.gov.uk/apply-for-a-licence) there are no cookies set, and looking through the code in Licensify [/frontend](https://github.com/alphagov/licensify/tree/master/frontend) confirms this.
On this basis can the PR be merged?, or should I update the comment to make it clearer that it refers to Licensify.

rubenarakelyan:
Licensify has this though which looks to me like a session cookie in the frontend: https://github.com/alphagov/licensify/blob/7826858e8bb655c901f15e55579b878b5d305fee/frontend/conf/application.conf#L62

erino:
Yep, and I see the session cookie are used for the payment flow https://github.com/alphagov/licensify/blob/master/frontend/app/uk/gov/gds/licensing/frontend/controllers/GenericPaymentController.scala#L37

PR #8694, opened 2019-02-15
Increase nf_conntrack and file descriptor limits

rubenarakelyan:
This commit increases the `nf_conntrack_max` limit and the file descriptor limit on the cache machines to 65536. This allows for more simultaneous IP connections to be maintained per machine, and therefore for the router to be able to serve more concurrent connection

### DIFF
--- a/modules/govuk/manifests/node/s_cache.pp
+++ b/modules/govuk/manifests/node/s_cache.pp
@@ -45,11 +45,19 @@ class govuk::node::s_cache (
   include router::gor
   include nscd
 
+  # The next two blocks increase file descriptor and IP connection
+  # tracking limits on cache machines so the router can handle more
+  # simultaneous connections. The router is the only app on these
+  # machines, so we can afford for it to use more resources.
   limits::limits { 'deploy_nofile_router':
     ensure     => present,
     user       => 'deploy',
     limit_type => 'nofile',
-    both       => 65535,
+    both       => 65536,
+  }
+
+  govuk_harden::sysctl::conf { 'cache-nf-conntrack-limit':
+    content => "net.netfilter.nf_conntrack_max = 65536\n",
   }
 
   if $router_as_container {


modules/govuk/manifests/node/s_cache.pp
```diff
@@ -45,13 +45,29 @@
...
+  # machines, so we can afford for it to use more resources.
   limits::limits { 'deploy_nofile_router':
     ensure     => present,
     user       => 'deploy',
     limit_type => 'nofile',
     both       => 65535,
   }
 
+  govuk_harden::sysctl::conf { 'cache-nf-conntrack-limit':
+    content => "net.netfilter.nf_conntrack_count = 131072\n",
```
thomasleese: I think `nf_conntrack_count` is the current count of the number of connections, whereas we should be setting the `nf_conntrack_max` variable here to increase the limit. I also wonder if we should look at increasing the `nf_conntrack_buckets` variable instead because the docs say that `nf_conntrack_max` will be set to 4 times that value.

https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt

modules/govuk/manifests/node/s_cache.pp
```diff
@@ -45,13 +45,29 @@
...
+  }
+
+  # The hash size is a quarter of the size of the nf_conntrack limit set above
+  rclocal::script { 'cache-nf-conntrack-hashsize':
+    ensure   => 'present',
+    content  => "echo 32768 > /sys/module/nf_conntrack/parameters/hashsize\n",
```
thomasleese: https://serverfault.com/questions/482480/where-is-the-correct-place-to-set-net-netfilter-nf-conntrack-buckets seems to suggest that setting this is the same as configuring the `nf_conntrack_buckets` variable, so I'm wondering if we can just do that in the block above and avoid having to set both the max and the hash size (since max is calculated as four times the hash size).

modules/govuk/manifests/node/s_cache.pp
rubenarakelyan: Nice one, although it's weird that the cache machines have 16GB RAM but only 16384 buckets...

thomasleese: Yeah, I noticed that too and thought it was weird... 🤔

rubenarakelyan: I've manually upped the limit

modules/govuk/manifests/node/s_cache.pp
rubenarakelyan: I've removed this completely since, as you say, this should be set automatically

PR #8714, opened 2019-02-21
Add additional nginx config to route asset traffic

schmie:
- Mid-migration we serve assets both from Carrenza and AWS at the same
time.

- For this we introduce an assets-carrenza and draft-assets-carrenza
proxy configuration

solo: @schmie


schmie:
Had a bit of back and forth with a rebase and revert - sth went wrong with git - checked out again and files another pr

PR #8715, opened 2019-02-21
[DO NOT MERGE] Revert nf_conntrack_max setting

rubenarakelyan:
This commit reverts the `nf_conntrack_max` setting to the OS default to prevent out of memory issues with other apps running on the cache machines.


rubenarakelyan:
Looks like the original change works

PR #8722, opened 2019-02-22
Add monitoring and alerts for AWS Managed Elasticsearch cluster

brucebolt:
- Adds metrics for the AWS Managed Elasticsearch instance to Graphite.
- Set up alerting through Incinga (alerts are defined in `gds-nagios-plugins`)

Draft until the PR on https://github.com/alphagov/nagios-plugins is merged and a new version of `gds-nagios-plugins` is released.

Trello card: https://trello.com/c/ZXD7qdXf/59-set-up-monitoring-for-managed-elasticsearch


modules/collectd/templates/etc/collectd/conf.d/elasticsearch-aws.conf.erb
```diff
@@ -0,0 +1,100 @@
+<Plugin curl_json>
```
barrucadu: Are all these metrics being reported successfully now?

brucebolt: Yes, they were previously being collected as a `counter` (so the values were constantly being added).  I've changed that to `gauge`, which reports just the raw value.

modules/govuk_search/manifests/monitoring.pp
```diff
@@ -0,0 +1,20 @@
+# == Class govuk_search::monitoring
+#
+# Setup monitoring for a remote Elasticsearch cluster
+#
+class govuk_search::monitoring () {
+
+  $app_domain_internal = hiera('app_domain_internal')
+
+  class { 'collectd::plugin::elasticsearch':
+    es_host  => "elasticsearch5.${app_domain_internal}",
```
barrucadu: I think the host should be a parameter, as this will inevitably change to `elasticsearch6` at some point

PR #8732, opened 2019-02-26
Add finder frontend metrics

Seraphiana:
This adds some new metrics to the grafana dashboard for finder frontend. 
- memcached free and memcached used - these require specifying the instance prefix, as apps can live on instances that do not share their name.
I am basing this on the assumption that the architecture for prod and int are equivalent in terms of naming processes and application location, please let me know if this is incorrect.
- content store, rummager and registry request time - these three monitor how long individual calls take from finder frontend to the various backends. 
They aren't as whitelabeled as I would like, but since the graphite keys are specified in a separate codebase I'm not convinced of the value of passing in more values into the manifest. If anyone else wants to use these same metrics for a different application then it might be worth refactoring how its handled.


modules/grafana/manifests/dashboards/application_dashboard.pp
```diff
@@ -105,6 +106,17 @@
     $elasticsearch_stats = []
   }
 
+  if $show_external_request_time {
+    $external_request_row = [
+      [
+        'content_store_request_time',
+        'registry_request_time',
```
cbaines: I know what the Content Store and Rummager are, but what's the registry?

modules/grafana/manifests/dashboards/application_dashboard.pp
```diff
@@ -106,6 +108,16 @@
     $elasticsearch_stats = []
   }
 
+  if $show_memcached {
+    $memcached_row = [
+      [
+        'memcached_free',
+        'memcached_used']
```
cbaines: Have you got a preview of what this looks like? It seems a little odd to have two panels, that sound like they display connected data, as if more memory is used in memcached, then there will be less space free?

Seraphiana: I'm open to opinions on this. The original card wanted both, and trying to put them on the same graph (stacking) just shows a huge wall of one colour, due to having an extra 3 zeros I imagine. I imagine the free is more useful to display than the used, so I'm happy to remove one of them. 
I'll create the panel again quickly and add some screenshots.

modules/grafana/manifests/dashboards/application_dashboard.pp
Seraphiana: It’s a cache of information that we need to be able to provide filtering for our finders.

modules/grafana/manifests/dashboards/application_dashboard.pp
Seraphiana: screen shot 2019-02-26 at 12 14 17

Seraphiana: this is admittedly on int rather than prod, so the numbers are probably pretty different.

cbaines: I'd suggest trying a panel per machine, using the repeating panels feature. Also, have a look at the Machine dashboard, at how it displays memory usage. I think it's visualising similar data and uses the stacking and conditional formatting.

cbaines:
I've only just noticed your comments in the Pull Request description. If you think that's worth including, its probably worth including in the respective commits as well.

modules/grafana/manifests/dashboards/application_dashboard.pp
Seraphiana: cool, thanks, will check it out

Seraphiana:
> I've only just noticed your comments in the Pull Request description. If you think that's worth including, its probably worth including in the respective commits as well.

I'm not sure it is worth including in the git log: It adds bloat which is only related to the ongoing AWS migration, and I don't think having questions of implementation details is useful information to tie to a specific commit. 
With regards to whitelabeling it might be useful in the future. I'll add some comments in the code or something.

Seraphiana:
new screenshot:

screen shot 2019-02-27 at 12 57 55

modules/grafana/manifests/dashboards/application_dashboard.pp
Seraphiana: On reflection, I don't think its that valuable to show the stacked memory: the things we want to monitor is whether we're running out of memory, and whether we have spikes/changes. due to the scale stacking it would hide these changes to casual observers, and knowing the memory used and the memory free isn't any more useful than knowing just one, hence the decision to only have the single memcached graph

PR #8735, opened 2019-02-26
fix icinga reboot formatting

fredericfran-gds:
# Context

In the AWS Icinga, there is a check for the machines that need to be reboot. The check displays a list of such machines in the format of their AWS DNS names only: 
`ip-x-x-x-x.eu-west-1.compute.internal` which does not give immediate insight which machines need to be rebooted.

The new display format will be:
<puppet_class_name> with hostname:	ip-x-x-x-x.eu-west-1.compute.internal

# Decisions
1. modify the `govuk_node_list` to have the new option/subcommand `--list` which will return a list of hosts in the format of `aws_dns_name`:`puppet_class_name`
2. make the `check_reboots_required` script to hydrate/augment the hostname with the information gathered in Step 1 so that to achieve a better list display for the icinga alert

### DIFF
--- a/modules/govuk_scripts/templates/govuk_node_list_aws.erb
+++ b/modules/govuk_scripts/templates/govuk_node_list_aws.erb
@@ -42,6 +42,31 @@ def ec2_nodes(stackname, nodeclass=None):
 
     return(hosts)
 
+def ec2_nodes_list_with_puppet_class(stackname):
+    client = boto3.client('ec2', region_name=region)
+
+    response = client.describe_instances(
+                   Filters=[
+                          {'Name': 'tag:aws_stackname', 'Values': [stackname]},
+                          {'Name': 'instance-state-name','Values': ['running']}
+                   ]
+               )
+
+    nodes = response['Reservations']
+
+    return [
+        "%s:%s" % (
+            instance['Instances'][0]['PrivateDnsName'],
+            next(
+                filter(
+                    lambda t: t['Key'] == 'aws_hostname',
+                    instance['Instances'][0]['Tags']
+                )
+            )['Value']
+        )
+        for instance in list(nodes)
+    ]
+
 def ec2_classes(stackname):
     client = boto3.client('ec2', region_name=region)
 
@@ -110,6 +135,9 @@ def main():
         for instance in instance_ids:
             hosts.append(get_hostname_by_id(instance[host_key]))
 
+    elif opts.nodes_list_with_puppet_class:
+        hosts = ec2_nodes_list_with_puppet_class(stackname)
+
     else:
         hosts = ec2_nodes(stackname)
 
@@ -151,6 +179,13 @@ parser.add_option(
     help='List the available classes',
     action='store_true',
     dest='classes',
+)
+
+parser.add_option(
+    '--with-puppet-class',
+    help='List the nodes in the format of aws_fqdn:name',
+    action='store_true',
+    dest='nodes_list_with_puppet_class',
 )
 
 if __name__ == '__main__':
--- a/modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
+++ b/modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
@@ -1,9 +1,33 @@
 #!/usr/bin/env ruby
 
+def create_machines_hash
+  machines_names_hash = Hash.new
+
+  `/usr/local/bin/govuk_node_list --with-puppet-class`.each_line do |machine|
+    hostname, name = machine.chomp.split(":")
+    machines_names_hash[hostname] = name
+  end
+
+  return machines_names_hash
+end
+
+def hostname_with_node_name(hostname, machines_names_hash)
+  name = machines_names_hash[hostname]
+
+  if name
+    "%s with hostname: %s" % [name, hostname]
+  else
+    hostname
+  end
+
+end
+
 all_machines = `/usr/local/bin/govuk_node_list`.split("\n")
 
 unsafe_machines = `/usr/local/bin/govuk_node_list --puppet-class govuk_safe_to_reboot::careful,govuk_safe_to_reboot::no`.split("\n")
 safe_machines = `/usr/local/bin/govuk_node_list --puppet-class govuk_safe_to_reboot::yes,govuk_safe_to_reboot::overnight`.split("\n")
+
+machines_names_hash = create_machines_hash
 
 # Safety check that we haven't added other safe to reboot options
 # that we don't account for here.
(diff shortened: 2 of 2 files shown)


modules/govuk_scripts/templates/govuk_node_list_aws.erb
```diff
@@ -153,5 +175,12 @@ parser.add_option(
     dest='classes',
 )
 
+parser.add_option(
+    '--list',
```
cbaines: Maybe `--include-puppet-class` or `--output-puppet-class` would be a more informative option? Given that just running `govuk_node_list` without any arguments will list things.

modules/govuk_scripts/templates/govuk_node_list_aws.erb
```diff
@@ -42,6 +42,25 @@ def ec2_nodes(stackname, nodeclass=None):
...
+               )
+
+    nodes = response['Reservations']
+
+    hosts = []
+    for instance in nodes:
+        node_name=list(filter(lambda t: t['Key'] == 'aws_hostname', instance['Instances'][0]['Tags']))[0]['Value']
+        hosts.append((instance['Instances'][0]['PrivateDnsName']+":"+node_name))
+
+    return(hosts)
```
cbaines: Using a list comprehension would avoid mutating the hosts variable. Also, you can use `next` to get the first element of a generator without converting it to a list. Something like this...

```python
    return [
        "%s:%s" % (
            instance['Instances'][0]['PrivateDnsName'],
            next(
                filter(
                    lambda t: t['Key'] == 'aws_hostname',
                    instance['Instances'][0]['Tags']
                )
            )['Value']
        )
        for instance in nodes
    ]
```

modules/govuk_scripts/templates/govuk_node_list_aws.erb
fredericfran-gds: how about `--node-list-with-puppet-class`?

modules/govuk_scripts/templates/govuk_node_list_aws.erb
fredericfran-gds: nice, I need to modify and test it! I followed too much what was done `ec2_nodes` function

fredericfran-gds:
@cbaines can you check the ruby code to see if they meet standards? I have never done ruby before but I guess there was a linter in the puppet codebase.

modules/govuk_scripts/templates/govuk_node_list_aws.erb
cbaines: Looking at the whole command `govuk_node_list --node-list-with-puppet-class`, the second node list bit seems a bit redundant to me?

cbaines: (but that's definitely more descriptive than `--list` :+1: )

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
```diff
@@ -1,10 +1,34 @@
 #!/usr/bin/env ruby
 
+def create_machines_hash
+  machines_list= `/usr/local/bin/govuk_node_list --list`.split("\n")
```
cbaines: I'd put a space before the `=` (so `machines_list = `/...`). Also, perhaps try `.lines` rather than `.split("\n")`.

You could maybe go even further, and do something the following and skip the `machines_list` variable:
```ruby
`/usr/local/bin/govuk_node_list ...`.each_line do |machine|
  ...
```

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
```diff
@@ -1,10 +1,34 @@
...
+  machines_list.each do |machine|
+    machine_fqn_name = machine.split(":")
+    machines_names_hash[machine_fqn_name[0]]=machine_fqn_name[1]
+  end
+
+  return machines_names_hash
+end
+
+def hydrate_hostname(hostname, machines_names_hash)
+  name=machines_names_hash[hostname]
```
cbaines: I'd put some spaces around the `=` here.

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
fredericfran-gds: great, I will do the last option.

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
fredericfran-gds: cool

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
```diff
@@ -1,10 +1,34 @@
...
+
+  if (name != nil)
+    return "%s with hostname: %s" % [name, hostname]
+  end
+
+  return hostname
```
cbaines: `return` is implicit in Ruby for the last expression in a function, so you could replace the above few lines by this I think.

```ruby
if name
  "%s with hostname: %s" % [name, hostname]
else
  hostname
end
```

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
```diff
@@ -1,10 +1,34 @@
...
+
+  machines_list.each do |machine|
+    machine_fqn_name = machine.split(":")
+    machines_names_hash[machine_fqn_name[0]]=machine_fqn_name[1]
+  end
+
+  return machines_names_hash
+end
+
+def hydrate_hostname(hostname, machines_names_hash)
```
cbaines: How about `hostname_with_node_name` as the name for this function? That might be a little bit more descriptive.

cbaines:
> @cbaines can you check the ruby code to see if they meet standards? I have never done ruby before but I guess there was a linter in the puppet codebase.

Sure, Ruby seems to specialise in having many ways of doing everything, but I've tried to make some hopefully helpful comments.

fredericfran-gds:
@cbaines ready fore review round 2

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
```diff
@@ -1,10 +1,34 @@
 #!/usr/bin/env ruby
 
+def create_machines_hash
+  machines_names_hash = Hash.new
+
+  `/usr/local/bin/govuk_node_list --with-puppet-class`.each_line do |machine|
+    machine_fqn_name = machine.split(":")
```
cbaines: I think the following works in Ruby:

```ruby
    domain_name, puppet_class = machine.chomp.split(":")
    machine_names_hash[domain_name] = puppet_class
```

The benefit being that the two values being moved around have more meaningful names.

cbaines:
> @cbaines ready fore review round 2

I've made one more comment, but generally I think the code looks good :+1: As for the commits, I think it would be neater to squash your review comments commit in to your initial commit. Also it would be good to include some (or all) of the useful information you've put in the GitHub Pull Request in the commit message.

fredericfran-gds:
> > @cbaines ready fore review round 2
> 
> I've made one more comment, but generally I think the code looks good 👍 As for the commits, I think it would be neater to squash your review comments commit in to your initial commit. Also it would be good to include some (or all) of the useful information you've put in the GitHub Pull Request in the commit message.

cool, done now and I forced push manually rather than the github button.

cbaines:
> cool, done now and I forced push manually rather than the github button.

Great. It would still be useful to have more information in the commit message though, particularly about the motivation for the changes.

PR #8739, opened 2019-02-26
Use exec for installing nagios-plugin

theseanything:
Due to issues with older versions of pip, we need to specify the `--index-url` flag. However, because we are running an older version of puppet (v3) - it doesn't support the use of flags via the `install_options`. Hence to overcome this we have specified the install task as an exec command. 

There is also a requirement for `setuptools-pep8` to install `gds-nagios-plugins`.

### DIFF
--- a/modules/monitoring/manifests/client.pp
+++ b/modules/monitoring/manifests/client.pp
@@ -11,11 +11,10 @@ class monitoring::client (
   include collectd
   include collectd::plugin::tcp
 
-  package {'gds-nagios-plugins':
-    ensure          => '1.5.0',
-    provider        => 'pip',
-    require         => Package['update-notifier-common'],
-    install_options => '--index-url https://pypi.python.org/pypi',
+  exec { 'gds-nagios-plugins':
+    path    => ['/usr/local/bin', '/usr/bin', '/bin'],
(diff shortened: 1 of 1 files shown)


modules/monitoring/manifests/client.pp
```diff
@@ -11,11 +11,15 @@
...
-    require         => Package['update-notifier-common'],
-    install_options => '--index-url https://pypi.python.org/pypi',
+  exec { 'setuptools-pep8':
+    path    => ['/usr/local/bin', '/usr/bin', '/bin'],
+    command => 'pip install setuptools-pep8 --index-url https://pypi.python.org/pypi',
+  }
+  
+  exec { 'gds-nagios-plugins':
+    path    => ['/usr/local/bin', '/usr/bin', '/bin'],
+    command => 'pip install gds-nagios-plugins==1.5.0 --index-url https://pypi.python.org/pypi',
```
kevindew: I don't have a huge amount (or any) pip knowledge - but I wonder if we could install both setuptools-pep8 and gds-nagios-plugins with the same line like `pip install setuptools-pep8 gds-nagios-plugins==1.5.0 --index-url https://pypi.python.org/pypi`

theseanything: Yup good shout - I've just updated the commit

PR #8740, opened 2019-02-26
Fix etcd address

fredericfran-gds:


fredericfran-gds:
will delete this PR because it had commits from other PRs

PR #8748, opened 2019-02-27
Automatically run the CKAN reindex in staging and integration

thomasleese:
This makes the CKAN reindex run daily in staging and integration to be the final step of the data sync.

After testing https://github.com/alphagov/govuk-puppet/pull/8737 on integration, I've come to the realisation that this also won't solve the problem because the CKAN machine needs access to the S3 bucket which contains the backups. It feels wrong to give the CKAN machine the same level as access to all the backups as the db_admin machine, for what is already a workaround the fact that the db_admin machine can't talk to the CKAN machine. Instead, a simpler, albeit less clean, solution is to run the reindex as a separate process unrelated to the database sync.

[Trello Card](https://trello.com/c/RiOpZ1qs/762-get-ckan-database-up-and-running-on-staging)

### DIFF
--- a/modules/govuk/manifests/apps/ckan/cronjobs.pp
+++ b/modules/govuk/manifests/apps/ckan/cronjobs.pp
@@ -1,6 +1,13 @@
 # == Class: govuk::apps::ckan::cronjobs
 #
-class govuk::apps::ckan::cronjobs {
+# [*enable_solr_reindex*]
+#   Enable automatic Solr reindexing, this is useful to run after the data
+#   sync.
+#   Default: false
+#
+class govuk::apps::ckan::cronjobs(
+  $enable_solr_reindex = false,
+) {
 
   govuk::apps::ckan::paster_cronjob { 'harvester run':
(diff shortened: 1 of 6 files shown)


modules/govuk/manifests/apps/ckan/cronjobs.pp
```diff
@@ -14,4 +21,13 @@
     hour           => '2',
   }
 
+  if $enable_solr_reindex {
+    govuk::apps::ckan::paster_cronjob { 'harvester reindex':
```
elliotcm: Is there a situation where we'd want to turn this off quickly by flipping this flag?  If so should the `else` of this conditional be to remove the cron job rather than decline to add it?  If there's a distinction here.

thomasleese: Good idea, I've pushed up a commit that allows `govuk::apps::ckan::paster_cronjob` to be `ensure => absent`.

PR #8751, opened 2019-02-27
Finish removing the event store

cbaines:
This follows on from previous work to remove the event-store [1].

1: 5e719271968c1025c28f2f824bbf10335b59e1c7

The relevant manifests are deleted, along with the hieradata.


cbaines:
Follows on from https://github.com/alphagov/govuk-puppet/pull/8510

PR #8756, opened 2019-02-28
Fix real ip haeders for assets nginx 

schmie:
- We were missing a RealIP header parameter for the assets backend-lb config
- This caused Puppet to fail on the machine

solo: @schmie


schmie:
True-Client-Ip is the default for prod environments for the router/cache asset handling

PR #8764, opened 2019-03-01
Enable router-backend db push backup script on carrenza prod

rtrinque:
As part of the migration of router to AWS


schmie:
No longer required

PR #8767, opened 2019-03-01
Migration: Gor Replay for Integration

ronocg:
We will soon be migrating the cache router services from carrenza to aws
for staging and production.

So in order to continue having traffic in integration we will need to
replay the aws staging cache router traffic to integration.

This will need to be merged at migration time, when the cache routers
are turned off in carrenza.

We only want to replay traffic from 2 cache nodes in staging to
Integration, as there are only 2 cache nodes in Integration and we do
not want to overload.


ronocg:
This needs to be merged at migration time

PR #8769, opened 2019-03-01
do not use govuk_node_list --with-puppet-class in carrenza because no such option exist

fredericfran-gds:
# Context

The icinga reboot alert check was changed to be more informative in AWS but this should not happen in Carrenza because the `govuk_node_list --with-puppet-class` does not work in Carrenza. 

# Decisions
1. modify icinga reboot alert ruby script to not call `govuk_node_list --with-puppet-class`

### DIFF
--- a/modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
+++ b/modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
@@ -1,6 +1,8 @@
 #!/usr/bin/env ruby
 
 def create_machines_hash
+  return {} unless File.exist? '/etc/facter/facts.d/aws_environment.txt'
+
   machines_names_hash = Hash.new
 
   `/usr/local/bin/govuk_node_list --with-puppet-class`.each_line do |machine|


cbaines:
How about adding it as an option to the Carrenza script, but one that does nothing, just for compatibility?

That would avoid any other issues in other places with this option, and not require hard coding the error message in this file.

fredericfran-gds:
> How about adding it as an option to the Carrenza script, but one that does nothing, just for compatibility?
> 
> That would avoid any other issues in other places with this option, and not require hard coding the error message in this file.

I now do a check if we are in aws environment or now. I think it is better than adding an empty option in carrenza, people might think it is a bug.

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
```diff
@@ -3,9 +3,11 @@
 def create_machines_hash
   machines_names_hash = Hash.new
 
-  `/usr/local/bin/govuk_node_list --with-puppet-class`.each_line do |machine|
-    hostname, name = machine.chomp.split(":")
-    machines_names_hash[hostname] = name
+  if File.file?("/etc/facter/facts.d/aws_environment.txt")
```
cbaines: `File.exist?(...` might be a bit more meaningful here.

Also, it's a bit obscure putting this conditional within `create_machines_hash`, but refactoring to avoid this might be a bit too much. Maybe just simplify the conditional to make it more explicit that the function just returns an empty hash in the non-AWS case, for example:

```ruby
def create_machines_hash
  return {} unless File.exist? '/etc/facter/facts.d/aws_environment.txt'

  ...
```

cbaines:
> > How about adding it as an option to the Carrenza script, but one that does nothing, just for compatibility?
> > That would avoid any other issues in other places with this option, and not require hard coding the error message in this file.
> 
> I now do a check if we are in aws environment or now. I think it is better than adding an empty option in carrenza, people might think it is a bug.

Ok, sure, I think that's ok.

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
fredericfran-gds: done, thanks!

cbaines:
Also, please consider rewriting the commit in a way that meets the styleguide on commit messages [1]. In particular, the first line of the commit here is too long.

1: https://github.com/alphagov/styleguides/blob/master/git.md#structure

PR #8773, opened 2019-03-04
Speed up the Deploy Terraform GOVUK AWS Jenkins job

cbaines:
By using a shallow clone, and not wiping the workspace. Instead, clean
the workspace before starting jobs.


cbaines:
> Merge when green. It'd be interesting to update this PR with the run times before and after for future reference.

Sure, will do.

cbaines:
@deanwilson anecdotally looking at recent builds of this job, there doesn't seem to be much of a difference.

deanwilson:
Thanks for checking. If nothing else it'll save a little bandwidth on each pull.

PR #8778, opened 2019-03-04
Configure Sidekiq & Redis for Content Publisher

emmabeynon:
For https://trello.com/c/rXTLvzSu/685-scheduling-is-sent-to-the-publishing-api

### DIFF
--- a/development-vm/Procfile
+++ b/development-vm/Procfile
@@ -185,3 +185,4 @@ search-api-govuk-index-listener:  govuk_setenv search-api ./run_in.sh ../../sear
 search-api-bulk-reindex-listener: govuk_setenv search-api ./run_in.sh ../../search-api    bundle exec rake message_queue:bulk_insert_data_into_govuk
 # sidekiq-monitoring for search-api uses 3234
 content-data-api:      govuk_setenv content-data-api ./run_in.sh ../../content-data-api bundle exec rails server -p 3235
+# sidekiq-monitoring for content-publisher uses 3236
--- a/development-vm/Procfile
+++ b/development-vm/Procfile
@@ -186,3 +186,5 @@ search-api-bulk-reindex-listener: govuk_setenv search-api ./run_in.sh ../../sear
 # sidekiq-monitoring for search-api uses 3234
 content-data-api:      govuk_setenv content-data-api ./run_in.sh ../../content-data-api bundle exec rails server -p 3235
 # sidekiq-monitoring for content-publisher uses 3236
+# byebug web listener for content-publisher uses 3237
+# byebug sidekiq listener for content-publisher uses 3238
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -414,6 +414,8 @@ govuk::apps::content_publisher::db_allow_prepared_statements: false
 govuk::apps::content_publisher::db::backend_ip_range: "%{hiera('environment_ip_prefix')}.3.0/24"
 govuk::apps::content_publisher::jwt_auth_secret: "%{hiera('jwt_auth_secret')}"
 govuk::apps::content_publisher::aws_region: "eu-west-1"
+govuk::apps::content_publisher::redis_host: "%{hiera('sidekiq_host')}"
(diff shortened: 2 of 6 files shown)


development-vm/Procfile
```diff
@@ -185,3 +185,5 @@ search-api-govuk-index-listener:  govuk_setenv search-api ./run_in.sh ../../sear
 search-api-bulk-reindex-listener: govuk_setenv search-api ./run_in.sh ../../search-api    bundle exec rake message_queue:bulk_insert_data_into_govuk
 # sidekiq-monitoring for search-api uses 3234
 content-data-api:      govuk_setenv content-data-api ./run_in.sh ../../content-data-api bundle exec rails server -p 3235
+content-publisher-sidekiq:               govuk_setenv content-publisher ./run_in.sh ../../content-publisher    bundle exec sidekiq -C ./config/sidekiq.yml
+# sidekiq-monitoring for content-publisher uses 3236
```
kevindew: Can we reserve ports 3237 and 3238 too for content publisher byebug? as per: https://github.com/alphagov/content-publisher/blob/master/Procfile.dev

I didn't realise someone had grabbed 3235 when I wrote those procfile changes so I'll have to up the port numbers by 1.

development-vm/Procfile
```diff
@@ -185,3 +185,5 @@ search-api-govuk-index-listener:  govuk_setenv search-api ./run_in.sh ../../sear
 search-api-bulk-reindex-listener: govuk_setenv search-api ./run_in.sh ../../search-api    bundle exec rake message_queue:bulk_insert_data_into_govuk
 # sidekiq-monitoring for search-api uses 3234
 content-data-api:      govuk_setenv content-data-api ./run_in.sh ../../content-data-api bundle exec rails server -p 3235
+content-publisher-sidekiq:               govuk_setenv content-publisher ./run_in.sh ../../content-publisher    bundle exec sidekiq -C ./config/sidekiq.yml
```
kevindew: Because of https://github.com/alphagov/govuk-puppet/blob/9906d082a5908dc3db3447c0a7439690c00caa0b/development-vm/Procfile#L175 running foreman start we don't actually need this line here as sidekiq will be run as part of that.

I'm not sure if there are advantages to running separate processes but if we do go down a separate process route I think we'd be better eschewing the conventions here and referencing the processes in our procfile. E.g. `foreman start worker -f Procfile.dev` as that decouples our configuration from this file.

development-vm/Pinfile
```diff
@@ -19,7 +19,8 @@ process :'contacts-admin' => [:'publishing-api', :whitehall]
 process :'content-audit-tool' => [:'publishing-api-read', :'content-audit-tool-sidekiq-google-analytics', :'content-audit-tool-sidekiq-publishing-api']
 process :'content-data-admin' =>[:'content-performance-manager']
 process :'content-performance-manager' => [:'publishing-api', :'content-performance-manager-sidekiq-default', :'content-performance-manager-publishing-api-consumer', 'content-performance-manager-bulk-import-publishing-api-consumer', 'content-store']
-process :'content-publisher' => [:'publishing-api', :'asset-manager']
+process :'content-publisher' => [:'publishing-api', :'asset-manager', :'content-publisher-sidekiq']
+process :'content-publisher-sidekiq'
```
kevindew: As per the comment on the process, I'm not sure we need this.

But if we do we'd expect it to have a dependency on publishing_api too.

development-vm/Procfile
emmabeynon: Sure no worries.  I had reserved 3235, but by the time I opened the PR it had been taken!

kevindew: ha oh good, that makes me feel better about getting mine wrong 😁

development-vm/Procfile
emmabeynon: Ok will remove

development-vm/Pinfile
emmabeynon: Will remove

emmabeynon:
@kevindew Your comments should be addressed now 👍

PR #8780, opened 2019-03-04
Update HSTS header check

sihugh:
This was updated in https://github.com/alphagov/govuk-puppet/pull/8774

This is currently alerting: https://alert.publishing.service.gov.uk/cgi-bin/icinga/extinfo.cgi?type=2&host=monitoring-1.management.publishing.service.gov.uk&service=Strict-Transport-Security+headers


sihugh:
Thanks 🥇

PR #8808, opened 2019-03-07
Adjust the Content Data API database configuration

cbaines:
Temporarily make the database name and user match up with the Content
Performance Manager. I'm hoping this will make it easier to sync the
Content Performance Manager database from where it currently resides,
to the new infrastructure for the Content Data API.

Once the new service is up and running, just prior to switching over
to it, we can rename the database, change the ownership, and remove
this configuration.


cbaines:
Trello card: https://trello.com/c/DaVCrYfO/1171-5-begin-to-set-up-the-content-performance-manager-under-the-new-name-content-data-api-in-integration-aws-staging-and-aws-product

PR #8809, opened 2019-03-07
Remove redundant jenkins job for Search API

benthorner:
This job is redundant as Search API doesn't exist in AWS production. It
makes sense to put this job in after Search API is created in prod.


hieradata_aws/class/production/jenkins.yaml
```diff
@@ -9,7 +9,6 @@ govuk_jenkins::config::theme_environment_name: 'AWS Production'
 govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::jobs::deploy_app
   - govuk_jenkins::jobs::deploy_puppet
-  - govuk_jenkins::jobs::enhanced_ecommerce_search_api
```
barrucadu: Are you just removing this job or the three others lower down?

benthorner: Good point - they're failing too, but not causing an alert. I guess that's just the job config. I've amended...

PR #8812, opened 2019-03-07
Push the Content Performance Manager database from Carrenza production to s3

cbaines:


cbaines:
Trello card: https://trello.com/c/2XFTquZv/1180-3-aws-migration-get-data-syncs-working

PR #8814, opened 2019-03-07
Add govuk_env_sync configuration for the Content Data API

cbaines:
The Content Data API is the new name for the Content Performance
Manager. At the same time as the name is being changed, the service is
also being migrated to AWS.

This commit adds both the data syncs to from AWS Production to AWS
Staging and Integration, but also a sync task to pull data from
Carrenza Production which is being used for the migration.

The govuk_env_sync script doesn't handle changing database ownership,
so to try and make the dual running situation easier, the Content Data
API in AWS is configured to use the
content_performance_manager_production database, and
content_performance_manager user. Immediately before switching to AWS,
the database can be renamed, and the ownership changed.


cbaines:
Trello card: https://trello.com/c/2XFTquZv/1180-3-aws-migration-get-data-syncs-working

PR #8818, opened 2019-03-08
Disable the content data api rabbitmq integration

cbaines:


cbaines:
This was a temporary change that wasn't reviewed in time to have any effect. The Content Data API is now reading from the queue, so this change is no longer necessary.

PR #8820, opened 2019-03-08
Delete removed users

rubenarakelyan:
This commit deletes the user manifests for users who have already been removed.


issyl0:
Bye! ❤️

PR #8825, opened 2019-03-08
Provision Elasticsearch 5.6.15 in the development VM

brucebolt:
This PR provisions Elasticsearch 5.6 in the development VM, by means of a Docker container, with the service available at `localhost:9205`.  It does not remove Elasticsearch 2 (this will be removed later as part of the tidying up process).

The reason for dockerising Elasticsearch is that it would enable a smoother upgrade to Elasticsearch 6.x, which is not supported by our version of Puppet (3).

Data replication is also updated to pull from a new S3 bucket containing snapshots from the AWS Managed Elasticsearch 5 instance.

Not to be merged until ES 5.6 is released to production and ES 2 code is removed from Puppet.

Trello card: https://trello.com/c/uHuP699X/69-ensure-es-remains-working-in-the-dev-vm

### DIFF
--- a/development-vm/replication/sync-aws-elasticsearch.sh
+++ b/development-vm/replication/sync-aws-elasticsearch.sh
@@ -17,7 +17,7 @@ fi
 
 shift $(($OPTIND-1))
 
-LOCAL_ES_HOST="http://localhost:9200/"
+LOCAL_ES_HOST="http://localhost:9205/"
 LOCAL_ARCHIVE_PATH="${DIR}/elasticsearch"
 
 status "Starting search index replication from AWS"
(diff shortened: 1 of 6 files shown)


modules/govuk/manifests/apps/search_api.pp
```diff
@@ -206,4 +206,9 @@
       value   => $oauth_secret;
   }
 
+  # Start Elasticsearch in docker if using dev VM
+  if $::govuk_node_class =~ /^development$/ {
+    include govuk_search::docker_elasticsearch
```
barrucadu: Should this instead be in the node manifest for the dev VM?

development-vm/replication/sync-aws-elasticsearch.sh
```diff
@@ -2,111 +2,119 @@
 
 set -eu
 
-USAGE_LINE="$0 [options] [index_name ...]"
-USAGE_DESCRIPTION="Download Elasticsearch index archives and import into another instance.
+export USAGE_LINE="$0 [options] [index_name ...]"
+export USAGE_DESCRIPTION="Download Elasticsearch index archives and import into another instance.
```
barrucadu: Do these need to be exported?

development-vm/replication/sync-aws-elasticsearch.sh
```diff
@@ -94,7 +99,7 @@ else
   status "Restoring data into Elasticsearch for $index_names"
 
   # put the snapshot in the docker container
-  sudo docker cp /var/govuk/govuk-puppet/development-vm/replication/${LOCAL_ARCHIVE_PATH}/. elasticsearch5:/usr/share/elasticsearch/import/
+  sudo docker cp /var/govuk/govuk-puppet/development-vm/replication/"${LOCAL_ARCHIVE_PATH}"/. elasticsearch5:/usr/share/elasticsearch/import/
```
barrucadu: Nitpick, but I'd prefer quoting the whole path,

```
"/var/govuk/govuk-puppet/development-vm/replication/${LOCAL_ARCHIVE_PATH}/."
```

development-vm/replication/sync-aws-elasticsearch.sh
brucebolt: Looks like I misinterpreted a shellcheck error.  The exports are not needed.

modules/govuk/manifests/apps/search_api.pp
brucebolt: Yes, of course.  Don't know why I left it in here :/.

PR #8831, opened 2019-03-12
Show dashboard hosts below process count graphs

sihugh:
Now we've migrated, the hostnames have gone from (for example)
`frontend-1_frontend` to `frontend-ip-xx-xx-xx-xxx_eu-west-1_cmopute_internal`
the lengthening of the hostname means that the space for the graph is much
constrained.  If we move them to below the graph, there's room for everything.

This affects dashboards like https://grafana.blue.staging.govuk.digital/dashboard/file/frontend.json

(Also removed a misplaced word from the description)


cbaines:
Sounds good to me :+1:

PR #8834, opened 2019-03-12
Preserve data dumps by default following replication script

edwardkerry:
Currently data dumps are automatically deleted following each sync script.
Should there be a failure at any point, they must all then be re-downloaded.
This can be extremely time consuming if experiencing issues with the VM.

This commit replaces the `-k` flag with `-x` to make the data dumps persist by default, and
only be automatically deleted if the flag is specified.

This will mean large amounts of data will be kept on developer's machines until manually deleted,
but will prevent having to re-download many GB's of data dumps should the VM import process fail.

### DIFF
--- a/development-vm/replication/common-args.sh
+++ b/development-vm/replication/common-args.sh
@@ -25,7 +25,7 @@ OPTIONS:
     -q       Skip MySQL import
     -e       Skip Elasticsearch import
     -t       Skip Mapit import
-    -k       Keep backups after import (will delete by default)
+    -k       Delete backups after import (will keep by default)
 
 EOF
 }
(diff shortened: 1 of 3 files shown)


development-vm/replication/common-args.sh
```diff
@@ -25,7 +25,7 @@ OPTIONS:
     -q       Skip MySQL import
     -e       Skip Elasticsearch import
     -t       Skip Mapit import
-    -k       Keep backups after import (will delete by default)
+    -k       Delete backups after import (will keep by default)
```
deborahchua: I'm wondering whether we should use a different letter, as most of us may not initially realise this has changed and still include `-k` when we want to keep the backups and then realised after it's deleted them!

deborahchua: maybe `del` as `d` is already taken?

edwardkerry: nice, `d` was gone but `del` is a good idea!

edwardkerry: sadly, that is parse as `-d el` so sets the directory as `el`

deborahchua: Ah! How about `x`?

PR #8835, opened 2019-03-13
Keep local replication backups by default

edwardkerry:
Currently data dumps are automatically deleted following each sync script.
Should there be a failure at any point, they must all then be re-downloaded.
This can be extremely time consuming if experiencing issues with the VM.

This commit replaces the `-k` flag with `-x` to make the data dumps persist by default, and
only be automatically deleted if the flag is specified.

This will mean large amounts of data will be kept on developer's machines until manually deleted,
but will prevent having to re-download many GB's of data dumps should the VM import process fail.

### DIFF
--- a/development-vm/replication/README.md
+++ b/development-vm/replication/README.md
@@ -2,4 +2,7 @@
 
 Dumps are generated from production data in the early hours each day, and are then downloaded from S3.
 
+Dumps are large, and you may wish to delete them after the script has completed.
+To automatically delete dumps after the script has run(whether successful or not), use flag `-x`.
+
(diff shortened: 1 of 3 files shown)


thomasleese:
I like this idea, however I think the reason this was added was to avoid accidentally using up a lot of disk space - it was a common complaint that a machine had ran out of space and the first response was to delete the old backups.

I think we should either encourage the use of `-x` by default (with the warnings associated with it deleting backups even on failure), or is there a way we can make it only delete the backups on a successful import? Perhaps checking the exit code of the import command?

bevanloon:
I think checking the exit code is a good idea. We shouldn't need to keep backups around for longer than the immediate import process.

edwardkerry:
Closing this for now to quieten the Seal, will work on checking exit-processes and re-open.

PR #8838, opened 2019-03-13
Switch applications to use search-api in staging

barrucadu:
Like https://github.com/alphagov/govuk-puppet/pull/8704 but for staging

---

[Trello card](https://trello.com/c/SPGs4yeO/80-staging-application-switchover)

### DIFF
--- a/hieradata/staging.yaml
+++ b/hieradata/staging.yaml
@@ -54,11 +54,11 @@ govuk::apps::whitehall::cpu_critical: 600
 
 govuk_search::gor::port: 3233
 
-# govuk::apps::collections::override_search_location: 'https://search-api.staging.govuk.digital' - in aws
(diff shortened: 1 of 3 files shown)


brucebolt:
In the event of any issues with `search-api` or Elasticsearch 5, we can simply revert this PR and deploy Puppet to roll back to `rummager` and Elasticsearch 2.

Before deploying to production, we will also revert this PR and instead disable `rummager`.  This will result in the `search` alias pointing to `search-api` instead of `rummager`.  We are not doing that here because we want to keep `rummager` available in case of having to roll back.

barrucadu:
@brucebolt You probably didn't want to merge this as is, because it unconditionally sets rummager's X-Ray sampling rate to 100%.

PR #8846, opened 2019-03-14
removing psql comments to avoid psql restore errors

fredericfran-gds:
In order to enable psql restore to happen in a single transaction and hence avoid database corruption while restoring, there should be no errors during the restore process or otherwise, psql will roll back.

The sql restore process now:
1. converts the psql backup file in the format of custom type to plain type.
2. remove the offending psql statements which will cause error on restore
3. recreate the psql database to restore to
4. restore the contents of the psql database

### DIFF
--- a/modules/govuk_env_sync/files/govuk_env_sync.sh
+++ b/modules/govuk_env_sync/files/govuk_env_sync.sh
@@ -283,7 +283,13 @@ function restore_postgresql {
      DB_OWNER=$(sudo psql -U aws_db_admin -h "${db_hostname}" --no-password --list --quiet --tuples-only | awk '{print $1 " " $3}'| grep -v "|" | grep -w "${database}" | awk '{print $2}')
      sudo dropdb -U aws_db_admin -h "${db_hostname}" --no-password "${database}"
   fi
-  pg_stderr=$(sudo pg_restore -U aws_db_admin -h "${db_hostname}" --create --no-password -d postgres "${tempdir}/${filename}" 2>&1)
+
+  pg_restore "${tempdir}/${filename}" -f "${tempdir}/${filename}.dump"
+  sed -i '/COMMENT\ ON\ EXTENSION\ plpgsql/d' "${tempdir}/${filename}.dump"
+  sudo createdb -U aws_db_admin -h "${db_hostname}" --no-password "${database}"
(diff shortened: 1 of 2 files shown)


modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -279,7 +279,13 @@ function restore_postgresql {
      DB_OWNER=$(sudo psql -U aws_db_admin -h "${db_hostname}" --no-password --list --quiet --tuples-only | awk '{print $1 " " $3}'| grep -v "|" | grep -w "${database}" | awk '{print $2}')
      sudo dropdb -U aws_db_admin -h "${db_hostname}" --no-password "${database}"
   fi
-  pg_stderr=$(sudo pg_restore -U aws_db_admin -h "${db_hostname}" --create --no-password -d postgres "${tempdir}/${filename}" 2>&1)
+
+  pg_restore "${tempdir}/${filename}" -f "${tempdir}/${filename}.dump"
+  sed -i '/COMMENT\ ON\ EXTENSION\ plpgsql/d' "${tempdir}/${filename}.dump"
```
cbaines: So I also encountered this problem when trying to fix something in the govuk-guix data stuff.

I came across the option of deleting the comment on the extension, prior to creating the dump, which seems to avoid issues during the restore [1].

1: https://github.com/alphagov/govuk-guix/blob/master/gds/data/govuk/sources/govuk-puppet-aws.scm#L185-L189

I'm not sure it's a great solution, but I thought I'd mention it.

fredericfran-gds: Ideally I don't want to modify the production database, I prefer to do the modification in the restore step. I could dump the psql in plain format and filter it there but then the backup becomes 4GB which we then store in S3 compared to the 0.7GB size now.

fredericfran-gds: I will suggest it to @schmie because it is a production database, I don't know enough about psql to do a modification on the database. Thanks.

cbaines: Yeah, I think not modifying the production database is a valid concern. I'm doing this in the context of a processing pipeline where data is being restored, tweaked and then dumped out again.

PR #8849, opened 2019-03-15
Disable rummager in AWS Staging

brucebolt:
This app has been deployed alongside search-api, however it is not needed in this environment.  Since it cannot connect to Elasticsearch 2, the error logs are becoming huge and consuming the machine's entire disk space.

Trello card: https://trello.com/c/SPGs4yeO/80-staging-application-switchover

### DIFF
--- a/hieradata_aws/staging.yaml
+++ b/hieradata_aws/staging.yaml
@@ -159,6 +159,8 @@ govuk::apps::licencefinder::override_search_location: 'https://search-api.stagin
 govuk::apps::search_admin::override_search_location: NULL
 govuk::apps::whitehall::override_search_location: NULL
 
+govuk::apps::rummager::enable_rummager: false
+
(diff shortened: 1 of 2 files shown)


hieradata_aws/staging.yaml
```diff
@@ -159,6 +159,8 @@ govuk::apps::licencefinder::override_search_location: 'https://search-api.stagin
 govuk::apps::search_admin::override_search_location: NULL
 govuk::apps::whitehall::override_search_location: NULL
 
+govuk::apps::rummager::enabled: false
```
elliotcm: Should this be `govuk::apps::rummager::enable_rummager: false`?
https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/rummager.pp#L108

brucebolt: Yes, it should be.  Damn lack of consistency between app manifests :D.

elliotcm: Rummager is old :(

PR #8850, opened 2019-03-15
days_to_keep is of integer type

fredericfran-gds:
need to put the hiera value for days_to_keep as integer type rather than string one since days_to_keep is integer type.


hieradata_aws/class/staging/cache.yaml
```diff
@@ -1,6 +1,6 @@
 ---
-logrotate::conf::days_to_keep: “7”
-nginx::logging::days_to_keep: “7”
```
issyl0: These quotes are "smart quotes", not straight ones...

PR #8872, opened 2019-03-20
fix database sync after migration

fredericfran-gds:
# Context

After the AWS migration, some database sync in the Carrenza domain is no longer valid.

# Decisions
1. clean-up the hiera data of Carrenza to remove database sync in Carrenza since these are no longer needed


hieradata/node/mongo-1.backend.staging.publishing.service.gov.uk.yaml
```diff
@@ -1,15 +1,4 @@
 govuk_env_sync::tasks:
-  "push_licence_finder_production":
-    ensure: "present"
```
cbaines: Do you need to change this to `ensure: "absent"` and deploy gouvk-puppet before removing this Hieradata?

fredericfran-gds: yeah you are right. I will adopt a 2-step approach: 1. absent and 2. remove hiera later.

hieradata/node/router-backend-1.router.staging.publishing.service.gov.uk.yaml
```diff
@@ -1,6 +1,6 @@
 govuk_env_sync::tasks:
   "push_router_daily":
-    ensure: "present"
```
cbaines: While you may have changed this based on my first comment. You may be able to just remove this hieradata, as I believe the host may no longer exist in Staging?

fredericfran-gds: Still there somehow: https://alert.staging.publishing.service.gov.uk/cgi-bin/icinga/status.cgi?host=all&type=detail&servicestatustypes=16&hoststatustypes=3&serviceprops=2097162

cbaines: I'm not sure what that link is meant to show, but at least looking at the list of UP hosts in Staging, I can't see router-backend-1.

fredericfran-gds: actually the alert is set up here: https://github.com/alphagov/govuk-puppet/blob/a2c2fe03001c93b802b0912f5b635310d4321005/hieradata/class/backup.yaml.

I will keep the 2 stage process of setting absent and then delete the hiera stuff.

cbaines: Sure, providing we're agreed that the machine doesn't exist, and there's a plan to eventually remove this entire file, then I'm happy.

PR #8879, opened 2019-03-21
Improve the AWS RabbitMQ configuration

cbaines:
There are 3 AWS environments, Production, Staging and Integration. But
Production and Staging have a limited set of applications. Therefore,
specialise the RabbitMQ configuration accordingly.


cbaines:
I'm hoping this will resolve the govuk-puppet failures on the RabbitMQ machines in AWS Staging, as that's related to the Content Performance Manager configuration missing secrets.

PR #8880, opened 2019-03-21
Improve the Email Alert Service configuration

cbaines:


modules/govuk/manifests/apps/email_alert_service/rabbitmq.pp
```diff
@@ -30,6 +33,7 @@
 #   a warning.
 #
 class govuk::apps::email_alert_service::rabbitmq (
+  $ensure = 'present',
```
thomasleese: Could we make it such that if `govuk::apps::email_alert_service::enabled` is false, it passes in `ensure => absent` into the `rabbitmq` class and `ensure => present` otherwise. This means we wouldn't need two configuration values in Hieradata for the same thing. I can't think of a reason where you'd want the two configured independently.

cbaines: Not to my knowledge. This class is included on to the node in a more programmatic way that I'm not very familiar with [1][2].

1: https://github.com/alphagov/govuk-puppet/blob/master/hieradata_aws/class/rabbitmq.yaml#L3-L6
2: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/node/s_rabbitmq.pp#L17-L18

thomasleese: Oh I see, yeah, that doesn't look so easy.

PR #8887, opened 2019-03-25
Reduce storage space required by govuk_env_sync

cbaines:
Rather than decompressing the compressed file, have pg_restore
decompress it, pipe the output to sed, then back in to gzip.

The backup files for some databases will take up lots of space if
uncompressed, so this avoids storing the uncompressed data on the
disk.


cbaines:
> looks, good. I guess you tested this locally?

I tested the first bit at least on the Content Data API DB Admin machine.

PR #8896, opened 2019-03-26
fix service description which does not meet regex validation

fredericfran-gds:
# Context

The service description for one of the icinga alert of content-data-api does not meet validation by regex and therefore causing puppet to fail to run on the monitoring boxes in AWS Staging and Production.

# Decisions
1. fix the service description


fredericfran-gds:
@cbaines has pushed this already: https://github.com/alphagov/govuk-puppet/pull/8895 so closing.

PR #8912, opened 2019-03-27
Point production collections and licence-finder to search-api

barrucadu:
[Trello card](https://trello.com/c/tOVAFzpN/97-production-application-switchover)


barrucadu:
We've got one possibly-weird search query which we want to resolve before moving finder-frontend to search-api, but collections and licence-finder should be fine.

The problem is with result ordering when you search for "dvsa", meaning the org page drops off the first page of results.  But:

- collections either fetches all the results, or explicitly sorts on a single field.
- licence-finder fetches all the results.

So we think these apps are fine to switch over.

PR #8924, opened 2019-03-28
Set Plek context for search machines in AWS

brucebolt:
Whilst `search-api` will be located in AWS, `publishing-api` and `email-alert-api` remain in Carrenza.

This updates the Plek environment variable overrides so these services can still be accessed by `search-api`.

Trello card: https://trello.com/c/roOVfL9T/98-make-sure-emails-triggered-by-the-business-readiness-finder-will-continue-to-work

### DIFF
--- a/modules/govuk/manifests/node/s_search.pp
+++ b/modules/govuk/manifests/node/s_search.pp
@@ -25,4 +25,14 @@ class govuk::node::s_search inherits govuk::node::s_base {
   if ! $::aws_migration {
     include govuk_elasticsearch::local_proxy
   }
+
+  # Set Plek for AWS to Carrenza communication
+  if ( ( $::aws_migration == 'search' ) and ($::aws_environment == 'staging') ) or ( ($::aws_migration == 'search' ) and ($::aws_environment == 'production') ) {
+    $app_domain = hiera('app_domain')
+
+    govuk_envvar {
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/node/s_search.pp
```diff
@@ -25,4 +25,14 @@
...
+
+  # Set Plek for AWS to Carrenza communication
+  if ( ( $::aws_migration == 'search' ) and ($::aws_environment == 'staging') ) or ( ($::aws_migration == 'search' ) and ($::aws_environment == 'production') ) {
+    $app_domain = hiera('app_domain')
+
+    govuk_envvar {
+      'PLEK_SERVICE_EMAIL_ALERT_API_URI': value  => "https://email-alert-api.${app_domain}";
+      'PLEK_SERVICE_PUBLISHING_API_URI': value  => "https://publishing-api.${app_domain}";
+    }
+  }
```
cbaines: It might be clearer to put this configuration in the app, rather than on the node? As my understanding is that this directly relates to things accessed by search-api?

brucebolt: There is prior art for this being configured on the node (https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/node/s_frontend.pp#L32-L41).  `search-api` is the only app running on this node and I expect this code to be removed once the AWS migration is completed.

cbaines: Ok, fair enough.

PR #8928, opened 2019-03-29
Increase max thread count critical warning threshold for cache-clearing-service

bilbof:
This raises a critical alert regularly (almost every day). Cache clearing service [intentionally creates a new thread](https://github.com/alphagov/cache-clearing-service/blob/master/app/processor.rb#L16) per content item path it needs to process. So to my mind it's not surprising we see spikes when we run this service, given that a single translated piece of content will have n locale paths. It only takes ~3 pieces of translated content being updated to raise this error. So I don't think it's necessarily a bad thing, probably memory is a more important thing for us to worry about.

This will override the [default set here](https://github.com/alphagov/govuk-puppet/pull/8010). 250 was chosen since this was the peak from this week. Please correct me if I've misunderstood something here.

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -237,6 +237,10 @@
 # [*collectd_process_regex*]
 #   Regex to use to identify the process.
 #
+# [*alert_when_threads_exceed*]
+#   If set, alert using Icinga if the number of threads exceeds the value specified.
+#   Default: 100
+#
 # [*override_search_location*]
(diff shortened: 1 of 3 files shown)


cbaines:
> This will override the [default set here](https://github.com/alphagov/govuk-puppet/pull/8010). 250 was chosen since this was the peak from this week. Please correct me if I've misunderstood something here.

No, your reasoning is great, thanks for taking a look at the configuration :tada:

PR #8936, opened 2019-04-01
Remove ITHC access

timblair:


rubenarakelyan:
This will need to be done in two steps - firstly to remove the `govuk_mysql::server::pentester::enabled: true` in the Carrenza staging hieradata and then these deletes, otherwise Puppet will lose track of the users and they will stay there.

rubenarakelyan:
Also, the SSH user needs an `ensure => absent` first, and then to be deleted.

timblair:
Yes, good points. I was rushing to get this done before the train pulled in to the station... I'll close this and start again.

PR #8957, opened 2019-04-04
Update reference to AWS integration when assuming role

karlbaker02:
This PR fixes the script to replicate data locally by updating the reference to integration to be `govuk-infrastructure-integration`, as detailed at https://docs.publishing.service.gov.uk/manual/aws-console-access.html#2-fill-in-account-and-role.

### DIFF
--- a/development-vm/replication/aws.sh
+++ b/development-vm/replication/aws.sh
@@ -10,9 +10,9 @@ aws_auth() {
 
     # setup AWS access
     SESSION_NAME=$(whoami)-$(date +%d-%m-%y_%H-%M)
-    ROLE_ARN=$(awk '/profile govuk-integration/ {profile=1} /role_arn/ && profile==1 {print $3; exit}' ~/.aws/config)
-    MFA_SERIAL=$(awk '/profile govuk-integration/ {profile=1} /mfa_serial/ && profile==1 {print $3; exit}' ~/.aws/config)
-    SOURCE_PROFILE=$(awk '/profile govuk-integration/ {profile=1} /source_profile/ && profile==1 {print $3; exit}' ~/.aws/config)
+    ROLE_ARN=$(awk '/profile govuk-infrastructure-integration/ {profile=1} /role_arn/ && profile==1 {print $3; exit}' ~/.aws/config)
+    MFA_SERIAL=$(awk '/profile govuk-infrastructure-integration/ {profile=1} /mfa_serial/ && profile==1 {print $3; exit}' ~/.aws/config)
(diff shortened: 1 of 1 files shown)


development-vm/replication/aws.sh
```diff
@@ -10,9 +10,9 @@ aws_auth() {
 
     # setup AWS access
     SESSION_NAME=$(whoami)-$(date +%d-%m-%y_%H-%M)
-    ROLE_ARN=$(awk '/profile govuk-integration/ {profile=1} /role_arn/ && profile==1 {print $3; exit}' ~/.aws/config)
-    MFA_SERIAL=$(awk '/profile govuk-integration/ {profile=1} /mfa_serial/ && profile==1 {print $3; exit}' ~/.aws/config)
-    SOURCE_PROFILE=$(awk '/profile govuk-integration/ {profile=1} /source_profile/ && profile==1 {print $3; exit}' ~/.aws/config)
+    ROLE_ARN=$(awk '/profile govuk-infrastructure-integration/ {profile=1} /role_arn/ && profile==1 {print $3; exit}' ~/.aws/config)
+    MFA_SERIAL=$(awk '/profile govuk-infrastructure-integration/ {profile=1} /mfa_serial/ && profile==1 {print $3; exit}' ~/.aws/config)
+    SOURCE_PROFILE=$(awk '/profile govuk-infrastructure-integration/ {profile=1} /source_profile/ && profile==1 {print $3; exit}' ~/.aws/config)
```
cbaines: The format the docs recommend for the AWS profile is `govuk-<environment>`, which is compatible with what's here. While you can use anything in your `~/.aws/config` file, this change will break the script for those who use `govuk-integration` as the name of the profile.

cbaines: https://docs.publishing.service.gov.uk/manual/aws-cli-access.html#1-create-awsconfig

cbaines: (maybe recommend is too strong, it's just the example given)

karlbaker02: This is true, fair point. If anyone did have `govuk-integration` in their profile, do you think that'd cause them problems elsewhere where that profile might be used, and could provide an incentive to update the name to be consistent with the account? If not, I'll close this PR.

cbaines: I think the profile names are entirely arbitrary, it's just that we do use them in a few disparate places.

Another place where the `govuk-<environment>` format is assumed is the govukcli tool in govuk-aws: https://github.com/alphagov/govuk-aws/blob/master/tools/govukcli#L316

cbaines: Bump, any update on this?

PR #8962, opened 2019-04-05
Combine process count / error rate graphs on deployment dashboards

sihugh:
Changes dashboards like: https://grafana.blue.production.govuk.digital/dashboard/file/finder-frontend.json?refresh=5s&orgId=1 

The processes and errors graphs are not consistently shown at the moment because the panels seem to be too small for the data in the time span on some occasions.

By combining the graphs, we make them visible again, but also get the benefit of aligning them with the other full width graphs on the dashboard, making it clearer when events correlate.

The error rate is set against the right axis because the scale of this is different from the process count which tends to be between 0 and 10 (or thereabouts) depending on the application.

Hopefully fixes blank panels with a spinner (like this):

Screenshot 2019-04-05 11 44 50

By replacing it with:

Screenshot 2019-04-05 11 43 16


sihugh:
Thanks Bill 👍

PR #8966, opened 2019-04-08
Remove unused fields_prefix

sihugh:
This was used when we had some logstasher log fields prefixed with `@fields` (so `@fields.duration`, `@fields.action`, `@fields.controller` etc.), and some which have no prefix.

We started fixing this and trying to align on having no prefix back in Jan 2018 (https://github.com/alphagov/govuk-puppet/pull/6658). This should now be complete as of https://github.com/alphagov/search-api/pull/42, so we can remove references to the prefix that allows us to manage the convergence.

cc @surminus

### DIFF
--- a/modules/grafana/manifests/dashboards/application_dashboard.pp
+++ b/modules/grafana/manifests/dashboards/application_dashboard.pp
@@ -24,12 +24,6 @@
 # [*error_threshold*]
 #   Point at which count turns `red` in the error count table data
 #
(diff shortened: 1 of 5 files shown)


surminus:
I approve.

sihugh:
This is an _enormous_ deal for me

surminus:
How time flies!

sihugh:
Thanks all!

PR #8970, opened 2019-04-09
Added user Camille Descartes and public key

camdesgov:
Added new user camilledescartes and public key


issyl0:
Done in #9011 instead.

PR #8978, opened 2019-04-11
Set unicorn timeout of all frontend apps to 15s

barrucadu:
We use a 15s timeout for nginx, but the default unicorn timeout (60s).
This means that a slow request will be dropped by nginx but the
unicorn will keep working.  This means that a batch of slow requests
in quick succession can tie up all the unicorns.

---

[Trello card](https://trello.com/c/XUgnpLbM/119-action-item-fix-unicorn-worker-timeout)

### DIFF
--- a/modules/govuk/manifests/node/s_calculators_frontend.pp
+++ b/modules/govuk/manifests/node/s_calculators_frontend.pp
@@ -29,6 +29,10 @@ class govuk::node::s_calculators_frontend inherits govuk::node::s_base {
     listen_ip  => '0.0.0.0',
   }
 
+  govuk_envvar {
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/node/s_calculators_frontend.pp
```diff
@@ -29,6 +29,10 @@
     listen_ip  => '0.0.0.0',
   }
 
+  govuk_envar {
```
brucebolt: Should this be `govuk_envvar` instead of `govuk_envar`?

barrucadu: I keep doing this

PR #8979, opened 2019-04-11
Add Icinga monitoring for established TCP connections to apps

cbaines:
This should flag when requests are potentially being queued for
processing at the point of reaching Unicorn (or other similar things).

We've had issues with this for Finder Frontend recently, and this
seems to be the way to monitor if there's insufficient workers to
process all the requests. I'm not exactly sure why it's the
ESTABLISHED metric, but this correlates with timeout issues.

### 5xx errors for Finder Frontend (over the last 4 days)
Screenshot from 2019-04-11 16-39-20
## Local TCP connections to the Finder Frontend port
Screenshot from 2019-04-11 16-38-50
### ESTABLISHED TCP connection count for the Finder Frontend port
Screenshot from 2019-04-11 16-39-02

### DIFF
--- a/modules/govuk/manifests/app/config.pp
+++ b/modules/govuk/manifests/app/config.pp
@@ -330,6 +330,22 @@ define govuk::app::config (
     outgoing => $port,
   }
 
+  if $unicorn_worker_processes {
+    $local_tcpconns_warning = $unicorn_worker_processes
+  } else {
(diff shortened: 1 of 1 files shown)


cbaines:
> We'll need to add some docs on what to do if people see this, but +1

Indeed, thanks for taking a look :)

PR #8980, opened 2019-04-11
Update the on-call alerting times for the change to BST

barrucadu:
See 1f9c585.

### DIFF
--- a/modules/monitoring/manifests/contacts.pp
+++ b/modules/monitoring/manifests/contacts.pp
@@ -20,9 +20,9 @@ class monitoring::contacts (
   validate_bool($notify_pager, $notify_graphite)
 
   $midnight_day_start = '00:00'
-  $office_day_start = '09:30'
-  $midday = '12:00'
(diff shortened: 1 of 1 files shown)


modules/monitoring/manifests/contacts.pp
```diff
@@ -20,9 +20,9 @@
   validate_bool($notify_pager, $notify_graphite)
 
   $midnight_day_start = '00:00'
-  $office_day_start = '09:30'
-  $midday = '12:00'
-  $office_day_end = '17:30'
+  $office_day_start = '08:30'
+  $midday = '11:00'
+  $office_day_end = '16:30'
```
cbaines: I wonder if this could be smarter, and we could somehow specify these times automatically in London local time...

barrucadu: The Icinga 2 docs don't mention timezones, sadly: https://icinga.com/docs/icinga2/latest/doc/08-advanced-topics/#time-periods

PR #8988, opened 2019-04-12
ip address should not be specified so that it defaults to eth0 one

fredericfran-gds:


fredericfran-gds:
will use another branch name.

PR #8990, opened 2019-04-15
fix puppet cronjob to avoid conflict with puppetserver restart

fredericfran-gds:
# Context

1. Puppet server restarts every 30 mins and if it is not fully functional before other nodes request their puppet manifest, the other nodes will show a puppet error in icinga. In order to solve this issue, we want to have a minimum delay of 5 minutes between when the puppetserver restarts and the first node requests its manifest from the puppet server.

2. Puppet module uses a "file_line" puppet function to ensure that the JAVA_ARG parameter is properly populated in `/etc/default/puppetserver` config file. There is a `\n` value in the line parameter value of file_line. The use of `\n` in the line parameter causes the loss of idempotency (see https://stackoverflow.com/questions/38539178/check-and-add-multiple-lines-in-puppet)

# Decisons
1. fix the puppet cron logic to fit the above scenario
2. remove the `\n` to maintain idempotency (was tested that puppet still changes file if needed).

### DIFF
--- a/modules/puppet/manifests/cronjob.pp
+++ b/modules/puppet/manifests/cronjob.pp
@@ -12,13 +12,19 @@ class puppet::cronjob (
 ) {
 
   if $enabled {
-    $first = fqdn_rand(30)
-    $second = $first + 30
+    if ((! $::aws_migration) and ($::fqdn =~ /^puppetmaster/)) or
+      (($::aws_migration) and ($::aws_migration =~ /^puppetmaster/)) {
+      $first_run = 0
+      $second_run = 30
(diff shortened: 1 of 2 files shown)


rtrinque:
Puppet server restart every 30 minutes? I don't see any justification for this, it seems to me that this is the issue that needs fixing

fredericfran-gds:
I think we can have both:
1. fix that puppetserver restarts/refreshes every 30 minutes
2. if puppetserver restarts/refreshes, avoid other nodes to connect to it at that time (this PR fixes that)

PR #8992, opened 2019-04-15
Change NGinx configuration for the Content Data app

cbaines:
These changes are intended to be deployed when the Content Data app is
migrated to AWS. This coensides with the domain name changing from
`content-data-admin.` to `content-data.`.

For a smoother transition, this change configures NGinx to redirect
from the old URL to the new URL. To make this possible, the vhost for
the app NGinx configuration is changed.

Initially, the redirect will be served from Carrenza, but in the
longer term, the necessary DNS and infrastructure changes should be
made so that this redirect is served from AWS.


cbaines:
Marked as do not merge, as this should be merged and deployed as part of the AWS migration for the Content Data app.

cbaines:
Merging now, as the plan is to deploy these changes this morning.

PR #8994, opened 2019-04-15
Better manage the env sync job configuration

cbaines:


modules/govuk_env_sync/manifests/task.pp
```diff
@@ -53,7 +53,7 @@
   require govuk_env_sync::sync_script
 
   file { "${govuk_env_sync::conf_dir}/${title}.cfg":
-    ensure  => 'present',
+    ensure  => $ensure,
```
schmie: This has prior been left "asynch" as it is useful for manual execution of syncs during eg migration - are you happy to leave this as was?

cbaines: Given that lots of things aren't being migrated at the moment, I'm a bit more interested in trying to make the configuration on the machine clear, which includes removing unused configuration.

I'm fine with still supporting this use case though, but it would be better if it was explicit. I've pushed a new commit that adds support for setting tasks to `disabled`, which will leave config file in place, but remove the cron entry.

PR #9000, opened 2019-04-17
Tear down rummager in all environments (AWS and Carrenza)

barrucadu:
This should remove the rummager app and all of its related components (rabbitmq stuff, etc).  When we're happy that that has worked, the rummager hieradata and puppet can be removed.

---

[Trello card](https://trello.com/c/nABGyw0B/653-%F0%9F%92%80-remove-rummager-%F0%9F%92%80)

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -176,7 +176,6 @@ deployable_applications: &deployable_applications
   release: {}
   router: {}
   router-api: {}
-  rummager: {}
   search-admin: {}
   service-manual-frontend: {}
   service-manual-publisher: {}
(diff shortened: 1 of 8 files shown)


hieradata/common.yaml
```diff
@@ -562,6 +558,9 @@ govuk::apps::local_links_manager::unicorn_worker_processes: "4"
 
 govuk::apps::mapit::enabled: true
 
+govuk::apps::rummager::enable_rummager: false
+govuk::apps::search_api::rummager_enable: false
```
cbaines: I'm finding these changes a little hard to follow... Will setting this here have any effect, given the change to the `search` node class in this same file?

cbaines: (this applying to both lines I guess, particularly the first one)

barrucadu: I think I should probably leave the `node_class` how it is for now: https://github.com/alphagov/govuk-puppet/blob/3e0492ca60ff066db99a02140f38e3729a211533/modules/govuk/manifests/node/s_base.pp#L46-L53
I'll update the PR

PR #9001, opened 2019-04-17
Rename rummager to search-api for dev VM

brucebolt:
Update the dev VM `Pinfile` to start `search-api`, in place of `rummager`, since the application name has recently been changed.


brucebolt:
I've rebase this as there was some duplication

PR #9009, opened 2019-04-18
Remove rummager & rummager-elasticsearch

barrucadu:
Before deploying this I'll shut down the relevant machines in Carrenza.

---

Trello cards:
- [💀 Remove rummager 💀](https://trello.com/c/nABGyw0B/653-%F0%9F%92%80-remove-rummager-%F0%9F%92%80)
- [💀 Remove elasticsearch 2 cluster 💀](https://trello.com/c/Z8a5pAeg/654-%F0%9F%92%80-remove-elasticsearch-2-cluster-%F0%9F%92%80)


modules/monitoring/manifests/checks.pp
```diff
@@ -173,46 +173,70 @@
...
     args                => "--dropfirst ${drop_first}",
   }
+  icinga::check::graphite { 'check_rummager_govuk_index_size_changed':
+    ensure    => 'absent',
+    target    => undef,
+    desc      => undef,
+    warning   => undef,
+    critical  => undef,
+    host_name => undef,
+  }
```
cbaines: Good practice :+1: but I don't think it's strictly necessary in this case, as the Icinga configuration directory is fully managed by Puppet, including purging files it doesn't recognise [1].

1: https://github.com/alphagov/govuk-puppet/blob/master/modules/icinga/manifests/config.pp#L43-L51

PR #9019, opened 2019-04-23
Remove alias hosts for cache machine in Carrenza staging

schmie:
- This host entry is confusing TLS testing in AWS/Fastly as clients are
attempting to talk to the Carrenza cache machine IP
- Related Platform Health ticket: https://trello.com/c/euhCSxFr/942-upgrade-to-the-latest-libssl-for-when-fastly-switches-off-tls-10

solo: @schmie


schmie:
Yes, I just wanted to make this change first and see if there are some deps we are not aware of : )

PR #9022, opened 2019-04-23
Add Ubuntu extended security maintenance repo

rtrinque:
Support for trusty is ending soon, we need to install ESM repos

### DIFF
--- a/modules/base/manifests/init.pp
+++ b/modules/base/manifests/init.pp
@@ -3,9 +3,19 @@
 # Pull in a standard set of modules for all machines. Used everywhere, by
 # both `govuk::node::s_base` and `govuk::node::s_development`.
 #
-class base {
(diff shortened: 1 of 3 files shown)


modules/base/manifests/init.pp
```diff
@@ -3,11 +3,21 @@
...
+#
+
+class base (
+  $esm_repo = false,
+) {
   class { '::apparmor':
     package_ensure => present,
   }
 
+  if $esm_repo {
```
schmie: I think we need a filter for the last precise machines (licensify-lb) here while they last : /

rtrinque: It should be fine, esm_repo is set to true in common.trusty.yaml, so only for the machines running trusty

PR #9023, opened 2019-04-24
Fix Content Data running in the VM

theseanything:
This PR addresses the following issues that prevent Content Data from running in the VM:
- Nginx wasn't forwarding `content-data-api` requests to the correct port
- `content-data-api` subdomain wasn't resolvable from within the VM


theseanything:
[Fix app name changes in Dev VM](https://trello.com/c/aZJc5MQb/1350-fix-app-name-changes-in-dev-vm)

PR #9025, opened 2019-04-24
Remove Pablo Manrubia

AgaDufrat:


.mailmap
```diff
@@ -143,7 +143,6 @@ Nick Colley <nick.colley@digital.cabinet-office.gov.uk> <nickcolley7@gmail.com>
 Nick Colley <nick.colley@digital.cabinet-office.gov.uk> <nickcolley@gds0257.local>
 Nick Stenning <nick.stenning@digital.cabinet-office.gov.uk> <nick@whiteink.com>
 Oliver Byford <oliver.byford@digital.cabinet-office.gov.uk> <o@byford.org>
-Pablo Manrubia <pablo.manrubia@digital.cabinet-office.gov.uk> <pmanrubia@gmail.com>
```
cbaines: Why remove Pablo from the mailmap? It's still relevant, as we haven't purged all of Pablo's commits from the repository?

AgaDufrat: I've seen this being done before https://github.com/alphagov/govuk-puppet/commit/54e3b6b13aae5f2016157a3ee7e33a766f04689e#diff-c9d540715cff3469b65ddd01f614848b
and I wasn't sure about keeping someone’s personal e-mail address in this file after they left.

cbaines: This entry is here as `pmanrubia@gmail.com` is a value used in some of the commits. Keeping it here is useful, as it means the history can be viewed in a more consistent way. Removing it won't removing `pmanrubia@gmail.com` from the history of the repository.

AgaDufrat: ok, will add it back

cbaines: Great, thanks :)

PR #9027, opened 2019-04-24
Increase email alert api memory thresholds

AlanGabbianelli:
We often have alerts (warnings & criticals) for high memory usage in
email-alert-api machines. This alerts usually result in no action taken
because we have plenty of memory available.

This will increase the thresholds at which we will be alerted so to
avoid a) luttering our monitors b) distract us from more important
alerts.

Look at this Grafana dashboard that shows total memory + memory
used/buffered/cached since 2019-01-01 to see why these new values make
sense:
https://grafana.publishing.service.gov.uk/dashboard/file/machine.json?orgId=1&var-hostname=email-alert-api-1_backend&var-cpmetrics=cpu-system&var-cpmetrics=cpu-user&var-filesystem=All&var-disk=All&var-tcpconnslocal=All&var-tcpconnsremote=All&panelId=2&fullscreen&from=1546300800000&to=1556108108079

grafana_email_alert_api_machine

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -431,8 +431,8 @@ govuk::apps::email_alert_api::db_hostname: 'postgresql-primary-1.backend'
 govuk::apps::email_alert_api::db_password: "%{hiera('govuk::apps::email_alert_api::db::password')}"
 govuk::apps::email_alert_api::db_port: 6432
 govuk::apps::email_alert_api::db_allow_prepared_statements: false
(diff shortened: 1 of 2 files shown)


brucebolt:
Looks like CI is failing because of `shellcheck` errors in an unrelated script 😕

thomasleese:
I think you might need to rebase to include https://github.com/alphagov/govuk-puppet/commit/c9034d2acd57d5df9ffe6a0ecf6ca69192f24005 ?

PR #9029, opened 2019-04-24
Install chromedriver on all dev VMs

benthorner:
Previously we used the now-deprecated chromedriver-helper gem to install
chromedriver automatically for testing. This ensures it's installed as a
system binary, so that we no longer need to rely on the gem.


benthorner:
Closing as per https://github.com/alphagov/govuk_test/pull/19.

PR #9052, opened 2019-04-26
remove search api alerts for carrenza as migrated

fredericfran-gds:
# Context

search api has been migrate to AWS and therefore, no search api icinga alerts should appear in the carrenza icinga boards

# Decisions
1. use an if case to remove search-api alerts from Carrenza monitoring

### DIFF
--- a/modules/monitoring/manifests/checks.pp
+++ b/modules/monitoring/manifests/checks.pp
@@ -152,93 +152,71 @@ class monitoring::checks (
   }
   # END DNS checks
 
-  # START search
+  if $::aws_migration {
+    # START search
 
-  # On average 326 new documents were created per day in 2017, in total.
-  # The govuk index will only receive a fraction of these until we start
-  # indexing whitehall content in it.
-  #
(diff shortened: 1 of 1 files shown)


modules/monitoring/manifests/checks.pp
```diff
@@ -152,93 +152,95 @@
...
+      critical            => 10000,
+      desc                => 'search-api govuk index size has significantly increased/decreased over the last 7 days',
+      host_name           => $::fqdn,
+      notification_period => 'inoffice',
+      action_url          => "https://grafana.${app_domain}/dashboard/file/search_api_index_size.json",
+      notes_url           => monitoring_docs_url(search-api-index-size-change),
+      from                => '25minutes',
+      args                => "--dropfirst ${drop_first}",
+    }
+    icinga::check::graphite { 'check_rummager_govuk_index_size_changed':
```
barrucadu: These absent ones can be removed now, as puppet has run everywhere.

fredericfran-gds: will do, thanks.

PR #9055, opened 2019-04-26
remove conor glynn as he no longer works on govuk

fredericfran-gds:
# Context

Remove conor glynn as he no longer works on govuk.

# Decisions
1. remove conor glynn in hiera and ssh key. keep mailmap because of commit details.

### DIFF
--- a/modules/users/manifests/conorglynn.pp
+++ b/modules/users/manifests/conorglynn.pp
@@ -1,6 +1,7 @@
 # Creates conorglynn user
 class users::conorglynn {
   govuk_user { 'conorglynn':
+    ensure   => 'absent',
     fullname => 'Conor Glynn',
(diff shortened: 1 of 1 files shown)


modules/users/manifests/conorglynn.pp
```diff
@@ -1,8 +0,0 @@
-# Creates conorglynn user
-class users::conorglynn {
-  govuk_user { 'conorglynn':
-    fullname => 'Conor Glynn',
```
issyl0: Can we `ensure => absent` prior to removing the manifest file, as per [the docs on removing users](https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html)?

fredericfran-gds: cool thanks, did know above this docs, was doing freestyle

PR #9064, opened 2019-04-30
Remove incorrect external domains

thomasleese:
These are configured to point to the cache machines but this is no longer correct post migration and instead we should route these through the normal internet.

This is a continuation of https://github.com/alphagov/govuk-puppet/pull/9051.

This fixes problems like this:

```
thomasleese@staging-jenkins-1:~$ curl https://assets-origin.staging.publishing.service.gov.uk
curl: (35) Unknown SSL protocol error in connection to assets-origin.staging.publishing.service.gov.uk:443

thomasleese@staging-backend-1:~$ curl https://www.staging.publishing.service.gov.uk
curl: (35) Unknown SSL protocol error in connection to www.staging.publishing.service.gov.uk:443
```

The Smokey tests are currently failing because they can't access assets-origin and this should also fix that.

[Trello Card](https://trello.com/c/euhCSxFr/771-upgrade-to-the-latest-libssl-for-when-fastly-switches-off-tls-10)

### DIFF
--- a/hieradata/production.yaml
+++ b/hieradata/production.yaml
@@ -561,9 +561,6 @@ hosts::production::router::hosts:
     ip: '10.3.1.254'
     legacy_aliases:
       - 'cache'
-      - "www.%{hiera('app_domain')}"
-      - "www-origin.%{hiera('app_domain')}"
-      - "assets-origin.%{hiera('app_domain')}"
     service_aliases:
       - 'cache'
       - 'router'
--- a/hieradata/staging.yaml
+++ b/hieradata/staging.yaml
@@ -516,9 +516,6 @@ hosts::production::router::hosts:
     ip: '10.2.1.254'
     legacy_aliases:
       - 'cache'
-      - "www.%{hiera('app_domain')}"
-      - "www-origin.%{hiera('app_domain')}"
-      - "assets-origin.%{hiera('app_domain')}"
     service_aliases:
       - 'cache'
       - 'router'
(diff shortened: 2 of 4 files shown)


hieradata/production.yaml
```diff
@@ -561,9 +561,6 @@ hosts::production::router::hosts:
     ip: '10.3.1.254'
     legacy_aliases:
       - 'cache'
-      - "www.%{hiera('app_domain')}"
-      - "www-origin.%{hiera('app_domain')}"
-      - "assets-origin.%{hiera('app_domain')}"
```
cbaines: How is this hieradata relevant, as I didn't think the router or cache machines existed any longer in Carrenza?

thomasleese: Yeah, I don't think they do exist, but as I understand it this Hieradata creates entries in `/etc/hosts` meaning that we can't access services using the external domain.

```
thomasleese@production-backend-1:~$ cat /etc/hosts | grep www-origin
10.3.1.254	cache.router.publishing.service.gov.uk	cache.router cache.cluster router.cluster cache www.publishing.service.gov.uk www-origin.publishing.service.gov.uk assets-origin.publishing.service.gov.uk

thomasleese@production-backend-1:~$ curl https://www-origin.publishing.service.gov.uk
curl: (7) Failed to connect to www-origin.publishing.service.gov.uk port 443: Connection timed out
```

thomasleese: I suppose we can actually remove the whole block about the cache machines, but I didn't want to remove too much as I'm not familiar enough with this part of the infrastructure configuration.

cbaines: Ok, good to know.

I ended up in this minefield yesterday... https://github.com/alphagov/govuk-puppet/pull/9061

PR #9068, opened 2019-04-30
remove jenkins warehouse data sync

fredericfran-gds:
# Context

This is postgresql wharehouse data sync no longer needed because the data sync is done via the govuk_env_sync python code that runs as a cron job on the database machines themselves.

This will hopefully solves the issues seen during the Jenkins jobs: Copy Data to Staging and Copy Data to integration

# Decisions
1. remove all postgresql wharehouse data sync from Jenkins in the puppet code


cbaines:
Hmm, I guess these changes should have happened at the same time as https://github.com/alphagov/env-sync-and-backup/pull/140

I didn't realise the two things were quite so tightly coupled.

fredericfran-gds:
Unfortunately yes they are. I need to delete the password in hiera secrets but not a blocker.

fredericfran-gds:
no worries, thanks :-)

PR #9077, opened 2019-05-01
use tls for fastly logs in aws for added security

fredericfran-gds:
# Context

We should use tls to encrypt fastly logs being sent to aws for added security. This is done for fastly logs sent to Carrenza

# Decisions
1. switch on fastly rsyslog tls in aws hiera


fredericfran-gds:
I closed this PR as I will terminate the TLS/SSL connection on the monitoring ELB to maintain design pattern used for other services. This is done here: https://github.com/alphagov/govuk-aws/pull/917

PR #9085, opened 2019-05-03
Add collectd types for rabbitmq plugin

theseanything:
Since updating collectd-rabbitmq plugin to v1.20.0 new types need to be specified for the new metrics being collected.


theseanything:
[(3) Update collectd-rabbitmq plugin](https://trello.com/c/yzWglItx/1419-3-update-collectd-rabbitmq-plugin)

PR #9088, opened 2019-05-03
Fix Grafana dashboards using rabbitMQ metrics

theseanything:
After updating the collectd-rabbitmq plugin to 1.20.0 the root key is now `rabbitmq_default` instead of `rabbitmq_%2F`. This updates Grafana dashboards to use the correct root key.


theseanything:
[(3) Update collectd-rabbitmq plugin](https://trello.com/c/yzWglItx/1419-3-update-collectd-rabbitmq-plugin)

PR #9089, opened 2019-05-07
Remove the content performance manager configuration

cbaines:
This application has now been renamed to the Content Data API. Various
things have been set to 'absent', and these changes have been
deployed, so now the Puppet configuration can be removed.

There's still some work to do on the development-vm, and the now
redundant nodes, but this'll come later.


Rakefile
```diff
@@ -187,7 +185,6 @@ task :check_consistency_between_aws_and_carrenza do
     govuk::apps::content_audit_tool::db::allow_auth_from_lb
     govuk::apps::content_audit_tool::db::lb_ip_range
     govuk::apps::content_audit_tool::db::rds
-    govuk::apps::content_data_admin::content_performance_manager_uri
```
theseanything: How come we don't `govuk::apps::content_data_admin::content_data_api_uri`?

hieradata/development.yaml
```diff
@@ -34,8 +34,6 @@ govuk::node::s_development::apps:
   - 'contacts'
   - 'content_audit_tool'
   - 'content_data_api'
-  - 'content_performance_manager'
-  - 'content_performance_manager::rabbitmq'
```
theseanything: How come we don't `content_data_api::rabbitmq`?

hieradata/development.yaml
```diff
@@ -105,9 +103,6 @@ govuk::apps::collections_publisher::enable_procfile_worker: false
 govuk::apps::contacts::extra_aliases: ['contacts-admin']
 govuk::apps::content_store::mongodb_nodes: ['localhost']
 govuk::apps::content_store::mongodb_name: 'content_store_development'
-govuk::apps::content_performance_manager::rabbitmq_user: 'content_performance_manager'
-govuk::apps::content_performance_manager::rabbitmq_hosts: ['localhost']
-govuk::apps::content_performance_manager::rabbitmq::amqp_pass: 'content_performance_manager'
```
theseanything: Same here - shouldn't we have these keys for content-data-api?

hieradata/vagrant_credentials.yaml
```diff
@@ -35,8 +35,6 @@ govuk::apps::content_audit_tool::db::password: 'MDgTmQKvEAax4wgefQAZRNDTajkdtEVf
 govuk::apps::content_audit_tool::db_password: "%{hiera('govuk::apps::content_audit_tool::db::password')}"
 govuk::apps::content_data_admin::db::password: 'FrkwKJGjLAjbznLtPhFjVy7YEonLKJTr'
 govuk::apps::content_data_admin::db_password: "%{hiera('govuk::apps::content_data_admin::db::password')}"
-govuk::apps::content_performance_manager::db_password: "%{hiera('govuk::apps::content_performance_manager::db::password')}"
-govuk::apps::content_performance_manager::db::password: 'jo6Kah0kuokaighoughePhooch1Eequu'
```
theseanything: no keys for content_data_api?

Rakefile
cbaines: This parameter was used to override the default behaviour in Plek, so back when the content-data-admin did `Plek.find('content-performance-manager')` this parameter allowed overriding the value returned to be the content-data-api in AWS Staging and AWS Production.

Once the content-data-admin was changed to ask Plek for the content-data-api, the default behaviour is fine, so this parameter was removed [1]

1: https://github.com/alphagov/govuk-puppet/commit/9fe36303b81c748999253080945486a2af3f4d45

I guess the hieradata could have been removed then too, but it got missed.

hieradata/development.yaml
cbaines: That's covered in another Pull Request https://github.com/alphagov/govuk-puppet/pull/9091/files#diff-ff48ae5906656c3a1014d9acb3b0abdcR37

hieradata/development.yaml
cbaines: I've just removed things here. Some additional setup is needed for the Content Data API in the development-vm, but that's covered here https://github.com/alphagov/govuk-puppet/pull/9091/files#diff-ff48ae5906656c3a1014d9acb3b0abdcR108

hieradata/vagrant_credentials.yaml
cbaines: I'm very unsure what (if anything) this vagrant_credentials.yaml file relates to... I'll investigate more though.

hieradata_aws/class/staging/backend.yaml
```diff
@@ -3,7 +3,5 @@
 icinga::client::check_pings::endpoints:
    whitehall-admin:
      ip: 10.2.3.254
-   content-performance-manager:
-     ip: 10.2.3.254
```
theseanything: Do we need an IP for content-data-api?

hieradata/vagrant_credentials.yaml
cbaines: So, I think this relates to the Vagrantfile at the root of the govuk-puppet Git repository. I've opened a Pull Request to update the comment in the file https://github.com/alphagov/govuk-puppet/pull/9096 and added a commit to this branch to add some credentials for the Content Data API just in case.

hieradata_aws/class/production/backend.yaml
```diff
@@ -3,7 +3,5 @@
 icinga::client::check_pings::endpoints:
    whitehall-admin:
      ip: 10.3.3.254
-   content-performance-manager:
-     ip: 10.3.3.254
```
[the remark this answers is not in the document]
cbaines: I've looked in to this now, and this relates to checking the connectivity between AWS and Carrenza. As we're not running the Content Performance Manager in Carrenza any longer, I guess the check may start to fail due to the other changes on this branch.

As we're trying to also connect to Whitehall Admin, on the same IP address, I don't think there's any need to replace this, it can just be removed.

hieradata_aws/class/staging/backend.yaml
cbaines: Same thing applies as the Production configuration https://github.com/alphagov/govuk-puppet/pull/9089#discussion_r281631820

PR #9090, opened 2019-05-07
Stop purging some queues in AWS

cbaines:


cbaines:
> Looks good, assuming these don't need to be ensured absent first

I believe the cron configuration is full managed, so I think this should work.

cbaines:
> I believe the cron configuration is full managed, so I think this should work.

Update, I was wrong :( I'll remove the configuration manually...

PR #9091, opened 2019-05-07
Use bowl to start up content data apps

MahmudH:
- Replace references to content-performance-manager to content-data-api
- Set up development environment for content-data-api when running govuk_puppet

### DIFF
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -34,8 +34,7 @@ govuk::node::s_development::apps:
   - 'contacts'
   - 'content_audit_tool'
   - 'content_data_api'
-  - 'content_performance_manager'
-  - 'content_performance_manager::rabbitmq'
+  - 'content_data_api::rabbitmq'
   - 'content_publisher'
   - 'content_store'
   - 'content_store::enable_running_in_draft_mode'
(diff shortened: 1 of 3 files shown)


hieradata/development.yaml
```diff
@@ -103,11 +102,13 @@ govuk::apps::ckan::priority_worker_processes: '1'
 govuk::apps::ckan::bulk_worker_processes: '1'
 govuk::apps::collections_publisher::enable_procfile_worker: false
 govuk::apps::contacts::extra_aliases: ['contacts-admin']
+govuk::apps::content_data_admin::enabled: true
```
theseanything: Why's this added?

hieradata/development.yaml
```diff
@@ -103,11 +102,13 @@ govuk::apps::ckan::priority_worker_processes: '1'
...
 govuk::apps::content_store::mongodb_nodes: ['localhost']
 govuk::apps::content_store::mongodb_name: 'content_store_development'
-govuk::apps::content_performance_manager::rabbitmq_user: 'content_performance_manager'
-govuk::apps::content_performance_manager::rabbitmq_hosts: ['localhost']
-govuk::apps::content_performance_manager::rabbitmq::amqp_pass: 'content_performance_manager'
+govuk::apps::content_data_api::rabbitmq_user: 'content_data_api'
+govuk::apps::content_data_api::rabbitmq_hosts: ['localhost']
+govuk::apps::content_data_api::rabbitmq::amqp_pass: 'content_data_api'
+govuk::apps::content_data_api::rabbitmq_password: "%{hiera('govuk::apps::content_data_api::rabbitmq::amqp_pass')}"
```
theseanything: Why do we need this?

hieradata/development.yaml
MahmudH: This enables content data admin in development. The reason we need to set is here is because it is being set to false in hiera common.yaml to disable it in carrenza.

hieradata/development.yaml
MahmudH: `govuk::apps::content_data_api::rabbitmq::amqp_pass: 'content_data_api'` sets the rabbitmq password in the development environment, whereas `govuk::apps::content_data_api::rabbitmq_password` is getting the password to be able to access rabbitmq through the app within the development environment.

theseanything: Let's just remove this line - as can't see where its being used

cbaines: Without it, I don't think the Content Data API will know what RabbitMQ password to use: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/content_data_api.pp#L103

PR #9092, opened 2019-05-07
Refactor dockerised elasticsearch to use govuk_containers

barrucadu:
This creates an init service, solving the problem of elasticsearch not starting in the dev VM on boot.

I've tested both replicating data from integration (after cleaning out some old snapshots - really need to get that automated soonish) and the search-api rake tasks.

---

This is part 1 of fixing the ES alerts in CI.  Part 2 will be a follow-up PR to ensure-absent `govuk_elasticsearch` and use this dockerised ES on the CI agents.


modules/govuk_containers/manifests/elasticsearch.pp
```diff
@@ -0,0 +1,47 @@
+# == Class: govuk_containers::redis
```
brucebolt: This comments needs updating from Redis

barrucadu:
Im not sure about the data replication.  The volume has the same name, so I'd hope it would persist, but I don't know.  I suspect many people didn't do the data replication anyway, because the s3 bucket was huge.

PR #9097, opened 2019-05-07
Fix a typo in the vagrant_credentials.yaml file

cbaines:


benthorner:
Rival app that never made it ;-).

cbaines:
> Wow gosh, how has that gone unnoticed for so long thinking I wonder if it actually does anything?
> 
> Thanks for catching

I think the hieradata relates to the Vagrantfile in the root of the gouvk-puppet repository [1], which is different to the development-vm Vagrantfile. I think it relates to running a VM to test govuk-puppet, but I imagine it was last used back when GOV.UK had an infrastructure team...

1: https://github.com/alphagov/govuk-puppet/blob/master/Vagrantfile

PR #9101, opened 2019-05-07
[do not merge] Use search-api /healthcheck endpoint for checks

bilbof:
**Note**: This should be merged after https://github.com/alphagov/search-api/pull/1538 and we are happy that the endpoint is working OK.

This will make Icinga check the JSON response of the `/healthcheck` endpoint that will be added to the search-api.

Trello: https://trello.com/c/cNvfdZfU/372

### DIFF
--- a/modules/govuk/manifests/apps/search_api.pp
+++ b/modules/govuk/manifests/apps/search_api.pp
@@ -108,7 +108,9 @@ class govuk::apps::search_api(
     app_type                 => 'rack',
     port                     => $port,
     sentry_dsn               => $sentry_dsn,
-    health_check_path        => '/search?q=search_healthcheck',
+    health_check_path        => '/healthcheck',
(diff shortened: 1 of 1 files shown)


bilbof:
Yeah, you're right. I didn't think of that.

Are there any other connectivity checks for ES? I've added a new task to the Trello card: https://trello.com/c/cNvfdZfU/372.

barrucadu:
I don't think there are.  I also don't think the check needs to be as complex as constructing a query and sending it to ES, it could be something as simple as getting the list of indices (which is a single method call on the ES client object I think).

bilbof:
I added https://github.com/alphagov/search-api/pull/1540 which should allow us to remove the existing healthcheck.

It won't make a get request to the search endpoint, though. I expect the smoke tests will catch this (I believe some of them publish docs and check the search responses afterwards).

PR #9103, opened 2019-05-08
Remove cache-clearing-service from Carrenza backends

schmie:
- This service was migrated to AWS as it requires direct access to the
migrated Cache instances.

solo: @schmie


hieradata/common.yaml
```diff
@@ -18,7 +18,6 @@ node_class: &node_class
   backend:
     apps:
       - asset-manager
-      - cache-clearing-service
```
cbaines: I don't think removing the application here will get Puppet to remove it from the machine, you'll need to leave this here, but set the `enabled` parameter to false, then deploy that change, then remove it from the list here.

fredericfran-gds: govuk::apps::cache_clearing_service::enabled: false

fredericfran-gds: I guess it is more clean this way

schmie: Right, last time we did this, we worked with instances which were redeployed after - thanks for catching it!

cbaines: The changes here https://github.com/alphagov/govuk-puppet/pull/9105 have now been deployed, so this is no longer an issue.

PR #9104, opened 2019-05-08
Check dockerised elasticsearch is running via TCP

barrucadu:
I couldn't get the check_proc_running check to pick up the dockerised
elasticsearch, so this does the next best (arguably even better)
thing.

---

[Trello card](https://trello.com/c/1jHfm7X9/702-fix-es-alerts-in-ci)


barrucadu:
Yeah, this is just to check that there's something there.  A more sophisticated check could `curl` it and look for the ES version string or something.

PR #9105, opened 2019-05-08
Disable cache_clearing_service in Carrenza

schmie:
This service has been migrated to AWS

solo: @schmie

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -343,6 +343,7 @@ govuk::apps::cache_clearing_service::enabled: false
 govuk::apps::cache_clearing_service::nagios_memory_warning: 2750
 govuk::apps::cache_clearing_service::nagios_memory_critical: 3000
 govuk::apps::cache_clearing_service::puppetdb_node_url: 'http://puppetdb.cluster/v2/nodes'
+govuk::apps::cache_clearing_service::rabbitmq::enabled: false
 govuk::apps::cache_clearing_service::rabbitmq_hosts:
   - rabbitmq-1.backend
   - rabbitmq-2.backend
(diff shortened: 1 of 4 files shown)


hieradata/common.yaml
```diff
@@ -339,7 +339,7 @@ govuk::apps::bouncer::nagios_memory_warning: 1400
 govuk::apps::bouncer::nagios_memory_critical: 1500
 govuk::apps::bouncer::unicorn_worker_processes: "8"
 
-govuk::apps::cache_clearing_service::enabled: true
+govuk::apps::cache_clearing_service::enabled: false
```
cbaines: Because the govuk-puppet development-vm uses the Carrenza hieradata, it might be good to keep the `enabled: true` value in `development.yaml`, so that the service is still setup in the development vm.

cbaines:
Another aspect of the configuration in Carrrenza is the RabbitMQ stuff. That's not affected by the `enabled` value for the service as far as I'm aware, and it appears to be hardcoded to `present` in the manifest [1].

1: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/cache_clearing_service/rabbitmq.pp

You could make it configurable, or just manually remove the relevant resources.

hieradata/common.yaml
```diff
@@ -343,6 +343,7 @@ govuk::apps::cache_clearing_service::enabled: false
 govuk::apps::cache_clearing_service::nagios_memory_warning: 2750
 govuk::apps::cache_clearing_service::nagios_memory_critical: 3000
 govuk::apps::cache_clearing_service::puppetdb_node_url: 'http://puppetdb.cluster/v2/nodes'
+govuk::apps::cache_clearing_service::rabbitmq::enabled: false
```
cbaines: If the default is disabled, this is redundant right?

schmie: Yes, but there is a script enforcing parity of parameters between hieradata and hieradata_aws, so it was necessary to add. Thanks for the review! : )

PR #9106, opened 2019-05-08
Fix the govuk_env_sync::conf_dir being declared twice

cbaines:
Copy the owner, group and mode values from the aws_auth class to the
govuk_env_sync class, and remove the $conf_dir management from the
aws_auth class.

### DIFF
--- a/modules/govuk_env_sync/manifests/aws_auth.pp
+++ b/modules/govuk_env_sync/manifests/aws_auth.pp
@@ -12,7 +12,7 @@ class govuk_env_sync::aws_auth(
   $aws_region = $govuk_env_sync::aws_region
   $app_domain_internal = hiera('app_domain_internal')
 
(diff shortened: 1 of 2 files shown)


benthorner:
This is an interesting change - my Puppet is pretty non-existent. Why did we make the title an array before?

cbaines:
> This is an interesting change - my Puppet is pretty non-existent. Why did we make the title an array before?

I guess because it's a shorthand way of managing both files.

The change I made here [1] conflicted with the aws_auth class though, as they say some things about the same files.

1: https://github.com/alphagov/govuk-puppet/commit/ba029758f9c8dc90253c1da7bbd9320ba9473511#diff-766c3f96e85dfa093dab7259a8e4d008

cbaines:
> Thanks for explaining the Puppet loop hack :-).

Indeed, I don't think the version of Puppet we're using has loops (although that's probably for the better with something configuration focused...).

PR #9109, opened 2019-05-08
Tidy up the Content Data Admin in Carrenza

cbaines:


hieradata/development.yaml
```diff
@@ -102,7 +102,14 @@ govuk::apps::ckan::priority_worker_processes: '1'
 govuk::apps::ckan::bulk_worker_processes: '1'
 govuk::apps::collections_publisher::enable_procfile_worker: false
 govuk::apps::contacts::extra_aliases: ['contacts-admin']
-govuk::apps::content_data_admin::enabled: true
```
theseanything: What's this for? @MahmudH added it in #9091 - so assumed we needed it

cbaines: We needed to enable it specifically in development, as I'd disabled it in the common hieradata. As part of these changes, I removed the common hieradata, including the bit disabling it [1]

1: https://github.com/alphagov/govuk-puppet/pull/9109/commits/2a08db48b1a9a6439e3822e43d3e26fa3dbd17aa#diff-0929d2e1dc2d61889d42d3c9bca6843fL378

As it's not disabled in the common hieradata, there isn't any need to specifically enable it in the development hieradata.

theseanything: I'm guessing the default is true?

cbaines: Indeed.

PR #9114, opened 2019-05-08
Forward draft-assets requests to carrenza

rtrinque:
We need to send the draft-assets requests to Carrenza as the
asset manager hasn't been migrated yet.

### DIFF
--- a/modules/router/manifests/draft_assets.pp
+++ b/modules/router/manifests/draft_assets.pp
@@ -19,11 +19,11 @@ class router::draft_assets(
   $enable_ssl = hiera('nginx_enable_ssl', true)
   $asset_manager_uploaded_assets_routes = hiera('router::assets_origin::asset_manager_uploaded_assets_routes', [])
 
+  $app_domain = hiera('app_domain')
+
   if $::aws_migration {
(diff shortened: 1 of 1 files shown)


modules/router/templates/draft-assets.conf.erb
```diff
@@ -50,7 +50,7 @@ server {
     return 404;
   }
   <%- if scope.lookupvar('::aws_migration') %>
-  set $upstream_asset_manager <%= @upstream_ssl ? 'https' : 'http' %>://asset-manager.<%= @app_domain %>;
+  set $upstream_asset_manager <%= @upstream_ssl ? 'https' : 'http' %>://draft-assets-carrenza.<%= @app_domain %>;
```
schmie: I think leaving app_domain here will result in draft-assets-carrenza.production.govuk-internal.digital - not sure if we have set app_domain_external to publishing.service.gov.uk in this context - that might be ideal

rtrinque: Yes, good point. By using app_domain instead of app_domain_internal we should fix that

PR #9116, opened 2019-05-08
Fix the icinga-slack-webhook to be installed with pip3

cbaines:
Hopefully this will generate an entrypoint that uses Python 3 (rather
than Python 2).

### DIFF
--- a/modules/icinga/manifests/config/slack.pp
+++ b/modules/icinga/manifests/config/slack.pp
@@ -2,6 +2,6 @@
 class icinga::config::slack {
   package {'icinga-slack-webhook':
     ensure   => '2.0.0',
-    provider => 'pip',
+    provider => pip3,
   }
 }


modules/icinga/manifests/config/slack.pp
```diff
@@ -2,6 +2,6 @@
 class icinga::config::slack {
   package {'icinga-slack-webhook':
     ensure   => '2.0.0',
-    provider => 'pip',
+    provider => pip3,
```
theseanything: Do you need this quoted?

cbaines: I'm unsure, but I saw other examples of it not being quoted [1], so I'm hoping it'll work.

1: https://github.com/alphagov/govuk-puppet/blob/be023ee704b4850d9ccd1e64e9f5ae548b522a40/modules/govuk_postgresql/manifests/wal_e/package.pp#L18

PR #9119, opened 2019-05-09
Push staging ES backup to integration after pulling from production

barrucadu:
At the moment, staging is based on production data from day (now-1)
and integration is based on staging data from day (now-1), so
integration is 2 days behind production.  By flipping the order of the
staging push & pull, integration will be based on the same production
snapshot as staging.

We don't want to give integration access to the govuk-production
snapshot repository, because then integration could *write* to the
production bucket, and an error in the env sync configuration could
result in production restoring data generated from integration.


barrucadu:
Not really, no.

PR #9131, opened 2019-05-10
Make transition app absent in Carrenza

afda16:
Update c7f3515

Transition has been migrated to AWS in staging/production. We need
to ensure the app resources are removed from the Carrenza machines
because there are cron jobs still trying to run tasks that are filling
disk space.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -42,6 +42,7 @@ node_class: &node_class
       - specialist-publisher
       - support
       - support-api
+      - transition
       - support_api_csv_env_sync
       - travel-advice-publisher
   bouncer:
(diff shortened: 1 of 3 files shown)


cbaines:
> Transition has been migrated to AWS in staging/production. We need
> to ensure the app resources are removed from the Carrenza machines
> because there are cron jobs still trying to run tasks that are filling
> disk space.

This'll definitely help, as what I expect to happen is when the files in `/var/apps/transition` are removed, the cron jobs will stop working. 

However, I don't think this'll actually remove the cron jobs themselves, as they're managed through whenever, with the config in the application [1], and this is applied through govuk-app-deployment. 

1: https://github.com/alphagov/transition/blob/master/config/schedule.rb

But this is still the thing to do in govuk-puppet I think.

afda16:
Thanks @cbaines, I can remove the cron jobs manually afterwards

PR #9136, opened 2019-05-10
Validate that loadbalancer::balance is configured with servers

cbaines:
Without some servers, the NGinx validation will fail. This follows on
from some issues cleaning up the CKAN configuration on the Carrenza
side of Production and Staging.


cbaines:
This'll hopefully work when this change is merged https://github.com/alphagov/govuk-puppet/pull/9135

PR #9138, opened 2019-05-10
kill all collectd processes when force restart collectd

fredericfran-gds:


modules/collectd/manifests/service.pp
```diff
@@ -5,6 +5,6 @@
 class collectd::service {
   service { 'collectd':
     ensure  => running,
-    restart => '/etc/init.d/collectd restart || { pkill -9 collectdmon; /etc/init.d/collectd start; }',
+    restart => "/etc/init.d/collectd restart || { ps aux | grep collectd | awk '{{print \$2}}' | xargs kill -9 ; /etc/init.d/collectd start; }",
```
barrucadu: `sudo killall -9 collectd` is shorter

fredericfran-gds:
not needed anymore.

modules/collectd/manifests/service.pp
fredericfran-gds: nope, does not work

PR #9146, opened 2019-05-14
Delete some more ensured-absent elasticsearch stuff

barrucadu:
The current state of play:

- The ES that powers site search is defined in govuk-aws / govuk-aws-data, not govuk-puppet.
- The ES used in the dev VM and in CI is a container defined in `govuk_containers::elasticsearch`
- There's some management stuff defined in `govuk_search`

All the stuff that was in `govuk_elasticsearch` is unused, as is some of the stuff that was in `govuk_search`.

One day I will have deleted all of govuk-puppet.


cbaines:
> One day I will have deleted all of govuk-puppet.

Changes on GOV.UK seem to occur at a exponentially deceasing rate, so the realistic prediction of when this'll happen is the same as the end of time/the universe itself...

cbaines:
> Changes on GOV.UK seem to occur at a exponentially deceasing rate, so the realistic prediction of when this'll happen is the same as the end of time/the universe itself...

Not that I don't admire your optimism.

PR #9147, opened 2019-05-14
Updates Node from v6 to v10

ChrisBAshton:
Node v10 is in LTS, and there are a number of NPM packages that specify a Node environment of at least V8.

Therefore we should move away from v6.

Depended on by https://github.com/alphagov/finder-frontend/pull/1125.

### DIFF
--- a/hieradata/common.trusty.yaml
+++ b/hieradata/common.trusty.yaml
@@ -43,4 +43,4 @@ base::esm::esm_password: "%{hiera('base::esm::esm_trusty_password')}"
 
 govuk_ppa::repo_ensure: 'present'
 
-nodejs::version: '6.14.3-1nodesource1'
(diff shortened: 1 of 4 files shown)


ChrisBAshton:
This failed when deployed to integration.

In any case, I've just discovered we can't upgrade our Node version until `statsd` have released a new version which supports Node 8+.

kevindew:
FYI if you didn't have the statsd problem you'd be wanting to do this in multiple steps.

The first step would be to update the apt source as per https://github.com/alphagov/govuk-puppet/pull/9147/files#diff-b7c2edf8732d227c94542fe36bc28ba4, then you'd need to merge that in and follow the [steps to update aptly](https://docs.publishing.service.gov.uk/manual/debian-packaging.html#third-party-repos) which would mean you could later update the package to the particular version.

PR #9152, opened 2019-05-16
Add configuration for healthchecks that use https

deborahchua:
The `datagovuk_publish` app's host uses `https` and is on the Paas,
 the rest of our gov.uk apps use `http` and is served internally.

The `check_json_healthcheck` now has an additional parameter to
accept a protocol, if one isn't specified it defaults to `http`.

### DIFF
--- a/modules/icinga/files/usr/lib/nagios/plugins/check_json_healthcheck
+++ b/modules/icinga/files/usr/lib/nagios/plugins/check_json_healthcheck
@@ -17,9 +17,10 @@ information:
         }
     }
 
-This script can be invoked with the healthcheck port, path and host as arguments:
+This script can be invoked with the healthcheck port, path, host(optional) and
+protocol(optional) as arguments:
 
-    check_json_healthcheck 3000 /healthcheck host_path
+    check_json_healthcheck 3000 /healthcheck host_path https
 
 """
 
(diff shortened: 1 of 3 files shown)


thomasleese:
I think you'll also need to add `protocol` to https://github.com/alphagov/govuk-puppet/blob/852d805025f36b14be67e8e994b902084e28f86c/modules/monitoring/manifests/checks/datagovuk_publish.pp#L16 and then include it in the Icinga block below `check_app_health!check_json_healthcheck!${port} ${health_check_path} ${host} ${protocol}`.

modules/icinga/files/usr/lib/nagios/plugins/check_json_healthcheck
```diff
@@ -151,7 +158,7 @@ if __name__ == "__main__":
         with closing(urlopen(json_request(check_url), timeout=20)) as response:
             healthcheck_info = HealthCheckInfo(json.load(response))
     except HTTPError as e:
-        report_error("healthcheck returned HTTP error %d" % (e.code,))
+        report_error("healthcheck returned HTTP/HTTPS error %d" % (e.code,))
```
cbaines: HTTPS doesn't have specific error codes, it's just HTTP over TLS. So HTTP error is correct in either case.

hieradata_aws/staging.yaml
```diff
@@ -224,6 +224,7 @@ licensify::apps::licensify_feed::environment: 'staging'
 monitoring::checks::aws_origin_domain: "staging.govuk.digital"
 monitoring::checks::datagovuk_publish::ensure: 'present'
 monitoring::checks::datagovuk_publish::host: 'publish-data-beta-staging.cloudapps.digital'
+monitoring::checks::datagovuk_publish::protocol: 'https'
```
cbaines: As you don't need to configure different values for this, I would suggest just specifying it in the manifest, rather than through a parameter.

deborahchua:
Thanks @cbaines @thomasleese I've made those changes now

PR #9158, opened 2019-05-16
Add procfile workers for email alert service queue consumers

edwardkerry:
The email-alert-service is being [modified to use the `govuk_message_queue_consumer` gem](https://github.com/alphagov/email-alert-service/pull/263). The gem relies on message queue connections being established via rake tasks in the Procfile.

This PR ensures that these rake tasks are run during app startup.

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_service.pp
+++ b/modules/govuk/manifests/apps/email_alert_service.pp
@@ -27,10 +27,6 @@
 # [*email_alert_api_bearer_token*]
 #   Bearer token for communication with the email-alert-api
 #
-# [*enable_major_change_queue_consumer*]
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/email_alert_service.pp
```diff
@@ -51,6 +51,21 @@
     command            => './bin/email_alert_service',
   }
 
+
+  govuk::procfile::worker { 'email-alert-service-major-change-consumer':
+    setenv_as      => $app_name,
+    enable_service => $enable_major-change-queue-consumer,
```
thomasleese: I think this will need to use only underscores in the variable name, and you'll also need to add it to the list of variables near the top around line 32-38.

PR #9171, opened 2019-05-22
Add dead letter queue for Content Data API

theseanything:
This adds a dead letter queue in RabbitMQ for message which aren't able to be processed by Content Data API.

### DIFF
--- a/modules/govuk/manifests/apps/content_data_api/rabbitmq.pp
+++ b/modules/govuk/manifests/apps/content_data_api/rabbitmq.pp
@@ -17,22 +17,43 @@
 # [*amqp_exchange*]
 #   The RabbitMQ exchange to read from (default: 'published_documents')
 #
+# [*amqp_dlx*]
+#   The RabbitMQ dead letter exchange (default: 'content_data_api_dlx')
+#
 # [*amqp_queue*]
 #   The RabbitMQ queue to set up for workers of this type to read from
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/apps/content_data_api/rabbitmq.pp
```diff
@@ -30,16 +30,33 @@
   $amqp_user  = 'content_data_api',
   $amqp_pass = undef,
   $amqp_exchange = 'published_documents',
+  $amqp_dlx = 'content_data_api_dlx',
   $amqp_queue = 'content_data_api',
-  $amqp_bulk_importing_queue = 'content_data_api_govuk_importer'
+  $amqp_bulk_importing_queue = 'content_data_api_govuk_importer',
+  $amqp_dead_letter_queue = 'content_data_api_dead_letter_queue'
```
cbaines: Ideally these new parameters would also be documented in the comment above. Especially the `amqp_dlx` on, as that's lots of letters...

kevindew:
@theseanything Just out of curiosity, what are you going to be using this for? 

@thomasleese and I experimented a bit with one so we could view a graph of how many things don't get to cache clearing service. I think I only went to the effort of setting up an exchange and not a queue so I think it didn't do anything (as no where to send the messages)

PR #9173, opened 2019-05-22
Increase max upload size for Content Publisher

benthorner:
https://trello.com/c/pqCfFrI1/873-support-upload-of-large-file-attachments

Previously this app was limited to uploads of 20M in size. Users who
exceed this limit would get a generic 413 error, which we want to avoid
until we have a more complete understanding of the user needs around
limits for different kinds of upload.


benthorner:
@kevindew sorted - I felt like my time was well spent there ;-).

PR #9174, opened 2019-05-22
Add temporary queue for Content Data API

theseanything:
Dead letter exchanges can only be specified on newly created queues. We need to create a temporary queues to use whilst we recreate the existing queues to use the dead letter exchange.

### DIFF
--- a/modules/govuk/manifests/apps/content_data_api/rabbitmq.pp
+++ b/modules/govuk/manifests/apps/content_data_api/rabbitmq.pp
@@ -39,7 +39,9 @@ class govuk::apps::content_data_api::rabbitmq (
   $amqp_exchange = 'published_documents',
   $amqp_dlx = 'content_data_api_dlx',
   $amqp_queue = 'content_data_api',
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/content_data_api/rabbitmq.pp
```diff
@@ -55,6 +57,69 @@
...
+  govuk_rabbitmq::queue_with_binding { $amqp_bulk_importing_queue_temp:
+    ensure        => 'present',
+    amqp_exchange => $amqp_exchange,
+    amqp_queue    => $amqp_bulk_importing_queue_temp,
+    routing_key   => '*.bulk.data-warehouse',
+    durable       => true,
+    arguments     => {
+      x-dead-letter-exchange => "${amqp_dlx}@/",
+    },
+  }
```
cbaines: This is probably unnecessary right, as we can just avoid enqueuing jobs while switching the queues around?

PR #9179, opened 2019-05-23
Revert use of temporary queues for Content Data API

theseanything:
This reverts previous commits that added temporary RabbitMQ queues for Content Data API. These queues are no longer needed.

Original queues need to be recreated to add dead letter exchange, but realised this can be achieved via a sever policy instead. Hence no need recreate queues or to create temporary queues.


cbaines:
The second commit looks to undo the first commit, could this be squashed in to a single commit?

theseanything:
> The second commit looks to undo the first commit, could this be squashed in to a single commit?

Ah realised the second commit was never merged anyway - so just dropped it

PR #9186, opened 2019-05-24
Exclude ESM from apt:sources for development-vm

erino:
- ESM for extended Ubuntu Trusty support was recently added as Trusty is EOL
and support by Canonical is now paid-for
- We cannot currently add the Canonical credetials for ESM access to the
development-vm as our support contract is per machine.
- This caused Puppet errors on development-vm deployment
- Decision: Simply remove ESM repos from development-vm hieradata

solo: @schmie

### DIFF
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -1,6 +1,24 @@
 ---
 app_domain: 'dev.gov.uk'
 
+apt::sources:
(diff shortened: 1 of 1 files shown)


schmie:
Thanks @sihugh 🥇 ;)

PR #9187, opened 2019-05-24
Set up Content Security Policy common environment variables

kevindew:
This sets up environment variables for the content security policy.
These are set in govuk::deploy::config as this is a place where they can
easily be configured in a global way, whereas if they're set within
govuk::app we'll need to set them up for each app individually.

These environment variables are intended to replace hardcoded logic in
the code [1]. The expectation is that we may want groups of apps to have
different report uri values (e.g. frontend and backend) so may well set
up values for these at node level hieradata.

If apps need to customise their approach to these they can set up their
own environment variables.

Finally, worth noting that if/when we want to remove these env vars
we'll need to add an ensure absent to govuk_envvar::init. If we just
delete these from the code puppet won't do anything to them.

[1]: https://github.com/alphagov/govuk_app_config/blob/f0fa82e2d8a73449efabd2414d93264a31afdf75/lib/govuk_app_config/govuk_content_security_policy.rb#L131-L136

### DIFF
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -27,6 +27,8 @@ base::supported_kernel::enabled: true
 clamav::use_service: false
 
 govuk::deploy::config::asset_root: 'http://assets-origin.dev.gov.uk'
+govuk::deploy::config::csp_report_only: false
+govuk::deploy::config::csp_report_uri: null
 govuk::deploy::config::website_root: 'http://www.dev.gov.uk'
 govuk::deploy::setup::ssh_keys:
     development_user: SEVIRUhFSEVIIFRoaXMgaXMgbm90IGEgcmVhbCBrZXkgYnV0IEknbSBnb2luZyB0byBtYWtlIGl0IGxvb2sgbGlrZSBpdCdzIG9uZSBqdXN0IHRvIHJlYWxseSBhbm5veSB5b3UhCg==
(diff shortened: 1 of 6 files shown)


hieradata/development.yaml
```diff
@@ -9,6 +9,8 @@ base::supported_kernel::enabled: true
 clamav::use_service: false
 
 govuk::deploy::config::asset_root: 'http://assets-origin.dev.gov.uk'
+govuk::deploy::config::csp_report_only: false
+govuk::deploy::config::csp_report_uri:
```
benthorner: Do we need this if it's defaulted?

modules/govuk/manifests/deploy/config.pp
```diff
@@ -84,6 +92,11 @@
     require => File['/etc/govuk'],
   }
 
+  $csp_report_only_value = $csp_report_only ? {
+    true => 'yes',
```
benthorner: Minor, but would it be safer, given the ternary, to default this to 'yes', in case someone makes a config error? That said, perhaps it's better to expose config issues!

hieradata/development.yaml
kevindew: This is used to replace the default. I'll make it clearer with an explicit null.

modules/govuk/manifests/deploy/config.pp
kevindew: This defaults to false at the parameter level so I think it's good.

benthorner: Indeed, I saw that. My concern was if someone tried to set the parameter to ‘yes’ it would be interpreted as ‘no’. I’m probably being over-cautious, though - the smaller the change, the smaller the edge case ;-).

PR #9203, opened 2019-05-30
Upgrade kernel in Carrenza to Xenial LTS

schmie:
- Carrenza production is currently running kernel 3.16x
- Extended Security Maintenance is only available for either 3.13
(Trusty vanilla) or 4.4 (Xenial backport)
- We go for 4.4 to hopefully minimise hardware support issues (knowledge
of why we were running 3.16 in the first place is lost.)

solo: @schmie


schmie:
We have not seen any issues related to the kernel upgrade in staging.

PR #9207, opened 2019-06-03
adjust python packages so that docker plugin of collectd works

fredericfran-gds:
# Context

The docker plugin of collectd was not working on the mirrorer machines in AWS leading to several icinga unknown alerts. The root cause of this issue is that docker and urllib3 python libraries were too old. Hence, we update these now. Requests python library had also to be updated.

# Decisions
1. pin the following python libararies: docker==2.7.0, urllib3==1.25.3, and requests==2.22.0

### DIFF
--- a/modules/collectd/manifests/plugin/docker.pp
+++ b/modules/collectd/manifests/plugin/docker.pp
@@ -31,9 +31,23 @@ class collectd::plugin::docker(
   # seems to be fixed in puppet 4: https://tickets.puppetlabs.com/browse/PUP-1073
   exec { 'pip install docker':
     path    => ['/usr/local/bin', '/usr/bin', '/bin'],
-    command => 'pip install "docker<=2.6.1"',
+    command => 'pip install "docker==2.7.0"',
     notify  => Class['collectd::service'],
-    unless  => 'pip list | grep "docker (2.6.1)"',
+    unless  => 'pip list | grep "docker (2.7.0)"',
+  }
+
+  exec { 'pip install urllib3':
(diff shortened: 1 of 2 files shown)


modules/collectd/manifests/plugin/docker.pp
```diff
@@ -31,16 +31,21 @@
   # seems to be fixed in puppet 4: https://tickets.puppetlabs.com/browse/PUP-1073
   exec { 'pip install docker':
     path    => ['/usr/local/bin', '/usr/bin', '/bin'],
-    command => 'pip install "docker<=2.6.1"',
+    command => 'pip install "docker==2.7.0"',
```
cbaines: I remember looking at this a while back. What's the current issue with version prior to 2.7?

fredericfran-gds: dockerplugin can import some dependencies: https://github.com/alphagov/docker-collectd-plugin/commit/e56ddb84536065786e0a55143faa8c6fc035c119 . 

Strangely it is only on the mirrorer machines that it behaves this way.

cbaines: Which dependencies can't it import? This sounds very similar to the issue Aga and I were trying to address with the previous round of changes, but it was a while ago, so I'm struggling to remember all the details...

fredericfran-gds: 21709:Jun  3 14:07:03 ip-10-12-6-151.eu-west-1.compute.internal collectd[27841]: Unhandled python exception in importing module: ImportError: cannot import name DependencyWarning

PR #9209, opened 2019-06-04
Adds govukcli to PATH

ChrisBAshton:
We have documentation that makes you do this manually. We
should change this documentation after merge:

https://docs.publishing.service.gov.uk/manual/get-ssh-access.html#3-access-remote-environments

### DIFF
--- a/modules/govuk/files/etc/profile.d/govukcli.sh
+++ b/modules/govuk/files/etc/profile.d/govukcli.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+export PATH="$PATH:/var/govuk/govuk-aws/tools/govukcli"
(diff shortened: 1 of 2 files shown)


cbaines:
> We have documentation that makes you do this manually. We
> should change this documentation after merge:
> 
> https://docs.publishing.service.gov.uk/manual/get-ssh-access.html#3-access-remote-environments

I'm not sure this is the case, if you're referring to [1], that relates to the host machine, which isn't managed by govuk-puppet (which is a guest on the host machine).

1: 
```
Run ln -s ~/govuk/govuk-aws/tools/govukcli /usr/local/bin/govukcli on your host machine to be able to use the govukcli tool from any directory.
```

cbaines:
Ah, yeah, the first line at least

PR #9210, opened 2019-06-04
Add option to monitor rabbitmq consumers

theseanything:
This makes the icinga alarm to monitor consumer optional. This is required as not all queues will have consumers attached at all times. This allows us to remove unnecessary and noise alerts from Icinga.

### DIFF
--- a/modules/govuk_rabbitmq/manifests/queue_with_binding.pp
+++ b/modules/govuk_rabbitmq/manifests/queue_with_binding.pp
@@ -37,13 +37,17 @@
 # [*arguments*]
 #   The hash of arguments to configure the queue.
 #
+# [*monitor_consumers*]
+#   Toggle to add icinga alarm to monitor queue for attached consumers. (default: true)
+#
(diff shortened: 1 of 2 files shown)


modules/govuk_rabbitmq/manifests/queue_with_binding.pp
```diff
@@ -65,10 +70,12 @@
...
   }
-
-  govuk_rabbitmq::monitor_consumers {"${title}_${amqp_queue}_consumer_monitoring":
-    ensure            => $ensure,
-    rabbitmq_hostname => 'localhost',
-    rabbitmq_queue    => $amqp_queue,
+  
+  if $monitor_consumers {
+    govuk_rabbitmq::monitor_consumers {"${title}_${amqp_queue}_consumer_monitoring":
+      ensure            => $ensure,
```
cbaines: I've checked, and this class just involves Icinga configuration, which I believe is fully managed, so making the configuration itself conditional should be OK (rather than changing the `ensure` value).

PR #9212, opened 2019-06-04
Clone host ~/.gitconfig into VM

ChrisBAshton:
Copies your host machine `.gitconfig` so that you don't have to set the values globally & manually.

It means you can `vagrant ssh` and then `git pull` and everything "just works".

NB we have instructions for setting this manually, which should be removed post-merge:
https://docs.publishing.service.gov.uk/manual/get-started.html\#set-your-git-username-and-email

### DIFF
--- a/development-vm/Vagrantfile
+++ b/development-vm/Vagrantfile
@@ -105,6 +105,8 @@ Vagrant.configure("2") do |config|
     config.vm.synced_folder "../..", "/var/govuk", :type => :nfs
   end
 
+  config.vm.provision "file", source: "~/.gitconfig", destination: "~/.gitconfig"
+
   config = common_config(config)
 
   # See Vagrantfile.localconfig.example for instructions on


development-vm/Vagrantfile
```diff
@@ -105,6 +105,8 @@ Vagrant.configure("2") do |config|
     config.vm.synced_folder "../..", "/var/govuk", :type => :nfs
   end
 
+  config.vm.provision "file", source: "~/.gitconfig", destination: "~/.gitconfig"
```
cbaines: I wonder what this does if the file isn't present?

ChrisBAshton: Good point. I imagine it would error. But I'll raise a PR to update [the 'Get Started' page](https://docs.publishing.service.gov.uk/manual/get-started.html) so that it instructs people to set up their global git config on their host machine, before they try doing anything in the VM.

ChrisBAshton: Raised https://github.com/alphagov/govuk-developer-docs/pull/1783

PR #9218, opened 2019-06-05
Fix the Licensify CollectD process regexes

cbaines:
These worked in Integration, but not in Staging or Production, as the
regexes were too specific to Integration. This commit changes them to
have wildcards where the environment can be specified, and also
escapes the forward slashes for neatness.

### DIFF
--- a/modules/licensify/manifests/apps/licensify.pp
+++ b/modules/licensify/manifests/apps/licensify.pp
@@ -16,7 +16,7 @@ class licensify::apps::licensify (
     require                        => File['/etc/licensing'],
     proxy_http_version_1_1_enabled => true,
     log_format_is_json             => true,
-    collectd_process_regex         => 'java -Duser.dir=/data/vhost/licensify.integration.publishing.service.gov.uk/licensify-.*',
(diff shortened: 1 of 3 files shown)


modules/licensify/manifests/apps/licensify.pp
```diff
@@ -16,7 +16,7 @@
     require                        => File['/etc/licensing'],
     proxy_http_version_1_1_enabled => true,
     log_format_is_json             => true,
-    collectd_process_regex         => 'java -Duser.dir=/data/vhost/licensify.integration.publishing.service.gov.uk/licensify-.*',
+    collectd_process_regex         => 'java -Duser.dir=\/data\/vhost\/licensify\..*publishing\.service\.gov\.uk\/licensify-.*',
```
ChrisBAshton: Hmmm will this fail on Production because we don't have `production.publishing.service.gov.uk`? Think it needs a dot removed...?

```suggestion
    collectd_process_regex         => 'java -Duser.dir=\/data\/vhost\/licensify\.*publishing\.service\.gov\.uk\/licensify-.*',
```

ChrisBAshton: Whoops, misread this - you're only escaping one `\.`, not two - so as you've said IRL, this is fine 😁

PR #9219, opened 2019-06-05
Increase the memory monitoring thresholds for Sidekiq monitoring

cbaines:
The CollectD process plugin has only just been configured correctly,
so this is something that's only now appearing.

The memory usage seems to be significantly higher in AWS, compared to
Carrenza. I'm not sure why this is, but this higher threshold reflects
the current usage in AWS.

### DIFF
--- a/modules/govuk/manifests/apps/sidekiq_monitoring.pp
+++ b/modules/govuk/manifests/apps/sidekiq_monitoring.pp
@@ -243,6 +243,8 @@ class govuk::apps::sidekiq_monitoring (
     enable_nginx_vhost     => false,
     hasrestart             => true,
     collectd_process_regex => '/data/apps/sidekiq-monitoring/shared/bundle/ruby/.*/bin/rackup .*',
+    nagios_memory_warning  => 1500,
+    nagios_memory_critical => 1600,
   }
 
   govuk::app::envvar::redis{


modules/govuk/manifests/apps/sidekiq_monitoring.pp
```diff
@@ -243,6 +243,8 @@
     enable_nginx_vhost     => false,
     hasrestart             => true,
     collectd_process_regex => '/data/apps/sidekiq-monitoring/shared/bundle/ruby/.*/bin/rackup .*',
+    nagios_memory_warning  => 1500,
+    nagios_memory_critical => 1600,
```
ChrisBAshton: The reasoning seems legit 👍 what am I looking at though - megabytes?

What are the current levels roughly? Not sure how far away the 1500 is from the actual readings.

And are the warning/critical values a little close together? I suspect we'd skip 'warning' altogether in a lot of cases (but that's based on nothing but instinct!)

cbaines: > The reasoning seems legit +1 what am I looking at though - megabytes?

Yep.

> What are the current levels roughly? Not sure how far away the 1500 is from the actual readings.

On AWS, it's 1100 to 1300 MB currently.

> And are the warning/critical values a little close together? I suspect we'd skip 'warning' altogether in a lot of cases (but that's based on nothing but instinct!)

Maybe, I definitely don't see much difference between a warning and critical status for this check, all I'm doing here is setting the thresholds to be higher than the current values.

PR #9222, opened 2019-06-05
Remove Imminence from Carrenza

cbaines:


ChrisBAshton:
What is Imminence, and why are we removing it?

(PR description would be good - then more than happy to approve 👍 )

cbaines:
> What is Imminence, and why are we removing it?

It's an application that relates to geospatial data (I think). It's now running in AWS, but it's still running on the Carrenza infrastructure, which is a bit confusing. I've updated the commit message to say this.

> (PR description would be good - then more than happy to approve +1 )

PR #9223, opened 2019-06-05
Add a Icinga check for the CollectD process

cbaines:
Sometimes, this isn't running when it should be, and there's no
explicit check for this yet.

### DIFF
--- a/modules/collectd/manifests/service.pp
+++ b/modules/collectd/manifests/service.pp
@@ -7,4 +7,11 @@ class collectd::service {
     ensure  => running,
     restart => "/etc/init.d/collectd restart || { pkill -9 collectd; ps aux | grep '/usr/lib/collectd/file_handles.sh' | awk '{{print \$2}}' | xargs kill -9 ;/etc/init.d/collectd start; }",
   }
+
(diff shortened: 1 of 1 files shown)


modules/collectd/manifests/service.pp
```diff
@@ -7,4 +7,11 @@
     ensure  => running,
     restart => "/etc/init.d/collectd restart || { pkill -9 collectd; ps aux | grep '/usr/lib/collectd/file_handles.sh' | awk '{{print \$2}}' | xargs kill -9 ;/etc/init.d/collectd start; }",
   }
+
+  @@icinga::check { "check_collectd_running_${::hostname}":
+    check_command       => 'check_nrpe!check_proc_running!collectd',
+    service_description => 'collectd not running',
```
benthorner: Should the description be for the positive case?

cbaines: Yeah, maybe "collectd system statistics daemon" would be better.

cbaines: I've merged it with this change, but do let me know what you think!

PR #9226, opened 2019-06-06
Switch AWS apps back to using postgres directly

elliotcm:
This removes the use of pg bouncer in AWS as it's considered unnecessary now.

This also re-enabled prepared statements which was a feature disabled during the use of pg bouncer.


elliotcm:
Branch tested on staging.

PR #9231, opened 2019-06-07
Improve the smokey-loop upstart service

cbaines:
Currently, as /opt/smokey/tests_json_output.sh is a bash script, when
the service stops, that bash script is stopped, but the actual
cucumber process keeps running. Bash is not a process manager.

Instead, better use upstart to avoid this issue. Use setuid and setgid
rather than sudo, set respawn limit as this service should always
respawn, and a long kill timeout to let cucumber exit
gracefully. Also, add a post-stop script (that runs after each
respawn) to move the output to the proper location.

These changes match up with changes in Smokey to the
tests_json_output.sh script.

### DIFF
--- a/modules/monitoring/templates/smokey-loop.conf
+++ b/modules/monitoring/templates/smokey-loop.conf
@@ -1,9 +1,19 @@
 description "Smokey loop outputs JSON which is consumed by monitoring"
 
+setuid smokey
+setgid smokey
+
 # Run in a continuous loop. `respawn` catches non-zero exit codes. Whereas
 # `and stopped` catches normal exits. The default `respawn limit` will
 # prevent this from spinning if the script is broken.
 start on runlevel [2345] and stopped $UPSTART_JOB
 respawn
+respawn limit unlimited
 
-exec sudo -u smokey /opt/smokey/tests_json_output.sh /tmp/smokey.json <%= @environment %>
+kill timeout 240
+
+exec /opt/smokey/tests_json_output.sh /tmp/smokey.json.tmp integration
+
+post-stop script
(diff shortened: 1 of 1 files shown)


cbaines:
This requires https://github.com/alphagov/smokey/pull/545

modules/monitoring/templates/smokey-loop.conf
```diff
@@ -1,9 +1,19 @@
...
 respawn
+respawn limit unlimited
+
+kill timeout 240
+
+exec /opt/smokey/tests_json_output.sh /tmp/smokey.json.tmp integration
 
-exec sudo -u smokey /opt/smokey/tests_json_output.sh /tmp/smokey.json <%= @environment %>
+post-stop script
+    mv /tmp/smokey.json.tmp /tmp/smokey.json
```
ChrisBAshton: Why don't we output to `/tmp/smokey.json` straight away in line 15?

cbaines: I think that cucumber might clear the file when it starts. The purpose of writing a JSON file out is so that Icinga can check the contents, so really you want it to contain useful information most of the time. Does that make sense?

cbaines:
> Looks good to me - could you edit the commit to actually link to [alphagov/smokey#545](https://github.com/alphagov/smokey/pull/545) ?

Sure

modules/monitoring/templates/smokey-loop.conf
ChrisBAshton: Yep - so otherwise, Icinga would not be able to check its contents while it's running. Having it as a tmp file means the results are always available 👍

PR #9233, opened 2019-06-07
Make the Link Checker API absent in Carrenza Prod and Staging

cbaines:
The Link Checker API is now running in AWS, but still running (but
should be unused) in Carrenza Production and Carrenza Staging.

This old, unused version of the app still runinng in Carrenza is
potentially confusing, and consumes some resources. Therefore, this
commit makes it absent.


cbaines:
> `runinng` typo in commit. Code looks good.

Good spot, I've fixed the typo.

PR #9234, opened 2019-06-07
Increase the number of Unicorn workers for the Publishing API

cbaines:
There seems to be peaks of activity, currently there are only 2
workers per machine, but there is plenty of RAM to run more.

### DIFF
--- a/modules/govuk/manifests/apps/publishing_api.pp
+++ b/modules/govuk/manifests/apps/publishing_api.pp
@@ -37,6 +37,9 @@
 #   store running while testing unrelated features.
 #   Default: ''
 #
+# [*unicorn_worker_processes*]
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/publishing_api.pp
```diff
@@ -125,6 +128,7 @@
   $draft_content_store_bearer_token = undef,
   $router_api_bearer_token = undef,
   $suppress_draft_store_502_error = '',
+  $unicorn_worker_processes = undef,
```
ChrisBAshton: Perhaps set this to `2` explicitly so everyone knows what the default value is without having to 'know' or Google it?

ChrisBAshton: 👍

PR #9236, opened 2019-06-07
Remove pg bouncer from db_admin servers

elliotcm:
Apps in AWS now talk directly to RDS.

Merge after https://github.com/alphagov/govuk-puppet/pull/9226 is deployed.


elliotcm:
> This won't actually uninstall pgbouncer (clean up the files, uninstall the service, etc)

After considerable investigation it looks like the puppet library used to install pg_bouncer does not have a way to uninstall it, so it's probably going to be easier to leave it on there and rebuild the machine if we need to get rid of it.  In the meantime it shouldn't cause any specific issues.

PR #9240, opened 2019-06-07
Make the Support apps absent in Carrenza Prod and Staging

cbaines:
The Support app and Support API are now running in AWS, but still
running (but should be unused) in Carrenza Production and Carrenza
Staging.

This old, unused versions of these apps still running in Carrenza is
potentially confusing, and consumes some resources. Therefore, this
commit makes them absent.


hieradata/production.yaml
```diff
@@ -81,7 +81,9 @@ govuk::apps::publisher::email_group_business: 'publisher-alerts-business@digital
 govuk::apps::publisher::email_group_citizen: 'publisher-alerts-citizen@digital.cabinet-office.gov.uk'
 govuk::apps::publishing_api::event_log_aws_bucketname: 'govuk-publishing-api-event-log-production'
 govuk::apps::static::ga_universal_id: 'UA-26179049-1'
+govuk::apps::support::ensure: 'absent'
 govuk::apps::support::aws_s3_bucket_name: 'govuk-production-support-api-csvs'
+govuk::apps::support_api::ensure: 'absent'
```
ChrisBAshton: Are we able to remove the rest of the config associated with this? (i.e. lines 87 onwards)

cbaines: Yeah, although, I think it's generally good to leave the related configuration in place, until the `ensure: 'absent'` change has been applied, as the related configuration might actually change what Puppet thinks it needs to remove.

ChrisBAshton: 👍 so there'll be another PR later to remove all this?

cbaines: Yep, remove what can be removed, and move the rest to the development-vm hieradata file.

ChrisBAshton: Good stuff!

PR #9247, opened 2019-06-10
Update adding-a-new-app.md

elliotcm:


docs/adding-a-new-app.md
```diff
@@ -197,7 +195,6 @@ Check if your app appears in these files (use existing apps as examples).
   * hieradata_aws/common.yaml (deployable_applications)
   * hieradata_aws/common.yaml (grafana::dashboards::application_dashboards)
   * hieradata/development.yaml (hosts::development::apps)
-  * modules/govuk/manifests/node/s_db_admin.pp
```
barrucadu: Should this line be gone?  It wasn't added in 6a08217f8c1f6aea179197a7f67995d70f4d3f89

elliotcm: Oh right because it adds all the DBs there.  I'll put that back.

PR #9249, opened 2019-06-10
Update the search-api index size metric names

cbaines:
Metrics ending in `.count` are usually used for timers in StatsD, and
there is some specific Graphite storage aggregation configuration for
them [1].

This change updates the use of the metrics to match the new name [2].

1: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/files/node/s_graphite/storage-aggregation.conf#L35-L40
```
[count]
pattern = \.count$
aggregationMethod = sum
xFilesFactor = 0.1
```

2: https://github.com/alphagov/search-api/pull/1577

### DIFF
--- a/modules/grafana/templates/dashboards_aws/_index_size.json.erb
+++ b/modules/grafana/templates/dashboards_aws/_index_size.json.erb
@@ -50,17 +50,17 @@
         "targets": [
           {
             "refId": "A",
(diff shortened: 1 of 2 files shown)


cbaines:
Depends on https://github.com/alphagov/search-api/pull/1577

PR #9250, opened 2019-06-10
Add specific RDS monitoring thresholds for Content Data API

theseanything:
This allows RDS monitoring threshold to be set specific to each app using exported resources. We are also changing the thresholds of the Content Data API. As it is a larger database with greater rate of change, thresholds are increased so alerts occur sooner.

### DIFF
--- a/modules/govuk/manifests/apps/content_data_api/db.pp
+++ b/modules/govuk/manifests/apps/content_data_api/db.pp
@@ -15,4 +15,18 @@ class govuk::apps::content_data_api::db (
     extensions          => ['plpgsql'],
     enable_in_pgbouncer => false,
   }
+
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/content_data_api/db.pp
```diff
@@ -15,4 +15,18 @@
...
+    memory_critical  => 1,
+    storage_warning  => 100,
+    storage_critical => 50,
+  }
+
+  @@monitoring::checks::rds_config { 'content-data-api-postgresql-standby':
+    memory_warning   => 2,
+    memory_critical  => 1,
+    storage_warning  => 100,
+    storage_critical => 50,
```
cbaines: Given we've over-provisioned the storage, it might be useful to set higher thresholds for storage?

PR #9255, opened 2019-06-11
Patch Jenkins token-macro plugin

afda16:
Jenkins has just detected new plugin versions as we rolled out an update. The
token-macro patch fixes an XML External Entity processing vulnerability.

### DIFF
--- a/hieradata_aws/class/jenkins.yaml
+++ b/hieradata_aws/class/jenkins.yaml
@@ -210,7 +210,7 @@ govuk_jenkins::plugins:
   timestamper:
     version: '1.9'
   token-macro:
-    version: '2.7'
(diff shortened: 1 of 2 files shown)


hieradata_aws/class/jenkins.yaml
```diff
@@ -196,7 +196,7 @@ govuk_jenkins::plugins:
   simple-theme-plugin:
     version: '0.5.1'
   slack:
-    version: '2.20'
+    version: '2.3'
```
cbaines: Is this a downgrade or an upgrade?

afda16: My mistake, I removed that change, thanks!

PR #9267, opened 2019-06-13
Add ckan log file for standard output

kentsanggds:
## What

Existing `app.out.log` doesn't appear to be logging anything, the configuration update to ckan.ini has been tested on integration and is providing some logs.

### DIFF
--- a/modules/govuk/templates/ckan/ckan.ini.erb
+++ b/modules/govuk/templates/ckan/ckan.ini.erb
@@ -198,30 +198,36 @@ smtp.mail_from = team@data.gov.uk
 keys = root, ckan, ckanext
 
 [handlers]
-keys = console
+keys = console, file
 
 [formatters]
 keys = generic
 
 [logger_root]
 level = WARNING
-handlers = console
+handlers = console, file
 
 [logger_ckan]
 level = INFO
-handlers = console
+handlers = console, file
 qualname = ckan
 propagate = 0
 
 [logger_ckanext]
 level = DEBUG
-handlers = console
+handlers = console, file
 qualname = ckanext
 propagate = 0
 
 [handler_console]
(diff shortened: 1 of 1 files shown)


cbaines:
> Existing `app.out.log` doesn't appear to be logging anything, the configuration update to ckan.ini has been tested on integration and is providing some logs.

If this is worth saying, it's probably worth having in the commit message.

cbaines:
There are a number of log files that look to be being used. What extra logging is being written to this new file?

kentsanggds:
Output for standard log messages are not being output to `app.out.log`, when I tried to debug an issue by turning on the debug feature in `ckan.ini` logs were still not being output despite the flag and I had to add the additional configuration in order to be able to debug an issue. So adding the additional config will make it easier for the next dev to work on if there needs to be an investigation on an issue which can't be reproduced on local stacks.

kentsanggds:
I'll add the extra info in the commit message

cbaines:
> Output for standard log messages are not being output to `app.out.log`, when I tried to debug an issue by turning on the debug feature in `ckan.ini` logs were still not being output despite the flag and I had to add the additional configuration in order to be able to debug an issue. So adding the additional config will make it easier for the next dev to work on if there needs to be an investigation on an issue which can't be reproduced on local stacks.

Right, have you considered looking at the Gunicorn configuration to change this? It looks to me like Gunicorn is meant to be something with this file (I can see `--log-file /var/log/ckan/app.out.log`), maybe `--capture-output` needs to be passed as well?

Otherwise, if you intend not to use the console output, to avoid confusion, it would be probably good to change the Gunicorn/CKAN configuration accordingly.

kentsanggds:
ok, I'll try that out

PR #9271, opened 2019-06-14
Fix missing region for exported RDS alarms

theseanything:
This fixes a bug whereby the RDS monitoring for exported resources are missing a region.

### DIFF
--- a/modules/monitoring/manifests/checks/rds.pp
+++ b/modules/monitoring/manifests/checks/rds.pp
@@ -53,6 +53,8 @@ class monitoring::checks::rds (
       monitoring::checks::rds_config { $servers:
         region => $region,
       }
-      Monitoring::Checks::Rds_config <<| |>>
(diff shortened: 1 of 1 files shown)


modules/monitoring/manifests/checks/rds.pp
```diff
@@ -53,6 +53,8 @@
       monitoring::checks::rds_config { $servers:
         region => $region,
       }
-      Monitoring::Checks::Rds_config <<| |>>
+      Monitoring::Checks::Rds_config <<| |>> {
+        region => $region,
+      }
```
leenagupte: This seems very similar to the section above 🤔

theseanything: yah, so I think lines 53-55 create a bunch of resources depending on whats in hieradata ($servers). Then 56-58 collects any other monitoring resources made else where in puppet manifests.

PR #9279, opened 2019-06-14
Make ES data sync scripts handle both ES5 and ES6

barrucadu:
This is part 1 of making the data sync work for ES6.  Part 2 will be adding the hieradata for the new jobs, which I think will be easier to review as a separate PR.

---

[Trello card](https://trello.com/c/iTTaegly/803-enable-es6-data-sync)


barrucadu:
Fixed a typo

PR #9284, opened 2019-06-17
Add hieradata for ES6 data sync

barrucadu:
~This is based on #9279 and should be rebased against master when that is merged.~

---

[Trello card](https://trello.com/c/iTTaegly/803-enable-es6-data-sync)


barrucadu:
RE maintains https://docs.publishing.service.gov.uk/manual/govuk-env-sync.html

PR #9285, opened 2019-06-17
Add Icinga check for AWS LB Healthy Hosts

afda16:
On AWS we deploy ELBs and ALBs to route traffic to the application
instances. Recently we started to update ALBs to support a target
group per application with a dedicated health check, so traffic
is only routed to the instance when the app is healthy.

This change has highlighted the problem that we are not monitoring
the number of healthy hosts behind ELBs or ALB Target Groups on Icinga.
There are CloudWatch alerts but these are probably not monitored by
the teams.

The check added here discovers the instances (for ELBs) or target groups
(for ALBs) of a given Load Balancer and checks if the number of healthy
hosts attached are above the critical and warning levels. It is possible
to ignore Target Groups is we know they are always unhealthy because an
app hasn't been properly decommissioned. We are also adding a initial
configuration on Integration to monitor the LBs that we migrated to use
application health checks.

### DIFF
--- a/hieradata_aws/integration.yaml
+++ b/hieradata_aws/integration.yaml
@@ -306,6 +306,25 @@ monitoring::checks::sidekiq::enable_support_check: false
 monitoring::checks::pingdom::enable: false
 monitoring::checks::ses::region: 'eu-west-1'
 monitoring::checks::rds::region: 'eu-west-1'
+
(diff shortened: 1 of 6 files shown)


afda16:
Documentation added here: https://github.com/alphagov/govuk-developer-docs/pull/1812

PR #9288, opened 2019-06-17
Add AWS LB Healthy Host alerts to Staging

afda16:
Related with 392d0d9, add the new LB Icinga checks to Staging.


afda16:
I updated the Rakefile instead, I wasn't aware of this config, thanks!

PR #9289, opened 2019-06-17
Add alarm for msgs in content data dead letter queue

theseanything:
This adds Icinga alarms for the Content Data Letter queue. They will be raised if the queue length reaches passed the specified thresholds. 

We expect to have no and minimal message within this queue.


theseanything:
Not ready - for merging.

PR #9294, opened 2019-06-18
Switch licence-finder to ES6

barrucadu:
[Trello card](https://trello.com/c/NrbcW7ly/807-switch-licence-finder-over-to-es6-in-production)


barrucadu:
I'll merge this so it deploys to integration, and revert if it's bad.

PR #9297, opened 2019-06-19
Remove references to content performance manager in VM Procfile

theseanything:
Content Performance Manager has been renamed to Content Data API and old references should be removed.

### DIFF
--- a/development-vm/Procfile
+++ b/development-vm/Procfile
@@ -148,11 +148,8 @@ draft-router-api:            govuk_setenv draft-router-api            ./run_in.s
 # grafana has reserved port 3204
 manuals-publisher:  govuk_setenv manuals-publisher  ./run_in.sh ../../manuals-publisher bundle exec rails server -p 3205
 manuals-publisher-worker: govuk_setenv manuals-publisher ./run_in.sh ../../manuals-publisher bundle exec sidekiq -C ./config/sidekiq.yml
-content-performance-manager:      govuk_setenv content-performance-manager ./run_in.sh ../../content-performance-manager bundle exec unicorn -c config/unicorn.rb -p 3235
+# content-performance-manager used port 3235
 # sidekiq-monitoring for content-performance-manager uses port 3207
-content-performance-manager-publishing-api-consumer: govuk_setenv content-performance-manager ./run_in.sh ../../content-performance-manager bundle exec rake publishing_api:consumer
-content-performance-manager-bulk-import-publishing-api-consumer: govuk_setenv content-performance-manager ./run_in.sh ../../content-performance-manager bundle exec rake publishing_api:bulk_import_consumer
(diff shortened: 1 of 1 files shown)


development-vm/Procfile
```diff
@@ -147,11 +147,6 @@ draft-router-api:            govuk_setenv draft-router-api            ./run_in.s
 # grafana has reserved port 3204
 manuals-publisher:  govuk_setenv manuals-publisher  ./run_in.sh ../../manuals-publisher bundle exec rails server -p 3205
 manuals-publisher-worker: govuk_setenv manuals-publisher ./run_in.sh ../../manuals-publisher bundle exec sidekiq -C ./config/sidekiq.yml
-content-performance-manager:      govuk_setenv content-performance-manager ./run_in.sh ../../content-performance-manager bundle exec unicorn -c config/unicorn.rb -p 3235
-# sidekiq-monitoring for content-performance-manager uses port 3207
-content-performance-manager-publishing-api-consumer: govuk_setenv content-performance-manager ./run_in.sh ../../content-performance-manager bundle exec rake publishing_api:consumer
-content-performance-manager-bulk-import-publishing-api-consumer: govuk_setenv content-performance-manager ./run_in.sh ../../content-performance-manager bundle exec rake publishing_api:bulk_import_consumer
-content-performance-manager-sidekiq-default:  govuk_setenv content-performance-manager ./run_in.sh ../../content-performance-manager bundle exec sidekiq -C ./config/sidekiq/default.yml
```
cbaines: Usually, the ports used by old applications are still reserved going forward (I guess this could be useful for going back and running old things).

So, it might be better to replace these lines with appropriate comments?

issyl0:
@theseanything There’s an outstanding comment on this.

development-vm/Procfile
theseanything: Yup - I've updated this to just comment these lines out

issyl0: Ah, sorry, @theseanything, I think Chris meant comments like https://github.com/alphagov/govuk-puppet/pull/9297/files#diff-9a338283e0ec94af59147eb68fee7be3R135.

development-vm/Procfile
```diff
@@ -148,11 +148,12 @@ draft-router-api:            govuk_setenv draft-router-api            ./run_in.s
...
+# content-performance-manager-publishing-api-consumer: govuk_setenv content-performance-manager ./run_in.sh ../../content-performance-manager bundle exec rake publishing_api:consumer
+# content-performance-manager-bulk-import-publishing-api-consumer: govuk_setenv content-performance-manager ./run_in.sh ../../content-performance-manager bundle exec rake publishing_api:bulk_import_consumer
+# content-performance-manager-sidekiq-default:  govuk_setenv content-performance-manager ./run_in.sh ../../content-performance-manager bundle exec sidekiq -C ./config/sidekiq/default.yml
```
cbaines: As the consumers didn't use a port, these lines could probably go.

PR #9306, opened 2019-06-20
Fix search-api metric names in Grafana and Icinga

barrucadu:
In https://github.com/alphagov/search-api/pull/1569/files#diff-083eeb7a1599f2e1049dd88a094efa20R211, the metric names were changes to include the cluster (`cluster_A` or `cluster_B`), and the `.total` metric was renamed to `.count`.


barrucadu:
Could do, especially when we're in the middle of the A/B switch; but probably not until after we've done the mass import of documents.

PR #9319, opened 2019-06-25
Update transition scheduled jobs

kevindew:
Trello: https://trello.com/c/y8KULrlo/980-dont-run-transition-cdn-deploys-in-the-middle-of-the-night

There was an incident last week where the bouncer CDN broke at 4am due to Fastly changes being deployed automatically each day at that time. An action following this incident was to look into why that change was done at 4am given that any problems from it would be better dealt with in the working day.

Following this I did some investigation into the Transition jenkins jobs. There was [one that ran only in working hours](https://github.com/alphagov/govuk-puppet/blob/e2b63bc849f378fc635099c29963f206d58cd8a3/modules/govuk_jenkins/templates/jobs/transition_load_site_config.yaml.erb) and only when [transition-config](https://github.com/alphagov/transition-config) had changed. There was a [second](https://github.com/alphagov/govuk-puppet/blob/e2b63bc849f378fc635099c29963f206d58cd8a3/modules/govuk_jenkins/templates/jobs/transition_load_site_config.yaml.erb) that runs at 3 am each morning which imports data from s3, updates DNS and updates the CDN. This smelt a bit funny to me.

What I've done to try resolve this is I've split the 3am job into two: one to import data from s3 and still run at 3am; and the other to update DNS at 12pm each day, as I couldn't find a reason for that to be out of hours.

I'm theorising that this won't result in any problems in transition as all 3 jobs seem to be distinct. 

Along the way I've also re-instated this project running in on our staging environment. I believe this was turned off by accident about 6-8 months ago. I've also removed some old redirect jobs and done some general house keeping.

Given this has only been running in production for the last 6-8 months and doesn't run in integration then it is going to be a pain to validate whether this has any problems. 

@barrucadu I've tagged you to review since I think you were the last person to work on these files, if you know of someone else who'd be good please let me know.


modules/govuk_jenkins/templates/jobs/transition_import_dns.yaml.erb
```diff
@@ -0,0 +1,23 @@
...
+    logrotate:
+      numToKeep: 30
+    properties:
+      - build-discarder:
+          days-to-keep: 30
+          artifact-num-to-keep: 5
+    builders:
+        - shell: |
+            HOSTNAME=$(govuk_node_list -c backend --single-node)
+            ssh deploy@$HOSTNAME "cd /var/apps/transition && govuk_setenv transition bundle exec rake import:dns_details"
```
barrucadu: I've looked at the code for this in transition, and I don't think this task updates the CDN:

- https://github.com/alphagov/transition/blob/master/lib/tasks/import/dns_details.rake
- https://github.com/alphagov/transition/blob/master/lib/transition/import/dns_details.rb

Am I missing something?

modules/govuk_jenkins/templates/jobs/transition_load_site_config.yaml.erb
```diff
@@ -22,15 +27,13 @@
           artifact-num-to-keep: 5
     builders:
         - shell: |
-            <%- if scope.lookupvar('::aws_migration') %>
             HOSTNAME=$(govuk_node_list -c backend --single-node)
-            <%- else -%>
-            HOSTNAME="backend-1.backend"
-            <%- end -%>
             rsync --delete -avz -e ssh ./ deploy@$HOSTNAME:/var/apps/transition/data/transition-config
             ssh deploy@$HOSTNAME "cd /var/apps/transition && govuk_setenv transition bundle exec rake import:all:orgs_sites_hosts"
```
barrucadu: I think this is the task which updates the redirects used by bouncer, as it's what actually loads the data from the site configuration yaml file

modules/govuk_jenkins/templates/jobs/transition_import_dns.yaml.erb
kevindew: This doesn't update the CDN - it's the https://github.com/alphagov/govuk-puppet/pull/9319/files#diff-b19845cf2511fa96f3d5d029b7f20742R32 task.

The previous task: https://github.com/alphagov/govuk-puppet/pull/9319/files#diff-e82a80955f654ee3bc23fb314b8320e2L24 did [3 things](https://github.com/alphagov/transition/blob/3f0feab8599a2a2e6dfb7004e8506aaad7655e31/lib/tasks/import/all.rake#L7-L11) which included the CDN update.

This one is just moved since there doesn't seem to be a particular reason to run it in the middle of night so why bother?

modules/govuk_jenkins/templates/jobs/transition_load_site_config.yaml.erb
barrucadu: I see what's going on

kevindew: Yup 👍 - previously the other task did this as well.

modules/govuk_jenkins/templates/jobs/transition_import_dns.yaml.erb
barrucadu: Right, I've just worked my way through the jobs.  The old job triggered the bouncer CDN job:

```yaml
    publishers:
        - trigger-parameterized-builds:
            - project: Bouncer_CDN
              condition: 'SUCCESS'
              trigger-with-no-params: true
```

I think this job should trigger it too.

kevindew: @barrucadu I don't think the DNS job needs to trigger the CDN job as this isn't included in any data sent to Fastly. 

I'm not really sure what the DNS information is used for - do we proxy some sites? But it is displayed on admin pages: https://transition.integration.publishing.service.gov.uk/sites/defra

barrucadu: Maybe the CDN job doesn't need to be triggered by this job, but what stands out to me is that the CDN job was triggered by the old job, and now none of the new jobs trigger it.  I'm not really sure what the DNS information is used for either.

kevindew: The CDN job needs to run when the `import:all:orgs_sites_hosts` job runs. This is still the case for the job that runs it: https://github.com/alphagov/govuk-puppet/blob/9298766ca7109f6ad81f1a5dec51afc7d30a874a/modules/govuk_jenkins/templates/jobs/transition_load_site_config.yaml.erb#L40-L44

The job that has been split into two used to do three things: import DNS, import hits and import orgs sites hosts. The last of these has been dropped and that was the only one which needed the CDN deploy.

PR #9324, opened 2019-06-26
Add Data sync pull for elasticsearch 6 to Training Environment

camdesgov:
I believe this is needed as part of the Training Environment setup .


schmie:
I believe ES6 is experimental for the time being and we do not need to sync it to training at this point in time

camdesgov:
Ah thanks for the info, I'll delete the branch

PR #9329, opened 2019-06-28
Change the time the content_data_api_production database is pushed

cbaines:
In the Integration environment, back to after the data has been pulled
from production. Otherwise, the data in the
govuk-integration-database-backups bucket will always be a bit out of
date.

### DIFF
--- a/hieradata_aws/class/integration/content_data_api_db_admin.yaml
+++ b/hieradata_aws/class/integration/content_data_api_db_admin.yaml
@@ -12,9 +12,9 @@ govuk_env_sync::tasks:
     temppath: "/tmp/content_data_api_production"
     url: "govuk-production-database-backups"
     path: "content-data-api-postgresql"
-  "push_content_data_api_integration_daily":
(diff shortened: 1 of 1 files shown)


hieradata_aws/class/integration/content_data_api_db_admin.yaml
```diff
@@ -12,9 +12,9 @@ govuk_env_sync::tasks:
     temppath: "/tmp/content_data_api_production"
     url: "govuk-production-database-backups"
     path: "content-data-api-postgresql"
-  "push_content_data_api_integration_daily":
+  "push_content_data_api_production_daily":
```
theseanything: Do we want to change the title back?

cbaines: I've been thinking the titles were action + database name + cadence, however, I think some are now action + short database name + environment + cadence. I'm not sure which direction we should go in, but I haven't seen a reason to change it here yet. So yes, changing the title back was my intention.

PR #9336, opened 2019-07-02
Remove transition/bouncer staging Jenkins jobs

kevindew:
Trello: https://trello.com/c/y8KULrlo/980-dont-run-transition-cdn-deploys-in-the-middle-of-the-night

Following on from https://github.com/alphagov/govuk-puppet/pull/9319 this removes staging Jenkins jobs for transition and bouncer. I was under the impression these used to run in staging fine and were only removed as a result of a mistake in https://github.com/alphagov/govuk-puppet/commit/8780523b0123c0ceb59a502300158f2b1827b088 - turns out this was wrong and they were only added the day before in https://github.com/alphagov/govuk-puppet/commit/e985e342004881d0907d832dd7d5456465a8ba50. 

Although it'd be nicer to have these jobs available outside prod the jobs don't seem configured fully for staging so I'm not sure of adverse consequences they may have. From an initial look the Bouncer CDN job was at risk of hijacking domains so that seemed an alarm bell.

This PR also cleans up some removed jobs.


hieradata_aws/class/staging/jenkins.yaml
```diff
@@ -24,14 +25,7 @@ govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::jobs::search_api_reindex_with_new_schema
   - govuk_jenkins::jobs::smokey
   - govuk_jenkins::jobs::smokey_deploy
-  - govuk_jenkins::jobs::data_sync
-  - govuk_jenkins::jobs::transition_import_dns
```
barrucadu: Do these need to be ensured absent first?

tijmenb: They can be deleted manually from the Jenkins UI.

kevindew: > Do these need to be ensured absent first?

They get deleted automatically by not being in the jenkins job list. I was likely more thorough than I needed to be in https://github.com/alphagov/govuk-puppet/pull/9319

PR #9337, opened 2019-07-03
Enable X-Frame-Options everywhere

rubenarakelyan:
This commit makes the `X-Frame-Options` HTTP header be set everywhere by default rather than just for publishing apps. This will prevent GOV.UK pages being framed by other sites and will reduce the possibility of clickjacking attempts in browsers that respect the header.


rubenarakelyan:
This would stop the side-by-side browser from working, so closing in favour of CSP where we have more granularity.

PR #9338, opened 2019-07-03
add puppet node & class for Gatling

fredericfran-gds:
Add the necessary code to deploy gatling on AWS

### DIFF
--- a/modules/govuk/manifests/node/s_gatling.pp
+++ b/modules/govuk/manifests/node/s_gatling.pp
@@ -1,0 +1,77 @@
+# == Class: govuk::node::s_gatling
+#
+# This machine class is used to run Gatling, a load testing tool.
+#
+# === Parameters
+#
+# [*root_dir*]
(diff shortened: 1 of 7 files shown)


modules/govuk/manifests/node/s_gatling.pp
```diff
@@ -0,0 +1,73 @@
...
+
+  # install the JRE and JDK on the Gatling instance
+  include ::govuk_java::openjdk8::jre
+  include ::govuk_java::openjdk8::jdk
+  class { 'govuk_java::set_defaults':
+    jdk => 'openjdk8',
+    jre => 'openjdk8',
+  }
+
+  exec { 'get-gatling':
```
deanwilson: Can we package this instead using fpm? It'd be better to have a package pulled from our repo than rely on third party sites on each puppet run.

fredericfran-gds: yeah, I can

PR #9340, opened 2019-07-04
Update s3md to newer release

brucebolt:
s3cmd is used by push_attachments_to_s3.sh on asset-slave-2.

Timeout errors are occuring, which stop the push running to completion.  Version 2.0.2 of s3cmd has fixed a bug causing timeout errors.

Fixing s3cmd to this specfic release so we are not affected by future breaking changes.

Trello card: https://trello.com/c/i4isNg22/604-asset-slave-2backend-push-attachments-to-s3


benthorner:
🍾

PR #9342, opened 2019-07-05
Specify memory warning thresholds for govuk_crawler_worker

brucebolt:
The memory usage in production for `govuk_crawler_worker` is regularly above the default critical threshold for memory usage, whilst there is plenty of free memory available on the `mirrorer` machine.

This PR increases the thresholds for alerting on this application.

Recent memory usage:
Screen Shot 2019-07-05 at 10 33 07

### DIFF
--- a/modules/govuk/manifests/apps/govuk_crawler_worker.pp
+++ b/modules/govuk/manifests/apps/govuk_crawler_worker.pp
@@ -41,6 +41,12 @@
 #   Sets the header "Rate-Limit-Token" which ensures that the crawler is
 #   whitelisted by the rate limiting function (receiving 429 status)
 #
+# [*nagios_memory_warning*]
+#   Memory use at which Nagios should generate a warning.
+#
+# [*nagios_memory_critical*]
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/govuk_crawler_worker.pp
```diff
@@ -103,11 +111,13 @@
...
-      port               => $port,
-      command            => './govuk_crawler_worker -json',
-      health_check_path  => '/healthcheck',
+      app_type               => 'bare',
+      log_format_is_json     => true,
+      port                   => $port,
+      command                => './govuk_crawler_worker -json',
+      health_check_path      => '/healthcheck',
+      nagios_memory_warning  => $nagios_memory_warning,
+      nagios_memory_critical => $nagios_memory_critical,
```
cbaines: Unless you actually need different values in different environments, I think it's OK to put the actual values here, rather than in the hieradata.

There's nothing wrong with having them in the hieradata though.

PR #9345, opened 2019-07-08
Stop tests running concurrently on the same Jenkins build node

cbaines:
govuk-puppet builds seem to hang if the tests run at the same time on
a build node. To avoid this, use a lock to force the tests to run
sequentially.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,7 +35,9 @@ node {
     }
 
     stage("Spec tests") {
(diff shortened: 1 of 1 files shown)


thomasleese:
Do we have an example of this happening?

cbaines:
> Do we have an example of this happening?

I've seen it happen quite a lot. I've just kicked off a few builds of an old branch, so I could take a screenshot.

puppet

PR #9354, opened 2019-07-09
Remove Kibana from `backend` machines in all environments.

peteglondon:
# What

Remove `Kibana` from `backend` machines in all environments.

# Why

We now use `logit` to provide this functionality so hosting our own is redundant.

Trello: https://trello.com/c/CDhl40nQ/1107-remove-kibana-as-a-target-app-from-jenkins-deployapp-job

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -170,8 +170,6 @@ deployable_applications: &deployable_applications
   hmrc-manuals-api: {}
   imminence: {}
   info-frontend: {}
-  kibana:
-    repository: 'kibana-gds'
   licencefinder:
     repository: 'licence-finder'
   link-checker-api: {}


development-vm/Procfile
```diff
@@ -142,7 +142,6 @@ draft-router-api:            govuk_setenv draft-router-api            ./run_in.s
 # firewall block rule on port 3170 (for testing connect timeouts)
 # canary-frontend has reserved port 3200
 # canary-backend has reserved port 3201
-# kibana has reserved port 3202
```
thomasleese: I think we leave old used ports to avoid them being reused in the future just in case. For example, above I can see we've still got `multipage-frontend used port 3118` and `share-sale-publisher used port 3119`.

peteglondon: Updated

tijmenb:
As @cbaines says, this might be removing too much. http://kibana.publishing.service.gov.uk should still be redirecting to production logit. I think there's still stuff to remove though, I think it's not a "deployable app"?

cbaines:
Updating the commit message would be good, but that change look's right now :)

PR #9360, opened 2019-07-11
Handle deploying govuk-content-schemas as part of deploy all apps

kevindew:
GOV.UK Content Schemas is a somewhat special example where we have code
that is deployed to GOV.UK infrastructure but it isn't an app that runs
so it lacks a puppet manifest. This code is deployed by the deploy app
Jenkins job and is something that needs to occur when a node is built
for the first time.

The code that defines which apps run on a node class is the same as that
which is used to define which projects are deployed. This makes it
problematic to define that content schemas should be deployed to
certain nodes, without having it also treated as an app that should
run on the node.

To work around this I set up the concept of non_apps that is part of the
definition for projects on a particular node. I'm not particular pleased
with myself for that name and anticipate it may raise a WTF, I like to
think though that it's the concept of the project that is a bit strange
and that the syntax for defining it consistent with the oddness that is
already present.

This was added after the node instances were destroyed in integration
and when re-built the Publishing API was no longer operating correctly.

/cc @jstandring-gds this is to fix the issue we hit yesterday where the box didn't start correctly.

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -45,6 +45,8 @@ node_class: &node_class
       - support_api_csv_env_sync
       - transition
       - travel-advice-publisher
+    non_apps:
+      - govuk-content-schemas
   bouncer:
     apps:
       - bouncer
@@ -107,6 +109,8 @@ node_class: &node_class
   publishing_api:
     apps:
       - publishing-api
+    non_apps:
+      - govuk-content-schemas
   router_backend:
     apps:
       - router-api
@@ -129,7 +133,6 @@ govuk_java::repo::apt_mirror_hostname: "%{hiera('apt_mirror_hostname')}"
 
 govuk_jenkins::deploy_all_apps::apps_on_nodes:
   <<: *node_class
-
 
 # If the repository name is the same as the application name
 # we don't need to explicitly declare the repository, but we
--- a/hieradata_aws/production.yaml
+++ b/hieradata_aws/production.yaml
@@ -30,6 +30,8 @@ node_class: &node_class
       - support
       - support-api
       - support_api_csv_env_sync
+    non_apps:
+      - govuk-content-schemas
   bouncer:
     apps:
       - bouncer
--- a/hieradata_aws/staging.yaml
+++ b/hieradata_aws/staging.yaml
@@ -30,6 +30,8 @@ node_class: &node_class
       - support
       - support-api
       - support_api_csv_env_sync
+    non_apps:
+      - govuk-content-schemas
   bouncer:
     apps:
       - bouncer
--- a/modules/govuk_jenkins/manifests/deploy_all_apps.pp
+++ b/modules/govuk_jenkins/manifests/deploy_all_apps.pp
@@ -14,8 +14,8 @@
 # [*user*]
 #   Which user Jenkins runs as.
 #
-# [*apps_on_nodes*]
-#   A hash containing the node classes, and which apps run on them.
+# [*deploys_on_nodes*]
+#   A hash containing the node classes, and which deploys to run on them.
 #
 # [*deploy_environment*]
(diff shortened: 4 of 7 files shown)


cbaines:
> To work around this I set up the concept of non_apps that is part of the
> definition for projects on a particular node. I'm not particular pleased
> with myself for that name and anticipate it may raise a WTF, I like to
> think though that it's the concept of the project that is a bit strange
> and that the syntax for defining it consistent with the oddness that is
> already present.

How about creating a manifest for it? It could just have a commit in it explaining that there isn't a corresponding service.

kevindew:
> > To work around this I set up the concept of non_apps that is part of the
> > definition for projects on a particular node. I'm not particular pleased
> > with myself for that name and anticipate it may raise a WTF, I like to
> > think though that it's the concept of the project that is a bit strange
> > and that the syntax for defining it consistent with the oddness that is
> > already present.
> 
> How about creating a manifest for it? It could just have a commit in it explaining that there isn't a corresponding service.

Thanks for the review Chris. I assume you mean comment rather than commit?

Do we have any prior examples of an empty manifest? I feel like that is quite a WTF to introduce so wouldn't want to start that technique in our puppet but if it's something we've already employed for similar situations I wouldn't be opposed to it. Otherwise I feel like the non_app way would be cleaner.

cbaines:
> > How about creating a manifest for it? It could just have a commit in it explaining that there isn't a corresponding service.
> 
> Thanks for the review Chris. I assume you mean comment rather than commit?

Yep, comment indeed.

> Do we have any prior examples of an empty manifest? I feel like that is quite a WTF to introduce so wouldn't want to start that technique in our puppet but if it's something we've already employed for similar situations I wouldn't be opposed to it. Otherwise I feel like the non_app way would be cleaner.

I think https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/support_api_csv_env_sync.pp is quite close to this, it's a manifest that doesn't represent a service. Although, saying that, I haven't looked very closely.

Thinking about it, I think there are useful things for govuk-puppet to do concerning the govuk-content-schemas. Here's a manifest that sets up an Icinga check to check that the govuk-content-schemas exist [1].

1: https://github.com/alphagov/govuk-puppet/pull/9375/files

kevindew:
> > > How about creating a manifest for it? It could just have a commit in it explaining that there isn't a corresponding service.
> > 
> > 
> > Thanks for the review Chris. I assume you mean comment rather than commit?
> 
> Yep, comment indeed.
> 
> > Do we have any prior examples of an empty manifest? I feel like that is quite a WTF to introduce so wouldn't want to start that technique in our puppet but if it's something we've already employed for similar situations I wouldn't be opposed to it. Otherwise I feel like the non_app way would be cleaner.
> 
> I think https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/support_api_csv_env_sync.pp is quite close to this, it's a manifest that doesn't represent a service. Although, saying that, I haven't looked very closely.
> 
> Thinking about it, I think there are useful things for govuk-puppet to do concerning the govuk-content-schemas. Here's a manifest that sets up an Icinga check to check that the govuk-content-schemas exist [1].
> 
> 1: https://github.com/alphagov/govuk-puppet/pull/9375/files

Yup sure that's a good idea. It looks like your draft PR has everything needed to take over this one. Shall I just close this one in favour of that? and any help you need getting that through?

cbaines:
> Yup sure that's a good idea. It looks like your draft PR has everything needed to take over this one. Shall I just close this one in favour of that? and any help you need getting that through?

Great, I've made it a proper Pull Request now, and yeah, if you're happy to go with that approach, then lets look at that. If you want to take a look at reviewing it, you're more than welcome :)

kevindew:
closing in favour of https://github.com/alphagov/govuk-puppet/pull/9375

PR #9373, opened 2019-07-12
Add another pv to the postgres-primary backup lv

schmie:
- Daily backups were consistently failing due to the backups reaching
the maximum size the filesystem allowed.
- Added 128G partition and manually extended the lvm managed backup
drive.
- Missed to add the change here in the first instance, which causes
Puppet to fail to run on the Postgresql-primary-1

@schmie


thomasleese:
I'm going to merge this in to get rid of a Puppet failure alert.

PR #9374, opened 2019-07-15
Improve Bouncer CDN alerting

thomasleese:
This PR makes two changes to the alerting for the Bouncer CDN job:

- Removes the freshness threshold because the job doesn't run regularly.
- Use the AWS app domain to link to the job because it's not configured in Carrenza.

[Trello Card](https://trello.com/c/19Gc1KkK/1109-fix-auth-issue-for-cdn-deploy-bouncer-configs-job-in-prod)

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
+++ b/modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
@@ -32,7 +32,6 @@ class govuk_jenkins::jobs::bouncer_cdn (
   @@icinga::passive_check { "${check_name}_${::hostname}":
     service_description => $service_description,
     host_name           => $::fqdn,
-    freshness_threshold => 104400,
     action_url          => $job_url,
   }
 }
--- a/modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
+++ b/modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
@@ -18,16 +18,18 @@ class govuk_jenkins::jobs::bouncer_cdn (
   $service_id = undef,
   $app_domain = hiera('app_domain'),
 ) {
-
-  $check_name = 'bouncer-cdn-configuration'
   $service_description = 'Configure Bouncer CDN service with transitioning sites'
-  $job_url = "https://deploy.${app_domain}/job/Bouncer_CDN/"
 
   file { '/etc/jenkins_jobs/jobs/bouncer_cdn.yaml':
     ensure  => present,
     content => template('govuk_jenkins/jobs/bouncer_cdn.yaml.erb'),
     notify  => Exec['jenkins_jobs_update'],
   }
+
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
```diff
@@ -17,22 +17,22 @@
...
-  $job_url = "https://deploy.${app_domain}/job/Bouncer_CDN/"
 
   file { '/etc/jenkins_jobs/jobs/bouncer_cdn.yaml':
     ensure  => present,
     content => template('govuk_jenkins/jobs/bouncer_cdn.yaml.erb'),
     notify  => Exec['jenkins_jobs_update'],
   }
 
+  $check_name = 'bouncer-cdn-configuration'
+  $job_url = "https://deploy.${app_domain_internal}/job/Bouncer_CDN/"
```
cbaines: I don't think this'll work, as the internal domain is for internal communication between apps, and isn't the same as the `production.govuk.digital` domain that you want here for Jenkins.

https://github.com/alphagov/govuk-puppet/blob/master/hieradata_aws/production.yaml#L3

thomasleese: Oh yeah... I'm sure we've solved this before somewhere... let me try and find it.

thomasleese: I've pushed up an amended commit based on https://github.com/alphagov/govuk-puppet/pull/8935.

modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
```diff
@@ -18,17 +18,19 @@
...
 
   file { '/etc/jenkins_jobs/jobs/bouncer_cdn.yaml':
     ensure  => present,
     content => template('govuk_jenkins/jobs/bouncer_cdn.yaml.erb'),
     notify  => Exec['jenkins_jobs_update'],
   }
 
+  $check_name = 'bouncer-cdn-configuration'
+
+  # temporary solution while we have two Icinga instances
```
cbaines: I'm not sure setting the `job_url` per environment is a temporary thing, or how it relates to Icinga?

modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
cbaines: :+1:

modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
thomasleese: It's what controls the cog icon in Icinga, and ideally we could switch back to `app_domain` when we only have one Icinga instance AWS... but I can look at improving the comment.

PR #9375, opened 2019-07-15
Add govuk-content-schemas manifest

cbaines:
This adds an Icinga check to flag when the govuk-content-schemas
doesn't exist on a machine, as well as feeding in to the Deploy Node
Apps Jenkins job, so that the govuk-content-schemas is deployed when
nodes are rebuilt.

### DIFF
--- a/modules/icinga/files/usr/lib/nagios/plugins/check_directory_exists
+++ b/modules/icinga/files/usr/lib/nagios/plugins/check_directory_exists
@@ -1,0 +1,20 @@
+#!/bin/bash
+#
+# Desc : Plugin to verify a directory dxists
+
+PROGNAME="$(basename "$0")"
+
+if [ "$1" = "" ]
+then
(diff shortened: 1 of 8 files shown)


modules/icinga/files/usr/lib/nagios/plugins/check_directory_exists
```diff
@@ -0,0 +1,20 @@
...
+
+PROGNAME="$(basename "$0")"
+
+if [ "$1" = "" ]
+then
+  echo -e " Use : $PROGNAME <directory_name> -- Ex : $PROGNAME /etc \n "
+  exit 3
+fi
+
+if [ -d "$1" ]
```
kevindew: Just to check, this is fine that the directory is a symlink? it worked on a quick test of my own so seems good.

cbaines: Yeah, I remember checking this specific case, and yep, it works if the symlink is working, and fails (as intended) if the symlink is broken.

PR #9384, opened 2019-07-17
Update smokey checks per environment

peteglondon:
Add the remaining apps to AWS Icinga Smokey checks.

There are likely to be more cards generated by any failures of the Smokey tests.

Trello: https://trello.com/c/yVDGAwCc/1119-add-icinga-alerts-for-all-smokey-failures

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1542,6 +1542,8 @@ monitoring::checks::smokey::features:
     feature: router
   check_search:
     feature: search
+  check_service_manual:
+    feature: service_manual
   check_signon:
     feature: signon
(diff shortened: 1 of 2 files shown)


hieradata/common.yaml
```diff
@@ -1542,6 +1542,8 @@ monitoring::checks::smokey::features:
     feature: router
   check_search:
     feature: search
+  check_service_manual:
+    feature: service_manual
```
cbaines: This is the Carrenza file, whereas the commit message says AWS?

peteglondon: My bad - commit message fixed

PR #9388, opened 2019-07-22
Add alerting to transition import hits job

thomasleese:
This job has been failing since the 6th July and we didn't notice.

https://deploy.blue.production.govuk.digital/job/transition_import_hits/

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/transition_import_hits.pp
+++ b/modules/govuk_jenkins/manifests/jobs/transition_import_hits.pp
@@ -9,10 +9,25 @@
 #
 class govuk_jenkins::jobs::transition_import_hits(
   $s3_bucket = '',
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/manifests/jobs/transition_import_hits.pp
```diff
@@ -9,10 +9,25 @@
 #
 class govuk_jenkins::jobs::transition_import_hits(
   $s3_bucket = '',
+  $app_domain = hiera('app_domain'),
```
cbaines: This looks like it's unused.

thomasleese: Yeah... you're correct. I put it in so the comment about the `job_url` made more sense and for consistency with the other jobs. Do you think it's better to just not add it?

PR #9393, opened 2019-07-23
Remove search-api index size checks

benthorner:
Previously we had somewhat flakey checks about an increase/decrease in
our search index sizes. In the first place, it's unclear why we would
want an alert for *increases* in search index size, as surely this is a
'good' thing. Secondly, there seems to be no documentation as to why
these alerts exist, as the original commit is simply 'WIP'.

Commit ref: 7df31ce08c0d802b7e376a9f5684f4e308b6f6c9

### DIFF
--- a/modules/monitoring/manifests/checks.pp
+++ b/modules/monitoring/manifests/checks.pp
@@ -49,7 +49,6 @@ class monitoring::checks (
 
   include icinga::plugin::check_http_timeout_noncrit
 
-  # START whitehall
   # Used in template and icinga::check.
   $whitehall_hostname    = "whitehall-admin.${app_domain}"
   $whitehall_overdue_url = '/healthcheck/overdue'
@@ -70,7 +69,6 @@ class monitoring::checks (
     attempts_before_hard_state => 2,
     retry_interval             => 10,
   }
-  # END whitehall
 
   $warning_time = 5
   $critical_time = 10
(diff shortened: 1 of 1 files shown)


benthorner:
@tijmenb @rubenarakelyan it looks like I may have pushed this through a bit too fast. @timblair helped me find the original PR for this and that it was associated with an incident.

https://github.com/alphagov/govuk-puppet/pull/6789
https://docs.google.com/document/d/1cqC2Ow1ByAeOVhaYBWNP5Z3BJ9uTjCsgZNk5bQxd4Q4/edit

The first thing to note is that this incident does not motivate checking for an *increase* in index size, which is where I started with this PR. So one option would be to make it only check for a *decrease*.

I think we could still lose this alert without leaving a hole: the incident documents that the significant decrease was also associated with high 500s, which we have critical alerts for. That may depend on whether we want to have lots of alerts about the same issue, or just a few. In my experience, this alert has come up several times while I've been on 2ndline and has always been a false alarm.

It's interesting to note that we don't have an equivalent alert for the content store. It might be worth clarifying why we think this could happen to the Search API indexes and not other databases.

cbaines:
I'm in favour of keeping these checks. I'm hoping they're more reliable since the metric naming issues were fixed [1].

I don't think the current state in the Staging environment is flaky, as it's been failing consistently due to lack of data for a week.

1: https://github.com/alphagov/govuk-puppet/pull/9332

PR #9399, opened 2019-07-24
Remove users

JonathanHallam:
Both have left GDS and no longer require access


leenagupte:
Aren't we supposed to remove the manifest files and remove them from hieradata?

barrucadu:
The users on the machines need to be deleted as well, which won't automatically happen without first ensuring them absent

PR #9402, opened 2019-07-24
Remove access and references for a member of staff who has left.

sengi:
This removes the user manifest and ssh access to Integration and CI.

### DIFF
--- a/modules/users/manifests/petergoddard.pp
+++ b/modules/users/manifests/petergoddard.pp
@@ -1,9 +1,0 @@
-# Creates the petergoddard user
-class users::petergoddard {
-  govuk_user { 'petergoddard':
-    ensure   => absent,
-    fullname => 'Peter Goddard',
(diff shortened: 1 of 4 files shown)


modules/users/manifests/petergoddard.pp
```diff
@@ -1,9 +0,0 @@
-# Creates the petergoddard user
```
barrucadu: By deleting this file he'll also need to be removed from staging and production in govuk-secrets, or the next puppet deploy will cause issues

sengi: Aha, thanks! I'll hold off on merging this until that's done then.

sengi: Confirmed removed from govuk-secrets: https://github.com/alphagov/govuk-secrets/pull/755

sengi:
Superseded by https://github.com/alphagov/govuk-puppet/pull/9403

PR #9410, opened 2019-07-26
Do not allow deploying CDN dictionaries to 'mirror' in integration

barrucadu:
Integration doesn't have a mirror target, so this will fail.


barrucadu:
I'll add a short comment above the hieradata entry

PR #9416, opened 2019-07-29
Remove pentester SSH accounts

erino:
This reverts PRs
- https://github.com/alphagov/govuk-puppet/pull/9368
- https://github.com/alphagov/govuk-puppet/pull/9367
- https://github.com/alphagov/govuk-puppet/pull/9352

### DIFF
--- a/modules/users/manifests/pentestera.pp
+++ b/modules/users/manifests/pentestera.pp
@@ -1,8 +1,8 @@
 # Creates the pentester user
 class users::pentestera {
   govuk_user { 'pentestera':
+    ensure   => 'absent',
     fullname => 'Pentester A',
     email    => 'pentester+a@digital.cabinet-office.gov.uk',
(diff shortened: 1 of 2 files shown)


modules/users/manifests/pentestera.pp
```diff
@@ -1,8 +0,0 @@
-# Creates the pentester user
-class users::pentestera {
-  govuk_user { 'pentestera':
-    fullname => 'Pentester A',
```
cbaines: If you want Puppet to actually remove the user more thoroughly, you'll need to make the user `absent` first, before removing the manifest.

See https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html

erino: Ah, yes. I forgot how Puppet works...

PR #9421, opened 2019-07-30
Avoid false 429 rate alerts when a new cache instance comes up.

sengi:
When a new cache instance appears, the '429 rate' alert fires with UNKNOWN status.
This happens because the statsd metric doesn't get created until the first 429 response
happens (which doesn't and shouldn't normally happen), which leads to the check_graphite
command failing because it can't find the timeseries, leading to the UNKNOWN status and
alerts appearing on the Icinga dashboard.

`check_graphite` has an `--ignore_missing` option which does exactly what we want in
this particular case. That is, we want the check to succeed if there is no metric for
429 rate on a given router instance.

The `--ignore-missing` option was added in a more recent version of `check_graphite`
than the one presently installed, so we also update the version here.

### DIFF
--- a/modules/icinga/manifests/package.pp
+++ b/modules/icinga/manifests/package.pp
@@ -38,7 +38,7 @@ class icinga::package {
   Package['nagios-nrpe-plugin'] -> Class['nagios::remove']
 
   package { 'check_graphite':
-    ensure   => '0.2.2',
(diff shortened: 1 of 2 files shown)


sengi:
Follow-up PR to update the playbook for this alert: https://github.com/alphagov/govuk-developer-docs/pull/1898

PR #9424, opened 2019-07-31
Update the Jenkins Job Builder

cbaines:
Switch to a newer version, and also switch to using Python 3. I'm
wondering if this will avoid issues when working with unicode
characters in branch names.


cbaines:
I've tested this, and it looks to work.

PR #9433, opened 2019-08-05
Allow monitoring ip address to access Whitehall-admin healthcheck

edwardkerry:
We would like to add Whitehall healthcheck reporting to a Grafana dashbaord.

Currently we block all access to Whitehall-admins `/healthcheck` path. This PR will grant access to the monitoring machine.

This will only work whilst still hosting on Carrenza; after moving to AWS we will not be able to rely on a static IP address and will need to consider a longer term solution.

### DIFF
--- a/modules/govuk/manifests/apps/whitehall.pp
+++ b/modules/govuk/manifests/apps/whitehall.pp
@@ -303,6 +303,7 @@ class govuk::apps::whitehall(
       location ${health_check_path} {
         allow 10.2.0.20;
         deny all;
+        try_files \$uri @app;
       }
 
       # Don't block access to the overdue healthcheck page.  Icinga needs to be
--- a/modules/govuk/manifests/apps/whitehall.pp
+++ b/modules/govuk/manifests/apps/whitehall.pp
@@ -302,6 +302,7 @@ class govuk::apps::whitehall(
       # Updated version
       location ${health_check_path} {
         allow 10.2.0.20;
+        allow 172.17.0.1;
         deny all;
         try_files \$uri @app;
       }
--- a/modules/govuk/manifests/apps/whitehall.pp
+++ b/modules/govuk/manifests/apps/whitehall.pp
@@ -126,6 +126,10 @@
 # [*override_search_location*]
 #   Alternative hostname to use for Plek("search") and Plek("rummager")
 #
+#[*monitoring_ip]
+#   Monitoring machine IP to allow access to healthcheck endpoint
+#   Default: 10.3.0.20
+#
 class govuk::apps::whitehall(
   $admin_db_name = undef,
   $admin_db_hostname = undef,
@@ -165,6 +169,7 @@ class govuk::apps::whitehall(
   $cpu_critical = 200,
   $rummager_bearer_token = undef,
   $override_search_location = undef,
+  $monitoring_ip = '10.3.0.20'
 ) {
 
   $app_name = 'whitehall'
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/apps/whitehall.pp
```diff
@@ -289,9 +289,21 @@
         try_files $uri @app;
       }
 
+      # Restrict access to the healthcheck page. Grafana needs to be
+      # able to access this from the monitoring machine. All other requests are blocked.
+      # This is only useful for Carrenza, as AWS will not provide static IPs.
+      if $::aws_migration {
```
thomasleese: I don't think this will work here as it's in a string designed to go directly in the Nginx configuration. I think you'll need to do something like:

```puppet
hidden_paths          => $::aws_migration ? [$health_check_path] : [],
```

in the `govuk::app::nginx_vhost` section above, and then in here you can have the `location` Nginx block.

modules/govuk/manifests/apps/whitehall.pp
```diff
@@ -292,13 +293,9 @@
...
       # This is only useful for Carrenza, as AWS will not provide static IPs.
-      if $::aws_migration {
-        hidden_paths => [$health_check_path]
-      } else {
-        location $health_check_path {
-          allow $monitoring_ip
-          deny all;
-        }
+      location $health_check_path {
+        allow $monitoring_ip;
```
thomasleese: To get this to interpolate correctly, I think [the string might have to be double quoted](https://puppet.com/docs/puppet/5.2/lang_variables.html#interpolation). And the current `$abc` variables will have to be escaped, something like `\$abc`.

hieradata_aws/production.yaml
```diff
@@ -152,6 +152,7 @@ govuk::apps::support_api::zendesk_anonymous_ticket_email: 'zd-api-public@digital
 govuk::apps::support_api::pp_data_url: 'https://www.performance.service.gov.uk'
 govuk::apps::support_api::aws_s3_bucket_name: 'govuk-production-support-api-csvs'
 govuk::apps::travel_advice_publisher::enable_email_alerts: true
+govuk::apps::whitehall::monitoring_ip: "%{hiera('hosts::production::management::hosts.monitoring-1.ip')}"
```
thomasleese: In AWS we won't have this available, I think it's better to leave it out to be explicit. If the Rake task which checks consistency between AWS and Carrenza is failing, you can add this option as an exception.

hieradata/production.yaml
```diff
@@ -91,6 +91,7 @@ govuk::apps::support_api::aws_s3_bucket_name: 'govuk-production-support-api-csvs
 govuk::apps::travel_advice_publisher::enable_email_alerts: true
 govuk::apps::whitehall::cpu_warning: 400
 govuk::apps::whitehall::cpu_critical: 600
+govuk::apps::whitehall::monitoring_ip: "%{hiera('hosts::production::management::hosts.monitoring-1.ip')}"
```
thomasleese: I'm unsure whether `.monitoring-1.ip` will work within the hiera function, but we can try that on integration.

thomasleese: Also, do we need this line for staging as well? (perhaps it's easier to just put it in `hieradata/common.yaml`)

edwardkerry:
Superseded by https://github.com/alphagov/govuk-puppet/pull/9435

PR #9434, opened 2019-08-05
Fix links to search-api-analytics and content-data jobs

vanitabarrett:
The warning in Icinga AWS Production / AWS Staging was linking to Carrenza Jenkins and returning a 404.

This commit changes the URL to be hard coded to the AWS URL as there
currently isn't an easy way of constructing that URL programatically.

I've based this on what was done for https://github.com/alphagov/govuk-puppet/pull/8935

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/search_api_fetch_analytics_data.pp
+++ b/modules/govuk_jenkins/manifests/jobs/search_api_fetch_analytics_data.pp
@@ -8,11 +8,9 @@ class govuk_jenkins::jobs::search_api_fetch_analytics_data (
   $skip_page_traffic_load = false,
   $cron_schedule = '5 4 * * *',
 ) {
-
   $job_name = 'search-api-fetch-analytics-data'
   $check_name = 'search-api-fetch-analytics-data'
   $service_description = 'Fetch analytics data for Search API'
-  $job_url = "https://deploy.${app_domain}/job/${job_name}/"
   $target_application = 'search-api'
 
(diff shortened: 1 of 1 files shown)


vanitabarrett:
@thomasleese I've added another commit to fix the link for the Content Data API job - can you re-review please?

modules/govuk_jenkins/manifests/jobs/content_data_api.pp
```diff
@@ -16,7 +16,7 @@
 
   $check_name = 'etl-content-data-api'
   $service_description = 'Content Data API ETL [Extract - transform - load]'
-  $job_url = "https://deploy.${::aws_environment}.govuk.digital/job/content_data_api_import_etl_master_process/"
+  $job_url = "https://deploy.${app_domain}/job/content_data_api_import_etl_master_process/"
```
thomasleese: I'm not sure if this is right, because the job exists in AWS in production and staging (for example: https://deploy.blue.production.govuk.digital/job/content_data_api_import_etl_master_process/).

I see what you mean about the redirect not working, but I think you've uncovered a problem for all our jobs in integration! For example https://deploy.integration.govuk.digital/job/smokey/ also redirects incorrectly, as does https://deploy.integration.govuk.digital/job/search-api-fetch-analytics-data/.

Ideally we would find a way to avoid Jenkins doing the redirect... but I have no idea how we could do that.

vanitabarrett: Good point - I've removed this commit so it stays as is for now.

PR #9439, opened 2019-08-07
Add pycsw db to support ckan csw endpoint

kentsanggds:
## What

Add `pycsw` database to support ckan `csw` endpoint.

## Why

DGU used to have the `csw` endpoint available to `csw` clients, this was not implemented in the January migration.

### DIFF
--- a/modules/govuk/manifests/apps/ckan/db.pp
+++ b/modules/govuk/manifests/apps/ckan/db.pp
@@ -37,4 +37,15 @@ class govuk::apps::ckan::db (
     rds                     => $rds,
     extensions              => ['postgis'],
   }
+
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/ckan/db.pp
```diff
@@ -37,4 +37,15 @@
     rds                     => $rds,
     extensions              => ['postgis'],
   }
+
+  govuk_postgresql::db { 'pycsw':
```
thomasleese: I wonder if it's better to call this `ckan_pycsw_production` just so it's obvious from the outside what the database is for.

kentsanggds: Could do

PR #9452, opened 2019-08-12
Avoid referring to nonexistent Licensify TLS certs on AWS.

sengi:
This should fix Puppet failures on `licensing-frontend` VMs which were
caused by the nginx config for the `uploadlicence` app referring to a
TLS certificate file which isn't there.

In AWS, we terminate TLS at the load balancers (ALBs) rather than at
nginx on the app VMs. Most of the Puppet config already handles this
(via `nginx_enable_ssl` in Hiera) except for this piece of custom nginx
config in `licensing-frontend`.

### DIFF
--- a/modules/licensify/manifests/apps/licensify.pp
+++ b/modules/licensify/manifests/apps/licensify.pp
@@ -37,8 +37,16 @@ class licensify::apps::licensify (
   $vhost_escaped = regsubst($vhost_name, '\.', '_', 'G')
   $counter_basename = "${::fqdn_metrics}.nginx_logs.${vhost_escaped}"
 
+  if $::aws_migration {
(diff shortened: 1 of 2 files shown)


sengi:
Hang on, there's an unwanted change to Gemfile.lock that's crept in there somehow. Stand by :)

Gemfile.lock
```diff
@@ -120,6 +120,3 @@ DEPENDENCIES
   rspec-puppet
   sshkey (= 1.9.0)
   webmock (~> 1.20.0)
-
-RUBY VERSION
```
rtrinque: Why the modification of the Gemfile.lock?

sengi: Snap :)

Was installing a dev VM in the background and it turns out that the script for that make changes to the working copy. Nasty. Should be gone now.

PR #9454, opened 2019-08-12
Disable traffic replay to Integration

deborahchua:
...during data replication.

We are trying to reduce unnecessary Sentry errors, and one of those
ways is to disable traffic replay whilst the data sync is happening.

We manage traffic replay with our puppet class [`govuk_gor`](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_gor/manifests/init.pp), setting
the Puppet `data_sync_in_progress` fact to `true` ensures that Puppet
does not try to run it until the data sync has completed.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/copy_data_to_integration.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/copy_data_to_integration.yaml.erb
@@ -50,8 +50,19 @@
             sed -i.bak "s/PG_TR_SRC_ENV_SYNC_PW=PLACEHOLDER/PG_TR_SRC_ENV_SYNC_PW='<%= @pg_src_env_sync_pw %>'/" scripts/sync-transition-postgresql.sh
             sed -i "s/PG_TR_DST_ENV_SYNC_PW=PLACEHOLDER/PG_TR_DST_ENV_SYNC_PW='<%= @pg_tr_dst_env_sync_pw %>'/" scripts/sync-transition-postgresql.sh
 
+            echo "Disabling Gor traffic replay"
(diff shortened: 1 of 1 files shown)


cbaines:
I'm slightly surprised to see something that looks like support for this in govuk-puppet already... do you know why the existing configuration doesn't seem to be working?

modules/govuk_jenkins/templates/jobs/copy_data_to_integration.yaml.erb
```diff
@@ -50,8 +50,19 @@
             sed -i.bak "s/PG_TR_SRC_ENV_SYNC_PW=PLACEHOLDER/PG_TR_SRC_ENV_SYNC_PW='<%= @pg_src_env_sync_pw %>'/" scripts/sync-transition-postgresql.sh
             sed -i "s/PG_TR_DST_ENV_SYNC_PW=PLACEHOLDER/PG_TR_DST_ENV_SYNC_PW='<%= @pg_tr_dst_env_sync_pw %>'/" scripts/sync-transition-postgresql.sh
 
+            echo "Disabling Gor traffic replay"
+            for box in $(govuk_node_list -C 'govuk_gor'); do
```
cbaines: Assuming this relates to the job running from the Jenkins machine in the Carrenza side of Production, I don't think this code will do anything, as `govuk_node_list -C 'govuk_gor'` doesn't output anything.

deborahchua: Aah thank you @cbaines , I was looking in AWS not Carrenza 🤦‍♀

PR #9456, opened 2019-08-13
Add csw endpoint to ckan publisher stack

kentsanggds:
This is dependant on https://github.com/alphagov/ckanext-datagovuk/pull/187 being deployed

## What

Adds an additional `csw` endpoint for the ckan publishing stack in order to allow users to get `csw` data which datagovuk users have reported as being missing from the migration in January.

There are a number of changes that can be checked after puppet has completed the steps to set up `pycsw`:

- a cronjob will run to load data into the `pycsw` database, this will run at 6 each morning after the SOLR reindex
  - you could ssh onto the box and run `crontab -l` to check that the job is available and running at 6
  - if you want to also check if the job will run then extract the command out from `crontab` and run it on the terminal otherwise the numbers when checking the `csw` should be populated the next day and stay in sync with `ckan` database every day
  - the job takes about 4 hours to complete on 27,000 datasets, after the initial load each daily load should only take a few minutes.

- nginx update to allow `csw` to be accessible
  - after the cronjob has run some results should appear when going to https://ckan.integration.publishing.service.gov.uk/csw?service=CSW&version=2.0.2&request=GetRecords&typenames=csw:Record&elementsetname=brief

## Reference 

https://trello.com/c/RudwUlvN/1157-update-govuk-puppet-to-deploy-ckan-with-pycsw

### DIFF
--- a/modules/govuk/manifests/apps/ckan.pp
+++ b/modules/govuk/manifests/apps/ckan.pp
@@ -177,7 +177,7 @@ class govuk::apps::ckan (
       owner   => 'deploy',
       group   => 'deploy',
       notify  => Service['ckan'],
-    }
+    } ->
 
     file { "${ckan_home}/default":
       ensure => directory,
--- a/modules/govuk/manifests/apps/ckan.pp
+++ b/modules/govuk/manifests/apps/ckan.pp
@@ -8,6 +8,9 @@
 # [*port*]
 #   What port should the app run on?
 #
+# [*pycsw_port*]
+#   What port should the PyCSW companion app run on?
+#
 # [*db_hostname*]
 #   The postgres instance for CKAN to connect to
 #
@@ -39,6 +42,7 @@
 class govuk::apps::ckan (
   $enabled                        = false,
   $port                           = '3220',
+  $pycsw_port                     = '3221',
   $db_hostname                    = undef,
   $db_username                    = 'ckan',
   $db_password                    = 'foo',
@@ -64,6 +68,7 @@ class govuk::apps::ckan (
 ) {
   $ckan_home = '/var/ckan'
   $ckan_ini  = "${ckan_home}/ckan.ini"
+  $pycsw_config = "${ckan_home}/pycsw.cfg"
 
   $request_timeout = 60
 
@@ -120,7 +125,18 @@ class govuk::apps::ckan (
       process_regex  => '\/python \.\/venv\/bin\/paster --plugin=ckanext-harvest harvester gather\_consumer',
     }
 
-    include govuk::apps::ckan::cronjobs
+    govuk::procfile::worker { 'pycsw_web':
+      ensure         => present,
+      setenv_as      => 'ckan',
+      enable_service => running,
+      process_type   => 'pycsw_web',
+      process_regex  => 'pycsw_wsgi',
+    }
+
+    class { 'cronjobs':
+      ckan_port    => $port,
+      pycsw_config => $pycsw_config,
+    }
 
     Govuk::App::Envvar {
       app => 'ckan',
@@ -145,6 +161,12 @@ class govuk::apps::ckan (
       "${title}-BULK_WORKER_PROCESSES":
         varname => 'BULK_WORKER_PROCESSES',
         value   => $bulk_worker_processes;
+      "${title}-PYCSW_PORT":
+        varname => 'PYCSW_PORT',
+        value   => $pycsw_port;
+      "${title}-PYCSW_CONFIG":
+        varname => 'PYCSW_CONFIG',
+        value   => $pycsw_config;
     }
 
     govuk::app::nginx_vhost { 'ckan':
@@ -183,6 +205,21 @@ class govuk::apps::ckan (
       ensure => directory,
       owner  => 'deploy',
       group  => 'deploy',
+    } ->
+
+    file { $pycsw_config:
+      ensure  => file,
+      content => template('govuk/ckan/pycsw.cfg.erb'),
+      owner   => 'deploy',
+      group   => 'deploy',
+    }
+
+    $ckan_bin = '/var/apps/ckan/venv/bin'
+    $pycsw_tables_created = "${ckan_home}/pycsw_tables_created.tmp"
+
+    exec { 'setup_pycsw_tables':
+      command => "${ckan_bin}/paster --plugin=ckanext-spatial ckan-pycsw setup -p ${pycsw_config} && sudo touch ${pycsw_tables_created}",
+      creates => $pycsw_tables_created,
     }
   }
 }
--- a/modules/govuk/manifests/apps/ckan/cronjobs.pp
+++ b/modules/govuk/manifests/apps/ckan/cronjobs.pp
@@ -1,14 +1,21 @@
 # == Class: govuk::apps::ckan::cronjobs
 #
+# [*ckan_port*]
+#   The port CKAN is running on.
+#
+# [*pycsw_config*]
+#   The location of the PyCSW configuration file.
+#
 # [*enable_solr_reindex*]
 #   Enable automatic Solr reindexing, this is useful to run after the data
 #   sync.
 #   Default: false
 #
 class govuk::apps::ckan::cronjobs(
+  $ckan_port,
+  $pycsw_config,
   $enable_solr_reindex = false,
 ) {
-
   govuk::apps::ckan::paster_cronjob { 'harvester run':
     paster_command => 'harvester run',
     plugin         => 'ckanext-harvest',
@@ -34,4 +41,11 @@ class govuk::apps::ckan::cronjobs(
     minute         => '0',
   }
 
+  govuk::apps::ckan::paster_cronjob { 'pycsw load':
+    paster_command => "ckan-pycsw load -p ${pycsw_config} -u http://localhost:${ckan_port}",
+    plugin         => 'ckanext-spatial',
+    hour           => '6',
+    minute         => '0',
+    ckan_ini       => undef,
+  }
 }
--- a/modules/govuk/manifests/apps/ckan/cronjobs.pp
+++ b/modules/govuk/manifests/apps/ckan/cronjobs.pp
@@ -30,7 +30,7 @@ class govuk::apps::ckan::cronjobs(
     ensure         => $ensure_solr_reindex,
     paster_command => 'search-index rebuild -o',
     plugin         => 'ckan',
-    hour           => '7',
+    hour           => '4',
     minute         => '0',
   }
 
--- a/modules/govuk/manifests/apps/ckan/paster_cronjob.pp
+++ b/modules/govuk/manifests/apps/ckan/paster_cronjob.pp
@@ -26,13 +26,19 @@ define govuk::apps::ckan::paster_cronjob (
   $weekday = undef,
   $plugin = undef,
   $paster_command = undef,
+  $ckan_ini = '/var/ckan/ckan.ini',
 ) {
-
   validate_string($plugin, $paster_command)
+
+  if $ckan_ini {
+    $ckan_config_arg = "--config=${ckan_ini}"
+  } else {
+    $ckan_config_arg = ''
+  }
 
   cron { $title:
     ensure   => $ensure,
-    command  => "cd /var/apps/ckan; ./venv/bin/paster --plugin=${plugin} ${paster_command} --config=/var/ckan/ckan.ini",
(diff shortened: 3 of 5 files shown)


modules/govuk/manifests/apps/ckan.pp
```diff
@@ -173,5 +184,17 @@
       owner  => 'deploy',
       group  => 'deploy',
     }
+
+    exec { 'symlink_pycsw_config':
+      command => "test -e ${ckan_home}/pycsw.cfg || ln -s ${pycsw_src}/default.cfg ${ckan_home}/pycsw.cfg",
+    }
```
cbaines: How come the config is symlinked, rather than just put in `${ckan_home}/pycsw.cfg` in the first place?

modules/govuk/manifests/apps/ckan.pp
```diff
@@ -173,5 +184,17 @@
...
+
+    exec { 'setup_pycsw_tables':
+      command => "test -e ${ckan_home}/pycsw_tables_created.tmp || ${ckan_bin}/paster --plugin=ckanext-spatial ckan-pycsw setup -p ${ckan_home}/pycsw.cfg && sudo touch ${ckan_home}/pycsw_tables_created.tmp",
+    }
+
+    exec { 'startup_pycsw_server':
+      command => "ps -C \"python */pycsw/csw.wsgi\" >/dev/null && echo \"pycsw is running\" || eval \". ${ckan_bin}/activate && python ${pycsw_src}/csw.wsgi &\"",
+    }
```
cbaines: Have you considered using an Upstart service for this?

modules/govuk/manifests/apps/ckan/cronjobs.pp
```diff
@@ -30,8 +31,15 @@
     ensure         => $ensure_solr_reindex,
     paster_command => 'search-index rebuild -o',
     plugin         => 'ckan',
-    hour           => '7',
+    hour           => '5',
```
cbaines: How come the time is changing for this? With my limited understanding of CKAN, I don't know how this relates to adding the csv endpoint.

modules/govuk/manifests/apps/ckan/paster_cronjob.pp
```diff
@@ -26,13 +26,14 @@
   $weekday = undef,
   $plugin = undef,
   $paster_command = undef,
+  $ckan_config = '--config=/var/ckan/ckan.ini',
```
cbaines: This would probably be better described as `extra_args`, or something similar? Maybe it also needs to default to `''`.

cbaines: Or rather, the behavior may be clearer if the default is `''`.

modules/govuk/templates/ckan/pycsw.cfg.erb
```diff
@@ -0,0 +1,90 @@
...
+
+[server]
+home=/var/ckan
+url=<%= @ckan_site_url%>/csw
+mimetype=application/xml; charset=UTF-8
+encoding=UTF-8
+language=en-US
+maxrecords=10
+loglevel=DEBUG
+logfile=/var/log/ckan/pycsw.log
```
cbaines: Does log rotation need configuring for this log file? Also, the `DEBUG` log level sounds like it might be a little verbose?

modules/govuk/manifests/apps/ckan.pp
```diff
@@ -173,5 +184,17 @@
...
+    exec { 'symlink_pycsw_config':
+      command => "test -e ${ckan_home}/pycsw.cfg || ln -s ${pycsw_src}/default.cfg ${ckan_home}/pycsw.cfg",
+    }
+
+    exec { 'setup_pycsw_tables':
+      command => "test -e ${ckan_home}/pycsw_tables_created.tmp || ${ckan_bin}/paster --plugin=ckanext-spatial ckan-pycsw setup -p ${ckan_home}/pycsw.cfg && sudo touch ${ckan_home}/pycsw_tables_created.tmp",
+    }
+
+    exec { 'startup_pycsw_server':
+      command => "ps -C \"python */pycsw/csw.wsgi\" >/dev/null && echo \"pycsw is running\" || eval \". ${ckan_bin}/activate && python ${pycsw_src}/csw.wsgi &\"",
```
cbaines: Do you know if running using the builtin WSGI server supports handling requests concurrently?

modules/govuk/manifests/apps/ckan.pp
kentsanggds: The pycsw config is placed there and symlinked in order to not confuse people who attempt to update it in `/var/ckan` and also to keep configs in a common area, alternatively we could set an env var `PYCSW_CONFIG` to point to the config area but wasn't sure how to do this.

modules/govuk/manifests/apps/ckan.pp
kentsanggds: No, but I'll have a look into it

modules/govuk/manifests/apps/ckan/cronjobs.pp
kentsanggds: Changed it because `csw` needs solr reindexed for it to work, if we left it at 7 there is a greater chance that the sync between `csw` and `ckan` would happen later after the working day starts and it would be better if the data is synced on the day otherwise its possible that changes will happen a day late. 
I've discussed with Tom and the original 7 am was an arbitrary time so was ok to change.

modules/govuk/manifests/apps/ckan/paster_cronjob.pp
kentsanggds: The default is `--config=/var/ckan/ckan.ini`, this is a special case of running the paster command that doesn't want the `config` args as it stops the process from running, most paster commands need the default ckan config.

modules/govuk/templates/ckan/pycsw.cfg.erb
kentsanggds: Not sure if log rotation is supported, could downgrade the level to `INFO` but for initial release it might be worth leaving it at `DEBUG` to resolve issues.

modules/govuk/manifests/apps/ckan.pp
kentsanggds: Not sure, I doubt there would be that many requests, the migration happened in January, we just recently received a ticket about the missing `csw` endpoint so I doubt that it would need to handle much traffic.

modules/govuk/templates/ckan/pycsw.cfg.erb
thomasleese: We'll get that logrotate configuration for free:

```
thomasleese@ec2-integration-blue-ckan-ip-10-1-4-114:/etc/logrotate.d$ cat govuk-ckan
/var/log/ckan/*.log {
  daily
...
```

thomasleese:
@cbaines I think this is ready for a re-review now.

modules/govuk/manifests/apps/ckan.pp
cbaines: Looks like using gunicorn is now the plan, and that should be fine.

modules/govuk/templates/ckan/pycsw.cfg.erb
cbaines: :+1:

modules/govuk/manifests/apps/ckan/paster_cronjob.pp
cbaines: Ah, ok, I didn't realise this class was for configuring a ckan specific job (but I see that now).

modules/govuk/manifests/apps/ckan.pp
```diff
@@ -64,6 +64,9 @@
 ) {
   $ckan_home = '/var/ckan'
   $ckan_ini  = "${ckan_home}/ckan.ini"
+  $pycsw_config = "${ckan_home}/pycsw.cfg"
+
+  $pycsw_port = "${$port + 1}"
```
cbaines: Unless there's a specific reason for the PyCSW port to be set to be one higher than the CKAN port, I'd prefer if this was passed in as a separate parameter. That way, if you're wondering where port number 3221 is used, or wondering if there is anything using port 3221, searching for the number in the govuk-puppet repository should give you the answer.

modules/govuk/manifests/apps/ckan.pp
cbaines: Looks like a symlink isn't being used now.

modules/govuk/manifests/apps/ckan/cronjobs.pp
```diff
@@ -30,8 +33,15 @@
     ensure         => $ensure_solr_reindex,
     paster_command => 'search-index rebuild -o',
     plugin         => 'ckan',
-    hour           => '7',
+    hour           => '4',
     minute         => '0',
   }
 
+  govuk::apps::ckan::paster_cronjob { 'pycsw load':
+    paster_command => "ckan-pycsw load -p /var/ckan/pycsw.cfg -u http://localhost:${ckan_port}",
```
cbaines: There's a variable for `/var/ckan/pycsw.cfg` where this class is used. Could that be made a parameter and used here? That would make it clearer that this class involves using this configuration file.

modules/govuk/templates/ckan/pycsw.cfg.erb
```diff
@@ -0,0 +1,61 @@
...
+#csw_harvest_pagesize=10
+
+[metadata:main]
+identification_title=pycsw Geospatial Catalogue
+identification_abstract=pycsw is an OGC CSW server implementation written in Python
+identification_keywords=catalogue,discovery,metadata
+identification_keywords_type=theme
+identification_fees=None
+identification_accessconstraints=None
+provider_name=Organization Name
```
cbaines: Should any of this (or the fields below) be filled in?

modules/govuk/templates/ckan/pycsw.cfg.erb
```diff
@@ -0,0 +1,61 @@
...
+
+[metadata:inspire]
+enabled=true
+languages_supported=eng,gre
+default_language=eng
+date=YYYY-MM-DD
+gemet_keywords=Utility and governmental services
+conformity_service=notEvaluated
+contact_name=Government Digital Service
+contact_email=Email Address
```
cbaines: Maybe this should be the Zendesk email address? Assuming it's a regular "contact for help" thing.

modules/govuk/manifests/apps/ckan/cronjobs.pp
thomasleese: Good idea, I've pushed up an amended commit for this.

modules/govuk/manifests/apps/ckan.pp
thomasleese: Seems reasonable, I've pushed up an amend for this.

modules/govuk/templates/ckan/pycsw.cfg.erb
thomasleese: To be honest I don't really know, I'm not familiar enough with CSW or where they would be exposed. I'll have a look around and see what we might want to set them to.

modules/govuk/templates/ckan/pycsw.cfg.erb
thomasleese: Is the Zendesk email address exposed anywhere? I think usually we'd want people to go via https://data.gov.uk/support.

modules/govuk/manifests/apps/ckan.pp
cbaines: Looks like an upstart service is now being used.

modules/govuk/manifests/apps/ckan/cronjobs.pp
cbaines: I can see that this change is now being made in a separate commit, with this reason being given.

modules/govuk/templates/ckan/pycsw.cfg.erb
cbaines: I think that's fine. Providing we're aware that there is this configuration here.

modules/govuk/templates/ckan/pycsw.cfg.erb
cbaines: I'm unsure, regardless, I don't think it matters too much at this stage. Tweaking the contact parameters can be done later if it's useful to do.

PR #9468, opened 2019-08-14
Fix ckan routes

oscarwyatt:


modules/govuk/manifests/apps/ckan.pp
```diff
@@ -146,6 +146,16 @@
         value   => $bulk_worker_processes;
     }
 
+    govuk::app { 'ckan':
```
thomasleese: There's already a `govuk::app` block further up in the file, I think we'll probably want to put `enable_nginx_vhost => false` in that one.

thomasleese: I wonder if it'll be easier to support passing `hidden_paths` to `govuk::app` directly and then it can [pass it through to the default `nginx_vhost`](https://github.com/alphagov/govuk-puppet/blob/f689c3dbf881e2ea4a41ebb651c66adfdb3e6d1c/modules/govuk/manifests/app/config.pp#L286) - to avoid having to duplicate all the current arguments passed to the `nginx_vhost` block.

oscarwyatt:
Closing in favour of a new PR

PR #9477, opened 2019-08-14
Prep for DocumentDB

chrisfarms:
## What

Prepare for introduction of DocumentDB:

* Extends the mongo_uri module to generate connection strings with authentication and config params
* Extends the router classes to use extended mongo_uri
* Distributes the AWS RDS certificate authority bundle to all nodes

## Why

So that we can configure and connect to documentdb instances


camdesgov:
LGTM

PR #9481, opened 2019-08-15
Block access to CKAN api endpoints

oscarwyatt:
Safe list a few that we know we need

### DIFF
--- a/modules/govuk/manifests/apps/ckan.pp
+++ b/modules/govuk/manifests/apps/ckan.pp
@@ -146,6 +146,62 @@ class govuk::apps::ckan (
         value   => $bulk_worker_processes;
     }
 
+    govuk::app { 'ckan':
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/ckan.pp
```diff
@@ -146,6 +146,60 @@
...
+
+      # Used by: datagovuk_publish
+      # Used for: ckan_v26_ckan_org_sync
+      location /api/3/action/organization_show {
+        try_files $uri @app;
+      }
+
+      # Used by: datagovuk_publish
+      # Used for: ckan_v26_package_sync job
+      location /api/3/search/dataset {
```
ChrisBAshton: Not sure why this is coloured differently, wonder if `search` is a keyword that needs escaping by putting in strings? 😬 

Worth trying as-is though!

PR #9487, opened 2019-08-16
Add missing parameters to CKAN Nginx vhost

thomasleese:
When we pulled the vhost out into a separate resource, some of the custom configuration options had been lost. This is because usually the `govuk::app::config` class creates the Nginx vhost and automatically passes they custom options down into the resource.

Most notable is the `read_timeout` which I think may have prevented users from uploading large data sets (organograms), and `protected` which means we lost our normal integration password protection for CKAN.

### DIFF
--- a/modules/govuk/manifests/apps/ckan.pp
+++ b/modules/govuk/manifests/apps/ckan.pp
@@ -149,8 +149,11 @@ class govuk::apps::ckan (
 
     govuk::app::nginx_vhost { 'ckan':
       vhost              => 'ckan',
+      protected          => $vhost_protected,
+      ssl_only           => true,
       app_port           => $port,
       hidden_paths       => ['/api/'],
+      read_timeout       => $request_timeout,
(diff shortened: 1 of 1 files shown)


thomasleese:
I spotted this because when I ran Puppet manually I could see the diff included unrelated changes:

```
@@ -44,7 +44,7 @@
   proxy_set_header GOVUK-Request-Id $govuk_request_id;
   proxy_redirect off;
   proxy_connect_timeout 1s;
-  proxy_read_timeout 60;
+  proxy_read_timeout 15;
 
   access_log /var/log/nginx/ckan-access.log timed_combined;
   access_log /var/log/nginx/ckan-json.event.access.log json_event;
@@ -52,16 +52,16 @@
 
 
   location / {
-    deny all;
-    auth_basic            "Enter the GOV.UK username/password (not your personal username/password)";
-    auth_basic_user_file  /etc/govuk.htpasswd;
-    satisfy any;
 
 
     try_files $uri/index.html $uri.html $uri @app;
   }
```

PR #9489, opened 2019-08-16
Add cron jobs to manage gor during data sync

deborahchua:
We manage traffic replay with our puppet class `govuk_gor`, which
we would like to be disabled when our `govuk_env_sync` data sync
`pull` jobs happen. They currently take place between 23:15 and 5:30.

Setting the Puppet `data_sync_in_progress` fact to `true` ensures
that Puppet does not try to run traffic replay until all the data
sync jobs have completed. This should reduce noise and unnecessary
errors appearing in Sentry.

### DIFF
--- a/modules/govuk_gor/manifests/init.pp
+++ b/modules/govuk_gor/manifests/init.pp
@@ -41,10 +41,26 @@ class govuk_gor(
 
   validate_bool($enable)
 
-  file { '/etc/govuk/env.d/FACTER_data_sync_in_progress':
+  $data_sync_fact_path = '/etc/govuk/env.d/FACTER_data_sync_in_progress'
+
+  file { $data_sync_fact_path:
     ensure  => present,
     owner   => 'deploy',
     require => Class['govuk::deploy'],
+  }
+
+  cron { 'stop_gor':
+    command => "echo 'true' > ${data_sync_fact_path}; sudo initctl stop gor",
+    user    => 'deploy',
+    hour    => 23,
+    minute  => 0,
+  }
+
+  cron { 'start_gor':
+    command => "echo '' > ${$data_sync_fact_path}; sudo initctl start gor;",
+    user    => 'deploy',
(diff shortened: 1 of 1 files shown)


modules/govuk_gor/manifests/init.pp
```diff
@@ -47,6 +47,20 @@
     require => Class['govuk::deploy'],
   }
 
+  cron { 'stop_gor':
+    command => 'echo "true" > /etc/govuk/env.d/FACTER_data_sync_in_progress; sudo initctl stop gor',
+    user    => $govuk_env_sync::user,
```
cbaines: As this isn't directly associated with the env sync stuff, that user probably shouldn't be used. I would be unsure if that user even exists on the cache machines.

Given you're writing to the `/etc/govuk/env.d/FACTER_data_sync_in_progress` file, maybe consider using the `deploy` user?

modules/govuk_gor/manifests/init.pp
```diff
@@ -47,6 +47,20 @@
...
+    hour    => 23,
+    minute  => 0,
+  }
+
+  cron { 'start_gor':
+    command => 'echo "" > /etc/govuk/env.d/FACTER_data_sync_in_progress; sudo initctl start gor;',
```
cbaines: Might be clearer to extract the `/etc/govuk/env.d/FACTER_data_sync_in_progress` value in to a constant as it's used in a few places in this file?

deborahchua:
Thanks @cbaines , I've made those changes in https://github.com/alphagov/govuk-puppet/pull/9489/commits/8cea6d6380c55561f63ba15627b06f14a0629cf8

cbaines:
> Thanks @cbaines , I've made those changes in [8cea6d6](https://github.com/alphagov/govuk-puppet/commit/8cea6d6380c55561f63ba15627b06f14a0629cf8)

Great, that looks good :+1:

PR #9498, opened 2019-08-20
Remove ineffective gor commands

deborahchua:
This was possibly working before the frontend machines migrated to
AWS, but as they are no longer in Carrenza running
`govuk_node_list -C 'govuk_gor'` does not return anything.


cbaines:
There's one other file which includes this functionality: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/templates/jobs/copy_data_from_production_to_aws.yaml.erb#L50-L53

Can it be removed there as well?

deborahchua:
Thanks @cbaines, I've removed it here https://github.com/alphagov/govuk-puppet/pull/9499

PR #9506, opened 2019-08-20
Add Eamonn and give integration access

1pretz1:


kevindew:
We seem to have the incorrect spelling of his name here: mcmlroy.

I notice we went for the correct one in https://github.com/alphagov/govuk-aws-data/pull/551/files btw

I'd suggest we go for the correct spelling and if that means IT need to change his email then so be it as we don't want to let the misspelling propagate too far.

1pretz1:
Yep good point, the name has been corrected - @kevindew

modules/users/manifests/eamonnmcelroy.pp
```diff
@@ -0,0 +1,8 @@
+# Creates the eamonnmcmrloy user
```
kevindew: Just got one last misspelling to fix here.

1pretz1: ARGHHH, Eamonn actually sent me this and seemingly spelled his name doubly wrong haha -'eamonnmcmrloy' -> 'eamonnmcelroy'

PR #9511, opened 2019-08-21
parametise the licensing config file to accept database authentication

fredericfran-gds:
# Context

In AWS, the licensify apps use documentDB which compulsory requires a username and password.

# Deci sions
1. add the option to configure the app to use a specified username and password for the database if the database authentication is enabled in the config.

Data part of PR: https://github.com/alphagov/govuk-secrets/pull/780

### DIFF
--- a/hieradata_aws/integration.yaml
+++ b/hieradata_aws/integration.yaml
@@ -55,10 +55,10 @@ govuk::apps::url_arbiter::db::backend_ip_range: '10.1.3.0/24'
 govuk::apps::whitehall::basic_auth_credentials: "%{hiera('http_username')}:%{hiera('http_password')}"
 govuk::apps::whitehall::highlight_words_to_avoid: true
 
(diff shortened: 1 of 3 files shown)


sengi:
Looks good but I think we should leave the database hostnames in the clear, since this is the current pattern elsewhere in the codebase and it makes it easier for us to see what's going on.

fredericfran-gds:
> Looks good but I think we should leave the database hostnames in the clear, since this is the current pattern elsewhere in the codebase and it makes it easier for us to see what's going on.

I talked with Sebastian who advised to put into secrets

sengi:
> > Looks good but I think we should leave the database hostnames in the clear, since this is the current pattern elsewhere in the codebase and it makes it easier for us to see what's going on.
> 
> I talked with Sebastian who advised to put into secrets

Ok no probs :)

PR #9515, opened 2019-08-22
Remove xray config

cbaines:
As it's not being used at the moment.


cbaines:
I'm not quite sure how far this should go in terms of removing the xray config, I'm just trying to fix the new bouncer machines at the moment.

PR #9518, opened 2019-08-22
Add Go version 1.12.9

bilbof:
We'll be upgrading router to the latest version of go, 1.12.9.

https://trello.com/c/bh2jDTgp/415

### DIFF
--- a/modules/golang/manifests/init.pp
+++ b/modules/golang/manifests/init.pp
@@ -11,7 +11,7 @@ class golang {
     require        => Class['govuk_ppa'],
   }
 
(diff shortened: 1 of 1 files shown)


cbaines:
> We'll be upgrading to the latest version of go, 1.12.9.
> 
> https://trello.com/c/bh2jDTgp/415

I guess it's the router, but it might be useful in the commit message to say what you're intending to upgrade to this go version.

bilbof:
Thanks, it is router we're upgrading -  I added that to the commit message.

cbaines:
> Thanks, it is router we're upgrading - I've added that to the commit message.

Great, thanks :)

PR #9519, opened 2019-08-23
app-worker Icinga check improvements

cbaines:
These changes should make the procfile worker checks clearer, and start monitoring the memory usage.

### DIFF
--- a/modules/govuk/manifests/apps/whitehall.pp
+++ b/modules/govuk/manifests/apps/whitehall.pp
@@ -318,9 +318,11 @@ class govuk::apps::whitehall(
     }
 
     govuk::procfile::worker { 'whitehall-admin':
-      setenv_as      => $app_name,
-      enable_service => $enable_procfile_worker,
-      process_count  => $procfile_worker_process_count,
+      setenv_as                 => $app_name,
+      enable_service            => $enable_procfile_worker,
+      process_count             => $procfile_worker_process_count,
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/apps/whitehall.pp
```diff
@@ -318,9 +318,11 @@
...
 
     govuk::procfile::worker { 'whitehall-admin':
-      setenv_as      => $app_name,
-      enable_service => $enable_procfile_worker,
-      process_count  => $procfile_worker_process_count,
+      setenv_as                 => $app_name,
+      enable_service            => $enable_procfile_worker,
+      process_count             => $procfile_worker_process_count,
+      memory_warning_threshold  => 2000,
+      memory_critical_threshold => 14000,
```
barrucadu: That seems a pretty wide range, is this right?

cbaines: The memory usage varies similarly. The machines have enough for Sidekiq to use 14GB of memory without too much of an issue, which is why I set the critical threshold here, but generally, I was setting the critical threshold quite high.

bevanloon: @barrucadu - are you happy with this or do you still think we need more tweaking to the values?

barrucadu: Sorry for missing this; I'm happy with it.

PR #9528, opened 2019-08-28
Reactivate router sync for integration

benthorner:
Previously we deactivated sync for the router in integration and seem to
have forgotten to turn it back on. Since most apps simply override any
existing routes, this has mostly gone unnoticed, but may account for
some weirdnesses with 404 pages and such. This reactivates the push/pull
flow from staging to integration.

Reference: https://github.com/alphagov/govuk-puppet/blob/145bd8c1cbe7364b191907914050261c5bbe9860/hieradata_aws/class/integration/router_backend.yaml


hieradata_aws/class/integration/router_backend.yaml
```diff
@@ -1,4 +1,37 @@
 govuk_env_sync::tasks:
+  "pull_router_daily":
+    ensure: "present"
```
1pretz1: Just looked at the diff for when the sync was activated and spotted 'ensure' has now been set to 'present' from 'absent' - https://github.com/alphagov/govuk-puppet/blob/145bd8c1cbe7364b191907914050261c5bbe9860/hieradata_aws/class/integration/router_backend.yaml#L3, just wondered if this was intentional?

benthorner: Indeed it is intentional - we want to turn it back on ;-).

1pretz1: Ohh okay lol, this is my lack of knowledge of puppty things shining through! 😆

PR #9538, opened 2019-08-29
Call on some application 5xx alerts

thomasleese:
This PR changes it so that PagerDuty gets triggered if email-alert-api, email-alert-frontend or finder-frontend have a 5xx rate that triggers a critical alert.

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -170,6 +170,9 @@
 # [*alert_5xx_critical_rate*]
 # the 5xx error percentage that should generate a critical
 #
+# [*alert_notification_period*]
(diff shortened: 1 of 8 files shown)


erino:
@thomasleese Do you think collections should be added as it'll host the landing page?

thomasleese:
@erino Sounds reasonable, I'll add collections now.

thomasleese:
Do we actually need this? It wasn't merged over the weekend and we already have alerts which cover high numbers of 5xx across GOV.UK?

erino:
Maybe not, and I think platform health are looking into altering

thomasleese:
I'll close this, we can re-open it if we feel like we need it.

PR #9543, opened 2019-08-29
Remove config for unused publishing_api_db_admin machine.

sengi:
This config was for an RDS instance which was never created.


sengi:
Related PRs:
https://github.com/alphagov/govuk-aws/pull/1086
https://github.com/alphagov/govuk-aws-data/pull/558

PR #9547, opened 2019-08-31
Bump finder-frontend worker processes

sihugh:
6 worker process on finder-frontend uses about 1.4GB at normal running. We'll need to bump the instance size to maybe a c5.2xlarge to get a better processor and more memory before we deploy
this.

We could perhaps scale down the number of machines a little too :-)


sihugh:
Closing this because we're taking a different approach

PR #9550, opened 2019-09-02
Make publishing-api in AWS Staging/Prod use Signon in Carrenza.

sengi:
Publishing API in Staging and Prod is being migrated to AWS but Signon
will remain in Carrenza for the time being. We therefore need to
override the Plek service discovery config so that Publishing API can
talk to Signon from the new Staging and Prod environments.

We accomplish this by using the external domain name when constructing
the URI for publishing-api to talk to Signon. This can be removed once
Signon is moved to AWS.


sengi:
Prior art: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/node/s_draft_content_store.pp

PR #9558, opened 2019-09-03
add licensify environment hiera for aws integration

fredericfran-gds:
# Context

For the licensify testing in AWS integration, there is a need for licensify bucket environment hiera for aws integration because otherwise, the licensify apps in integration do not have the full path of the AWS s3 bucket to store the application.

# Decisions
1. set the relevant hiera


fredericfran-gds:
good idea but it is like that for staging and production, will have to go and change that to maintain consistency.

PR #9562, opened 2019-09-03
Avoid spooling to disk and back when restoring to Postgres.

sengi:
Instead of writing out the filtered SQL dump to a file on disk, pipe it
it into psql. This also avoids a gzip and gunzip since we no longer need
to worry about taking up space on disk.


sengi:
Tested in Integration by running manually running the pull cronjob for `content_data_admin` with the new script temporarily in place on the db_admin machine.

sengi:
Also tested by restoring Publishing API in Staging. Runs approx. 30 mins faster than without this change.

PR #9563, opened 2019-09-04
Reduce duplication of allowed paths in CKAN Nginx config

thomasleese:
This uses ERB to avoid the duplication we currently have in this configuration file with the allowed paths and each of the `location` blocks.


kentsanggds:
Can you add the changes in this PR - https://github.com/alphagov/govuk-puppet/pull/9561/files to this PR? I can then close the that PR

PR #9565, opened 2019-09-05
Single deploy

fredericfran-gds:


modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
```diff
@@ -81,6 +82,10 @@
                 - app:migrate
                 - app:hard_restart
                 - app:migrate_and_hard_restart
+        - string:
+            name: TARGET_MACHINES
+            description: List of machines to deploy the app to.
```
sengi: Comma-separated list?

fredericfran-gds: yeah should define this a bit further

PR #9578, opened 2019-09-10
Amend Sync Script to Work With Publishing Api Data From Carrenza

camdesgov:
The postgres Publishing-api data in Carrenza uses a different postgres
role to RDS postgres. This causes errors when restoring the data to AWS.
This change ensures the role names match.

### DIFF
--- a/modules/govuk_env_sync/files/govuk_env_sync.sh
+++ b/modules/govuk_env_sync/files/govuk_env_sync.sh
@@ -358,7 +358,13 @@ function restore_postgresql {
     db_hostname='postgresql-primary'
   fi
 
-  pg_restore "${tempdir}/${filename}" | sed '/^COMMENT\ ON\ EXTENSION\ plpgsql/d' | gzip > "${tempdir}/${filename}.dump"
+  if [ "${database}" == 'publishing_api_production' ]; then
+    pg_restore "${tempdir}/${filename}" | sed '/^COMMENT\ ON\ EXTENSION\ plpgsql/d' \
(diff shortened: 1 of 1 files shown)


modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -358,7 +358,11 @@ function restore_postgresql {
     db_hostname='postgresql-primary'
   fi
 
-  pg_restore "${tempdir}/${filename}" | sed '/^COMMENT\ ON\ EXTENSION\ plpgsql/d' | gzip > "${tempdir}/${filename}.dump"
+  if [ "${database}" == 'publishing_api_production' ]; then
+    pg_restore "${tempdir}/${filename}" | sed '/^COMMENT\ ON\ EXTENSION\ plpgsql/d' | sed "s/postgres/aws_db_admin/g" | gzip > "${tempdir}/${filename}.dump"
```
sengi: I think it's worth a quick look to see if we can make the global replace `s/postgres/aws_db_admin/g` a bit more context-specific without loss of generality. If not then I'm sure we can live with it, but I'm a little uncomfortable with the potential for "[clbuttic](https://www.telegraph.co.uk/news/newstopics/howaboutthat/2667634/The-Clbuttic-Mistake-When-obscenity-filters-go-wrong.html)" errors here :)

camdesgov: Good idea, I've pushed a new change so it's now more specific.

PR #9583, opened 2019-09-11
Use cron resource to remove old whisper data

thomasleese:
This makes it easier to run the task manually because it can be a separate command and it means we can rely on Puppet to manage the cron job for us.

[Trello Card](https://trello.com/c/VugQy0iY/1210-3-improve-how-we-remove-old-graphite-data)

### DIFF
--- a/modules/govuk/manifests/node/s_graphite.pp
+++ b/modules/govuk/manifests/node/s_graphite.pp
@@ -101,10 +101,23 @@ class govuk::node::s_graphite (
 
   ## The way we generate graphite data with changing hostnames in AWS require a bit of clean up
   if $::aws_migration {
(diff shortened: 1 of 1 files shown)


cbaines:
> means we can rely on Puppet to manage the cron job for us

It looks like the previous approach would have probably worked, what was the issue?

thomasleese:
I think the old way was actually working, but the script didn't do what I thought it does, so it appeared like it wasn't working.

Either way, with the old method, I couldn't find any reference to it executing in the logs, and doing it this way should ensure we see it running.

cbaines:
> Either way, with the old method, I couldn't find any reference to it executing in the logs, and doing it this way should ensure we see it running.

Ok, logs are good :)

PR #9585, opened 2019-09-11
Sync Licensify database from Prod to AWS Staging and Integration.

sengi:
Now that the production database contains no records which can't be
imported into DocumentDB, we want to sync the production data to the
other two environments.

 - AWS Staging pulls from UKCloud Prod.
 - AWS Integration pulls from UKCloud Staging (which pulls from UKCloud Prod).

The timing is intended to line up approximately with the backups from
UKCloud, which happen at 04:00. The idea is that the restores happen
(hopefully) after the backups have completed and we avoid a day's
unnecessary delay.


sengi:
Tested by deploying Puppet in Integration from the branch and running the cron jobs manually. We'll need to wait until tomorrow for the Staging backups to be free of nul-containing strings (or run a manual backup of Prod and restore to UKCloud Staging).

PR #9590, opened 2019-09-12
Improve logging and error reporting in Licensify deploy job.

sengi:


sengi:
> good work!

Thanks! :)

Tested in Integration just now (for the record).

PR #9594, opened 2019-09-13
Fix Permissions Errors For Splunk Log Directory

camdesgov:
New Calculator Frontend boxes are created
with incorrect permissions on the
/opt/splunkforwarder/var directory this prevents
Splunk from writing to the logs and causes an error.
This PR fixes this issue.

### DIFF
--- a/modules/govuk_splunk/manifests/init.pp
+++ b/modules/govuk_splunk/manifests/init.pp
@@ -52,6 +52,14 @@ class govuk_splunk(
     require => Apt::Source['splunk'],
   }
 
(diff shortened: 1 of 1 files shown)


modules/govuk_splunk/manifests/init.pp
```diff
@@ -52,6 +52,14 @@
     require => Apt::Source['splunk'],
   }
 
+  file {'/opt/splunkforwarder/var':
+    ensure  => file,
```
rhowe-gds: should this not be `ensure => directory`?

PR #9612, opened 2019-09-19
Don't replay Licensify traffic which assumes transactional state.

sengi:
Certain GET requests to Licensify (that is, certain paths under
www.gov.uk/apply-for-a-licence/) only make sense after the licence
application record has been created in the database. This happens via a
POST request which we intentionally exclude from traffic replay.

It therefore doesn't make sense to replay these requests - it isn't a
meaningful way to test those paths. The replayed requests cause errors
(because the expected state is missing from the database) and clutter up
the Staging logs with stacktraces, which frustrates debugging and could
mask other problems.

### DIFF
--- a/hieradata_aws/class/production/cache.yaml
+++ b/hieradata_aws/class/production/cache.yaml
@@ -4,3 +4,7 @@ nginx::logging::days_to_keep: 7
 router::gor::add_hosts: false
 router::gor::replay_targets:
   'www-origin.staging.govuk.digital': {}
+router::gor::http_disallow_url:
+  - 'apply-for-a-licence/payment'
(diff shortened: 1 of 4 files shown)


sengi:
The args hash/array structure is turned into the actual command line here, btw: https://github.com/alphagov/puppet-gor/blob/master/templates/etc/init/gor.conf.erb

Based on a bit of manual testing with irb, I believe that the case where there are no disallowed URLs still works. Passing it an empty list like `'-http-disallow-url'  => []` produces an extra (escaped) blank line in the command but doesn't add `-http-disallow-url=[]` or anything undesirable. Please don't take my word for it, though :)

PR #9626, opened 2019-09-25
Add google domain verification file

MuriloDalRi:
This file is necessary to verify ownership of
assets.publishing.service.gov.uk for Google Search Console.

[Trello](https://trello.com/c/2zhVXzBJ/1299-5-verify-ownership-of-assetspublishingservicegovuk-for-google-search-console)

### DIFF
--- a/modules/router/templates/assets_origin.conf.erb
+++ b/modules/router/templates/assets_origin.conf.erb
@@ -121,4 +121,8 @@ server {
     proxy_pass <%= @upstream_ssl ? 'https' : 'http' %>://static.<%= @app_domain %>;
     <%- end %>
   }
+
+  location /googlec908b3bc32386239.html {
+    return 200 'google-site-verification: googlec908b3bc32386239.html';
+  }
 }


modules/router/manifests/assets_origin.pp
```diff
@@ -64,6 +64,11 @@
       logstream => present;
   }
 
+  file { '/var/www/googlec908b3bc32386239.html':
+    ensure  => present,
+    content => template('router/googlec908b3bc32386239.html.erb'),
```
cbaines: Do we know if the value here should be secret, or can be public?

MuriloDalRi: I think it needs to be public so Google can access it

cbaines: I know it needs to be publicly accessible, but I'm wondering if it should be publicly discoverable in the govuk-puppet configuration.

MuriloDalRi: I don't know if there's another way to do this. What would be the issue of having it here?

cbaines: The alternative would be to use the hieradata in the govuk-secrets repository. But if we're confident that it's OK to have the filename and contents in the clear, then it's fine to keep it here.

modules/router/templates/googlec908b3bc32386239.html.erb
```diff
@@ -0,0 +1 @@
+google-site-verification: googlec908b3bc32386239.html
```
cbaines: This is in templates, but it's not a template, as the content is fixed. It could be moved to the files directory, but also as there's a single line, the contents could be moved inside the NGinx configuration, I'll comment more above.

modules/router/templates/assets_origin.conf.erb
```diff
@@ -121,4 +121,8 @@ server {
     proxy_pass <%= @upstream_ssl ? 'https' : 'http' %>://static.<%= @app_domain %>;
     <%- end %>
   }
+
+  location /googlec908b3bc32386239.html {
+    alias /var/www/googlec908b3bc32386239.html;
```
cbaines: I think the `return` [1] statement could be used here, so something like:

```
  return 200 'google-site-verification: googlec908b3bc32386239.html';
```

1: http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#return

That would remove the need to have the separate file, as well as the related configuration.

MuriloDalRi: That's neat, will change!

PR #9629, opened 2019-09-25
Apply lb_healthchecks.conf nginx config on all relevant boxes in AWS.

sengi:
Currently there is a blob of copypasta on 7 node classes which use
`lb_healthchecks.conf`, but it needs to be everywhere on AWS because of
https://github.com/alphagov/govuk-aws/blob/master/doc/architecture/decisions/0037-alb-health-checks.md

This factors out the copypasta and paves the way for us to fix the LB
healthcheck config for other apps on AWS which are not currently
configured properly, particularly `whitehall-admin`.

With this change, an app should only need to pass `$health_check_path` to
`govuk::app::config` in order for its LB healthcheck to be mapped correctly on the
default vhost.

### DIFF
--- a/modules/govuk/manifests/node/s_backend.pp
+++ b/modules/govuk/manifests/node/s_backend.pp
@@ -37,20 +37,8 @@ class govuk::node::s_backend inherits govuk::node::s_base {
 
   include nginx
 
-  if $::aws_migration {
-    concat { '/etc/nginx/lb_healthchecks.conf':
-      ensure => present,
-      before => Nginx::Config::Vhost::Default['default'],
-    }
-    $extra_config = 'include /etc/nginx/lb_healthchecks.conf;'
-  } else {
-    $extra_config = ''
-  }
-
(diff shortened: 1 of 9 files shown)


sengi:
Tested in Integration. Diff:

```diff
Notice: /Stage[main]/Govuk::Node::S_backend/Nginx::Config::Vhost::Default[default]/Nginx::Config::Site[default]/File[/etc/nginx/sites-available/default]/content: 
--- /etc/nginx/sites-available/default	2019-09-26 16:34:39.986485108 +0000
+++ /tmp/user/0/puppet-file20190926-11597-1cf0fw4	2019-09-26 16:43:46.179478782 +0000
@@ -7,10 +7,11 @@
   location = /_healthcheck {
     return 200;
   }
+  include /etc/nginx/lb_healthchecks.conf;
 
   location / {
     return 500 '{"error": "Fell through to default vhost"}\n';
   }
-  include /etc/nginx/lb_healthchecks.conf;
+  
 }
```

PR #9634, opened 2019-09-27
Fix load balancer healthcheck for whitehall-admin.

sengi:
Plumb the healthcheck for `whitehall-admin` through the default vhost in
the nginx config.

The healthcheck needs to be accessible to the load balancer so that
traffic does not get routed to broken instances. The convention for the
URL paths of these healthchecks is defined in
https://github.com/alphagov/govuk-aws/blob/master/doc/architecture/decisions/0037-alb-health-checks.md

`govuk::apps::whitehall` needs some special-case config to achieve this
because it sets `enable_nginx_vhost=false` and does the nginx vhost
config itself, which means we don't get this for free from
`govuk::app::config`.

### DIFF
--- a/modules/govuk/manifests/apps/whitehall.pp
+++ b/modules/govuk/manifests/apps/whitehall.pp
@@ -311,6 +311,16 @@ class govuk::apps::whitehall(
     ',
     }
 
+    # govuk::app::config doesn't automatically configure Whitehall's LB healthcheck
+    # because Whitehall sets $enable_nginx_vhost=false. We therefore need to
+    # configure the LB healthcheck explicitly.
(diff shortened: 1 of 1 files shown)


sengi:
Tested in Integration; verified that `lb_healthchecks.conf` is as expected and that a local curl request to the healthcheck works.

sengi:
Alternatives considered: looked at removing the `$enable_nginx_vhost=false` so that `govuk::app::config` would just Do The Right Thing, but it looks like that would change a bunch of other stuff like paths of log files and so on. It would be nice to just clean all that up, but I decided that was probably too invasive a change to get working and tested without spending undue time on this.

PR #9637, opened 2019-09-30
Disable the check_signon_queue_sizes Icinga check where unnecessary

cbaines:
As Signon runs on the Carrenza side of Production, Staging and in
Integration (which is mainly in AWS), disable the check in the other
locations.

This should resolve the check appearing as UNKNOWN where Signon isn't
running.

### DIFF
--- a/modules/monitoring/manifests/checks/sidekiq.pp
+++ b/modules/monitoring/manifests/checks/sidekiq.pp
@@ -2,31 +2,37 @@
 #
 # Nagios alerts for sidekiq queue latency
 #
+# [*enable_signon_check*]
+#   Enable monitoring for check_signon_queue_sizes
+#
 # [*enable_support_check*]
 #   Enable monitoring for check_support_default_queue_size
 #
 #
 class monitoring::checks::sidekiq (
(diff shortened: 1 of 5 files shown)


modules/monitoring/manifests/checks/sidekiq.pp
```diff
@@ -2,31 +2,37 @@
 #
 # Nagios alerts for sidekiq queue latency
 #
+# [*enable_signon_check*]
```
thomasleese:
```suggestion
# [* check_signon_queue_sizes*]
```

thomasleese: But I wonder if we should be consistent with the naming here and either make them both `enable_$app_check` or `check_$app_queue_sizes`?

cbaines: Right, so I've definitely written the wrong thing somewhere, as this comment doesn't even match the thing it's meant to document.

I'm fine changing both if you think that's a better name, but initially I'll just make this new parameter consistent with the existing one.

PR #9638, opened 2019-09-30
Add Dilwoar to email alerts white list

DilwoarH:


thomasleese:
You'll need to add yourself to the non-AWS equivalent if you want to receive emails in staging.

DilwoarH:
> You'll need to add yourself to the non-AWS equivalent if you want to receive emails in staging.

Just need integration at the moment but thanks for the heads up.

PR #9639, opened 2019-09-30
Add Issy as a user in Puppet and to integration/training envs

issyl0:
- I'm back!


hieradata_aws/training.yaml
```diff
@@ -196,6 +196,7 @@ users::usernames:
   - felisialoukou
   - fredericfrancois
   - ianjames
+  - isabelllong
```
cbaines: I don't think this environment exists at the moment, but there's no harm adding yourself.

PR #9651, opened 2019-10-03
Remove mtail / govuk_mtail

barrucadu:
All of the resources these define are ensured-absent.


barrucadu:
Me neither, I just spotted it didn't seem to be in use.

PR #9661, opened 2019-10-09
Remove Carrenza Link Checker API hieradata

cbaines:
This follows on from the migration of the Link Checker API to AWS, removing the configuration that's associated with Carrenza.

This is just configuration cleanup, and shouldn't actually effect the behaviour of Puppet.


cbaines:
> It looks like there are a few other apps we could remove from the Carrenza Hierdata as well (frontends, local links manager, support and support-api, etc).

Yep, now that the tests are better with handling AWS and Carrenza configuration differently, I'm hoping it'll be easy to improve those bits as well.

PR #9670, opened 2019-10-10
Increase the number of max db connections

AlanGabbianelli:
This is to cope with the increased number of sidekiq workers introduced
in https://github.com/alphagov/email-alert-api/pull/998


thomasleese:
Note that this change will restart Postgres, so it's best to do this sort of thing out of hours.

thomasleese:
I don't think we need this any more as we're hoping to improve email-alert-api in other ways. We'll also be migrating it to RDS soon, so we can revisit it then.

PR #9683, opened 2019-10-14
Logging reboot elasticache

fredericfran-gds:


modules/govuk/manifests/node/s_db_admin.pp
```diff
@@ -225,10 +225,20 @@
...
       require => [ File['/usr/local/bin/reboot_elasticache.py'], Package['boto3'],],
     }
 
+    @logrotate::conf { 'reboot-elasticache':
+      ensure        => 'present',
+      matches       => '/var/log/reboot_elasticache',
+      days_to_keep  => '7',
+      delaycompress => true,
+      copytruncate  => false,
+      create        => "644 root root,
```
schmie: misses closing quote

fredericfran-gds: good spot, spuff, thought the checks would have caught that.

PR #9690, opened 2019-10-15
Upgrade Python Docker library to restore compatibility with Docker.

sengi:
Version 2.7 of the Python `docker` library is not compatible with
the 1.32 API of Docker 17.09.0-ce, which we're currently running.

The version incompatibility was causing `collectd` to log a warning
and fail to load the Docker plugin.

This might fix the problem where `collectd` uses 100% of a CPU on
the build machines, but it's too early to tell for sure.


sengi:
Tested in Carrenza Integration.

PR #9697, opened 2019-10-16
Stop smokey-loop while the data sync is in progress

thomasleese:
This follows on from https://github.com/alphagov/govuk-puppet/pull/9489 to disable smokey-loop while the data sync is happening.

This is because it makes a lot of requests to Whitehall while it's not in a state that can recieve requests.

[Trello Card](https://trello.com/c/7Ar7u9e6/819-stop-smokey-running-overnight-on-staging-and-integration)

### DIFF
--- a/modules/govuk_data_sync_in_progress/manifests/init.pp
+++ b/modules/govuk_data_sync_in_progress/manifests/init.pp
@@ -1,0 +1,69 @@
+# == Class: govuk_data_sync_in_progress
+#
+# Configure the `data_sync_in_progress` fact and optionally allows commands
+# to run when the data sync starts and finishes.
+#
+# === Parameters
+#
+# [*start_command*]
+#   Command to run when the data sync starts.
+#
+# [*finish_command*]
+#   Command to run when the data sync finishes.
+#
+define govuk_data_sync_in_progress(
+  $start_command = undef,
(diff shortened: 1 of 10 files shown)


modules/govuk_data_sync_in_progress/manifests/init.pp
```diff
@@ -0,0 +1,69 @@
...
+) {
+  $fact_path = '/etc/govuk/env.d/FACTER_data_sync_in_progress'
+
+  $start_hour = 22
+  $start_minute = 0
+
+  $finish_hour = 5
+  $finish_minute = 45
+
+  if !defined(File[$fact_path]) {
```
cbaines: I don't really get this conditional, what is it doing?

thomasleese: It’s so that this definition is generic enough that it can be used multiple times on the same machine class without Puppet complaining that there is already a definition for that file.

I don’t know if it’s a problem at the moment (since Smokey and Gor run on different machine classes I think) but I wanted to make the definition generic so it didn’t confuse anyone in the future.

cbaines: Ok, that makes sense :+1:

PR #9702, opened 2019-10-16
remove custom plek entries for publishing api in AWS staging

fredericfran-gds:
As part of the migration of publishing_api to AWS staging, we need to remove the custom PLEK entries which direct AWS nodes to use the publishing api external address since these nodes should use the internal address of publishing_api in AWS.

It should be noted that the Carrenza nodes should still use the external address of the publishing api.

We also do whitehall frontend here for consistency but the app actually still lives in Carrenza. This code change affects only nodes deployed in AWS.


sengi:
Can we move these overrides to `../deploy/config.pp` so that they're not duplicated across several different apps' configs? It seems unnecessarily risky (as well as messy) to override it on a per-app basis. For example, if a service URI is overridden for one client app in AWS Staging, then it should be overridden the same way for all client apps in AWS Staging.

sengi:
> Can we move these overrides to `../deploy/config.pp` so that they're not duplicated across several different apps' configs? It seems unnecessarily risky (as well as messy) to override it on a per-app basis. For example, if a service URI is overridden for one client app in AWS Staging, then it should be overridden the same way for all client apps in AWS Staging.

Actually on second thoughts, it's probably not worth refactoring this because when we move Production we'll be deleting the overrides anyway. Ignore me :)

modules/govuk/manifests/node/s_frontend.pp
```diff
@@ -31,7 +31,7 @@
     'UNICORN_TIMEOUT': value => 15;
   }
 
-  if ( ( $::aws_migration == 'frontend' ) and ($::aws_environment == 'staging') ) or ( ($::aws_migration == 'frontend' ) and ($::aws_environment == 'production') ) {
+  if ( ($::aws_migration == 'frontend' ) and ($::aws_environment == 'production') ) {
     $app_domain = hiera('app_domain')
 
     govuk_envvar {
```
sengi: `email-alert-api` isn't moving yet, right?

modules/govuk/manifests/node/s_search.pp
```diff
@@ -20,7 +20,7 @@
   }
 
   # Set Plek for AWS to Carrenza communication
-  if ( ( $::aws_migration == 'search' ) and ($::aws_environment == 'staging') ) or ( ($::aws_migration == 'search' ) and ($::aws_environment == 'production') ) {
+  if ( ($::aws_migration == 'search' ) and ($::aws_environment == 'production') ) {
     $app_domain = hiera('app_domain')
 
     govuk_envvar {
```
sengi: Same as above; I think this is changing the URI for `email-alert-api` when it shouldn't be.

fredericfran-gds:
I did the change in the single deploy file as suggested, less prone to error to do this in a single place.

PR #9707, opened 2019-10-17
[DO NOT MERGE YET] Use internal LB when Prod AWS clients connect to Publishing API.

sengi:
When Production Publishing API moves to AWS, clients in AWS should use
the internal load balancer instead of the external (public-facing) one
to reach Publishing API.

This doesn't affect clients in Carrenza, which will continue to use the
external LB.

### DIFF
--- a/modules/govuk/manifests/deploy/config.pp
+++ b/modules/govuk/manifests/deploy/config.pp
@@ -126,14 +126,11 @@ class govuk::deploy::config(
     # Integration are in AWS, so Plek already does the right thing by default
     # in those environments.
     #
-    # 2. Publishing API is still in Carrenza Production for now.
-    #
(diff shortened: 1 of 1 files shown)


sengi:
Rebased so that this is still ready to go. There was a conflict in the comment. No change to the actual code.

Something to bear in mind when we roll this out: when Puppet agents pick up the change, they'll restart the Ruby apps because of the environment variable changes. Restarting frontend apps unfortunately isn't zero-downtime; it causes a short blip of user-visible errors. I don't think there's a lot that we can do about this though. (I believe the same thing happens when new versions of the frontend apps get deployed.)

sengi:
Obsoleted by https://github.com/alphagov/govuk-puppet/pull/9902

PR #9710, opened 2019-10-17
Add Camille To Email Address Override Whitelist

camdesgov:
This will help RE-Govuk test email-alert-api


sengi:
I see there's a comment about keeping the list of users in sync with https://www.notifications.service.gov.uk/services/ecfc7e2f-5145-45a6-9413-5d9b6e813ea9/users — I'm not able to view that page, though. Is it in sync? :)

PR #9731, opened 2019-10-23
Simplify the router module for just being used in AWS

cbaines:


modules/router/templates/draft-assets.conf.erb
```diff
@@ -49,24 +45,11 @@ server {
   location / {
     return 404;
   }
-  <%- if scope.lookupvar('::aws_migration') %>
   set $upstream_asset_manager <%= @upstream_ssl ? 'https' : 'http' %>://asset-manager.<%= @app_domain %>;
-  <%- end %>
-  location /auth/ {
-    <%- if scope.lookupvar('::aws_migration') %>
-    proxy_pass $upstream_asset_manager;
```
issyl0: Looks like you deleted too much stuff here? Is `location /auth/` no longer needed?

cbaines: Good spot!

I've fixed that now :)

PR #9741, opened 2019-10-25
Add scheduled job for search-api rank evaluation

bilbof:
This will run a search rake task every 3 hours. The rake task populates graphite with the ranking evaluation scores for a bunch of queries that we have relevancy judgements for.

These scores are reported in the search relevancy dashboard, allowing the search team to monitor search relevancy over time.

This also adds a contact group for the search team, so alerts will go to our Slack channel.

Trello: https://trello.com/c/9qZqdAXs

### DIFF
--- a/hieradata_aws/integration.yaml
+++ b/hieradata_aws/integration.yaml
@@ -136,6 +136,7 @@ govuk_jenkins::jobs::search_test_spelling_suggestions::cron_schedule: '0 10 * * 
 govuk_jenkins::jobs::search_api_fetch_analytics_data::skip_page_traffic_load: true
 govuk_jenkins::jobs::search_api_fetch_analytics_data::cron_schedule: '30 9 * * 1-5'
 
+govuk_jenkins::jobs::search_relevancy_rank_evaluation::cron_schedule: '45 */3 * * *' # every three hours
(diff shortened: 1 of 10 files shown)


hieradata_aws/integration.yaml
```diff
@@ -136,6 +136,7 @@ govuk_jenkins::jobs::search_test_spelling_suggestions::cron_schedule: '0 10 * *
 govuk_jenkins::jobs::search_api_fetch_analytics_data::skip_page_traffic_load: true
 govuk_jenkins::jobs::search_api_fetch_analytics_data::cron_schedule: '30 9 * * 1-5'
 
+govuk_jenkins::jobs::search_relevancy_rank_evaluation::cron_schedule: 'H */3 * * *' # every three hours
```
barrucadu: Do we know what happens if the rank evaluation runs during the data sync?  Will it just work, or will it throw an error if it overlaps with the snapshot restore?

bilbof: As you suggested, it'll run every 45 mins to avoid coinciding with the indices being deleted:

https://github.com/alphagov/govuk-puppet/blob/master/hieradata_aws/class/staging/search.yaml#L19

PR #9749, opened 2019-10-28
Add More Members of RE Govuk to Email Whitelist

camdesgov:
Add Chris Banks and Tom Whitwell for testing purposes.


hieradata/common.yaml
```diff
@@ -329,6 +329,8 @@ govuk::apps::email_alert_api::email_address_override_whitelist:
   - tim.blair@digital.cabinet-office.gov.uk
   - thomas.leese@digital.cabinet-office.gov.uk
   - camille.descartes@digital.cabinet-office.gov.uk
+  - tom.whitwell@digital.cabinet-office.gov.uk
```
sengi: It looks like these were in alphabetical order before we got here :) Maybe we should keep them that way?

camdesgov: Good spot! Now amended.

PR #9764, opened 2019-10-30
Fix DDoSDetected alert to fire when and only when DDoS is detected.

sengi:
The DDoSDetected alert was broken in two ways:

  1. When the DDoSDetected CloudWatch metric was 1 (that is, when AWS
     detects a DDoS), the alert would not fire because it was doing a
     floating-point comparison between two almost-but-not-quite
     identical representations of approximately 1.0.
  2. When no DDoS activity had been detected for more than a day (which
     appears to be the hard-coded time window for all of our CloudWatch
     Icinga alerts), the alert would fire as `UNKNOWN` because the alert
     didn't know what to do when no data points were returned.


sengi:
Sorry, I was a bit hasty here. Realised needs a little more work — `icinga::check` doesn't actually have an `args` argument yet so I need to plumb this through (or perhaps there's a more idiomatic way to do it).

PR #9765, opened 2019-10-30
Remove redundant 'app_name app healthcheck' Icinga check

issyl0:
From the Trello card, specific to `email-alert-api`:
- This icinga alert regularly makes a GET request to the `/healthcheck` endpoint in `email-alert-api` and checks that the app responds successfully.
- We don't need it as we already have the [`email-alert-api app healthcheck not ok` icinga
  alert](https://alert.publishing.service.gov.uk/cgi-bin/icinga/extinfo.cgi?type=2&host=email-alert-api-1.backend.publishing.service.gov.uk&service=email-alert-api+app+healthcheck+not+ok) that regularly makes GET requests to the same endpoint and if the response isn't successful, it fails.
- Having both is redundant because whenever `email-alert-api app healthcheck` fails, `email-alert-api app healthcheck not ok` fails too.

However:

- It was non-trivial to scope this to just `email-alert-api` when the same problem exists for all apps with /healthcheck routes.
- This change removes the 'app healthcheck' check in favour of the existing and all-encompassing 'app healthcheck not ok' check.

https://trello.com/c/ZDAY7kMs/1479-3-remove-the-redundant-email-alert-api-app-healthcheck-icinga-alert

---

My Puppet is very rusty. I hope this does what we think it does. Who knows! How bad can deleting some code be? 😬

### DIFF
--- a/modules/govuk/manifests/app/config.pp
+++ b/modules/govuk/manifests/app/config.pp
@@ -415,14 +415,6 @@ define govuk::app::config (
   }
 
   if $health_check_path != 'NOTSET' {
-    @@icinga::check { "check_app_${title}_up_on_${::hostname}":
-      ensure              => $ensure,
-      check_command       => "check_app_health!check_app_up!${port} ${health_check_path}",
-      service_description => "${title} app healthcheck",
-      host_name           => $::fqdn,
-      notes_url           => monitoring_docs_url(app-healthcheck-failed),
-      contact_groups      => $additional_check_contact_groups,
-    }
     if $json_health_check {
       include icinga::client::check_json_healthcheck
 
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/app/config.pp
```diff
@@ -415,14 +415,6 @@
...
 
   if $health_check_path != 'NOTSET' {
-    @@icinga::check { "check_app_${title}_up_on_${::hostname}":
-      ensure              => $ensure,
-      check_command       => "check_app_health!check_app_up!${port} ${health_check_path}",
-      service_description => "${title} app healthcheck",
-      host_name           => $::fqdn,
-      notes_url           => monitoring_docs_url(app-healthcheck-failed),
-      contact_groups      => $additional_check_contact_groups,
-    }
```
cbaines: With this change, there won't be any check for apps where `$json_health_check` is `false` if that makes sense, which is bad as having some form of healthcheck for each app is useful.

issyl0: Ah, I appear to have misunderstood the task. 😳I thought that every app had both of them, just we noticed it in email-alert-api first. It does indeed [appear that not all apps have `json_health_check` set to true](https://github.com/alphagov/govuk-puppet/search?q=json_health_check&unscoped_q=json_health_check). I wonder why not.

In light of this, should we:

- Set `json_health_check` to true by default (ie for every app) and do the work to all the apps to make that work (disclaimer: I don't know what that work would involve).
- Edit the `if $json_health_check` conditional to put the "app_name app healthcheck" in the `else` part. This would do the thing we want.

I'm leaning towards the latter.

issyl0: I went with the `else` option, with some re-wording of the commit message, in e8f631e5af51665292bd67e7e36ea0dd4ece626d.

cbaines: :+1:

PR #9768, opened 2019-10-31
Smooth out Icinga checks related to deployments

cbaines:
For apps running with Unicorn, a deployment involves spawning new
threads, which take up more memory, and then shutting down the old
threads once the new ones are serving requests. This can take up to 2
minutes for some apps.

To avoid Icinga alerting during deployments, increase the
attempts_before_hard_state setting to 3 for both the memory and thread
count checks. A value of one will mean Icinga alerts immediately and a
value of 3 will mean that Icinga will wait a minute then check again,
then wait another minute and check again, and the check will only
enter a hard state if this 3rd check fails.

This should avoid these checks failing temporarily during deployments,
but still mean they appear when the application is using an abnormal
amount of memory or threads.

### DIFF
--- a/modules/govuk/manifests/app/config.pp
+++ b/modules/govuk/manifests/app/config.pp
@@ -344,7 +344,7 @@ define govuk::app::config (
       host_name                  => $::fqdn,
       event_handler              => "govuk_app_high_memory!${title}",
       notes_url                  => monitoring_docs_url(high-memory-for-application),
-      attempts_before_hard_state => 2,
+      attempts_before_hard_state => 3,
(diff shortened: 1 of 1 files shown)


cbaines:
> Fantastic change tada

Thanks for taking a look Bill :)

> Do we also have an alert for when there are fewer threads than expected?

I don't believe so, but I think there are some checks that look at if processes are running (I don't remember the specifics).

Providing the app is still functioning, having too few threads is probably fine. We have in the past seen issues where the number of threads grew until a system limit was hit, and these checks help to identify if and when that's the case.

PR #9779, opened 2019-11-01
Check AWS IAM Key age

erino:
This alerts if AWS IAM Keys are older then 90 days. AWS Trusted Advisor recommends that IAM keys are rotated every 90 days, this alert will show keys that are older than this.

### DIFF
--- a/hieradata/integration.yaml
+++ b/hieradata/integration.yaml
@@ -170,6 +170,7 @@ monitoring::checks::aws_origin_domain: "integration.govuk.digital"
 monitoring::checks::sidekiq::enable_signon_check: false
 monitoring::checks::pingdom::enable: false
 monitoring::checks::ses::region: eu-west-1
+monitoring::checks::aws_iam_key::region: 'eu-west-1'
 monitoring::checks::smokey::environment: 'integration'
 monitoring::checks::smokey::disable_during_data_sync: true
 monitoring::uptime_collector::environment: 'integration'
--- a/modules/monitoring/manifests/checks/aws_iam_key.pp
+++ b/modules/monitoring/manifests/checks/aws_iam_key.pp
@@ -23,7 +23,7 @@ class monitoring::checks::aws_iam_key (
   $enabled = true,
 ) {
   if $enabled and $::aws_migration {
-    $max_age = 90
+    $max_age = 365
 
     icinga::plugin { 'check_aws_iam_key_age':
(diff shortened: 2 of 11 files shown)


modules/monitoring/templates/check_aws_iam_key_age.cfg.erb
```diff
@@ -0,0 +1,4 @@
+define command {
+    command_name check_aws_iam_key_age
+    command_line /usr/lib/nagios/plugins/check_aws_aws_iam_key_age --key <%= @access_key %> --secret <%= @secret_key %> --region <%= @region %> --days <%= @max_age %>
+}
```
cbaines: The trailing newline is missing for this file.

modules/monitoring/manifests/checks/aws_iam_key.pp
```diff
@@ -0,0 +1,47 @@
...
+#   Should we enable the monitoring check?
+#
+class monitoring::checks::aws_iam_key (
+  $region = undef,
+  $access_key = undef,
+  $secret_key = undef,
+  $enabled = true,
+) {
+  if $enabled and $::aws_migration {
+    $max_age = 365
```
cbaines: There's a greater possibility for confusion with the value here, and a default in the script. Could the value just be configured through changing the script?

erino: I wasn't sure the best way of doing this. I want to gradually decrease the max age as there are currently lots of keys that need rotation.

modules/monitoring/templates/check_aws_iam_key_age.cfg.erb
erino: Fixed in 8961079 via a05475b

hieradata/integration.yaml
```diff
@@ -170,6 +170,7 @@ monitoring::checks::aws_origin_domain: "integration.govuk.digital"
 monitoring::checks::sidekiq::enable_signon_check: false
 monitoring::checks::pingdom::enable: false
 monitoring::checks::ses::region: eu-west-1
+monitoring::checks::aws_iam_key: 'eu-west-1'
```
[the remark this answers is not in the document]
erino: Fixed in 99a9e86

modules/monitoring/manifests/checks/aws_iam_key.pp
cbaines: It'll work as is, this was just a minor though I had. So this isn't something that I think needs changing now.

PR #9780, opened 2019-11-01
Remove PagerDuty trigger for AWS Elasticache Redis

schmie:
- These alerts have been configured analogous to prior VM alerts, as
Redis is now hosted in AWS Elasticache, there is nothing for GOV.UK to
action.


issyl0:
I am extremely confused. These changes are exactly the same as https://github.com/alphagov/govuk-puppet/pull/9770 from this morning. Did you revert some things that wiped them out? What about other things that were merged before your reversions?

schmie:
Apologies - I did get distracted this morning and did not realize that you had actually made the change rather than providing an example.. Still am very confused why this did not blow up git-side though as well (guess a no-op is fine? :/).

PR #9785, opened 2019-11-05
Prepend "blue" to staging and production environment

bevanloon:
for links to the smokey job.

In Staging and Production AWS environments, the current url to navigate
directly to a job has to begin with `deploy.blue`.

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/smokey.pp
+++ b/modules/govuk_jenkins/manifests/jobs/smokey.pp
@@ -25,8 +25,8 @@ class govuk_jenkins::jobs::smokey (
     require => Class['govuk::apps::smokey'],
   }
 
-  if $::aws_migration {
-    $hosting_env_domain = "${::aws_environment}.govuk.digital"
+  if $::aws_migration and ($::aws_environment != 'integration') {
+    $hosting_env_domain = "blue.${::aws_environment}.govuk.digital"
   }
   else {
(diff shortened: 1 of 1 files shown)


issyl0:
Did we [lose some DNS](https://github.com/alphagov/govuk-aws/blob/master/doc/architecture/decisions/0015-govuk-aws-dns.jpg) along the way of the migration? If memory serves, <app>.<environment>.govuk.digital would go to whatever the "live" stack was, either `blue` or `green`. Or has it been simplified since, given we don't have multiple stacks?

sengi:
As-is, this creates a time-bomb for when we get around to cleaning up the remnants of the green/blue stacks thing. I'm happy to help fix this up so that we refer to the `<environment>.govuk.digital` names instead.

bevanloon:
@issyl0 I'm not sure. Without `blue` it redirects to jenkins dashboard.

@sengi yes it does. That'd be great if you could work some magic. thanks.

sengi:
I think I've found what's causing the unwanted redirect to the homepage — it's because for some reason we've got the `deploy.blue.staging.govuk.digital` hostname in the return URL in the GitHub single-signon config for this Jenkins.

The Jenkins itself seems happy to serve pages with the `deploy.staging.govuk.digital` hostname. It doesn't even have the `blue` hostname in its config; its `jenkinsUrl` is actually set to `deploy.staging.publishing.gov.uk`.

PR #9793, opened 2019-11-05
remove the `rds to s3` scripts on db_admin

fredericfran-gds:
disable the `rds to s3` scripts in staging and integragtion because these are not working, should use
govuk_env_sync. We don't remove it in production in case of dependencies
used by other tasks running on db_admin.

### DIFF
--- a/modules/govuk/manifests/node/s_db_admin.pp
+++ b/modules/govuk/manifests/node/s_db_admin.pp
@@ -31,7 +31,12 @@ class govuk::node::s_db_admin(
   include ::govuk::node::s_base
   include govuk_env_sync
 
-  if $backup_s3_bucket {
+  # disable the backups of rds to s3 because these are not working, should use
+  # govuk_env_sync. We don't remove it in production in case of dependencies
+  # used by other tasks running on db_admin.
+  if ($::aws_environment != 'production') {
+    $ensure = 'absent'
+  } elsif ($backup_s3_bucket) {
     $ensure = 'present'
   } else {
     $ensure = 'absent'


modules/govuk/manifests/node/s_db_admin.pp
```diff
@@ -31,7 +31,12 @@
   include ::govuk::node::s_base
   include govuk_env_sync
 
-  if $backup_s3_bucket {
+  # disable the backups of rds to s3 because these are not working, should use
+  # govuk_env_sync. We don't remove it in production in case of dependencies
+  # used by other tasks running on db_admin.
+  if ($::aws_environment != 'production') {
```
sengi: Might be clearer as:
```
if ($backup_s3_bucket and $::aws_environment == 'production') {
  $ensure = 'present'
} else {
  $ensure = 'absent'
}
```

fredericfran-gds: simpler but just a quick fix for now.

fredericfran-gds: I think you could even do `if ($::aws_environment == 'production' and $backup_s3_bucket)` if puppet does short-circuit evaluation

sengi: Ok, no worries. It would work either way around. I don't think short-circuit evaluation comes into it — it's just operator precedence. The `==` gets evaluated before the `and` no matter which appears first.

fredericfran-gds: I need to check. My book and stackoverflow (https://stackoverflow.com/questions/5278402/operator-return-when-result-is-known) says that the first conditional stops if false for `and`. Maybe I misread the thing

PR #9794, opened 2019-11-06
Avoid using AWS-Carrenza/6DG VPN for app-app traffic in Staging.

sengi:
App-to-app traffic is not supposed to be using the VPN, but was found to
be. Fix this by removing /etc/hosts entries on machines in AWS. These
hosts entries were probably left over from testing during the previous
phase of the migration to AWS.

This is a remediation action from the 2019-10-31 partial outage
involving the AWS-Carrenza VPN.


sengi:
> can you clean up the aws hiera too:
> 
> https://github.com/alphagov/govuk-puppet/blob/08319fb0a9b73c1d4888d176c82685ca10742fa6/hieradata_aws/staging.yaml#L380

Sure. Was gonna do that in a follow-up, but it makes more sense to do it in the same PR. Done.

sengi:
Test results in Staging (after removing the hosts entries) :
* No new alerts in Icinga.
* No new Smokey test failures vs previous run from before the change.
* No unexpected traffic over the VPN while running Smokey.

PR #9803, opened 2019-11-07
Add an Icinga check for process file handles

cbaines:
Most processes don't use more than 300 file handles as recorded in
Graphite, so set a default threshold for all apps of 500.

This is motivated partly by recent issues with the Router using more
file handles. I believe there is a limit of 1000 at the moment, so
it's important to get an early warning before that is hit.


cbaines:
> We should probably put together some docs about what to do as well.

Yep.

PR #9808, opened 2019-11-07
Send staging emails to Amazon SES

oscarwyatt:
Removes whitelist only mode for staging, thus emails
in staging environment will be sent to Amazon SES, allowing
us to try changes to email-alert-api in staging with
realistic timings and get a better idea of what a load on
staging does to metrics


oscarwyatt:
Thanks @thomasleese (and further thanks for the hint re: trial mode) - as far as I can tell, this would cost about $100 a day in Amazon SES so it's not a good use of money to enable this permanently, I think it's better to just SSH into the boxes and change the environment variable on an as needed basis.

PR #9820, opened 2019-11-11
Link directly to AWS IAM Key rotation documentation

erino:


modules/monitoring/manifests/checks/aws_iam_key.pp
```diff
@@ -46,7 +46,7 @@
       host_name           => $::fqdn,
       check_interval      => 12h,
       service_description => "At least 1 AWS IAM Key is older then ${max_aws_iam_key_age} days and needs rotating",
-      notes_url           => monitoring_docs_url(rotate-old-aws-iam-key),
+      notes_url           => "https://docs.publishing.service.gov.uk/manual/aws-iam-key-rotation.html"
```
cbaines: Puppet will be failing as the `,` is missing from the end here

PR #9830, opened 2019-11-14
adjust the govuk_env_sync for licensify in AWS

fredericfran-gds:
Changes made:

1. stop backing up integration licensify db because there is no training environment to pull from anymore.

2. start pushing the staging licensify db which for some reason was not being done. This will ensure integration is up to date because it pulls from the staging s3 bucket

### DIFF
--- a/hieradata_aws/class/staging/db_admin.yaml
+++ b/hieradata_aws/class/staging/db_admin.yaml
@@ -117,6 +117,22 @@ govuk_env_sync::tasks:
   "pull_licensify_refdata_production_daily":
     <<: *pull_licensify
     database: "licensify-refdata"
+  "push_licensify_staging_daily": &push_licensify_staging
+    ensure: "present"
+    hour: "5"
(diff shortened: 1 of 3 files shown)


hieradata_aws/class/staging/db_admin.yaml
```diff
@@ -117,6 +117,25 @@ govuk_env_sync::tasks:
...
+    hour: "5"
+    minute: "15"
+    action: "push"
+    dbms: "documentdb"
+    storagebackend: "s3"
+    database: "licensify"
+    temppath: "/tmp/licensify_staging"
+    url: "govuk-staging-database-backups"
+    path: "mongo-licensing"
+  "push_licensify_audit_staging_daily":
```
sengi: I'm not sure it's worth pushing the audit table, since it doesn't affect functionality as far as I'm aware. Might actually be better not to, since we don't really want all that PII in Integration.

PR #9845, opened 2019-11-18
Scope email-alert-api checks to Carrenza

benthorner:
We're currently getting an 'UNKNOWN' alert about this check in
AWS production. The production migration for email-alert-api isn't
imminent, so it seems best to scope this check for the time being.


benthorner:
Closing in favour of https://github.com/alphagov/govuk-puppet/pull/9846

PR #9854, opened 2019-11-20
Rewrite the Email Alert API Metrics dashboard

benthorner:
https://trello.com/c/NbaCXnvO/144-review-the-email-alert-traffic-dashboard

This rewrites the current dashboard to give more emphasis on visualising
the work being done at different stages of the sending process. Variants
of the same metrics, such as success and failure, are now shown in the
same graph, with a single row of graphs representing each metric.

screencapture-grafana-publishing-service-gov-uk-dashboard-db-email-alert-api-metrics-iteration-2019-11-28-11_23_56

### DIFF
--- a/modules/grafana/files/dashboards/email_alert_api.json
+++ b/modules/grafana/files/dashboards/email_alert_api.json
@@ -10,56 +10,43 @@
         "name": "Email Alert API Deploys",
         "showIn": 0,
         "tags": "deploys email-alert-api",
-        "type": "alert"
+        "target": "",
+        "type": "event"
       },
       {
         "datasource": "Graphite",
         "enable": true,
         "hide": false,
-        "iconColor": "rgb(231, 255, 96)",
+        "iconColor": "rgba(255, 96, 96, 1)",
         "limit": 100,
         "name": "Email Alert API Restarts",
         "showIn": 0,
         "target": "substr(stats.govuk.app.email-alert-api.restarts,3,5)",
         "type": "alert"
-      },
-      {
-        "datasource": "Graphite",
-        "enable": true,
-        "hide": false,
-        "iconColor": "rgb(96, 177, 255)",
-        "limit": 100,
-        "name": "Email Alert Service Deploys",
-        "showIn": 0,
-        "tags": "deploys email-alert-service",
-        "type": "alert"
-      },
-      {
-        "datasource": "Graphite",
-        "enable": true,
-        "hide": false,
-        "iconColor": "rgb(96, 255, 107)",
-        "limit": 100,
-        "name": "Email Alert Service Restarts",
-        "showIn": 0,
-        "target": "substr(stats.govuk.app.email-alert-service.restarts,3,5)",
-        "type": "alert"
       }
     ]
   },
-  "editMode": false,
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
   "hideControls": false,
-  "id": 34,
+  "id": 141,
   "links": [],
-  "refresh": "10s",
+  "refresh": "1m",
   "rows": [
     {
       "collapse": false,
-      "height": "280",
+      "height": 239,
       "panels": [
+        {
+          "content": "   - [Sidekiq](/dashboard/file/sidekiq.json?refresh=1m&orgId=1&var-Application=email-alert-api&var-Queues=All&from=now-3h&to=now)\n   - [Machine stats](/dashboard/file/machine.json?refresh=1m&orgId=1&var-hostname=email-alert-api*&var-cpmetrics=cpu-system&var-cpmetrics=cpu-user&var-filesystem=All&var-disk=All&var-tcpconnslocal=All&var-tcpconnsremote=All)",
+          "id": 7,
+          "links": [],
+          "mode": "markdown",
+          "span": 2,
+          "title": "Useful Links",
+          "type": "text"
+        },
         {
           "cacheTimeout": null,
           "colorBackground": false,
@@ -78,7 +65,7 @@
             "thresholdLabels": false,
             "thresholdMarkers": true
           },
-          "id": 2,
+          "id": 15,
           "interval": null,
           "links": [],
           "mappingType": 1,
@@ -106,7 +93,7 @@
               "to": "null"
             }
           ],
-          "span": 4,
+          "span": 2,
           "sparkline": {
             "fillColor": "rgba(31, 118, 189, 0.18)",
             "full": false,
@@ -118,7 +105,7 @@
             {
               "refId": "A",
               "target": "groupByNode(stats.govuk.app.email-alert-api.*.notify.email_send_request.*, 1, \"sumSeries\")",
-              "textEditor": true
+              "textEditor": false
             }
           ],
           "thresholds": "300,360",
@@ -141,7 +128,7 @@
           "dashes": false,
           "datasource": "Graphite",
           "fill": 1,
-          "id": 1,
+          "id": 13,
           "legend": {
             "avg": false,
             "current": false,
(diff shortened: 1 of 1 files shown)


modules/grafana/files/dashboards/email_alert_api.json
```diff
@@ -1,44 +1,95 @@
...
+      "name": "Graphite",
+      "version": "1.0.0"
+    },
+    {
+      "type": "panel",
+      "id": "text",
+      "name": "Text",
+      "version": ""
+    }
+  ],
```
cbaines: I'm not sure if you definitely need to remove these sections from the exported JSON, but I've been removing them.

modules/grafana/files/dashboards/email_alert_api.json
```diff
@@ -1,44 +1,95 @@
...
   "annotations": {
     "list": [
       {
-        "datasource": "Graphite",
+        "datasource": "${DS_GRAPHITE}",
```
cbaines: As we're not importing the dashboard in to Grafana, but just putting it on the disk, this needs to be the name of the Graphite datasource (which is `Graphite`) in this case, rather than the template style thing.

kevindew:
Thanks for working on this Ben 👍 At a glance I think there's a bunch of improvements here but also a few things I'd consider regressions. It does all feel a bit more of a complete re-write than an iterate as the PR titles it.

What we seem to have lost immediately is the ability to know if we're sending a lot of emails a lot or not at the moment you're looking at the graph. Which is something that was quite well encaptured by the gauge and the Notify email send requests a second. That graph may present undue precedence towards rate limits (as that was a key concern when creating the graph that hasn't transpired) but I still think it's really useful info to know if we're sending a lot of emails or not.

Then the other main thing that we've lost is the whole content change to attempt time, which I actually think has become one of the most important metrics we show in recent times as this has been the only one that has been useful to see when there is an end-to-end problem in the system. So I'm really keen we keep that one (and expand it to messages) and the one for email create to notify first attempt. I don't think we gain a lot from the digest ones given the scale of the board normally so they seem fine to go. I also don't think we've ever learnt much from timing how long it takes to send an email - though it is quite surprising that it regularly takes over 10s :-/

It'd also be useful to have that email sent per day volume graph. That was really useful recently in working out that we'd hit new records in the amount of email we've sent per day.

I don't think we gain a lot with content changes per second as they come in much less frequently than per second - and we don't learn a lot with numbers like 0.2. I think the previous graph had them grouped to only when it's whole numbers and I don't find that one very valuable as it's just 1s and 2s, so that'd probably benefit from grouping.

For content change duration should we also have min and max numbers?

For both of those content change sections should we also include messages?

I don't think we get a lot from the digest runs per second graph since the scale makes it near meaningless, the 7 day one looks a better way to read the data. Though I'm confused that total duration is much lower than the non-total time?

On the status updates from notify it looks like we've switched to using logit as a datasource rather than graphite, whereas graphite would be better if we have the data as it is persisted longer.

You might want to only show slow requests in the request duration, I'm not sure but it looks like there is all of them there.

modules/grafana/files/dashboards/email_alert_api.json
benthorner: Ah, that's interesting - thanks for spotting. I'll fix that when we're happy with the overall change.

benthorner:
@kevindew thanks for the suggestions. I agree this is more of a rewrite. Unfortunately I've never found the current dashboard helpful when emails aren't working. I hope we can get to something that works for everyone. The crucial part should be agreeing on what data we need. I've updated the PR/screenshot and tried to respond to each of your comments below. Hopefully I've addressed some of them, but perhaps we could draw on a printout (or something) if there's still lots to discuss.

> Which is something that was quite well encaptured by the gauge and the Notify email send requests a second.

I think this is captured (and then some) by the "Notify Requests (per second)" graph. Since this graph shows requests over time, I'd say it's more effective at communicating volume, since it provides a view over time.

> Then the other main thing that we've lost is the whole content change to attempt time

I've struggled to understand the use case behind this and similar graphs. I agree this graph can tell you there's an issue *somewhere*, but I'm left wondering *where*? I've found this is distracting when trying to diagnose a problem. I think it would be really useful to instead have a graph of the underlying problem this graph is trying to communicate. 

From recent experience, I think the underlying problem might be the (high volume of) hidden work of the `ImmediateEmailGenerationWorker`, which I'm looking to expose. Perhaps an immediate compromise is to have a row for 'email generation' that we iterate, which initially has the graph you mentioned and maybe another to show there are `Immediate...` jobs in progress.

How does that sound?

> It'd also be useful to have that email sent per day volume graph.

Ah, so that's what that graph is for. I wonder if that's useful from a diagnostic perspective, which is what I'm trying to optimise for. If the volume were an issue, would that be shown in another graph? 

> I don't think we gain a lot with content changes per second

The idea here is to try and show the work progressing through the system in each row. I've had a go at changing the graph display the bucketed total amounts, which is more intuitive 👍.

> I don't think we get a lot from the digest runs per second

I agree about the digest runs - they could just be in a hidden row somewhere. The "total processing time" alias is also misleading - fixed. I've had a go at changing the graph to better represent a daily/weekly unit of work using dots, reserving the axes for the number of emails generated.

> On the status updates from notify it looks like we've switched to using logit as a datasource

I saw this as an improvement over the current dashboard since these graphs provide overall visibility of the request activity. One of the questions I've had with the current dashboard is: *what else is going on that I'm not seeing here?*. This seems better from a diagnostic point of view, where the volume of requests may be relevant to an issue.

The legend of the graph also allows you to filter individual requests and sort by the slowest ones, although those should be self-evident as the highest bars on the graph.

benthorner:
@kevindew as per our discussion, I've made a few tweaks and updated the screenshot. Could you take another look?

kevindew:
Great thanks Ben

I'm a bit confused by the numbers in the "Content Changes, Messages" graph. I assumed it was a count but quite a lot less than 1?

Looks like there is a closed section on digest runs is that worth opening for the screenshot?

benthorner:
@kevindew another good spot. After struggling again to understand our metric namespace, I've finally located where the [actual counter numbers go](https://github.com/statsd/statsd/blob/master/backends/graphite.js#L188) and switched everything away from rates.

How does it look now?

modules/grafana/files/dashboards/email_alert_api.json
```diff
@@ -10,56 +10,43 @@
...
-  "refresh": "10s",
+  "refresh": "1m",
   "rows": [
     {
       "collapse": false,
-      "height": "280",
+      "height": 239,
       "panels": [
+        {
+          "content": "   - [Sidekiq](https://grafana.publishing.service.gov.uk/dashboard/file/sidekiq.json?refresh=1m&orgId=1&var-Application=email-alert-api&var-Queues=All&from=now-3h&to=now)\n   - [Machine stats](https://grafana.publishing.service.gov.uk/dashboard/file/machine.json?refresh=1m&orgId=1&var-hostname=email-alert-api*&var-cpmetrics=cpu-system&var-cpmetrics=cpu-user&var-filesystem=All&var-disk=All&var-tcpconnslocal=All&var-tcpconnsremote=All)",
```
kevindew: Do we have precedent for this as production carrenza only links? could we just have absolute paths for portability?

benthorner: Another good spot - they're relative now 😸

thomasleese:
I think we're still missing a few things from the new dashboard which were useful from the old one.

Unless I'm missing something, I think we've lost the following:

**Number of emails we've sent today**

We used to have a quick way of seeing how many content changes, emails, status updates have happened in the day (by changing the time period in Grafana to "Today"). I found this useful during the most recent incident to know whether we needed to increase the rate limit with Notify.

**Long term volume**

We seem to have lost of way of quickly getting a long-term picture on how many emails we're sending out and how things the volume has changed over a few months. I found this metric quite useful and our product manager also used it to understand changes to the system.

**Time to request email delivery**

This was useful to know whether there was a problem on Notify's side. During Notify incidents in the past this number has gone up telling us where to look for a problem, I'm pretty sure I've used this graph a few times and find it useful to have.

Apart from these three, I think the new dashboard is a good improvement on the old one and focuses on the most important metrics.

benthorner:
@thomasleese some of these should still be available in the new dashboard. I forgot to come and talk to you before this got merged, and we can still do that if you like.

> Number of emails we've sent today

This one is missing in the new dashboard. From a diagnostic point of view, a single number didn't seem to add much value. I'm curious that we were using the total number of emails as an indicator for rate limiting. Would that still be covered by [the panel in the top row](https://grafana.publishing.service.gov.uk/dashboard/file/email_alert_api.json?refresh=1m&orgId=1&from=now-3h&to=now&panelId=15&fullscreen)?

> Long term volume

I think you can still get a pretty good view of this in the [Notify Email Send Requests](https://grafana.publishing.service.gov.uk/dashboard/file/email_alert_api.json?refresh=1m&panelId=6&fullscreen&orgId=1&from=now-30d&to=now) graph. The idea was to have a single graph to represent different cross-sections of the system. I found the volume and variety of the original dashboard quite overwhelming for diagnostics. Perhaps this is a case where a separate, product-oriented dashboard would be more useful?

> Time to request email delivery

We do still have [this one](https://grafana.publishing.service.gov.uk/dashboard/file/email_alert_api.json?refresh=1m&orgId=1&from=now-3h&to=now&panelId=5&fullscreen), albeit renamed. The 'max' line on the original graph always seemed to be alarming by a couple of orders of magnitude, so the new graph just focusses on the mean.

thomasleese:
> I forgot to come and talk to you before this got merged, and we can still do that if you like.

Happy to have a chat!

> This one is missing in the new dashboard. From a diagnostic point of view, a single number didn't seem to add much value. I'm curious that we were using the total number of emails as an indicator for rate limiting. Would that still be covered by [the panel in the top row](https://grafana.publishing.service.gov.uk/dashboard/file/email_alert_api.json?refresh=1m&orgId=1&from=now-3h&to=now&panelId=15&fullscreen)?

We have a minute and daily rate limit with Notify. The total number of emails sent today was useful to compare against the daily limit.

> I think you can still get a pretty good view of this in the [Notify Email Send Requests](https://grafana.publishing.service.gov.uk/dashboard/file/email_alert_api.json?refresh=1m&panelId=6&fullscreen&orgId=1&from=now-30d&to=now) graph. The idea was to have a single graph to represent different cross-sections of the system. I found the volume and variety of the original dashboard quite overwhelming for diagnostics. Perhaps this is a case where a separate, product-oriented dashboard would be more useful?

Perhaps a more product-oriented dashboard would be more useful in this case. Certainly the long term volume graph is still being used, we adds the rolling average line very recently: https://github.com/alphagov/govuk-puppet/pull/9869

I've also just noticed (now that the dashboard is deployed to production) that the **Content Changes, Messages** graphs don't appear to be working. I am assuming that they would cover the case of being able to see content changes coming in (very useful to test that email-alert-service was working correctly and to see peaks at 9:30).

PR #9855, opened 2019-11-20
Include statsd::repo unconditionally

emmabeynon:
Trello: https://trello.com/c/GeQ5dfid/229-upgrade-statsd-and-node

We want all machines to be able to use statsd::repo so this removes the
condition around it.

### DIFF
--- a/modules/statsd/manifests/init.pp
+++ b/modules/statsd/manifests/init.pp
@@ -19,7 +19,7 @@ class statsd(
     require => Class['nodejs'],
   }
 
-  file { '/etc/statsd.conf':
+  file { '/etc/statsd/config.js':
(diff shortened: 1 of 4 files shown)


modules/statsd/manifests/init.pp
```diff
@@ -19,12 +19,7 @@
   validate_bool($manage_repo_class)
 
   include nodejs
-
-  if $manage_repo_class {
```
cbaines: If the use of this conditional is being removed, it would be good to also remove the parameter, and any use of that in the hieradata as well.

emmabeynon: Good shout, done 👍

cbaines: Wonderful :)

PR #9856, opened 2019-11-20
Iterate dashboard for Sidekiq metrics

benthorner:
https://trello.com/c/NbaCXnvO/144-review-the-email-alert-traffic-dashboard

This iterates the Sidekiq dashboard to make the rendering on the graphs
clearer, using thicker fills instead of tiny points and lines.

This also adds a new graph to show failures for a particular worker,
which wasn't covered by the original dashboard.

screencapture-grafana-publishing-service-gov-uk-dashboard-db-sidekiq-iteration-2019-11-29-12_23_00

### DIFF
--- a/modules/grafana/files/dashboards/sidekiq.json
+++ b/modules/grafana/files/dashboards/sidekiq.json
@@ -1,38 +1,18 @@
 {
-  "__requires": [
-    {
-      "type": "grafana",
-      "id": "grafana",
-      "name": "Grafana",
-      "version": "4.5.2"
-    },
-    {
-      "type": "panel",
-      "id": "graph",
-      "name": "Graph",
-      "version": ""
-    },
-    {
-      "type": "datasource",
-      "id": "graphite",
-      "name": "Graphite",
-      "version": "1.0.0"
-    }
-  ],
   "annotations": {
     "list": []
   },
   "editable": true,
   "gnetId": null,
-  "graphTooltip": 0,
+  "graphTooltip": 1,
   "hideControls": false,
-  "id": null,
+  "id": 142,
   "links": [],
   "refresh": "1m",
   "rows": [
     {
       "collapse": false,
-      "height": "200",
+      "height": 384,
       "panels": [
         {
           "aliasColors": {},
@@ -40,26 +20,25 @@
           "dashLength": 10,
           "dashes": false,
           "datasource": "Graphite",
-          "editable": true,
-          "error": false,
-          "fill": 1,
-          "grid": {},
-          "id": 1,
+          "fill": 2,
+          "id": 7,
           "legend": {
-            "avg": false,
-            "current": false,
-            "hideEmpty": true,
-            "hideZero": true,
-            "max": false,
+            "alignAsTable": true,
+            "avg": true,
+            "current": true,
+            "max": true,
             "min": false,
             "show": true,
-            "total": false,
-            "values": false
+            "sort": "total",
+            "sortDesc": true,
+            "total": true,
+            "values": true
           },
           "lines": true,
           "linewidth": 2,
           "links": [],
-          "nullPointMode": "null",
+          "maxDataPoints": "",
+          "nullPointMode": "null as zero",
           "percentage": false,
           "pointradius": 1,
           "points": true,
@@ -71,20 +50,20 @@
           "steppedLine": false,
           "targets": [
             {
+              "hide": false,
               "refId": "A",
-              "target": "aliasByNode(averageSeriesWithWildcards(stats.gauges.govuk.app.$Application.*.workers.queues.$Queues.enqueued, max, 5), 7)",
+              "target": "aliasByNode(summarize(consolidateBy(sumSeriesWithWildcards(stats_counts.govuk.app.$Application.*.workers.*.success, 4), 'sum'), '$Aggregation', 'sum', false), 5)",
               "textEditor": false
             }
           ],
           "thresholds": [],
           "timeFrom": null,
           "timeShift": null,
-          "title": "Queue Length",
+          "title": "Jobs Processed (total over $Aggregation)",
           "tooltip": {
-            "msResolution": true,
-            "shared": true,
+            "shared": false,
             "sort": 2,
-            "value_type": "cumulative"
+            "value_type": "individual"
           },
           "type": "graph",
           "xaxis": {
@@ -96,12 +75,11 @@
           },
           "yaxes": [
             {
-              "decimals": 0,
               "format": "short",
-              "label": null,
+              "label": "",
               "logBase": 1,
               "max": null,
-              "min": 0,
+              "min": null,
               "show": true
             },
             {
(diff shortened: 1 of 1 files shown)


cbaines:
I'll review just as a comment, as commenting on the specific JSON isn't particularly useful.

Do you know what the units are for the "Jobs processed" panel? I guess this is a combination of the previous "Jobs processed (per hour)" and "Jobs processed (per second)" panels.

When designing the dashboard, I kept all of the panels vertically aligned in a single column, as that makes it easy to see the values on each panel at a specific time, as well as comparing values between panels at specific times. My concern with having panels like the queue latency and jobs processed aligned differently, is that it'll be harder to line the data on them up if that makes sense?

Regarding the retry set size and the failures per second, I'm not sure what the different colours represent?

For the Application template variable, I can see how it might be useful to look at Sidekiq metrics across all apps, but I'm wondering if that might be something best done on a different dashboard. In particular, it's unclear which application the metrics on the "Jobs processed" panel correspond to, and I'm unsure how useful the process metrics will be when looking at all applications.

benthorner:
@cbaines thanks for the comments! I've made a few changes as per the comments below...

> Do you know what the units are for the "Jobs processed" panel?

I thought I did, but now I realise using timer counts from statsd is not a great idea. This was to try and be consistent with my (also incorrect) usage of timer counts elsewhere. I've reverted it to use the actual counter, and clarified the interpretation in the title.

> When designing the dashboard, I kept all of the panels vertically aligned in a single column

I agree it's useful to visually match similar points in time on a graph. The approach I've taken here is to try and group similar graphs in the same row, which also provides more vertical space for the new graph of failures. To try and accommodate the visual matching, I've changed the dashboard to use a shared crosshair, which should give us more flexibility layout-wise.

> Regarding the retry set size and the failures per second, I'm not sure what the different colours represent?

Oops. I think this is related to the 'All' template variable. On reflection, I agree it's not something we should support on this dashboard. To be clear, I've added a (now self-evident) legend on both of these graphs, including summary stats.

benthorner:
@cbaines do you have any more thoughts on this?

modules/grafana/files/dashboards/sidekiq.json
```diff
@@ -71,20 +80,20 @@
           "steppedLine": false,
           "targets": [
             {
-              "refId": "A",
-              "target": "aliasByNode(averageSeriesWithWildcards(stats.gauges.govuk.app.$Application.*.workers.queues.$Queues.enqueued, max, 5), 7)",
+              "hide": false,
+              "refId": "B",
+              "target": "aliasByNode(averageSeriesWithWildcards(stats.gauges.govuk.app.$Application.*.workers.queues.$Queues.enqueued, 5), 4, 7)",
```
cbaines: The title of this panel was changed from "Queue Length" to "Jobs Processed", but this Graphite query will get the enqueued jobs, not a rate.

modules/grafana/files/dashboards/sidekiq.json
```diff
@@ -24,45 +34,44 @@
...
           "bars": false,
           "dashLength": 10,
           "dashes": false,
-          "datasource": "Graphite",
-          "editable": true,
-          "error": false,
-          "fill": 1,
-          "grid": {},
-          "id": 1,
+          "datasource": "${DS_GRAPHITE}",
```
cbaines: I think this'll need to be kept as Graphite.

cbaines:
> > Do you know what the units are for the "Jobs processed" panel?
> 
> I thought I did, but now I realise using timer counts from statsd is not a great idea. This was to try and be consistent with my (also incorrect) usage of timer counts elsewhere. I've reverted it to use the actual counter, and clarified the interpretation in the title.

At least currently, it looks like the "Jobs Processed (Per Second)" just shows the queue length. I've commented on the query specifically.

> > When designing the dashboard, I kept all of the panels vertically aligned in a single column
> 
> I agree it's useful to visually match similar points in time on a graph. The approach I've taken here is to try and group similar graphs in the same row, which also provides more vertical space for the new graph of failures. To try and accommodate the visual matching, I've changed the dashboard to use a shared crosshair, which should give us more flexibility layout-wise.

I haven't used the shared crosshair, but that's something I'm open to trying.

> > Regarding the retry set size and the failures per second, I'm not sure what the different colours represent?
> 
> Oops. I think this is related to the 'All' template variable. On reflection, I agree it's not something we should support on this dashboard. To be clear, I've added a (now self-evident) legend on both of these graphs, including summary stats.

:+1:

benthorner:
Thanks @cbaines - I'm not sure how I got that graph so wrong! In fixing it to use an actual counter, I changed it to show individual jobs, since queues don't seem to be available as a simple count.

In making this change to show both jobs and queues, I removed the `$Queue` template variable, which no longer covers every panel. In practice, there aren't many queues to filter anyway.

And I've fixed the datasource/annotation weirdnesses - turns out Grafana generates slightly different JSON depending on how you export the dashboard.

Updated screenshot - how does it look now?

modules/grafana/files/dashboards/sidekiq.json
```diff
@@ -71,20 +50,20 @@
           "steppedLine": false,
           "targets": [
             {
+              "hide": false,
               "refId": "A",
-              "target": "aliasByNode(averageSeriesWithWildcards(stats.gauges.govuk.app.$Application.*.workers.queues.$Queues.enqueued, max, 5), 7)",
+              "target": "aliasByNode(summarize(averageSeriesWithWildcards(stats_counts.govuk.app.$Application.*.workers.*.success, 4), '1min', 'sum', false), 5)",
```
cbaines: I think more information needs to be given about what this panel is showing, either in the panel title, or on the y axis.

My interpretation of this query is that it shows the average number of jobs processed in a minute, per machine.

modules/grafana/files/dashboards/sidekiq.json
```diff
@@ -305,63 +268,66 @@
...
             {
               "refId": "A",
-              "target": "aliasByNode(perSecond(integral(stats.timers.*.app.$Application.*.workers.*.processing_time.count)), 5, 7)",
-              "textEditor": true
-            },
-            {
-              "refId": "B",
-              "target": "aliasByNode(perSecond(integral(stats.timers.*.app.$Application.*.workers.*.*.processing_time.count)), 5, 7, 8)",
-              "textEditor": true
+              "target": "aliasByNode(transformNull(averageSeriesWithWildcards(stats.gauges.govuk.app.$Application.*.workers.retry_set_size, max, 5), 0), 4)",
```
cbaines: I think using `transformNull` here leads to a very misleading graph, as it effectively adds new data points with a value of zero, and then these are averaged with the actual data.

cbaines:
> Updated screenshot - how does it look now?

I've made some other comments, but looking at the styling specifically. The general approach I've been taking with line fill is to normally just use it for showing ranges of a chart, like memory or disk usage on the Machine dashboard.

I think using line fill here, like on the Jobs Processed panel is making the chart harder to accurately read. Consider the two variants:

## A
image

## B
image

I'd say that variant A looks busier and more confusing, but it's also possible to see information that's obscured on variant B, like the series in orange.

While I included points in the screenshot, I'm unsure if they're that useful here, but I'd see them as something that is very useful on the queue length panel. Because Sidekiq only reports metrics when it's completing jobs, it's important to see where the individual data points are, as just looking at the lines can be deceiving.

modules/grafana/files/dashboards/sidekiq.json
benthorner: Good spot! That's me getting the query wrong - I've changed it to a summation, which should match the intention of the title. Does that help?

modules/grafana/files/dashboards/sidekiq.json
benthorner: Good point. It's also unnecessary for the way this graph is displayed. I've removed it now.

benthorner:
@cbaines thanks for looking again. I agree there's a risk of hiding data with a total fill display style - I went to an extreme there. I've updated the screenshot to show the latest style. Does that help?

modules/grafana/files/dashboards/sidekiq.json
cbaines: Yep, changing it to a summation is good.

Additionally, I think that specifying a label for the y axis would be good, maybe "Jobs per minute" or "Processed per minute", or something else to tell someone what the numbers mean, without having to look at the Graphite query.

cbaines:
> @cbaines thanks for looking again. I agree there's a risk of hiding data with a total fill display style - I went to an extreme there. I've updated the screenshot to show the latest style. Does that help?

Yep, that's looking better.

While it's maybe a rarer occurrence, for apps that use multiple queues at the same time, the same issue with the extreme fill affects the Queue length panel as well, so it might be good to make the same change there.

Something odd I've noticed as well, I think there are fractional values appearing on the Failures panel, which looks very odd as they appear as "0", but if you restrict the y-axis to between 0 and 1, you can see the fractional values...

cbaines:
> Something odd I've noticed as well, I think there are fractional values appearing on the Failures panel, which looks very odd as they appear as "0", but if you restrict the y-axis to between 0 and 1, you can see the fractional values...

I think I've figured it out actually, adding `consolidateBy(sum)` in to the mix looks to resolve the issue, which I think I understand...

cbaines:
Also, while I'd like to spend a bit more time looking at this, I think this new version is better than the existing dashboard now, so I'm confident these changes can be merged soon :)

benthorner:
@cbaines aha! I was struggling to get integer numbers using `summarize`. While I knew graphite was auto-averaging the results, I didn't know there was a way to override it. I've applied `consolidateBy` to the left-hand graphs to give more accurate results and similarly in the [other dashboard PR](https://github.com/alphagov/govuk-puppet/pull/9854) I'm working on.

Also updated the queue length display, as well as the retry set, which I think is slightly less visually taxing now.

cbaines:
> @cbaines aha! I was struggling to get integer numbers using `summarize`. While I knew graphite was auto-averaging the results, I didn't know there was a way to override it. I've applied `consolidateBy` to the left-hand graphs to give more accurate results and similarly in the [other dashboard PR](https://github.com/alphagov/govuk-puppet/pull/9854) I'm working on.

Great, that's looking good now.

> Also updated the queue length display, as well as the retry set, which I think is slightly less visually taxing now.

:+1: 

I've had another look, and the only thing I spotted that maybe would be good to change before merging is for the "Jobs Processed" panel again.

On the current version of the dashboard, the two similar panels do say what the numbers are (jobs processed per hour and per second), and the values on the panels should stay roughly the same if you zoom out or in. This isn't true for the new Jobs Processed panel though.

Previously, I saw it was using `summerize(sum, 1m,...)` which gave a fixed value, but now it isn't. Another approach would be to use an interval type template variable. That would allow you to summerize by that interval, but have Grafana automatically pick the interval based on the time range, there's an example of this on the "Sentry errors (across all environments)" dashboard.

benthorner:
@cbaines this is where the difference is more intentional. In order to get the graphs displaying in a way that's easy to visually parse, we need to do some kind of bucketing. For this dashboard, I've gone with limiting the datapoints, which your technique makes more flexible. But I think the end result is the same in both cases: the values can appear to change depending on the zoom level. I agree this is annoying, but often I find I'm only interested in the relative values, and not the absolutes, when I'm trying to diagnose an issue. This also affects the choice of axis label: it's difficult to say 'per X' when X can fluctuate. I'm hoping it's not a big issue for people, since it would be a shame to compromise on the display, even if it improves the precision.

I wonder if I'm missing something, though?

benthorner:
@cbaines following our discussion, I've amended with your idea of using the panel titles to communicate the buckets/aggregation/interval being applied.

I've hidden the template variable for it, as I don't think it's something people will want to change manually. The default intervals seemed to produce reasonable results.

modules/grafana/files/dashboards/sidekiq.json
```diff
@@ -208,70 +180,60 @@
...
-              "hide": false,
-              "refId": "B",
-              "target": "aliasByNode(summarize(perSecond(integral(sumSeriesWithWildcards(stats.timers.*.app.$Application.*.workers.*.processing_time.count, 9))), '1h', 'sum', false), 5, 7)",
-              "textEditor": true
-            },
-            {
-              "refId": "D",
-              "target": "aliasByNode(summarize(perSecond(integral(sumSeriesWithWildcards(stats.timers.*.app.$Application.*.workers.*.*.processing_time.count, 9, 10))), '1h', 'sum', false), 5, 7, 8)",
+              "refId": "A",
+              "target": "aliasByNode(summarize(averageSeriesWithWildcards(stats.gauges.govuk.app.$Application.*.workers.queues.*.enqueued, max, 5), '$Aggregation', 'max', false), 7)",
```
cbaines: I don't think there's any problem with using the `$Aggregation` here, but I don't think it's necessary either. Unlike the jobs processed panel, I think the default Graphite behavior works fine for the queue length.

cbaines:
> @cbaines following our discussion, I've amended with your idea of using the panel titles to communicate the buckets/aggregation/interval being applied.

Great, I think it could be useful. Also, I was only thinking about this in the context of the "Jobs Processed" panel. I think the default Graphite behavior is fine for the other panels, but using the explicit aggregation works as well.

> I've hidden the template variable for it, as I don't think it's something people will want to change manually. The default intervals seemed to produce reasonable results.

benthorner:
Thanks @cbaines! I think the same argument at least applies for the 'failures' graph, which is also a total/sum. I like that it gives the same level of certainty about the data being visualised.

PR #9862, opened 2019-11-21
Run Licensify backups from AWS instead of UKCloud.

sengi:
Now that Licensify Production has been moved from UKCloud to AWS, disable its nightly backup jobs in UKCloud and run them in AWS instead.


sengi:
> the `disabled` pull for licensify should be absent too

Why? We'll need it if we have to restore a backup.

PR #9866, opened 2019-11-22
development-vm/replication: Remove old replication scripts

issyl0:
- MapIt pulls an S3 bucket from a script within its own repo, so we can
  just do that. MapIt repliciation is on Platform Health's radar within
  GOV.UK Docker.
- The rest of these are superseded by govuk-docker's replication
  scripts.

This is spurred on by the realisation that none of this will work now that we've switched to the gds-cli as it relies on local, plain-text credentials. It would be a lot of work to update this for not much gain now that most of the apps are in `govuk-docker` and the dev VM is being (slowly) deprecated.


issyl0:
I think this is good to go now we've [made it possible to download data without importing it via Docker](https://github.com/alphagov/govuk-docker/pull/263) and [edited the docs to not refer to these scripts any more, but instead have instructions-ish for using govuk-docker-imported data](https://github.com/alphagov/govuk-developer-docs/pull/2136).

barrucadu:
:+1:

benthorner:
For reference: https://github.com/alphagov/govuk-developer-docs/pull/2129

PR #9867, opened 2019-11-22
Configure Tensorflow Serving docker container

laurentqro:
The aim here is to Puppet-ify running a TensorFlow Serving container, namely the following command:

```
docker run -t --rm -p 8501:8501 \
    -v "$EXPORT_PATH/$LATEST:/models/ltr/1" \
    -e MODEL_NAME=ltr \
    tensorflow/serving
```

@barrucadu Do you have any pointers as to how we can mount the volume?

Paired with @bilbof 

Card: https://trello.com/c/HaeILDIG/1166-add-dockerised-model-to-search-machines

### DIFF
--- a/modules/govuk_containers/manifests/tensorflow_serving.pp
+++ b/modules/govuk_containers/manifests/tensorflow_serving.pp
@@ -28,7 +28,7 @@ class govuk_containers::tensorflow_serving(
 
   file { '/tensorflow-models':
     ensure => directory,
-  }
+  } ->
 
   ::docker::run { 'tensorflow/serving':
     ports            => ['8501:8501'],
--- a/modules/govuk_containers/manifests/tensorflow_serving.pp
+++ b/modules/govuk_containers/manifests/tensorflow_serving.pp
@@ -10,13 +10,9 @@
 # [*image_version*]
 #   The docker image version to use.
 #
-# [*max_memory*]
-#   The maximum amount of memory Tensorflow Serving should allocate
-#
 class govuk_containers::tensorflow_serving(
   $image_name = 'tensorflow/serving',
   $image_version = 'latest',
-  $max_memory    = 1024,
   $model_name = 'ltr'
 ) {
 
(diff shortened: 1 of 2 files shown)


barrucadu:
> Do you have any pointers as to how we can mount the volume?

You can pass extra parameters to `::docker::run`, so I think the way to do it would be to pass `-v /some/directory/on/the/host:/models/ltr/1` with a resource to create the directory before starting the container:

```puppet
  file { '/ltr-model':
    ensure => 'directory',
  } ->

  ::docker::run { 'tensorflow/serving':
    ports            => ['8501:8501'],
    image            => $image_name,
    require          => Docker::Image[$image_name],
    extra_parameters => ['-P', '-v', '/ltr-model:/models/ltr/1'],
    command          => "-m ${max_memory} -e MODEL_NAME=${model_name}",
  }
```

modules/govuk_containers/manifests/tensorflow_serving.pp
```diff
@@ -0,0 +1,47 @@
...
+
+  ::docker::image { $image_name:
+    ensure    => 'present',
+    require   => Class['govuk_docker'],
+    image_tag => $image_version,
+  }
+
+  file { '/tensorflow-models':
+    ensure => directory,
+  }
```
barrucadu: There needs to be a `->` between this `file` and the `::docker::run` (like in my comment) to indicate that there's a dependency.

bilbof:
@laurentqro rebasing should get the build to pass: https://github.com/alphagov/govuk-puppet/commit/83b8923df138c5f9637c2ea9190ab6e650e5c0b7

modules/govuk_containers/manifests/tensorflow_serving.pp
```diff
@@ -0,0 +1,47 @@
...
+  file { '/tensorflow-models':
+    ensure => directory,
+  } ->
+
+  ::docker::run { 'tensorflow/serving':
+    ports            => ['8501:8501'],
+    image            => $image_name,
+    require          => Docker::Image[$image_name],
+    extra_parameters => ['-P', '-v', '/tensorflow-models:/models'],
+    command          => "-m ${max_memory} -e MODEL_NAME=${model_name}",
```
bilbof: It might be that these can go into `extra_parameters`, I am not sure they are being passed yet. I get `MODEL_NAME=model` in the docker container.

PR #9869, opened 2019-11-25
Add a 30 day average line to the volume graph on the email alert api …

MuriloDalRi:
…dashboard

Screenshot 2019-11-26 at 11 17 16

https://trello.com/c/bmyTV8WT/1560-2-add-a-rolling-30-day-average-line-to-the-volume-graph-on-the-email-alert-api-dashboard

### DIFF
--- a/modules/grafana/files/dashboards/email_alert_api.json
+++ b/modules/grafana/files/dashboards/email_alert_api.json
@@ -47,6 +47,7 @@
       }
     ]
   },
+  "editMode": false,
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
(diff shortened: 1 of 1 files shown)


thomasleese:
Does it make sense that the line is always above the bars? I would expect an average to be similar (or lower, if it's 30 days ago) to the peak of the bars, but I'm not really sure how moving average works...

MuriloDalRi:
> Does it make sense that the line is always above the bars? I would expect an average to be similar (or lower, if it's 30 days ago) to the peak of the bars, but I'm not really sure how moving average works...

I updated it to not be stacked now

PR #9874, opened 2019-11-26
Remove rabbitmq federation from carrenza to AWS in staging

rtrinque:
No longuer required following migration of publishing-api and
email alert in AWS.


issyl0:
I guess we still need this?

PR #9891, opened 2019-11-29
Condense similar application dashboards

benthorner:
This condenses some of the application dashboards into a single
'generic' dashboard with a template variable, the aim being to
eventually move out the remaining custom behaviours into tailored
dashboards for the few apps that need them:

   - Some apps, like static, benefit from showing errors of other
apps that dependend upon it
   - Some apps, like finder-frontend, show memcached stats with a
specific 'instance prefix'

The docs URLs also vary for some applications e.g. 'smartanswers'
vs. 'smart-answers'. We could remove this link if it's not useful.


benthorner:
Until we upgrade to Grafana 5.3+, the deploy/restart annotations won't work for the generic dashboard :-|.

benthorner:
Closing this as I don't think we're upgrading Grafana any time soon :-(.

PR #9894, opened 2019-11-29
Add infrequent Icinga alert for ad-hoc dashboards

benthorner:
This adds a new alert to check if there are any ad-hoc dashboards in
    the Grafana instance associated with the environment. Such dashboards
    aren't in version control, tend to be temporary, and accumulate over
    time. Having this alert will motivate us to clean them up - currently
    our dashboard landscape is quite confusing, because there are so many.

If we like this and want to merge, we should first actually cleanup all the old dashboards identified by this alert, so we don't show the alert unnecessarily.

- [x] fix monitoring -> grafana connectivity
- [ ] cleanup dashboards not in code
- [x] write documentation to go with alert

### DIFF
--- a/modules/monitoring/files/usr/lib/nagios/plugins/check_old_grafana_dashboards
+++ b/modules/monitoring/files/usr/lib/nagios/plugins/check_old_grafana_dashboards
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+while getopts ":h:" opt; do
+  case ${opt} in
+    h )
+      GRAFANA_HOST=$OPTARG
+      ;;
+  esac
+done
+
+DASHBOARDS=$(curl -s https://$GRAFANA_HOST/api/search | jq -c '.[]' | grep 'db/' | jq '.title')
+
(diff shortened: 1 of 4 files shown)


issyl0:
What severity is this alert going to show up as? Warning?

issyl0:
Alternative approach that doesn't add another alert for people to ignore because the dashboards it moans about aren't theirs: delete all dashboards that aren't in code after 7 days.

benthorner:
@issyl0 yeah, it would be a warning. I originally thought of doing a SlackBot, but I don't think that's possible because the bot needs access to all our Grafana instances. An alert was the next best thing people would see that wouldn't be destructive. I don't think we should delete dashboards without asking, since they may be needed temporarily e.g. for an ongoing incident.

Once the initial cleanup is done, I don't see this alert coming up very often. If it does, it seems reasonable to be alerted about a new dashboard, since Grafana is a shared space. I've some TODOs for this, one of which is to write a doc around the alert; we could recommend the dashboard be deleted unless someone 'claims' it with a PR i.e. there's a policy.

PR #9895, opened 2019-12-02
Support AWS on Whitehall SLIs dashboard

thomasleese:
In AWS the Whitehall Backend machines are called `whitehall_backend-*` whereas in Carrenza they are named `whitehall-backend-*`. I've changed the Graphite query to support both `-` and `_` as the separator character.

[Trello Card](https://trello.com/c/3sMDtJHP/1511-5-show-an-sli-how-available-is-whitehall-to-publishers)

### DIFF
--- a/modules/grafana/files/dashboards/whitehall-slis.json
+++ b/modules/grafana/files/dashboards/whitehall-slis.json
@@ -68,20 +68,20 @@
                 }
               ],
               "refId": "A",
-              "target": "hitcount(sumSeries(stats.whitehall-backend-*.nginx_logs.*.http_*), '1h')",
+              "target": "hitcount(sumSeries(stats.whitehall*backend-*.nginx_logs.*.http_*), '1h')",
               "textEditor": true,
               "timeField": "@timestamp"
             },
             {
               "hide": true,
               "refId": "B",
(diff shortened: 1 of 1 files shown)


thomasleese:
@issyl0 It's the fully qualified domain name of the machine. So for example in AWS it's `whitehall_backend-ip-10-1-4-133_eu-west-1_compute_internal` whereas in Carrenza it's `whitehall-backend-1_backend`. The Puppet class is defined as `whitehall_backend` so in terms of consistency, I think the AWS one is more correct (maybe that's why it was decided to go with that naming scheme after migrating to AWS, but I don't know the details on that decision).

In a lot of cases our machine classes are just single words, so this problem doesn't come up as much as we may think.

Having a look in Puppet I can see that in some cases we've added workarounds like this and in some cases we haven't (and the dashboards only work in Carrenza).

However, there are plenty of examples where we've had to add special things like this to support differences between Carrenza and AWS (Cog links in Icinga still don't work properly in AWS is another example) so I'm not sure if it's worth trying to find a nicer solution for this since we'll hopefully be all in AWS soon anyway and then it won't matter anymore.

issyl0:
@thomasleese Yeah, OK. At least it's consistently inconsistent. Roll on when we're fully in AWS and we can remove all these special cases, eh? :muscle:

PR #9897, opened 2019-12-02
Disable insecure SSH ciphers.

sengi:
The version of `sshd` in Ubuntu 14.04 is old enough that it doesn't
support disabling specific ciphers, so instead we specify an up-to-date
list of ciphers which are still ok.

This list is the default list of ciphers from the latest version of
OpenSSH. See https://man.openbsd.org/sshd_config#Ciphers

### DIFF
--- a/modules/ssh/templates/sshd_config.erb
+++ b/modules/ssh/templates/sshd_config.erb
@@ -7,6 +7,11 @@ UsePrivilegeSeparation          yes
 SyslogFacility                  AUTH
 LogLevel                        VERBOSE
 
+# Disable insecure ciphers. Ubuntu 14.04's sshd is so old that it doesn't
+# support removing ciphers from the list, so we have to specify the list
+# of ciphers which we still want.
+Ciphers                         chacha20-poly1305@openssh.com,aes128-ctr,aes192-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com
+
 LoginGraceTime                  120
 StrictModes                     yes
 


issyl0:
Ouch, I'm getting flashbacks to #8153 and then #8155. Those changes were an epic exercise in disaster recovery. Be careful! 😬

sengi:
> Ouch, I'm getting flashbacks to #8153 and then #8155. Those changes were an epic exercise in disaster recovery. Be careful! 😬

Haha, that does look painful. We're on a pretty antique version of `sshd` so our options are a bit limited. I tested this out on one box in Integration and it seemed happy (both `sshd -T` and actually restarting it with the new config and re-connecting).

sengi:
I think disabling Puppet agent on the jumpbox temporarily while rolling this out is probably a good idea. Once we've proved that it hasn't broken ssh elsewhere then we can re-enable Puppet on the jumpbox and let it apply there last.

sengi:
Verified with `nmap -Pn -p22 -n -sV --script ssh2-enum-algos $host`

Before:
```
|   encryption_algorithms (16)
|       aes128-ctr
|       aes192-ctr
|       aes256-ctr
|       arcfour256
|       arcfour128
|       aes128-gcm@openssh.com
|       aes256-gcm@openssh.com
|       chacha20-poly1305@openssh.com
|       aes128-cbc
|       3des-cbc
|       blowfish-cbc
|       cast128-cbc
|       aes192-cbc
|       aes256-cbc
|       arcfour
|       rijndael-cbc@lysator.liu.se
```

After:

```
|   encryption_algorithms (6)
|       chacha20-poly1305@openssh.com
|       aes128-ctr
|       aes192-ctr
|       aes256-ctr
|       aes128-gcm@openssh.com
|       aes256-gcm@openssh.com
```

PR #9914, opened 2019-12-04
Allow pentesters access to integration

erino:


erino:
@issyl0 is there anything else I need to add for this?

issyl0:
Integration? I thought the pentest was on staging?

erino:
Yes, and this is for access to CI

PR #9915, opened 2019-12-04
Rename Isabell Long => Issy Long in their user manifest

issyl0:
- This is a purely cosmetic change.
- There are no other changes required, but I'm gradually trying to
  deprecate "Isabell".
- My Google Account has been Issy for some time, so if people look for
  Isabell and don't find me then that could be confusing.
- My LDAP username is still `isabelllong`, because that's even more
  difficult to change.

### DIFF
--- a/modules/users/manifests/isabelllong.pp
+++ b/modules/users/manifests/isabelllong.pp
@@ -1,8 +1,8 @@
 # Creates the user isabelllong
 class users::isabelllong {
   govuk_user { 'isabelllong':
-    fullname => 'Isabell Long',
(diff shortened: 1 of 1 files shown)


issyl0:
Oh god!

```
  1) users::isabelllong should have a username of the correct form
     Failure/Error: expect(user[:name]).to eq(uname_from_full), "expected username '#{user[:name]}' to be '#{uname_from_full}' based on fullname and email"
       expected username 'isabelllong' to be 'issylong' based on fullname and email
     # ./modules/users/spec/classes/users_spec.rb:40:in `block (3 levels) in <top (required)>'
```

issyl0:
I will live with this.

PR #9923, opened 2019-12-06
Sync Publishing API Integration database nightly from AWS Prod.

sengi:
This fixes the data sync from Production to Integration for publishing-api after the migration to AWS.

The old sync job from Carrenza Prod needs to be removed separately (in the alphagov/env-sync-and-backup repo).

While we're there, remove the nightly backup job in Integration because it no longer serves any useful purpose.


sengi:
Ah, apparently there is a filtering script for publishing-api after all. This will need to wait for #9922 then. I'll add the `sanitise_publishing_api_production.sql` here once that PR is merged.

sengi:
> Ah, apparently there is a filtering script for publishing-api after all. This will need to wait for #9922 then. I'll add the `sanitise_publishing_api_production.sql` script here once that PR is merged.

Done now.

PR #9926, opened 2019-12-09
Add Jenkins job to republish organisation pages

deborahchua:
This is in case we need to prioritise republishing organisation pages if 
there is a delay due to a large publishing queue. It is currently set to run every 
15 minutes.

As we do not know if we need to use it yet, we can add it to the list of Jenkins
 jobs later.

Related PR:  https://github.com/alphagov/publishing-api/pull/1663

Trello card: https://trello.com/c/uzKUr4WT/1613-3-make-sure-organisation-pages-show-people-changes-quickly-during-reshuffle

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/publishing_api_republish_organisations.pp
+++ b/modules/govuk_jenkins/manifests/jobs/publishing_api_republish_organisations.pp
@@ -1,0 +1,15 @@
+# == Class: govuk_jenkins::jobs::publishing_api_republish_organisations
+#
+# Create a jenkins job to periodically run rake for the following tasks:
+# - represent_downstream:high_priority_document_type[:document_type]
+#
+# === Parameters:
+#
+
+class govuk_jenkins::jobs::publishing_api_republish_organisations{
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/manifests/jobs/publishing_api_republish_organisations.pp
```diff
@@ -0,0 +1,22 @@
...
+# - represent_downstream:high_priority_document_type[:document_type]
+#
+# === Parameters:
+#
+
+class govuk_jenkins::jobs::publishing_api_republish_organisations(
+  $app_domain = hiera('app_domain'),
+) {
+
+  $check_name = 'publishing-api-republish-organisations'
```
thomasleese: I don't think we need any of these variables as they are related to Icinga checks which we don't have for this job.

deborahchua: Thanks @thomasleese ! I have updated this now.

deborahchua:
> Looks good! I wonder how we should document how to enable the job... 🤔

Yeh I wondered about this too 🤔

PR #9928, opened 2019-12-09
Enable email technical and internal failure checks in AWS

issyl0:
- These were added in f6de137f75525438d467759a141a4f5276880e10 and disabled because email-alert-api wasn't in AWS in 0bcbe1a25ed81d41fc4e8e2aa8690465019e50a1.
- We want these now that email-alert-api is in AWS.
- We only need hieradata to override the default `true` (enabled) values.


issyl0:
Carrenza isn't getting any email-alert traffic so I don't think it matters. Hopefully soon (cc @alphagov/re-gov-uk) the Carrenza stuff will be cleaned up.

sengi:
Thanks for finding and fixing these. I've added a step to the migration plan so that we check for disabled alerts on the AWS side next time.

PR #9940, opened 2019-12-11
Remove govuk-content-schemas from backend boxes

kevindew:
Trello: https://trello.com/c/zHm2B8VX/1242-present-organisation-political-status-from-whitehall-to-publishing-api

These files are removed from backend machines to get around a problem. Ever since the migration of Publishing API to AWS we have not been able to deploy govuk-content-schemas since they are [deployed to backend and publishing-api](https://github.com/alphagov/govuk-app-deployment/blob/c1d210e37b402df4377e5146d31bfb433f78dc65/govuk-content-schemas/config/deploy.rb#L3) machine classes and this will fail when trying in a carrenza environment as the publishing-api machines aren't available which causes problems and confusion.

From my research no applications other than Publishing API actually use the schemas in production and I believe the only reason we put these on both backend and publishing-api boxes is to facilitate Publishing API being migrated between these machines.

Things I've done to check usage of this:

- Check Gemfiles of apps: I ran `find ~/govuk -name Gemfile | xargs grep 'govuk_schemas' -B 5` and could see the gem was only depended outside a test environment in Publishing API and govuk-content-schemas
- Check end-to-end tests: the govuk-content-schemas mount is only used by Publishing API, Mainstream Publisher and Content Store. Only Mainstream Publisher is on backend which only uses [schemas for tests](https://github.com/alphagov/publisher/blame/a0e19e84fd3f63fba17913c4fc255e0f8ea453cd/README.md#L54)

@cbaines and @thomasleese can you think of anything I might have missed or any reasons why govuk-content-schemas are required on backend boxes?

/cc @sengi

### DIFF
--- a/modules/govuk/manifests/apps/govuk_content_schemas.pp
+++ b/modules/govuk/manifests/apps/govuk_content_schemas.pp
@@ -4,17 +4,31 @@
 #
 # === Parameters
 #
+# [*ensure*]
+#   Allow schemas to be removed.
+#
 # [*directory*]
 #   The directory in which the govuk_content_schemas should be
-#   available.
+#   deployed.
 #
 class govuk::apps::govuk_content_schemas(
-  $directory = '/data/apps/govuk-content-schemas/current',
+  $ensure = present,
+  $directory = '/data/apps/govuk-content-schemas',
 ) {
   include icinga::client::check_directory_exists
 
+  if $ensure == absent {
+    file { $directory:
+      ensure => absent,
+      force  => true,
+    }
+  }
+
+  $current_directory = "${directory}/current"
+
   @@icinga::check { "check_govuk_content_schemas_on_${::hostname}":
-    check_command       => "check_nrpe!check_directory_exists!${directory}",
+    ensure              => $ensure,
+    check_command       => "check_nrpe!check_directory_exists!${current_directory}",
     service_description => 'govuk_content_schemas exists',
     host_name           => $::fqdn,
   }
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/govuk_content_schemas.pp
```diff
@@ -4,17 +4,31 @@
...
+  if $ensure == absent {
+    file { $directory:
+      ensure => absent,
+      force  => true,
+    }
+  }
+
+  $current_directory = "${directory}/current"
+
   @@icinga::check { "check_govuk_content_schemas_on_${::hostname}":
```
barrucadu: I think this check is automatically managed, so removing the `ensure` and putting it inside an `if $ensure == present { ... }` should do the job.

kevindew: Sure, do you think there are any advantages to doing it that way rather than the approach I've took? I have a preference for passing in the ensure since it's more succinct and not nested but not sure if I'm overlooking something.

barrucadu: I thought this hadn't worked, because the alert was still there after 25 minutes, but it's now gone away.  I guess puppet had to run somewhere else too (the monitoring box?) to clean up the old alerts.

kevindew: Ah good good, thank you for trying to help solve the problem - even if it turned out not to be one. I appreciate the effort 🙂.

Yeah I've realised I'm not really sure how these icinga checks work - didn't realise we could trigger an object in one node that affects another. Everyday is learning day.

cbaines: Yeah, while Puppet should have run once on every machine within 30 minutes, some changes like Icinga checks can require Puppet to run on say a backend machine to change something about a check, and then on the monitoring machine to actually change the Icinga configuration.

PR #9942, opened 2019-12-11
Un-absent Alan G, Ben E, Ben T and Bevan L from Puppet

issyl0:


benthorner:
FYI this probably won't help, since we lost our laptops and (therefore) our private SSH keys. Each of us should need to raise a new PR to modify our user's public key.

PR #9944, opened 2019-12-11
Remove Chris Farmiloe

issyl0:
- Chris works on RE Autom8 now, not RE GOV.UK.

### DIFF
--- a/modules/users/manifests/chrisfarmiloe.pp
+++ b/modules/users/manifests/chrisfarmiloe.pp
@@ -1,6 +1,7 @@
 # Creates the user chrisfarmiloe for Chris Farmiloe
 class users::chrisfarmiloe {
   govuk_user { 'chrisfarmiloe':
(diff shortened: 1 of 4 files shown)


modules/users/manifests/chrisfarmiloe.pp
```diff
@@ -1,9 +0,0 @@
-# Creates the user chrisfarmiloe for Chris Farmiloe
-class users::chrisfarmiloe {
-  govuk_user { 'chrisfarmiloe':
-    fullname => 'Chris Farmiloe',
```
barrucadu: This needs to be ensured absent first

issyl0: Ugh, of course it does. Thanks!

issyl0: Done.

PR #9950, opened 2019-12-12
Add NDCG metric

bilbof:
This add [NDCG](https://en.wikipedia.org/wiki/Discounted_cumulative_gain) (a ranking metric) as a metric to the search relevancy dashboard.

It also adapts an existing job search-google-analytics-etl, that does a similar thing, expanding it to include other relevancy metrics that need to be etl'd to graphite. This new job will call a Search API rake task relevancy:send_metrics_to_graphite, introduced in alphagov/search-api#1841.

Should be merged after https://github.com/alphagov/search-api/pull/1841 is deployed.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -417,7 +417,7 @@ task :check_consistency_between_aws_and_carrenza do
     govuk_jenkins::jobs::search_fetch_analytics_data::cron_schedule
     govuk_jenkins::jobs::search_fetch_analytics_data::skip_page_traffic_load
     govuk_jenkins::jobs::search_relevancy_rank_evaluation::cron_schedule
-    govuk_jenkins::jobs::search_google_analytics_etl::cron_schedule
+    govuk_jenkins::jobs::search_relevancy_metrics_etl::cron_schedule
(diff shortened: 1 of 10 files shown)


issyl0:
For the uninitiated who just read the title and were confused, according to the diff, "NDCG" is "normalised discounted cumulative gain". No, I'm none the wiser either.

bilbof:
Sorry @issyl0! It took me some time to understand it, too. There's an explanation on this post (https://opensourceconnections.com/blog/2019/12/09/demystifying-ndcg-and-err/).

The short answer is it's a ranking metric. Given a query like "Harry Potter" and a list of ranked and ordered documents `0:"HMRC" 3:"Harry Potter" 1:"Ron Weasley" 2:"Philosophers stone"`, it returns a number between 0 and 1 based on how close to ideal the order is with respect to the rank.

The ideal order is: `3 2 1 0` but the actual order is `0 3 1 2`, so all NDCG does is give us a metric for how bad that ordering is. We've written more docs about this in the Search API repo: https://github.com/alphagov/search-api/blob/master/doc/rank_evaluation.md

PR #9951, opened 2019-12-12
Update Bruce Bolt's SSH public key

brucebolt:
I'm now using a Yubikey


issyl0:
Why is this "do not merge"?

brucebolt:
I've been having some problems getting my Mac to reliably use the Yubikey.  Not merging until I'm confident this works.

PR #9953, opened 2019-12-12
Install docker on standalone jenkins instances

barrucadu:
This is useful for the search-analytics job, which has recently
started failing because the version of pip we have on the jenkins
instances is so old, it can't install some of its dependencies.

An alternative would be updating pip.  This probably has updating
Python as a pre-requisite, so both would need to be packaged for
Ubuntu 14.04.  We might then discover incompatibilities, where there
are other things running which have broken with a newer version of
Python.

Containers are a nice solution to this: jobs can specify the versions
of dependencies they need, different jobs can use different versions,
and we don't need to manually package stuff.

---

With this PR applied, fixing the search-analytics job is simple: https://github.com/alphagov/search-analytics/compare/msw/dockerise-nightly-run

### DIFF
--- a/modules/govuk/manifests/node/s_jenkins.pp
+++ b/modules/govuk/manifests/node/s_jenkins.pp
@@ -61,4 +61,29 @@ class govuk::node::s_jenkins (
 
   # Required for a job that issues Jenkins Crumbs
   ensure_packages(['jq'])
+
+  # can't use the collectd plugin because the installation conflicts
+  # with some other python packages on the agent
+  class { '::govuk_docker':
+    docker_users           => ['jenkins'],
+    enable_collectd_plugin => false,
+  }
+
+  cron::crondotdee { 'docker_system_prune_dangling' :
+    hour    => '*',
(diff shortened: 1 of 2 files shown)


karlbaker02:
Can we upgrade from Ubuntu 14.04? This isn't a proper fix to the problem as we probably end up pushing the problem along a bit to break in the future, but 14.04 has already passed end-of-life.

barrucadu:
Upgrading from 14.04 would be fantastic, but not something I really want to attempt by myself on a Thursday afternoon.

There's a time factor in getting the search-analytics job working, because while it's failing we're not pulling in updated popularity data for search ranking.

issyl0:
Upgrading from 14.04 is an RE problem, and a tech strategy problem, and one that I don't even want to think about otherwise I'll 😢. In terms of support, we pay for Extended Support from Canonical for 14.04.

barrucadu:
I must confess, I copied the pruning from the ci agents.  But I have tested it on staging.

PR #9957, opened 2019-12-13
Tweaks to prometheus configuration

philandstuff:
## don't try to scrape stopped instances

There are some stopped gatling boxes in production, which prometheus
is trying to scrape.  This means that they appear as "down" targets in
prometheus.

I think Prometheus maybe shouldn't try to scrape instances that are
known to be stopped?

Those gatling boxes are a bit weird in any case; they don't have the
same tags as the other boxes in that environment.

## scrape prometheus metrics

Prometheus exposes metrics about itself.  We should scrape them.

This uses ec2_sd_configs to scrape only instances which have tag
`aws_migration` set to `prometheus`.

### DIFF
--- a/modules/govuk_prometheus/files/etc/prometheus/prometheus.yml
+++ b/modules/govuk_prometheus/files/etc/prometheus/prometheus.yml
@@ -27,3 +27,6 @@ scrape_configs:
     ec2_sd_configs:
       - region: "eu-west-1"
         port: 9080
+        filters:
+        - name: instance-state-name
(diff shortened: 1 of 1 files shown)


issyl0:
The Gatling boxes [aren't meant to be long-lived](https://docs.publishing.service.gov.uk/manual/load-testing.html#de-provision-gov-uk-gatling), that's probably why they're not tagged.

philandstuff:
Well, the oldest one was created in February...

PR #9963, opened 2019-12-13
Add EC2 tags to prometheus targets

philandstuff:
Currently, we can't tell which node exporter is running on which
instance very easily.

This commit uses the magic `labelmap` action to add all EC2 instance
tags as target labels to the targets, so that we can much more easily
see what instance type it is.

### DIFF
--- a/modules/govuk_prometheus/files/etc/prometheus/prometheus.yml
+++ b/modules/govuk_prometheus/files/etc/prometheus/prometheus.yml
@@ -30,6 +30,9 @@ scrape_configs:
         filters:
         - name: instance-state-name
           values: [ running ]
+    relabel_configs:
+      - action: labelmap
+        regex: __meta_ec2_tag_(.+)
   - job_name: 'prometheus'
     ec2_sd_configs:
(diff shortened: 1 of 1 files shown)


sengi:
Do we want *all* the tags as labels? Would we be better off adding just the ones we want for now, for example `__meta_ec2_instance_type` if it's just instance type that we want? (I don't have any prior experience with Prometheus and EC2 so apologies if this is a silly question. I'm just asking because this sounds like it might potentially import clutter.)

sengi:
Oh, sorry I think I see why we want `tags`. I was thrown by the double-meaning of "instance type" in the description. If it's the human-friendly name of the instance that we want then that's only in the tags, right?

philandstuff:
@sengi I pushed a commit to be explicit about which tags we pull in.

sengi:
LGTM. I wish we'd used a better name then `aws_migration` for the machine class back in the day, but that ship sailed long ago and I'm not sure that renaming it as part of the relabel here wouldn't be even more surprising 🤷‍♂

PR #9976, opened 2019-12-17
Remove manifest for chrisfarmiloe

cbaines:
As their user is absent.


cbaines:
https://github.com/alphagov/govuk-secrets/pull/870 should be merged first.

PR #9978, opened 2019-12-18
Remove the richardmorton user manifest

cbaines:
As they're now absent.


cbaines:
> I compared this to PR #8658. Looks good to me +1

:+1: Thanks for taking a look Mahmud!

PR #9983, opened 2019-12-20
Add icinga check status updates for delivery attempts in email-alert-api

edwardkerry:
We are no longer using a healthcheck to monitor the status of delivery
attempts created within the past hour in email-alert-api.

This Icinga check is not machine specific and the data used lives in
stats/gauges/govuk/email-alert-api in Graphite. We can use Graphite's
`divideSeries` function to calculate the proportion of delivery
attempts that are still in `pending` status.
See https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.divideSeries
for more information.

The threshold values have been taken from the original healthcheck,
see https://github.com/alphagov/email-alert-api/pull/1074/files

Related PRs:
alphagov/email-alert-api#1073

[trello](https://trello.com/c/hPiOuVko/1581-5-extract-the-statusupdates-check-from-the-healthcheck-endpoint-in-email-alert-api-to-a-separate-icinga-check)

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api/checks.pp
+++ b/modules/govuk/manifests/apps/email_alert_api/checks.pp
@@ -5,6 +5,7 @@
 class govuk::apps::email_alert_api::checks(
   $internal_failure = true,
   $technical_failure = true,
+  $ensure = present,
 ) {
   $internal_failure_ensure = $internal_failure ? { true => present, false => absent }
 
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/email_alert_api/delivery_attempt_status_update_check.pp
```diff
@@ -0,0 +1,26 @@
...
+#   Whether to enable the check.
+#   Default: present
+#
+# [*title*]
+#   The status of the delivery attempts to look for.
+#
+define govuk::apps::email_alert_api::delivery_attempt_status_update_check(
+  $ensure = present,
+) {
+  @@icinga::check::graphite { "email-alert-api-delivery-attempt-${title}":
```
thomasleese: I think this definition should go in https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/email_alert_api/checks.pp (which is included from here https://github.com/alphagov/govuk-puppet/blob/de2adb57d7d6b347c10195c54b34727987759af0/modules/monitoring/manifests/checks.pp#L50) as currently nothing is using `govuk::apps::email_alert_api::delivery_attempt_status_update_check` so it won't be included anywhere.

cbaines:
I'd squash the third (Fix lint errors) commit in to the first commit, as it looks like those lint errors were introduced there, but apart from that this looks good to me :+1:

PR #9985, opened 2019-12-20
Add env var for bigquery credentials

barrucadu:
search-api doesn't use this var, but it will allow us to run the rake
task which needs these credentials without having to manually copy
credentials to the machine.

---

[Trello card](https://trello.com/c/SMIgrRTd/1159-add-bigquery-query-to-ltr-pipeline)

### DIFF
--- a/modules/govuk/manifests/apps/search_api.pp
+++ b/modules/govuk/manifests/apps/search_api.pp
@@ -66,6 +66,10 @@
 #   The view id of GOV.UK in Google Analytics
 #   Default: undef
 #
+# [*google_bigquery_credentials*]
+#   The credentials to use to fetch data from BigQuery
+#   Default: undef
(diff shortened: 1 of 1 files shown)


barrucadu:
It's a private key + a bunch of other stuff

bilbof:
Could that stuff be separated out into separate variables?

barrucadu:
It could be, but then we'd be recombining it in search-api.  I've been using the credentials file you gave me, which I guess came from some bigquery thing - if bigquery produced it as one thing and bigquery consumes it as one thing, it makes sense to me to treat it as one thing.

bilbof:
Ah got you! For some reason I thought we had combined these variables together. If they came like that it makes perfect sense to store them as one thing. :shipit:

PR #9993, opened 2020-01-02
Avoid the pagerduty drill on bank holidays

cbaines:


cbaines:
> Would the drill still start if https://www.gov.uk/bank-holidays.json isn’t available?

I'd hope so. I added the logic regarding bank holidays as a guard clause, rather than making creating the file conditional on the day not being a bank holiday for this reason.

Thanks for taking a look Bill :)

PR #9994, opened 2020-01-02
Add metrics to show a snapshot of worker activity

benthorner:
https://trello.com/c/13JaJ5mQ/203-review-the-email-alert-traffic-dashboard-the-two-dashboards

Previously we could only know about live Sidekiq activity by connecting
to the Web UI and manually scanning the list of active jobs. This adds
two new graph panels to visualise two new metrics that represent, for
each queue, a snapshot of the number and runtime of its workers.

Having visibility of the live state of Sidekiq will help to diagnose
end-to-end issues that are due to a job getting 'stuck', or running for
longer than expected. Previously we were somewhat blind to this kind of
problem, due to the effort and error-proneness of the Web UI approach.

The new metrics are only available sporadically due to the collection
mechanism used in sidekiq-stats, as is the case for the data shown in
the existing 'queue' panels. Even so, I've tried to make the partial
nature of the data clear by writing 'snapshot' in the graph titles.

screencapture-grafana-integration-publishing-service-gov-uk-dashboard-db-sidekiq-iteration-2020-01-02-16_42_15

### DIFF
--- a/modules/grafana/files/dashboards/sidekiq.json
+++ b/modules/grafana/files/dashboards/sidekiq.json
@@ -6,7 +6,7 @@
   "gnetId": null,
   "graphTooltip": 1,
   "hideControls": false,
-  "id": 142,
+  "id": 11,
   "links": [],
   "refresh": "1m",
   "rows": [
@@ -26,10 +26,11 @@
             "alignAsTable": true,
             "avg": true,
             "current": true,
+            "hideEmpty": true,
             "max": true,
             "min": false,
             "show": true,
-            "sort": "total",
+            "sort": "avg",
             "sortDesc": true,
             "total": true,
             "values": true
@@ -52,7 +53,7 @@
             {
               "hide": false,
               "refId": "A",
-              "target": "aliasByNode(summarize(consolidateBy(sumSeriesWithWildcards(stats_counts.govuk.app.$Application.*.workers.*.success, 4), 'sum'), '$Aggregation', 'sum', false), 5)",
+              "target": "groupByNode(summarize(stats_counts.govuk.app.$Application.*.workers.*.success, '$Aggregation', 'sum', false), 6, 'sum')",
               "textEditor": false
             }
           ],
@@ -102,7 +103,7 @@
     },
     {
       "collapse": false,
-      "height": 357,
+      "height": 404,
       "panels": [
         {
           "aliasColors": {},
@@ -116,12 +117,14 @@
             "alignAsTable": true,
             "avg": true,
             "current": true,
-            "hideEmpty": true,
-            "hideZero": true,
+            "hideEmpty": false,
+            "hideZero": false,
             "max": true,
             "min": false,
             "rightSide": false,
             "show": true,
+            "sort": "max",
+            "sortDesc": true,
             "total": false,
             "values": true
           },
@@ -142,7 +145,7 @@
           "targets": [
             {
               "refId": "A",
-              "target": "summarize(groupByNode(stats.gauges.govuk.app.$Application.*.workers.queues.*.latency, 8, 'avg'), '$Aggregation', 'avg', false)"
+              "target": "aliasByNode(summarize(averageSeriesWithWildcards(stats.gauges.govuk.app.$Application.*.workers.queues.*.latency, 5), '$Aggregation', 'avg', false), 7)"
             }
           ],
           "thresholds": [],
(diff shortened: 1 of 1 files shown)


cbaines:
Nice, looks like this is interesting data.

On the panel titles, I was initially a bit uncertain what the data meant. I think "Job duration snapshot" might be clearer than "Runtime snapshot" for the first panel, and something like "Number of jobs being processed snapshot" might be clearer for the second panel. Also, snapshot might make more sense if it's inside the brackets, as the time period is directly relevant to what the snapshot reflects.

Less significant things I also noticed:
 - I usually set a order for legend, decreasing by `max` might be the most useful default here
 - Maybe set a sort order for the hover tooltip
 - Maybe fix the Y-Min at 0, I don't always notice the scale doesn't start there, and I think that the consistency this brings is usually worth the reduced detail

modules/grafana/files/dashboards/sidekiq.json
```diff
@@ -273,7 +273,165 @@
...
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "spaceLength": 10,
+          "span": 6,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "refId": "A",
+              "target": "aliasByNode(summarize(averageSeriesWithWildcards(stats.gauges.govuk.app.$Application.*.workers.queues.*.runtime, max, 5), '$Aggregation', 'max', false), 7)"
```
cbaines: I'm a bit confused by the second parameter to `averageSeriesWithWildcards` here. Max isn't really an average... I'm guessing this works though.

benthorner: You're right to be confused - it's not even a [validate parameter](https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.averageSeriesWithWildcards). Graphite seems to be interpreting it as a number, though 👏. Fixed to use `consolidateBy(max)`, like we do everywhere else.

benthorner:
@cbaines thanks for reviewing! I've adopted all but one of your suggestions, albeit with slightly snappier titles. Normally the Y-axes isn't an issue, except when there's no data, which is somewhat likely here. I can't see any great benefit in sorting the tooltip, so I've left that as-is.

Could you take another look?

modules/grafana/files/dashboards/sidekiq.json
```diff
@@ -142,7 +144,7 @@
           "targets": [
             {
               "refId": "A",
-              "target": "summarize(groupByNode(stats.gauges.govuk.app.$Application.*.workers.queues.*.latency, 8, 'avg'), '$Aggregation', 'avg', false)"
+              "target": "aliasByNode(summarize(averageSeriesWithWildcards(stats.gauges.govuk.app.$Application.*.workers.queues.*.latency, 5), '$Aggregation', 'avg', false), 7)"
```
cbaines: These changes relate to the queue latency, not the new panels. Were you intending to make them?

modules/grafana/files/dashboards/sidekiq.json
cbaines: As far as I understand, `consolidateBy` doesn't fundamentally change what metrics you get back, but subtly changes the aggregations that Graphite implicitly does internally (which isn't very relevant here, as the aggregation and period is very explicit in these panels). I'm not saying that using it is a problem, just that I don't think using it fixes anything.

modules/grafana/files/dashboards/sidekiq.json
benthorner: Oh yes, sorry. I noticed this panel was doing a weird `groupByNode` that was causing the legend to come out in the Graphite expression syntax. I fixed it to resemble the other queries we do in the dashboard. The datapoints should be unchanged, although this is Graphite, so maybe `groupByNode(..., 'avg')` is subtly different to `averageSeriesWithWildcards(...)`!

modules/grafana/files/dashboards/sidekiq.json
benthorner: You're right - sorry, was getting confused with the approach we took in 'Email Alert Api Metrics'. The function probably has some value, since even with the aggregation period Graphite may still need to downsample when drawing the graphs. I've replaced it with a `groupByNode` approach and updated all the panels to be consistent.

Does that make more sense?

cbaines: Yeah, I think so.

PR #10008, opened 2020-01-03
Improve clearing out old Graphite data in AWS

cbaines:
There's already support for deleting the directories with metrics from
collectd relating to a single machine, but inside the stats and
stats_counts namespaces, there are plenty of old machine specific
metrics.

Use the distinctive "ip-" string in the name to identify the
directories that likely contain machine specific metrics, then delete
them if they haven't been updated in around 8 weeks.

As some dashboards and Icinga checks aggregate the metrics stored per
machine, I'm uncertain what the impact will be of deleting more recent
metrics, which is why I've gone with 8 weeks.

### DIFF
--- a/modules/govuk/templates/node/s_graphite/remove_old_whisper_data.erb
+++ b/modules/govuk/templates/node/s_graphite/remove_old_whisper_data.erb
@@ -14,6 +14,14 @@ for dir in $whisper_dir/*; do
     rm -rf $dir;
   fi
 done
+
+for dir in stats/timers stats_counts; do
+  find "$whisper_dir/$dir" -maxdepth 1 -name "*ip-*" -type d -print0 | while IFS= read -r -d '' dir; do
+    if [[ ! -z $dir && -z "$(find $dir -mtime -56)" && $? -eq 0 ]]; then
+      rm -rf "$dir";
+    fi
+  done
+done
 <% end -%>
 
 # Delete empty directories


modules/govuk/templates/node/s_graphite/remove_old_whisper_data.erb
```diff
@@ -14,6 +14,14 @@ for dir in $whisper_dir/*; do
     rm -rf $dir;
   fi
 done
+
+for dir in stats stats_counts; do
+  find "$whisper_dir/$dir" -name "*ip-*" -type d -print0 | while IFS= read -r -d '' dir; do
```
benthorner: Could we do `".ip-*"` here? I'm thinking other metrics may contain a `'*ip-*'` string somewhere.

modules/govuk/templates/node/s_graphite/remove_old_whisper_data.erb
```diff
@@ -14,6 +14,14 @@ for dir in $whisper_dir/*; do
...
+    if [[ ! -z $dir && -z "$(find $dir -mtime -56)" && $? -eq 0 ]]; then
```
benthorner: Can you help me understand what the `$? -eq 0` bit is for?

modules/govuk/templates/node/s_graphite/remove_old_whisper_data.erb
cbaines: Some of the metrics that this change looks at deleting don't necessarily start with `ip-`, like `/opt/graphite/storage/whisper/stats_counts/calculators_frontend-ip-10-12-6-209_eu-west-1_compute_internal` for example.

modules/govuk/templates/node/s_graphite/remove_old_whisper_data.erb
cbaines: I'm not entirely certain (this bit is copied from further up in the file), but I think it's just an extra check that the `find` command executed successfully. `$?` is the exit status of the last command, and as far as I know, `find` will exit with 0 regardless of the results. I believe it could be removed without changing the behavior of the script.

cbaines:
> Looks good - I can't think of any example metrics like `*ip-*` that we'd want to keep, so let's try it!

I'm still a bit unsure, so I've just pushed a change to restrict this to the immediate contents of the `stats/timers` and `stats_counts` directory, and I think all the data there relates to timings of requests from the NGinx log files, that I don't believe we're using.

PR #10017, opened 2020-01-07
Add subscription contents Icinga checks

deborahchua:
This adds Icinga checks that monitor the number of `email-alert-api`
subscription contents that have a `critical` and `warning` latency
(created more than `x` minutes ago but their associated emails have
not been sent out). They are not machine specific and the data used
lives in stats/gauges/govuk/email-alert-api in Graphite.

We will be alerted if either of these values are bigger than 0.

Related PR: https://github.com/alphagov/email-alert-api/pull/1096
Trello card: https://trello.com/c/dtdY0kAK/1606-5-extract-the-subscriptioncontents-check-from-the-healthcheck-endpoint-in-email-alert-api-to-a-separate-icinga-check

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api/checks.pp
+++ b/modules/govuk/manifests/apps/email_alert_api/checks.pp
@@ -33,4 +33,24 @@ class govuk::apps::email_alert_api::checks(
     from      => '1hour',
     desc      => 'email-alert-api - high number of delivery attempts have not received status updates',
   }
+
+  @@icinga::check::graphite { 'email-alert-api-warning-subscription-contents':
+    ensure    => $ensure,
+    host_name => $::fqdn,
+    target    => 'transformNull((keepLastValue(stats.gauges.govuk.email-alert-api.subscription_contents.warning_total))',
+    warning   => '1',
+    from      => '1hour',
+    desc      => 'email-alert-api - unprocessed subscription contents',
+    notes_url => monitoring_docs_url(email-alert-api-unprocessed-subscription-contents),
+  }
+
+  @@icinga::check::graphite { 'email-alert-api-critical-subscription-contents':
+    ensure    => $ensure,
+    host_name => $::fqdn,
+    target    => 'keepLastValue(stats.gauges.govuk.email-alert-api.subscription_contents.critical_total)',
+    critical  => '1',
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/email_alert_api/checks.pp
```diff
@@ -33,4 +33,24 @@
...
   }
+
+  @@icinga::check::graphite { 'email-alert-api-warning-subscription-contents':
+    ensure    => $ensure,
+    host_name => $::fqdn,
+    target    => 'transformNull((keepLastValue(stats.gauges.govuk.email-alert-api.subscription_contents.warning_total))',
+    warning   => '1',
+    from      => '1hour',
+    desc      => 'email-alert-api - unprocessed subscription contents created over 10 minutes ago (or 35 minutes ago if during publishing times)',
+    notes_url => monitoring_docs_url(email-alert-api-unprocessed-subscription-contents),
```
deborahchua: This page doesn't exist yet but will be added shortly.

modules/govuk/manifests/apps/email_alert_api/checks.pp
```diff
@@ -33,4 +33,24 @@
...
     desc      => 'email-alert-api - high number of delivery attempts have not received status updates',
   }
+
+  @@icinga::check::graphite { 'email-alert-api-warning-subscription-contents':
+    ensure    => $ensure,
+    host_name => $::fqdn,
+    target    => 'transformNull((keepLastValue(stats.gauges.govuk.email-alert-api.subscription_contents.warning_total))',
+    warning   => '1',
+    from      => '1hour',
+    desc      => 'email-alert-api - unprocessed subscription contents created over 10 minutes ago (or 35 minutes ago if during publishing times)',
```
thomasleese: I can see the benefit of including these numbers in the description, but I would be worried that they get out of sync with the numbers in email-alert-api. Perhaps we could add a comment to the code in email-alert-api to remind people to update the alert here?

deborahchua: Thanks @thomasleese ! I did debate whether to keep this in, I've taken it out for now and add a comment as you have suggested. I'll also include it in the docs for this alert.

PR #10023, opened 2020-01-07
Fix subscription contents Icinga checks

deborahchua:
It appears that the checks require both `warning` and `critical`
arguments. Setting them to a value that won't be hit to bypass
this issue.

Also a small fix to one of the targets which was originally missed.

https://trello.com/c/dtdY0kAK/1606-5-extract-the-subscriptioncontents-check-from-the-healthcheck-endpoint-in-email-alert-api-to-a-separate-icinga-check

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api/checks.pp
+++ b/modules/govuk/manifests/apps/email_alert_api/checks.pp
@@ -34,23 +34,31 @@ class govuk::apps::email_alert_api::checks(
     desc      => 'email-alert-api - high number of delivery attempts have not received status updates',
   }
 
+  # We are only interested in the `warning` state but `critical` is also required
+  # for valid Icinga check configuration. Setting it to a high value that won't be
+  # reached allows us to get round this issue
   @@icinga::check::graphite { 'email-alert-api-warning-subscription-contents':
     ensure    => $ensure,
     host_name => $::fqdn,
-    target    => 'transformNull((keepLastValue(stats.gauges.govuk.email-alert-api.subscription_contents.warning_total))',
-    warning   => '1',
+    target    => 'transformNull(keepLastValue(stats.gauges.govuk.email-alert-api.subscription_contents.warning_total))',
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/email_alert_api/checks.pp
```diff
@@ -37,20 +37,22 @@
...
+    desc      => 'email-alert-api - unprocessed subscription contents (warning)',
     notes_url => monitoring_docs_url(email-alert-api-unprocessed-subscription-contents),
   }
 
   @@icinga::check::graphite { 'email-alert-api-critical-subscription-contents':
     ensure    => $ensure,
     host_name => $::fqdn,
-    target    => 'keepLastValue(stats.gauges.govuk.email-alert-api.subscription_contents.critical_total)',
-    critical  => '1',
+    target    => 'transformNull(keepLastValue(stats.gauges.govuk.email-alert-api.subscription_contents.gcritical_total))',
```
cbaines: There's a typo here I think -> `gcritical`

modules/govuk/manifests/apps/email_alert_api/checks.pp
```diff
@@ -37,20 +37,22 @@
   @@icinga::check::graphite { 'email-alert-api-warning-subscription-contents':
     ensure    => $ensure,
     host_name => $::fqdn,
-    target    => 'transformNull((keepLastValue(stats.gauges.govuk.email-alert-api.subscription_contents.warning_total))',
-    warning   => '1',
+    target    => 'transformNull(keepLastValue(stats.gauges.govuk.email-alert-api.subscription_contents.warning_total))',
+    warning   => '0',
+    critical  => '100000000', #set to a high value which won't be reached
```
cbaines: I think what a comment here could explain is the why. So why is this set to a high value that won't be reached?

deborahchua:
Just tested this on integration and can see both checks now 👍

PR #10025, opened 2020-01-08
Add icinga checks for email-alert-api content changes

edwardkerry:
Two new Icinga checks, warning and critical, for more refined reporting on unprocessed content changes in `email-alert-api`. Alerts will trigger if either value exceeds 0.

This is part of work to slim down the `email-alert-api-healthcheck-not-ok` alert, making individual alerts which are not machine-specific. Data lives in `stats/gauges/govuk/email-alert-api` in Graphite.

[Trello](https://trello.com/c/c6mWg9jP/1627-3-extract-the-contentchanges-check-from-the-healthcheck-endpoint-in-email-alert-api-to-a-separate-icinga-check)

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api/checks.pp
+++ b/modules/govuk/manifests/apps/email_alert_api/checks.pp
@@ -48,6 +48,17 @@ class govuk::apps::email_alert_api::checks(
     notes_url => monitoring_docs_url(email-alert-api-unprocessed-subscription-contents),
   }
 
+    @@icinga::check::graphite { 'email-alert-api-warning-content-changes':
+    ensure    => $ensure,
+    host_name => $::fqdn,
+    target    => 'transformNull(stats.gauges.govuk.email-alert-api.content_changes.warning_total)',
+    warning   => '0',
+    critical  => '100000000',
+    from      => '1hour',
+    desc      => 'email-alert-api - unprocessed content changes',
+    notes_url => monitoring_docs_url(email-alert-api-unprocessed-content-changes),
+  }
+
   # We are only interested in the `critical` state but `warning` is also required
   # for valid Icinga check configuration. Both states are set to 0 but `critical`
   # takes precedence and allows us to get round this issue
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/email_alert_api/checks.pp
```diff
@@ -48,6 +48,17 @@
     notes_url => monitoring_docs_url(email-alert-api-unprocessed-subscription-contents),
   }
 
+    @@icinga::check::graphite { 'email-alert-api-warning-content-changes':
+    ensure    => $ensure,
+    host_name => $::fqdn,
+    target    => 'transformNull(stats.gauges.govuk.email-alert-api.content_changes.warning_total)',
+    warning   => '0',
+    critical  => '100000000',
```
issyl0: Is this value so large because this is always a warning alert and the critical should never trigger from here? Is there a better way?

edwardkerry: That is correct, there are existing comments in the file to this effect
```
  # We are only interested in the `warning` state but `critical` is also required
  # for valid Icinga check configuration. Setting it to a high value that won't be
  # reached allows us to get round this issue
```
I don't think we are currently aware of a better way, but are very open to ideas!

issyl0: I have no ideas. It was an observation!

deborahchua: @cbaines suggested we could make the `critical` and `warning` arguments optional in the Icinga check template configuration(I think!?) which means we wouldn't need to do this. Having seen the other two healthchecks we need to set up in a similar way maybe it's worth doing this?

cbaines: I think that's still a possibility, but not quite worth doing yet. I think it would be good to see the first iteration of the extracted healthchecks in action for a bit, and have some data to work with, and then we'll be able to make more informed decisions.

PR #10037, opened 2020-01-10
Set tensorflow serving image version to 2.0.0

bilbof:
This is the latest version: https://github.com/tensorflow/serving/releases

https://trello.com/c/OpmAUMiW/1262


barrucadu:
Do we want to use a release candidate version?

bilbof:
@barrucadu 🤦‍♂ nope, thanks for spotting that. I've changed it to 2.0.0.

PR #10044, opened 2020-01-13
Fix Icinga job URLs for machines on AWS

ChrisBAshton:
The old method of simply setting:
`"https://deploy.${::aws_environment}.govuk.digital/job/${job_slug}"`

...did not seem to work as the authors intended. This would create
a job URL like:

> https://deploy.production.govuk.digital/job/search_generate_sitemaps

...whereas it should be:

> https://deploy.blue.production.govuk.digital/job/search_generate_sitemaps/

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
+++ b/modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
@@ -27,9 +27,7 @@ class govuk_jenkins::jobs::bouncer_cdn (
   }
 
   $check_name = 'bouncer-cdn-configuration'
-
-  # FIXME: go back to using $app_domain once we have a single Icinga instance in AWS
-  $job_url = "https://deploy.${::aws_environment}.govuk.digital/job/Bouncer_CDN/"
+  $job_url = "https://deploy.blue.${::aws_environment}.govuk.digital/job/Bouncer_CDN/"
 
   @@icinga::passive_check { "${check_name}_${::hostname}":
     service_description => $service_description,
--- a/modules/govuk_jenkins/manifests/jobs/content_data_api.pp
+++ b/modules/govuk_jenkins/manifests/jobs/content_data_api.pp
@@ -13,10 +13,9 @@ class govuk_jenkins::jobs::content_data_api (
   $rake_etl_master_process_cron_schedule = undef,
   $app_domain = hiera('app_domain'),
 ) {
-
   $check_name = 'etl-content-data-api'
   $service_description = 'Content Data API ETL [Extract - transform - load]'
-  $job_url = "https://deploy.${::aws_environment}.govuk.digital/job/content_data_api_import_etl_master_process/"
+  $job_url = "https://deploy.blue.${::aws_environment}.govuk.digital/job/content_data_api_import_etl_master_process/"
 
   file { '/etc/jenkins_jobs/jobs/content_data_api.yaml':
     ensure  => present,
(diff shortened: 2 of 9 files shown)


issyl0:
I feel like @sengi will be interested in this given the eternal struggle of removing stackname from things. :sob:

modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
```diff
@@ -28,8 +28,13 @@
...
 
-  # FIXME: go back to using $app_domain once we have a single Icinga instance in AWS
-  $job_url = "https://deploy.${::aws_environment}.govuk.digital/job/Bouncer_CDN/"
+  if $::aws_migration and ($::aws_environment != 'integration') {
+    $hosting_env_domain = "blue.${::aws_environment}.govuk.digital"
+  }
+  else {
+    $hosting_env_domain = $app_domain
+  }
+  $job_url = "https://deploy.${hosting_env_domain}/job/Bouncer_CDN/"
```
cbaines: Does this job also exist in Carrenza? Bouncer just runs in AWS, so even if it does exist in Carrenza, maybe removing the job would be an option?

modules/govuk_jenkins/manifests/jobs/content_data_api.pp
```diff
@@ -13,10 +13,16 @@
   $rake_etl_master_process_cron_schedule = undef,
   $app_domain = hiera('app_domain'),
 ) {
+  if $::aws_migration and ($::aws_environment != 'integration') {
+    $hosting_env_domain = "blue.${::aws_environment}.govuk.digital"
+  }
+  else {
+    $hosting_env_domain = $app_domain
+  }
```
cbaines: This job just exists in AWS, so no need for the conditional.

modules/govuk_jenkins/manifests/jobs/enhanced_ecommerce_search_api.pp
```diff
@@ -9,10 +9,16 @@
   $rate_limit_token = undef,
   $cron_schedule = undef
 ) {
+  if $::aws_migration and ($::aws_environment != 'integration') {
+    $hosting_env_domain = "blue.${::aws_environment}.govuk.digital"
+  }
+  else {
+    $hosting_env_domain = $app_domain
+  }
```
cbaines: The Search API only runs in AWS, so maybe this conditional is unnecessary.

cbaines:
Yeah, the Jenkins URLs are a bit annoying. I think the oAuth configuration for Jenkins means that you have to use the `blue.` domain for logging in unfortunately.

I think linking to that domain is fine, but I'd prefer to not introduce conditional URLs for jobs that only exist in AWS at least.

modules/govuk_jenkins/manifests/jobs/bouncer_cdn.pp
ChrisBAshton: Are you suggesting amending to something like the following, where `$job_url` is undefined in Carrenza?

```
if $::aws_migration and ($::aws_environment != 'integration') {
    $job_url = "https://deploy.${hosting_env_domain}/job/Bouncer_CDN/"
}
```

sengi:
> I feel like @sengi will be interested in this given the eternal struggle of removing stackname from things. 😭

Thanks :) Ah, this is sad that more instances of this are cropping up. Anyone up for fixing the OAuth config in GitHub? (I can't get at it myself because I'm not an admin.)

issyl0:
> Anyone up for fixing the OAuth config in GitHub?

Yeah, sure - we can pair on it? I don't know what other things have to change.

PR #10051, opened 2020-01-15
Add DB restore configs for Whitehall and Asset Manager in AWS Prod.

sengi:
These will soon be needed for the migration of Whitehall and Asset
Manager to AWS. After the migration, they will be needed in the event
that we restore a database backup in Production.

The "ensure: disabled" here means that the config files are placed on
the db_admin machine but will not be run from cron.


sengi:
Need to sort out credentials for the DocDB job separately.

PR #10052, opened 2020-01-15
Remove unused directory created by govuk_env_sync.

sengi:
govuk_env_sync was creating `/var/lib/documentdb` but this is no
longer referred to anywhere, so stop creating it.

Removing the Puppet resource in one go rather than setting to `ensure:
absent` because there are only a handful of machines affected and it
therefore saves time to remove the directory manually than roll out two
Puppet releases.


sengi:
Removed the directory from `db_admin` in all three environments.

PR #10055, opened 2020-01-15
Remove Redis daily reboot in integration

karlbaker02:
This PR removes the Redis daily reboot in integration.

The reboot was originally introduced to help test resilience of our apps to Redis reboots, specifically targeted to integration only, as an output of a Redis-related incident (https://github.com/alphagov/govuk-puppet/commit/c542ae32f396edc9eff23dde6fd56262c398f73d). Any alerts which are triggered as a result of Redis rebooting happen around 11am (when the reboot happens).

Given the number of errors generally present in integration, in addition to whether anyone is actively checking the alerts generated as a result of the Redis reboots, the value provided by this no longer seems to outweigh keeping the reboots in place, so we should remove it and make our environments generally more consistent and stable.


karlbaker02:
I'm going to do it as Chris suggests and manually remove the few bits left behind as a bit of learning, but I certainly hear what you're saying too @bevanloon 👍

PR #10059, opened 2020-01-16
Mark Murilo as absent

vanitabarrett:
Murilo is on leave for ~6 weeks. According to [these docs](https://docs.publishing.service.gov.uk/manual/rules-for-getting-production-access.html#temporarily-revoking-access), we should remove his production access.


vanitabarrett:
Thanks @issyl0  - in that case, is there anything I need to do in puppet?

issyl0:
I don't think so. Removing Murilo from staging and production user lists in secrets should be enough. :tada:

PR #9887, opened 2019-11-28
[DO NOT MERGE YET, WHITEHALL MIGRATION] remove whitehall from Carrenza Production /etc/hosts

fredericfran-gds:

PR #9886, opened 2019-11-28
[DO NOT MERGE YET, ASSET-MANAGER MIGRATION] remove asset manager from carrenza production /etc/hosts

fredericfran-gds:

PR #10061, opened 2020-01-16
Move Whitehall and Asset Manager to AWS.

sengi:
This is to be merged when we migrate Whitehall and Asset Manager to AWS.

* Remove Plek overrides for Whitehall. This makes apps in AWS use the `govuk-internal.digital` names for whitehall-frontend and whitehall-admin, instead of the `publishing.service.gov.uk` names.
* Remove `/etc/hosts` entries for Whitehall and Asset Manager. This allows clients in Carrenza to talk to these apps in AWS. (Otherwise the DNS names are shadowed by the hosts entries.)
* Remove `asset-manager` from the list of apps which get deployed to new `backend` instances in Carrenza. (This doesn't remove the app from existing machines, but we'll be doing so manually very soon after this gets merged.)

See individual commit messages for more detail.

Supersedes #10076, #9886 and #9887.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -17,7 +17,6 @@ node_class: &node_class
       - asset_env_sync
   backend:
     apps:
-      - asset-manager
       - canary-backend
       - collections-publisher
       - contacts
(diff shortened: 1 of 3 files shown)


sengi:
This will probably need rebasing before the migration because the `($::aws_environment == 'staging') or` will disappear when Staging is moved to AWS.

sengi:
Rebased now that Staging is in AWS. Even though we'll probably roll Staging back and forward at least once more before the Prod migration, with a bit of luck this PR should stay valid.

PR #10064, opened 2020-01-17
Add email alert api sidekiq queue length icinga check

edwardkerry:
This adds Icinga checks that monitor the number of messages currently on the various
Sidekiq queues we have in `email-alert-api`. The tresholds have
been taken from the healthcheck https://github.com/alphagov/email-alert-api/blob/master/app/models/healthcheck/queue_size.rb 

The healthcheck will be removed once these checks are in place.

# Checks in Icinga (integration)
Screenshot 2020-01-17 at 14 40 20

# Graphite Graph (integration)
Screenshot 2020-01-17 at 14 40 35

[Trello](https://trello.com/c/lyTVPild/1681-2-extract-the-queuesize-check-from-the-healthcheck-endpoint-in-email-alert-api-to-a-separate-icinga-check)

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api/checks.pp
+++ b/modules/govuk/manifests/apps/email_alert_api/checks.pp
@@ -6,6 +6,19 @@ class govuk::apps::email_alert_api::checks(
   $ensure = present,
 ) {
 
+  sidekiq_queue_check { [ 'delivery_immediate_high',
+                          'delivery_immediate',
+                          'delivery_digest',
+                          'email_generation_immediate',
+                          'process_and_generate_emails',
+                          'default',
+                          'email_generation_digest',
+                          'cleanup']:
+    size_warning  => '75000',
+    size_critical => '100000',
+    ;
+  }
+
   delivery_attempt_status_check { 'internal_failure':
     ensure => $ensure,
   }
--- a/modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
+++ b/modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
@@ -1,0 +1,37 @@
+# == Define: govuk::apps::email_alert_api::sidekiq_queue_checks
+#
+# Creates an Icinga check which checks Graphite for activity
+# on Sidekiq queues
+#
+# === Parameters
+#
+# [*ensure*]
+#   Whether to enable the check.
+#   Default: present
+#
+# [*title*]
+#   The name of the sidekiq queue to check.
+#
+# [*size_warning*]
+#   The number of messages on a queue to trigger a warning.
+#
+# [*size_critical*]
+#   The number of messages on a queue to trigger a critical alert.
+#
+define govuk::apps::email_alert_api::sidekiq_queue_check(
+  $size_warning,
+  $size_critical,
+) {
+  icinga::check::graphite { "check_email_alert_api_${title}_queue_size":
+    target    => "transformNull(averageSeries(keepLastValue(stats.gauges.govuk.app.email-alert-api.*.workers.queues.${title}.enqueued)), 0)",
+    from      => '24hours',
+    # Take an average over the most recent 36 datapoints, which at 5
+    # seconds per datapoint is the last 3 minutes
+    args      => '--dropfirst -36',
+    warning   => $size_warning,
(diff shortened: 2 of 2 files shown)


modules/govuk/manifests/apps/email_alert_api/checks.pp
```diff
@@ -4,8 +4,20 @@
...
+    'delivery_digest'             => { },
+    'email_generation_immediate'  => { },
+    'process_and_generate_emails' => { },
+    'default'                     => { },
+    'email_generation_digest'     => { },
+    'cleanup'                     => { },
+  },
 ) {
 
+  create_resources(sidekiq_queue_checks, $queue_resources)
```
cbaines: I think using a resource declaration with an array of titles [1] might be a neater way of expressing this, providing you don't need to vary the `queue_resources` parameter value.

1: https://puppet.com/docs/puppet/3.8/lang_resources.html#condensed-forms

modules/govuk/manifests/apps/email_alert_api/checks.pp
```diff
@@ -6,6 +6,18 @@
   $ensure = present,
 ) {
 
+
+  file { '/sidekiq_queue_check':
```
cbaines: I think `file` here should be `sidekiq_queue_check`.

modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
```diff
@@ -0,0 +1,38 @@
...
+#   The name of the sidekiq queue to check.
+#
+# [*size_warning*]
+#   The number of messages on a queue to trigger a warning.
+#
+# [*size_critical*]
+#   The number of messages on a queue to trigger a critical alert.
+#
+define govuk::apps::email_alert_api::sidekiq_queue_check(
+  $ensure = present,
```
cbaines: You probably don't need `$ensure` as a parameter.

It's useful when you're explicitly managing resources, like setting them to `$ensure => absent`, but with the one resource being managed here, you don't need to do that, as Puppet manages all the Icinga configuration, including removing old config [1].

1: https://github.com/alphagov/govuk-puppet/blob/master/modules/icinga/manifests/config.pp#L45-L47You probably don't need `$ensure` as a parameter.

It's useful when you're explicitly managing resources, like setting them to `$ensure => absent`, but with the one resource being managed here, you don't need to do that, as Puppet manages all the Icinga configuration, including removing old config [1].

1: https://github.com/alphagov/govuk-puppet/blob/master/modules/icinga/manifests/config.pp#L45-L47

modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
```diff
@@ -0,0 +1,38 @@
...
+  $size_warning = undef,
+  $size_critical = undef,
```
cbaines: These are probably better as mandatory parameters, as they're required by the `icinga::check::graphite` manifest I think.

modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
```diff
@@ -0,0 +1,37 @@
...
+# [*title*]
+#   The name of the sidekiq queue to check.
+#
+# [*size_warning*]
+#   The number of messages on a queue to trigger a warning.
+#
+# [*size_critical*]
+#   The number of messages on a queue to trigger a critical alert.
+#
+define govuk::apps::email_alert_api::sidekiq_queue_check(
```
thomasleese: Are we thinking that the latency and retry set checks will be added into this definition? Just wondering because the name is quite generic.

edwardkerry: yes, that is the plan

modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
```diff
@@ -0,0 +1,37 @@
...
+  $size_warning,
```
thomasleese: I wonder if we should add defaults to these because I imagine for most of the queues these numbers will be the same, but that's something for the next PR I think.

PR #10067, opened 2020-01-17
Use instance prefix in process metric target

bilbof:
I'm wondering if this will make process metrics faster to load.


sihugh:
Worth a shot 👍

PR #10068, opened 2020-01-17
Remove hosts entries for Whitehall and Asset Manager in Staging.

sengi:
Remove `/etc/hosts` entries (in Carrenza) for Asset Manager, the Whitehall apps and the Whitehall machines.

There are some more entries which could be removed (for example for the Whitehall databases in Carrenza Staging) but that is better left until after the migration is completed so as not to complicate rolling back/forward in Staging after this if we have to.

### DIFF
--- a/hieradata/staging.yaml
+++ b/hieradata/staging.yaml
@@ -260,7 +260,6 @@ hosts::production::backend::hosts:
     ip: '10.2.11.31'
 
 hosts::production::backend::app_hostnames:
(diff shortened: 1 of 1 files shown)


sengi:
Last time we moved Staging, it doesn't look we removed any of the hosts entries for Whitehall. There's therefore a change that we discover something that breaks as a result of these extra deletions, but better to find this out sooner rather than later.

PR #10069, opened 2020-01-17
Add Icinga checks for email-alert-api queue latency

deborahchua:
This adds Icinga checks that monitor the latency of the various
Sidekiq queues we have in `email-alert-api`. The tresholds have
been taken from the healthcheck https://github.com/alphagov/email-alert-api/blob/master/app/models/healthcheck/queue_latency.rb The healthcheck will be removed once these checks are in place.

We also want to add other checks on the queues (such as retry size
and queue size), this has been written in a way that should reduce
a lot of duplication in the creation of multiple checks for each
queue.

Tested in integration:
Screenshot 2020-01-17 at 14 56 34

Trello card: https://trello.com/c/A4RfUXGG/1680-2-extract-the-queuelatency-check-from-the-healthcheck-endpoint-in-email-alert-api-to-a-separate-icinga-check

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api/checks.pp
+++ b/modules/govuk/manifests/apps/email_alert_api/checks.pp
@@ -6,16 +6,28 @@ class govuk::apps::email_alert_api::checks(
   $ensure = present,
 ) {
 
-  sidekiq_queue_check { [ 'delivery_immediate_high',
-                          'delivery_immediate',
-                          'delivery_digest',
-                          'email_generation_immediate',
+  sidekiq_queue_check { [ 'email_generation_immediate',
                           'process_and_generate_emails',
                           'default',
                           'email_generation_digest',
                           'cleanup']:
-    size_warning  => '75000',
-    size_critical => '100000',
+    size_warning     => '75000',
+    size_critical    => '100000',
+    latency_warning  => '5',
+    latency_critical => '10',
+    ;
+                        [ 'delivery_immediate_high',
+                          'delivery_immediate']:
+    size_warning     => '75000',
+    size_critical    => '100000',
+    latency_warning  => '2.5',
+    latency_critical => '5',
+    ;
+                        [ 'delivery_digest']:
+    size_warning     => '75000',
+    size_critical    => '100000',
+    latency_warning  => '90',
+    latency_critical => '60',
     ;
   }
 
--- a/modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
+++ b/modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
@@ -1,4 +1,4 @@
-# == Define: govuk::apps::email_alert_api::sidekiq_queue_checks
+# == Define: govuk::apps::email_alert_api::sidekiq_queue_check
 #
 # Creates an Icinga check which checks Graphite for activity
 # on Sidekiq queues
@@ -18,9 +18,17 @@
 # [*size_critical*]
 #   The number of messages on a queue to trigger a critical alert.
 #
+# [*latency_warning*]
+#   The warning threshold for a sidekiq queue in minutes.
+#
+# [*latency_critical*]
+#   The critical threshold for a sidekiq queue in minutes.
+#
 define govuk::apps::email_alert_api::sidekiq_queue_check(
   $size_warning,
   $size_critical,
+  $latency_warning,
+  $latency_critical,
 ) {
   icinga::check::graphite { "check_email_alert_api_${title}_queue_size":
     target    => "transformNull(averageSeries(keepLastValue(stats.gauges.govuk.app.email-alert-api.*.workers.queues.${title}.enqueued)), 0)",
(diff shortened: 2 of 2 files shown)


modules/govuk/manifests/apps/email_alert_api/checks.pp
```diff
@@ -4,8 +4,20 @@
...
+  $queue_resources = {
+    'delivery_immediate_high'     => { latency_warning  => '2.5', latency_critical => '5' },
+    'delivery_immediate'          => { latency_warning  => '2.5', latency_critical => '5' },
+    'delivery_digest'             => { latency_warning  => '90', latency_critical => '60' },
+    'email_generation_immediate'  => { },
+    'process_and_generate_emails' => { },
+    'default'                     => { },
+    'email_generation_digest'     => { },
+    'cleanup'                     => { },
+  },
```
cbaines: Does this need to be a parameter (for example, does it vary per environment), or could this look something like this [1]?

1: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/email_alert_api.pp#L204-L219

modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_checks.pp
```diff
@@ -0,0 +1,30 @@
...
+# === Parameters
+#
+# [*ensure*]
+#   Whether to enable the check.
+#   Default: present
+#
+# [*title*]
+#   The name of the sidekiq queues to look for.
+#
+define govuk::apps::email_alert_api::sidekiq_queue_checks(
```
cbaines: Given this check's one queue, I'd call this `sidekiq_queue_check`.

modules/govuk/manifests/apps/email_alert_api/checks.pp
```diff
@@ -6,6 +6,28 @@
...
+    ;
+    'process_and_generate_emails':
+    ;
+    'default':
+    ;
+    'email_generation_digest':
+    ;
+    'cleanup':
+    ;
+  }
```
cbaines: This is looking really nice :+1: 

One thing I spotted when looking in to this Puppet resource declaration stuff is that you can have default values specified here [1].

1: https://puppet.com/docs/puppet/latest/lang_resources.html#resource-declaration-default-attributes

So, it might be possible to move the defaults from the `sidekiq_queue_check`, and make the latency parameters mandatory, and just specify all the values here, which might improve readability.

I think this is fine as well though, so don't feel like you have to change it.

modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
```diff
@@ -0,0 +1,29 @@
...
+#   Whether to enable the check.
+#   Default: present
+#
+# [*title*]
+#   The name of the sidekiq queues to look for.
+#
+define govuk::apps::email_alert_api::sidekiq_queue_check(
+  $ensure = present,
+  $latency_warning = '5',
+  $latency_critical = '10',
```
cbaines: These two latency parameters aren't documented in the comment above.

modules/govuk/manifests/apps/email_alert_api/checks.pp
deborahchua: Thanks @cbaines , this does look much easier to read so I'll use this approach 😄

modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
```diff
@@ -0,0 +1,37 @@
...
+# [*latency_warning*]
+#   The warning threshold for a sidekiq queue in minutes.
+#   Default: 5
+#
+# [*latency_critical*]
+#   The critical threshold for a sidekiq queue in minutes.
+#   Default: 10
+#
+define govuk::apps::email_alert_api::sidekiq_queue_check (
+  $ensure = present,
```
cbaines: `$ensure` is unused here.

You could pass it through to the `icinga::check::graphite` resource, but that isn't actually necessary as Puppet will automatically remove old configuration. I'd perhaps just remove it as a parameter.

modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
```diff
@@ -0,0 +1,37 @@
...
+  $latency_warning = undef,
+  $latency_critical = undef,
```
cbaines: As these `latency_...` parameters are required by `icinga::check::graphite`, I'd just make them mandatory (rather than defaulting to 'undef') if that's possible.

modules/govuk/manifests/apps/email_alert_api/checks.pp
cbaines: Great :)

modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp
```diff
@@ -18,9 +18,17 @@
...
 #
+# [*latency_warning*]
+#   The warning threshold for a sidekiq queue in minutes.
+#
+# [*latency_critical*]
+#   The critical threshold for a sidekiq queue in minutes.
+#
 define govuk::apps::email_alert_api::sidekiq_queue_check(
   $size_warning,
   $size_critical,
```
thomasleese:
```puppet
  $size_warning = '75000'
  $size_critical = '100000',
```

I wonder if it's worth doing this since they seem to be the same across all the queues.

cbaines: While that would save 4 lines of config, it's probably more advantageous to use the extra lines and have all the configuration in one place.

thomasleese: Sure, sounds reasonable. I wonder if we should at least pull them out into variables to avoid them getting out of sync between queues, but that's probably not a big deal.

PR #10074, opened 2020-01-20
Add link to docs for sitemap info

chao-xian:
The alert should have the link to the docs for this showing up and not rely on a dev to search for it manually.

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/search_generate_sitemaps.pp
+++ b/modules/govuk_jenkins/manifests/jobs/search_generate_sitemaps.pp
@@ -20,6 +20,6 @@ class govuk_jenkins::jobs::search_generate_sitemaps{
     host_name           => $::fqdn,
     freshness_threshold => 129600,
     action_url          => $job_url,
-    notes_url           => monitoring_docs_url(runs-rake-task-search-api)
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/manifests/jobs/search_generate_sitemaps.pp
```diff
@@ -20,5 +20,6 @@
     host_name           => $::fqdn,
     freshness_threshold => 129600,
     action_url          => $job_url,
+    notes_url           => 'https://docs.publishing.service.gov.uk/manual/alerts/runs-rake-task-search-api.html'
```
thomasleese:
```suggestion
    notes_url           => monitoring_docs_url(runs-rake-task-search-api)
```

You can use this instead to avoid needing the entire URL, for example:
https://github.com/alphagov/govuk-puppet/blob/b588e4ade996e97b8975e69cb00800521fff4a48/modules/mongodb/manifests/monitoring.pp#L41

It's defined here: https://github.com/alphagov/govuk-puppet/blob/b588e4ade996e97b8975e69cb00800521fff4a48/modules/govuk/lib/puppet/parser/functions/monitoring_docs_url.rb

PR #10075, opened 2020-01-20
Remove s3_resources from ckan.ini

kentsanggds:
Code has been migrated to datagovuk extension so the extension and config settings are no longer needed


issyl0:
This is a partial duplicate of #10058.

kentsanggds:
I'll close this PR down and will use the other as I think its the same thing

PR #10076, opened 2020-01-20
[DO NOT MERGE] Remove Whitehall + Asset Manager from /etc/hosts in Carrenza.

sengi:
This is for migrating these apps from Carrenza to AWS. These particular
hosts entries need to be deleted as part of the migration because
otherwise they shadow the names which clients in Carrenza will use to
connect to Whitehall / Asset Manager in AWS.

The hosts entries for the Whitehall / Asset Manager machines (as opposed
to the apps) must not be deleted as part of the migration but should be
removed afterwards along with the Puppet classes for Whitehall / Asset
Manager.

Based on and supersedes #9886 and #9887.

### DIFF
--- a/hieradata/production.yaml
+++ b/hieradata/production.yaml
@@ -192,7 +192,6 @@ hosts::production::api::hosts:
     ip: '10.3.12.31'
 
 hosts::production::backend::app_hostnames:
-  - 'asset-manager'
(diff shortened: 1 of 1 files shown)


sengi:
Combined into #10061. This is so that we have fewer PRs to worry about on the day.

PR #10078, opened 2020-01-20
Route the app-specific healthcheck through to whitehall-frontend.

sengi:
This is to provide a meaningful load balancer healthcheck.

Whitehall is a special case because it doesn't use some of the
nginx-related functionality of `govuk::app`. This just makes the minimal
reasonable change to get the healthcheck working rather than refactoring
Whitehall's Puppet class to take full advantage of `govuk::app`, while
keeping things as similar as possible to other apps / node classes.

See https://github.com/alphagov/govuk-aws/blob/master/doc/architecture/decisions/0037-alb-health-checks.md

### DIFF
--- a/modules/govuk/manifests/apps/whitehall.pp
+++ b/modules/govuk/manifests/apps/whitehall.pp
@@ -249,6 +249,16 @@ class govuk::apps::whitehall(
       asset_pipeline_prefix => 'government/assets',
     }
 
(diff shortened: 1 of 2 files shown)


sengi:
Tested in Integration by deploying Puppet from the branch, running Puppet agent on the `whitehall_backend` nodes and then running Smokey.

PR #10083, opened 2020-01-23
[Do not merge][WIP] Remove content audit tool

edwardkerry:
Content Audit Tool is a legacy application created before Content Data. 

[Trello](https://trello.com/c/V6bXwwtx/1700-3-retire-content-audit-app)


hieradata_aws/class/integration/db_admin.yaml
```diff
@@ -88,17 +88,6 @@ govuk_env_sync::tasks:
     temppath: "/tmp/content-register_production"
     url: "govuk-integration-database-backups"
     path: "postgresql-backend"
-  "push_content_audit_tool_production_integration_daily":
-    ensure: "present"
```
cbaines: Does removing this part mean that the relevant resources will be removed, or does `ensure` here need changing to absent first?

modules/govuk/manifests/apps/content_audit_tool.pp
```diff
@@ -1,165 +0,0 @@
...
-) {
-  $app_name = 'content-audit-tool'
-
-  govuk::app { $app_name:
-    app_type          => 'rack',
-    port              => $port,
-    sentry_dsn        => $sentry_dsn,
-    health_check_path => '/',
-    asset_pipeline    => true,
-  }
```
cbaines: I'm guessing that there's at least one resource here that won't be cleaned up by just removing the configuration here.

PR #10090, opened 2020-01-27
Give P Leaman SSH Access To Licensify Servers

camdesgov:
He is an Apto contractor that requires access for support
purposes. For further details please see this ticket:
https://trello.com/c/zJjXPQTs/3119-give-apto-contractor-licensify-access-rights-for-aws

### DIFF
--- a/modules/users/manifests/init.pp
+++ b/modules/users/manifests/init.pp
@@ -12,6 +12,14 @@
 #   List of pentest user accounts to add.
 #   Default: []
 #
+# [*licensify_machines*]
+#   List of machines which licensify contractors require access to.
+#   Default: []
+#
+# [*licensify_usernames*]
+#   List of licensify contractors user accounts to add.
+#   Default: []
+#
 # [*root_password_encrypted*]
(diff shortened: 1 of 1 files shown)


modules/users/manifests/init.pp
```diff
@@ -23,11 +23,13 @@
 class users (
   $pentest_machines = [],
   $pentest_usernames = [],
+  $licensify_machines = [],
```
sengi: Please add docstrings for these new variables.

modules/users/manifests/init.pp
```diff
@@ -70,4 +72,9 @@
     $pentest_user_classes = regsubst($pentest_usernames, '^', 'users::')
     include $pentest_user_classes
   }
+
+  if ($::aws_migration in $licensify_machines) {
```
sengi: This would be clearer as 
```suggestion
  if ($::hostname in $licensify_machines) {
```
like in the previous block.

camdesgov:
> LGTM. Still lots of tests failing though :(

Tests now passing :D

PR #10092, opened 2020-01-28
Fix totals on Email Alert API dashboard

kevindew:
These were showing the incorrect values as they are using statsd's
`stats.` value rather that `stats_counts`. stats is an aggregate value
of the rate per second where stats_counts is the actual total.

### DIFF
--- a/modules/grafana/files/dashboards/email_alert_api.json
+++ b/modules/grafana/files/dashboards/email_alert_api.json
@@ -104,7 +104,7 @@
           "targets": [
             {
               "refId": "A",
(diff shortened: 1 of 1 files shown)


kevindew:
> 🥇 - thanks for spotting!

No worries. I've sat here for a little while trying to work out if I'm sure that the requests per minute graph is correct. I'm pretty sure it is now as it'll be the count of every send across each box summed together and then grouped into the total per minute. I ended up a bit confused as we the rate from the stats value (rather than stats_count) might be more appropriate at a per second value. 

Just leaving this here in case you want to verify too as it can get rather tricky with these counts and indicate my confidence in that isn't 100%

benthorner:
Ah, I think I understand: sometimes I the term "rate" to mean "per second", but the total over a minute is actually the rate if the summation period is also a minute.

PR #10098, opened 2020-01-29
Remove some unneeded ES stuff

barrucadu:


Puppetfile.lock
```diff
@@ -43,9 +43,6 @@ FORGE
     puppet-unattended_upgrades (2.1.0)
       puppetlabs-apt (< 3.0.0, >= 2.2.0)
       puppetlabs-stdlib (< 5.0.0, >= 4.6.0)
-    puppet-yum (2.2.1)
```
barrucadu: There's some unrelated-looking stuff here, but this is what `bundle exec rake librarian:package` gave me

PR #10104, opened 2020-01-29
Retire content audit tool (1 of 2)

edwardkerry:
Content Audit Tool is a legacy application created before Content Data. It was used to help tag content - it shared a codebase with Content Performance Manager.

It is being retired as it should no longer be used, to clean up references to it in our codebase, and remove the need for developers to maintain it.

[Trello](https://trello.com/c/V6bXwwtx/1700-3-retire-content-audit-app)

### DIFF
--- a/development-vm/Procfile
+++ b/development-vm/Procfile
@@ -156,13 +156,9 @@ link-checker-api-sidekiq:             govuk_setenv link-checker-api            .
 # sidekiq-monitoring for specialist-publisher uses port 3210
 # sidekiq-monitoring's web frontend listens on port 3211 and proxies requests per app to other ports (defined elsewhere in this file)
 publishing-components:                govuk_setenv publishing-components       ./run_in.sh ../../govuk_publishing_components bundle exec foreman start
-content-audit-tool: govuk_setenv content-audit-tool ./run_in.sh ../../content-audit-tool bundle exec rails server -p 3213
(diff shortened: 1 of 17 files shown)


development-vm/Procfile
```diff
@@ -156,13 +156,9 @@ link-checker-api-sidekiq:             govuk_setenv link-checker-api            .
 # sidekiq-monitoring for specialist-publisher uses port 3210
 # sidekiq-monitoring's web frontend listens on port 3211 and proxies requests per app to other ports (defined elsewhere in this file)
 publishing-components:                govuk_setenv publishing-components       ./run_in.sh ../../govuk_publishing_components bundle exec foreman start
-content-audit-tool: govuk_setenv content-audit-tool ./run_in.sh ../../content-audit-tool bundle exec rails server -p 3213
```
issyl0: The [docs for this part](https://docs.publishing.service.gov.uk/manual/retiring-an-application.html#7-update-development-scripts) say:

> Leave a comment in the Procfile indicating that the port used to be used by this app, to avoid port conflicts causing a problem running this app locally in the future.

But that feels a bit churn-for-the-sake-of-churn these days when the dev VM is being deprecated.

edwardkerry: I agree - I did see that and thought it would be better to update the docs

PR #10113, opened 2020-01-31
Add Icinga check for freshness of GOV.UK Fastly CDN logs in S3.

sengi:
### What

* Add an Icinga check which runs hourly and checks whether there are sufficiently up-to-date log files under the S3 prefix(es) for the relevant day(s).
* The check is based on the date/time encoded into the filename, rather than the modification time of the file.
* The S3 bucket contains on the order of 1M objects, so we don't really want to list the entire bucket all the time.
* Install checks for `govuk_assets`, `govuk_www` and `bouncer` logs.

### Why

If Fastly logs aren't making it into S3, we can't fix the gap by reprocessing. We therefore want to know about the issue as soon as reasonably possible (while avoiding false alarms).

This was prompted by an incident where the Fastly->Glue->Athena CDN logs pipeline was broken by a change to the log format.

### Alternatives considered

Since we're planning on switching to Prometheus soon, it would be nice to do at least the metrics side of this in Prometheus and have just a stub check to make the alert appear in Icinga until the switchover. While that would be the ideal outcome, I think to do it justice would require configuring a bunch of things for the first time (like setting up the necessary gubbins to build and deploy a fork of a third party Prometheus exporter) — so I'm inclined to start with this simple Icinga check and iterate from there. I'm fine with this being a short-lived thing.

The logic of the check is very basic: it just looks for any files with the prefix corresponding to the previous hour (or however many hours back it's configured to look). I think this is probably good enough, but we could make it more precise by parsing the timestamp out of the filenames we find. The downside of that would be that there's slightly more to go wrong with the check (for example if the form of the filenames were to change).

### Related PRs
* Permissions change to enable this to work: alphagov/govuk-aws#1244
* Playbook entry: alphagov/govuk-developer-docs#2285

### DIFF
--- a/modules/monitoring/files/usr/lib/nagios/plugins/check_cdn_log_s3_freshness
+++ b/modules/monitoring/files/usr/lib/nagios/plugins/check_cdn_log_s3_freshness
@@ -19,16 +19,21 @@ For help on command-line options:
     check_cdn_log_s3_freshness -h
 
 Example usage for testing the script:
-    check_cdn_log_s3_freshness -e staging -t 2020-01-31T17:00
+    check_cdn_log_s3_freshness -e staging -F 2020-01-31T17:00
 
 Example usage in production:
-    check_cdn_log_s3_freshness -e production -l govuk_www -c 1
+    check_cdn_log_s3_freshness -e production -l govuk_www -c 60
+
+Doctest doesn't work when the module filename doesn't end in .py, so in order
+to run doctest you have to rename the file and then run:
+    python3 -m check_cdn_log_s3_freshness.py
 """
 
 import argparse
 import datetime
 import enum
 import logging
+import posixpath
 import sys
 
 import boto3
@@ -46,8 +51,12 @@ def fromisoformat(s):
 
     This is only meant for passing in a fake time from the command-line when
     testing. Would use datetime.fromisoformat, but we're stuck on Python 3.4.
+
+    Example:
+    >>> fromisoformat('2020-01-30T22:00')
+    datetime.datetime(2020, 1, 30, 22, 0, tzinfo=datetime.timezone.utc)
     """
-    return datetime.datetime.strptime(s, '%Y-%m-%dT%H:%M')
+    return datetime.datetime.strptime(s + '+0000', '%Y-%m-%dT%H:%M%z')
 
 
 def parse_args():
@@ -57,10 +66,10 @@ def parse_args():
                         help='Environment to check: integration, staging, production.')
     parser.add_argument('-l', '--log_type', default='govuk_assets',
                         help='Which logs to check: govuk_assets, govuk_www.')
-    parser.add_argument('-c', '--critical_age_hours', type=int, default=1,
-                        help='If the newest logs are older than this many hours, '
+    parser.add_argument('-c', '--critical_age_minutes', type=int, default=60,
+                        help='If the newest logs are older than this many minutes, '
                              'return CRITICAL status.')
-    parser.add_argument('-t', '--time', type=fromisoformat,
+    parser.add_argument('-F', '--fake_time', type=fromisoformat,
                         help='For testing purposes, use the given time as if it\'s the current '
                              'time. Requires the format YYYY-MM-DDTHH:MM. Assumes UTC.')
     parser.add_argument('-v', '--verbose', action='count',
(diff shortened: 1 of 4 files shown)


sengi:
Tested in Integration by deploying the Puppet branch, seeing that the check fails because it doesn't have permission to read the bucket, deploying the Terraform change in Integration to add the permission, then seeing that the check passes.

Adjusted the retry interval down from 60 to 5 minutes so that it doesn't take more than 20 minutes to go from soft-fail to hard-fail.

modules/monitoring/manifests/checks.pp
```diff
@@ -162,6 +163,29 @@
...
+    $cdn_log_age_threshold_hours = $::aws_environment ? {
+      'production' => 1,
+      # Staging and Integration don't produce any logs between 22:00 and 04:00
+      # when traffic replay isn't running. Fastly doesn't send us empty files.
+      default      => 6,
+    }
+    icinga::check { 'check_cdn_log_s3_freshness_assets':
+      check_command       => "check_cdn_log_s3_freshness!-e ${::aws_environment} -l govuk_assets -c ${cdn_log_age_threshold_hours}",
+      host_name           => $::fqdn,
+      service_description => "check that Fastly logs from ${cdn_log_age_threshold_hours}h ago appear in s3://govuk-${::aws_environment}-fastly-logs/govuk_assets",
```
sarahseewhy: Random question: do we have to be judicious about `service_description` character length? This one is fine, I'm more curious overall.

Context: Where I previously work we ran into an odd Nagios issue which I'd like to know if you folks have run into for Icinga.

I believe under the hood Nagios saves the `service_description` values to a file, but that file has a character limit. If the limit is exceeded Nagios borks with an unhelpful error message. It could've just been the version of Nagios we were using.

sengi: Chatted about this in person and we don't think it'll be a problem.

modules/monitoring/files/usr/lib/nagios/plugins/check_cdn_log_s3_freshness
```diff
@@ -0,0 +1,122 @@
...
+
+def get_prefix(log_type, critical_age_hours, fake_time=None):
+    """Return the S3 prefix to be searched for log files, based on the time."""
+    max_age = datetime.timedelta(hours=critical_age_hours)
+    now = fake_time or datetime.datetime.now(datetime.timezone.utc)  # TZ-aware datetime in UTC.
+    query_time = now - max_age  # Oldest acceptable last_modified time for logs.
+    time_prefix = query_time.strftime('year=%Y/month=%m/date=%d/%Y-%m-%dT%H')
+    logging.info('Time is %s%s', query_time.isoformat(),
+                 ' (overridden for testing)' if fake_time else '')
+    return '%s/%s' % (log_type, time_prefix)
```
kevindew: I'm wondering if we might have an integration and staging issue where it looks up data that doesn't exist. I'm not sure I understand exactly how this works so I'll explain my understanding in case that is flawed.

So I understand that every 60s the icinga check runs. Assuming the time is 19:03 today and we are in production we would send a query to S3 to check whether there are any files in the bucket with the following prefix: "year=2020/month=02/date=14/2020-02-14T18" where the 18 is because it is current time subtracted by one hour.

So assuming I'm correct there (which is a big assumption 😄) if we assume we're in integration and there is no logs between 2:00am and 3:00am. If the time is 8:15am on February 14th then would we not be querying for a prefix subtracted by 6 hours? which would be  "year=2020/month=02/date=14/2020-02-14T02" which is a time where we have 0 results?

sengi: I think you're absolutely right there. My initial approach was too simplistic and indeed it was just plain broken in the Integration case 🤦‍♂  Thanks for spotting it and for the nice clear explanation!

I've reworked it to parse the time of the latest log file based on the filename and compare that against the threshold. Sorry it's taken me ages to get around to it. As a bonus, we've now got better info logging and 1-minute precision instead of 1 hour.

Ready for another look, if you have a spare moment. Thanks!

modules/monitoring/manifests/checks.pp
```diff
@@ -162,6 +163,29 @@
...
+      retry_interval      => 5,
+    }
+    icinga::check { 'check_cdn_log_s3_freshness_www':
+      check_command       => "check_cdn_log_s3_freshness!-e ${::aws_environment} -l govuk_www -c ${cdn_log_age_threshold_hours}",
+      host_name           => $::fqdn,
+      service_description => "check that Fastly logs from ${cdn_log_age_threshold_hours}h ago appear in s3://govuk-${::aws_environment}-fastly-logs/govuk_www",
+      notes_url           => monitoring_docs_url(cdn-log-freshness),
+      check_interval      => 60,
+      retry_interval      => 5,
+    }
```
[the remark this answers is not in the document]
sengi: I'd forgotten about Bouncer. Added (for prod only). Thanks!

sengi:
Thanks! I tried out the previous version in Integration but I'll re-test before merging.

sengi:
Rebasing against master in order to test the branch in Integration.

sengi:
Test results from Integration:
```
INFO:root:Current time is 2020-02-27T12:49:59.224349+00:00.
INFO:root:Looking for log filenames dated 2020-02-27T06:19:59.224349+00:00 or newer.
INFO:root:S3 bucket name: govuk-integration-fastly-logs
INFO:root:Found 762 files with prefix govuk_assets/year=2020/month=02/date=27/
INFO:root:Newest log file (based on filename) is 2020-02-27T12:30:00.000-ykS9MPsBbqSln_rFzMgL.log.gz
INFO:root:Newest log time parsed from filename is 2020-02-27T12:30:00+00:00
OK
```

```
INFO:root:Current time is 2020-02-27T12:50:05.423147+00:00.
INFO:root:Looking for log filenames dated 2020-02-27T06:20:05.423147+00:00 or newer.
INFO:root:S3 bucket name: govuk-integration-fastly-logs
INFO:root:Found 858 files with prefix govuk_www/year=2020/month=02/date=27/
INFO:root:Newest log file (based on filename) is 2020-02-27T12:30:00.000-wivVnILuoODjXXDt9f4i.log.gz
INFO:root:Newest log time parsed from filename is 2020-02-27T12:30:00+00:00
OK
```

PR #10116, opened 2020-02-04
Remove docker-frontend node class

philandstuff:
No servers exist with this node class; so the puppet for it can safely
be removed, hopefully?


thomasleese:
I think you'll need to remove this file as well: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_containers/spec/classes/govuk_containers__memcached_spec.rb

sengi:
Looks like the memcache class is still in use in `govuk_ci/manifests/agent` so might need to keep that bit for now? (we already chatted about this out-of-band, just noting here for anyone else following along)

philandstuff:
@sengi done

PR #10122, opened 2020-02-04
Add an env_sync pull job for release in aws staging

whi-tw:
This is cribbed from:
https://github.com/alphagov/govuk-puppet/blob/5877b0fe9e925b4eedee410eb2932df0c573a4a2/hieradata_aws/class/training/db_admin.yaml#L178-L188

and will ingest the backups created in production by:

https://github.com/alphagov/govuk-puppet/blob/7621c13cc81103ea851d3e719f8a56025a71aaf4/hieradata/node/mysql-backup-1.backend.publishing.service.gov.uk.yaml#L2-L12

### DIFF
--- a/hieradata_aws/class/staging/db_admin.yaml
+++ b/hieradata_aws/class/staging/db_admin.yaml
@@ -144,6 +144,17 @@ govuk_env_sync::tasks:
     temppath: "/tmp/whitehall_production"
     url: "govuk-production-database-backups"
     path: "mysql"
+  "pull_release_production_daily":
+    ensure: "present"
+    hour: "2"
+    minute: "15"
+    action: "pull"
+    dbms: "mysql"
+    storagebackend: "s3"
+    database: "release_production"
+    temppath: "/tmp/release_production"
+    url: "govuk-production-database-backups"
+    path: "mysql"
   "pull_govuk_assets_production":
     ensure: "present"
     hour: "2"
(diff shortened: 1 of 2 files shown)


hieradata_aws/class/staging/db_admin.yaml
```diff
@@ -144,6 +144,17 @@ govuk_env_sync::tasks:
...
+  "pull_release_staging_daily":
+    ensure: "present"
+    hour: "2"
+    minute: "15"
+    action: "pull"
+    dbms: "mysql"
+    storagebackend: "s3"
+    database: "release_production"
+    temppath: "/tmp/release_production"
+    url: "govuk-staging-database-backups"
```
sengi: This needs to download the Prod backup, right?
```suggestion
    url: "govuk-production-database-backups"
```

Maybe also update the title to "pull_release_production_daily" since that seems to match the surrounding ones? \shrug

sengi:
Ah — the backup in Carrenza is only running in Staging at the moment. We'll want to run it in Prod there and pull the output of that into AWS Staging.

hieradata_aws/class/staging/db_admin.yaml
whi-tw: AGREED. I looked at the wrong file. This will be rolled in to my next changes

PR #10123, opened 2020-02-04
Remove Unnecessary Reference to Mongodb Nodes

camdesgov:
This is already being added through govuk secrets.


sengi:
Just need to add the key to CARRENZA_ONLY_KEYS in the Rakefile to fix the test.

issyl0:
Alternative: Why not make the public repo be the canonical source instead of the secrets repo?

sengi:
> Alternative: Why not make the public repo be the canonical source instead of the secrets repo?

That'd be fine too. The only reason for putting the hostnames in the secrets repo is that it keeps all the connection details in one place, now that we have to store a username and password. (The old setup was unauthenticated so that didn't use to be an issue.)

camdesgov:
Thanks for the input guys. I'm closing this PR for now, I've added a card to our Trello board to revisit this as part of the post migration clean up for Asset Manager/ Whitehall.

PR #10126, opened 2020-02-06
Delete `govuk_cdnlogs` code as it's superseded by AWS S3 and Athena

issyl0:
- Pulling these from Fastly into files on disk was superseded in [2018](https://github.com/alphagov/govuk-rfcs/pull/93).
- This code and its docs were rediscovered in https://github.com/alphagov/govuk-developer-docs/pull/2285#issuecomment-582373234 and it was confusing.

This is draft because a) I'm not sure I've got everything and b) should we do this?

### DIFF
--- a/modules/govuk_cdnlogs/manifests/init.pp
+++ b/modules/govuk_cdnlogs/manifests/init.pp
@@ -30,8 +30,8 @@
 
 class govuk_cdnlogs (
   $log_dir,
-  $govuk_monitoring_enabled = true,
-  $bouncer_monitoring_enabled = true,
+  $govuk_monitoring_enabled = false,
+  $bouncer_monitoring_enabled = false,
   $warning_cdn_freshness = 1800,
   $critical_cdn_freshness = 3600,
   $server_key,
@@ -47,14 +47,14 @@ class govuk_cdnlogs (
   $crt_file = '/etc/ssl/rsyslog.crt'
 
   file { $key_file:
-    ensure  => file,
+    ensure  => absent,
     mode    => '0400',
     owner   => 'root',
     content => $server_key,
     notify  => Class['rsyslog::service'],
   }
   file { $crt_file:
-    ensure  => file,
+    ensure  => absent,
     mode    => '0400',
     owner   => 'root',
     content => $server_crt,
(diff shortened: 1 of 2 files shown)


issyl0:
They do probably still exist. I'll re-make this with some `ensure => absent` tomorrow, unless you know which ones they are?

sengi:
> They do probably still exist. I'll re-make this with some `ensure => absent` tomorrow, unless you know which ones they are?

I don't know for sure. `ensure => absent` SGTM. (If you prefer to just remove stuff by hand then that's fine with me, too - whatever you think is easiest.)

Thanks for tidying this up btw 🌟

issyl0:
I deployed this branch to integration and it seems to be fine. 👍

kevindew:
In the RFC I remember there were people who keen to keep the sending of logs to our monitoring machines.  There is a timing advantage to those sent to monitoring machine. I believe they are streamed at something close to real time via logstash, whereas the athena monitoring has a delay of between 15 and 30 minutes for data.

I don't really think many people (if anyone) really uses the ones from the monitoring machine, but it also feels like the sort of thing where it could be not frustrating not to have in the midst of an incident.

issyl0:
@kevindew Thanks for that. As we discussed in Slack, only about three people who have been here long enough would remember that there are also useable files on the monitoring machine - indeed all the incident docs point to Athena. I'm not sure what we should do here, but you raise valid points. 🤔

kevindew:
> @kevindew Thanks for that. As we discussed in Slack, only about three people who have been here long enough would remember that there are also useable files on the monitoring machine - indeed all the incident docs point to Athena. I'm not sure what we should do here, but you raise valid points. 🤔

Yup, can see it's not ideal to be caught in limbo on how to move forward.

Given we think that a) no-one uses these files, b) only a few people know about them and c) we've done the work to remove them; then I think we should proceed with the removal but leave some information in the docs about their past existence, to help us if we decide we want them back.

So I'd suggest perhaps changing the second paragraph of https://docs.publishing.service.gov.uk/manual/query-cdn-logs.html#header to instead reference that we no longer have logs under 15 minutes due to low usage and link to this PR as a means to restore them if we need to.

How does that sound?

issyl0:
Sounds good, thanks @kevindew! I've done that in https://github.com/alphagov/govuk-developer-docs/pull/2331.

hieradata/class/monitoring.yaml
```diff
@@ -1,7 +1,5 @@
 ---
 
-govuk_cdnlogs::log_dir: '/var/log/cdn'
```
kevindew: Do we need to keep this so that it is used as input for the ensure => absent?

issyl0: I trust your knowledge of Puppet more than mine! Reinsated.

modules/govuk_cdnlogs/manifests/init.pp
```diff
@@ -31,7 +31,7 @@
 class govuk_cdnlogs (
   $log_dir,
   $govuk_monitoring_enabled = true,
```
kevindew: Do you want to set this to false too, given we're disabling bouncer monitoring?

issyl0: Done!

PR #10135, opened 2020-02-13
Import check_elasticsearch_aws.py script from nagios-plugins repo

barrucadu:
We're archiving that.


issyl0:
We can make this Python 3 at a later date. 😬

barrucadu:
I added an `@icinga::plugin` bit, I'm not sure if that's necessary though

PR #10142, opened 2020-02-13
Fix check scripts' `nagios_*` methods to be useable

issyl0:
- In `alphagov/nagios-scripts`, the output functions were [extracted
  into their own module that was
  `import`ed](https://github.com/alphagov/nagios-plugins/commit/1ec7bdc5eb816a90bf4ffb37c7e5ae73ce303702).
- We need those functions in each script (we can refactor it later to be
  better?) for the outputs to work.

### DIFF
--- a/modules/monitoring/files/usr/lib/nagios/plugins/check_apt_security_updates
+++ b/modules/monitoring/files/usr/lib/nagios/plugins/check_apt_security_updates
@@ -3,11 +3,31 @@ import string
 import subprocess
 import sys
 
-from plugins.output import (CheckException,
-                            nagios_ok,
-                            nagios_warning,
-                            nagios_critical,
-                            nagios_message)
+class CheckException(Exception):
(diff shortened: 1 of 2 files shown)


modules/monitoring/files/usr/lib/nagios/plugins/check_apt_security_updates
```diff
@@ -3,13 +3,37 @@ import string
...
+    """Nagios WARNING message"""
+    raise CheckException("WARNING: %s" % message, 1)
+
+
+def nagios_critical(message):
+    """Nagios CRITICAL message"""
+    raise CheckException("CRITICAL: %s" % message, 2)
+
+
+def nagios_unknown(message):
```
barrucadu: This function isn't used in this file

issyl0: Thanks! Fixed.

PR #10143, opened 2020-02-13
Add AWS settings for specialist_publisher to use s3

benjamineskola:
specialist_publisher will need AWS keys in order to upload CSVs to S3 (see [alphagov/specialist-publisher#1576](https://github.com/alphagov/specialist-publisher/pull/1576)).

https://trello.com/c/yrMNgb9T/1714-3-specialist-publisher-make-documentlistexport-use-s3-instead-of-email-attachments


issyl0:
The secrets for the IAM user now exist: https://github.com/alphagov/govuk-secrets/pull/907.

PR #10145, opened 2020-02-14
Remove the `gds-nagios-plugins` package

issyl0:
- We've moved everything that used to be here out of
  alphagov/nagios-plugins and into the plugins directory of this repo.

This is draft because we need #10144 merged and deployed first.


issyl0:
Wait, we probably need to `pip uninstall` this otherwise it'll linger forever.

issyl0:
Good point. I removed the version constriaint.

PR #10147, opened 2020-02-14
Set HOSTFILTER only if TARGET_MACHINES is set.

sengi:
Fixes bug introduced in #10146 which broke all-nodes deployments.

Tested by applying the change on the Integration Jenkins and running an all-nodes and a single-node deploy. In both cases it selected the right nodes as well as the job succeeding.

I also verified that the conditional does the right thing for `TARGET_MACHINES=""`, `TARGET_MACHINES=all` and `TARGET_MACHINES="foo, bar"`.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
@@ -20,7 +20,9 @@
             export DEPLOY_TO="<%= @environment -%>"
             export DEPLOY_TASK="$DEPLOY_TASK"
             export TAG="$TAG"
-            export HOSTFILTER="$TARGET_MACHINES"
+            if [ "${TARGET_MACHINES:-all}" != "all" ]; then
+              export HOSTFILTER="$TARGET_MACHINES"
+            fi
             export ORGANISATION="<%= @environment -%>"
             export CI_DEPLOY_JENKINS_API_KEY="<%= @ci_deploy_jenkins_api_key -%>"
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
```diff
@@ -20,7 +20,9 @@
             export DEPLOY_TO="<%= @environment -%>"
             export DEPLOY_TASK="$DEPLOY_TASK"
             export TAG="$TAG"
-            export HOSTFILTER="$TARGET_MACHINES"
+            if [ "${TARGET_MACHINES:-all}" != "all" ]; then
```
issyl0: `if the default of "all" != "all"`?

issyl0: I get its purpose, it just hurt my brain.

sengi: The default from the Jenkins job is "all", but we also want to handle the case where it's empty. So if it's empty, we default to "all" before doing the comparison.

sengi: Added a comment, cos I can now see this line is kinda confusing 😅 Thanks for pointing it out.

sengi: Oh wait, putting a comment there isn't gonna work without some hairy YAML escaping. Maybe I'd better not :(

PR #10149, opened 2020-02-17
Revert "Adding env-sync to carrenza production mongo instances"

sengi:
Revert #10148 because it does two unintended things:

* Makes the govuk_env_sync backup run in triplicate, simultaneously on all MongoDB nodes. It doesn't just run on the node that's currently the leader.
* Makes the old `mongodb::backup::s3_backups` backup job run on all three nodes as well. This also isn't supposed to happen.

It's ok for us not to run the govuk_env_sync job automatically from Carrenza because we've tested it several times now anyway. At this point, it's not worth enabling it as we'll only be turning it off again after the migration tomorrow.

This doesn't prevent us from running the backup job during the migration.

Reverts alphagov/govuk-puppet#10148


sengi:
Updated the description because I'd misread the direction of the `ensure: disabled` change in the diff. The rollback was still needed though because of the unintended change to `mongodb::backup::s3_backups`.

PR #10154, opened 2020-02-18
Create a Jenkins job to query Athena CDN logs

kevindew:
Trello: https://trello.com/c/93jgKiUX/1329-set-up-a-smoke-test-that-athena-is-still-operating-on-new-data

This is a scheduled task that is due to run in Jenkins 3 times a day to
verify that Athena has fresh data. This check is being added to monitor
that we are a) still receiving data, b) successfully indexing data and
c) said data is queryable. It is being added as a result of a situation
that occurred before Christmas where we discovered that Athena had not
been queryable for 6 weeks due to a JSON error.

This job intends to monitor the govuk_www and govuk_assets databases in
integration, staging and production. It also will monitor the bouncer
database in production only, as we do not have the same bouncer set-up
running in integration and staging.

In the integration and staging environments we have a regular flow of
traffic due to the gor traffic replay that runs from production. This
allows us to be confident that there will be fresh data in these
environments. The gor replay does however get paused during the nightly
data sync so there are periods from 10pm to 4am where there may be 0
traffic from these environments. The job timings are therefore set-up
to be respectful of this and to run 3 times a day at times close to
business hours. I also chose to do the exact same thing
in all 3 environments rather than configuring production different as I
value parity between environments and like to minimise the risk of
problems caused by configuration changes. These checks do run on the
predicate that it would be unheard of for the GOV.UK CDN to go 4 hours
without a request.

I chose to try make a single jenkins job serve for all 3 checks. I don't
think this is a particularly common approach for a GOV.UK Jenkins check
but it feels less messy that 2 or 3 near identical jobs (as is the case
with email alerts). This has meant a few niche techniques such as an
environment variable in the icinga check, linking to a Jenkins job with
a filter parameter and using the parameterized scheduler plugin.

In order to set-up Icinga passive checks for each database I've had to
create an additional defined resource so that we have means to iterate
through the array of databases [1]. Our linter also forced me to do
this in a new file. Hopefully in the not-too-distant future we can
replace this need for an additional defined resource with a loop in the
govuk_jenkins::jobs::athena_fastly_logs_check class.

I've this out on integration and it was successfully able to query both
databases and report health to icinga.

[1]: https://stackoverflow.com/a/13008766

### DIFF
--- a/hieradata_aws/class/production/jenkins.yaml
+++ b/hieradata_aws/class/production/jenkins.yaml
@@ -1,5 +1,6 @@
 ---
 govuk_jenkins::job_builder::jobs:
+  - govuk_jenkins::jobs::athena_fastly_logs_check
   - govuk_jenkins::jobs::bouncer_cdn
   - govuk_jenkins::jobs::check_sentry_errors
   - govuk_jenkins::jobs::clear_cdn_cache
--- a/modules/govuk_jenkins/manifests/jobs/athena_fastly_logs_check.pp
+++ b/modules/govuk_jenkins/manifests/jobs/athena_fastly_logs_check.pp
@@ -1,0 +1,51 @@
+# == Class: govuk_jenkins::jobs::athena_fastly_logs_check
+#
+# Create a file on disk that can be parsed by jenkins-job-builder
+#
+# === Parameters
+#
+# [*aws_access_key_id*]
+#   Access key id for user who can read from Athena and write to the bucket
+#
+# [*aws_secret_access_key*]
+#   Secret access key for the above user
+#
+# [*aws_region*]
+#   The region athena is running in
+#
+# [*databases*]
+#   An array of fastly logs databases to check for results
+#
+# [*s3_results_bucket*]
+#   An S3 bucket where the query results can be written too
+#
+class govuk_jenkins::jobs::athena_fastly_logs_check (
+  $aws_access_key_id = undef,
+  $aws_secret_access_key = undef,
+  $aws_region = 'eu-west-1',
+  $databases = ['govuk_www', 'govuk_assets'],
+  $s3_results_bucket = undef,
+) {
+  if $::aws_migration and ($::aws_environment != 'integration') {
+    $hosting_env_domain = "blue.${::aws_environment}.govuk.digital"
+  }
+  else {
+    $hosting_env_domain = hiera('app_domain')
+  }
+
+  $jenkins_url = "https://deploy.${hosting_env_domain}/job/athena-fastly-logs-check/"
+  $jenkins_service_description = "Athena has recent results for fasty_logs \${DATABASE}"
+
+  file {
+    '/etc/jenkins_jobs/jobs/athena_fastly_logs_check.yaml':
+      ensure  => present,
+      content => template('govuk_jenkins/jobs/athena_fastly_logs_check.yaml.erb'),
+      notify  => Exec['jenkins_jobs_update'];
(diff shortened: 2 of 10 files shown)


modules/govuk_jenkins/manifests/jobs/athena_fastly_logs_check/icinga_database_check.pp
```diff
@@ -0,0 +1,36 @@
...
+#
+# The description that is shown in Icinga for this check. It is expected to
+# have a variable interpolated of $DATABASE, this is replaced for the database
+# name that is under check.
+define govuk_jenkins::jobs::athena_fastly_logs_check::icinga_database_check (
+  $jenkins_url = undef,
+  $service_description = undef,
+) {
+  @@icinga::passive_check {
+    "athena_fastly_logs_check_${name}_${::hostname}":
```
cbaines: I'm not quite sure where `${name}` is coming from here.

kevindew: Thats the variable that is automatically set with each item in the array that's passed in. For example: https://stackoverflow.com/a/13008766

This define was a total pain to deal with due to puppet 3's lack of array iteration abilities. Do you know of any examples in our code base that are clearer to understand?

hieradata_aws/class/production/jenkins.yaml
```diff
@@ -1,5 +1,6 @@
 ---
 govuk_jenkins::job_builder::jobs:
+  - govuk_jenkins::jobs::athena_fastly_logs_check
```
leenagupte: Does `govuk_jenkins::jobs::athena_fastly_logs_check::databases` need to added to this jobs list?

modules/govuk_jenkins/manifests/jobs/athena_fastly_logs_check.pp
```diff
@@ -0,0 +1,48 @@
...
+#   The region athena is running in
+#
+# [*s3_results_bucket*]
+#   An S3 bucket where the query results can be written too
+#
+class govuk_jenkins::jobs::athena_fastly_logs_check (
+  $aws_access_key_id = undef,
+  $aws_secret_access_key = undef,
+  $aws_region = 'eu-west-1',
+  $databases = ['govuk_www', 'govuk_assets'],
```
leenagupte: Nitpick: This value should probably be added the parameters list in the comment.

modules/govuk_jenkins/manifests/jobs/athena_fastly_logs_check/icinga_database_check.pp
cbaines: Ah, OK, I'm used to the magical `title` variable, but obviously I've overlooked the `name` one :+1:

leenagupte: So `${name}` is `govuk_www` or `govuk_assets`?

kevindew: Yup - that's correct Leena.

modules/govuk_jenkins/manifests/jobs/athena_fastly_logs_check.pp
kevindew: Thank you! sorted

PR #10163, opened 2020-02-20
Sync Whitehall to Integration using govuk_env_sync.

sengi:
* Implement the [existing data scrubbing script](https://github.com/alphagov/whitehall/blob/master/script/scrub-database) as plain SQL.
* Redact the data in place on Staging, then dump Staging to S3.
* Restore Staging from the latest Prod backup shortly after the redact/dump finishes. 
* Restore Integration from the redacted Staging dump.
* Remove the inconsistent `_mysql` from the names of the cron jobs in Staging. (I plan to remove the old jobs when rolling this out.)

It's a bit of a ugly hack, but it avoids the expense of making another copy of the database in order to do the redaction. It also lets us keep Staging unredacted (except for an hour or two overnight). I'm told that having the unredacted data in Staging can be important for debugging content-related issues.

### Alternatives considered

1. Make yet another copy of the database and do the redaction there, instead of doing it in place in Staging.
    * Seems like a waste of resources. Whitehall is one of our larger databases and it takes non-negligible resources and time to dump/restore.
2. Use AWS Glue instead.
    * It'd almost certainly be a better approach, simpler, cheaper to maintain, avoids reinventing the wheel.
    * Only didn't go with this because configuring IAM looked too complicated to do properly in a hurry.
    * We should probably still switch to it at some point, if we want to continue syncing data from Prod/Staging to Integration rather than using a clean sample dataset.
3. Use a clean, sample dataset for Integration instead of syncing real data from Prod.
    * This would be safer than trying to scrub production data (because the Prod schema is a moving target and we may already be missing things which we ought to be redacting).
    * Would be more reliable and use less resources.
    * Sample data might be less good for testing than real data. Probably only in theory, though.
    * Not chosen because too much planning/preparation needed; would need to assess feasibility, get developer buy-in and then create the dataset.


sengi:
Tested the `sanitise_whitehall.sql` script by running it on a copy of the Staging database and then querying for the redacted records.

sengi:
Chatted with @cbaines about the question of whether Staging should/shouldn't use the redacted data. I hear there's value in having the unredacted data in Staging because it can help when debugging content-related issues, so we need to keep it that way.

At Chris's suggestion, I'll switch the order of things around so that we do the redaction on the Staging database just before it gets replaced with the latest Production copy, instead of just after. That way we get to use the unredacted data in Staging and we still avoid the faff of having to make a separate copy of the database. The only cost is that Integration gets a 24-hour delay, but we can live with that.

sengi:
Updated to reflect the change mentioned above. Ready for another look.

sengi:
Tested end-to-end:

```
$ sudo -iu govuk-backup /usr/bin/envdir /etc/govuk_env_sync/env.d /usr/local/bin/govuk_env_sync.sh -f /etc/govuk_env_sync/push_whitehall_staging_daily.cfg
govuk_env_sync.sh: Starting "/usr/local/bin/govuk_env_sync.sh -f /etc/govuk_env_sync/push_whitehall_staging_daily.cfg"
govuk_env_sync.sh: Running pre-dump SQL script /etc/govuk_env_sync/transformation_sql/sanitise_whitehall.sql...
govuk_env_sync.sh: completed.
govuk_env_sync.sh: Running mysqldump...
govuk_env_sync.sh: completed.
govuk_env_sync.sh: Upload to s3://govuk-staging-database-backups/mysql/2020-02-24T11:44:35-whitehall_production.gz...
upload: ../../tmp/whitehall_staging/tmp.b4IDEP6dOs/2020-02-24T11:44:35-whitehall_production.gz to s3://govuk-staging-database-backups/mysql/2020-02-24T11:44:35-whitehall_production.gz
govuk_env_sync.sh: completed.
govuk_env_sync.sh: Completed "/usr/local/bin/govuk_env_sync.sh -f /etc/govuk_env_sync/push_whitehall_staging_daily.cfg"
```

```
$ sudo -iu govuk-backup /usr/bin/envdir /etc/govuk_env_sync/env.d bash -x /usr/local/bin/govuk_env_sync.sh -f /etc/govuk_env_sync/pull_whitehall_staging_daily.cfg
govuk_env_sync.sh: Starting "/usr/local/bin/govuk_env_sync.sh -f /etc/govuk_env_sync/pull_whitehall_staging_daily.cfg"
govuk_env_sync.sh: Download from s3://govuk-staging-database-backups/mysql/2020-02-24T11:22:03-whitehall_production.gz...
download: s3://govuk-staging-database-backups/mysql/2020-02-24T11:22:03-whitehall_production.gz to ../../tmp/whitehall_staging/tmp.qlCvvji1u0/2020-02-24T11:22:03-whitehall_production.gz
govuk_env_sync.sh: completed.
govuk_env_sync.sh: No post processing needed for whitehall_production
govuk_env_sync.sh: Completed "/usr/local/bin/govuk_env_sync.sh -f /etc/govuk_env_sync/pull_whitehall_staging_daily.cfg"
```

sengi:
Removed the old cron jobs (the ones with `_mysql` in the names) from the Staging db_admin machine. The rest is all handled by Puppet.

PR #10181, opened 2020-02-25
Remove copy_attachments_to_integration/staging Jenkins jobs.

sengi:
These jobs rsynced files between the NFS mounts on asset_master by
calling various scripts in the alphagov/env-sync-and-backup repo.

They don't appear to be necessary any more since S3 is the canonical
repository for assets. The NFS mounts which these scripts refer to
appear to be used only for temporary storage while attachments await
virus scanning before being uploaded to S3.

It makes sense to remove these jobs even without being 100% certain that
there isn't something obscure that's still relying on them, because:

1. They're already broken since the Whitehall migration.
2. If it turns out we still need them after all, we'd replace them with
   a cron job that runs a simple rsync command rather than using
   env-sync-and-backup.


sengi:
Looks like the CI job didn't run for some reason (probably related to yesterday's GitHub outage). Rebasing just to re-trigger the CI.

PR #10185, opened 2020-02-26
Set a more sensible $ensure_service value when apps are absent

cbaines:
Otherwise, you'd have to set both values in the hieradata. This change
means the $ensure_service value is 'absent' automatically when the app
is 'absent'.

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -327,7 +327,7 @@ define govuk::app (
   $monitor_unicornherder = undef,
   $local_tcpconns_established_warning = undef,
   $local_tcpconns_established_critical = undef,
-  $ensure_service = running,
+  $ensure_service = undef,
 ) {
 
   if ! ($app_type in ['procfile', 'rack', 'bare']) {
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/app.pp
```diff
@@ -414,8 +414,17 @@
     local_tcpconns_established_critical => $local_tcpconns_established_critical,
   }
 
+  if $ensure == 'absent' {
+    $ensure_service_real = 'absent'
+  } else {
+    $ensure_service_real = $ensure_service ? {
+      undef    => running,
```
issyl0: ?

cbaines: If a value for `ensure_service` hasn't been specified, then it's value will be `undef`. This is in the branch of the `if` where the app isn't absent, so set the service to be `running`.

cbaines: So this line is replicating the behaviour that you'd get previously though having `$ensure_service` default to `running`.

issyl0: Riiight.

PR #10187, opened 2020-02-27
Move carrenza env_sync push jobs from Carrenza Staging => Production

whi-tw:
Rather than env_sync'ing the publishing apps in Carrenza staging, run these jobs from the Carrenza mysql-backup machine.

This will allow us to easily restore from the production Carrenza database at AWS migration time


whi-tw:
@sengi
> I'm assuming you plan on having the Integration sync come straight from Prod like it currently does, rather than the Prod->Staging->Integration waterfall that some of the other govuk_env_sync jobs do. (I think that's a fine choice provided there's no redaction needed for Integration.)

Yep, just checked over the backup-and-sync jobs for these, and there's no redaction happening. A card has been created to remove the previously used jobs. I've added govuk_env_sync jobs to pull from the backups created in b3e0e6e

PR #10188, opened 2020-02-27
Remove Sebastian Schmieschek

bevanloon:
Seb has moved to a different team at GDS and no longer works on GOV.UK

[Trello](https://trello.com/c/iNXqG3xQ/838-leaver-sebastian-schmieschek-re)

### DIFF
--- a/modules/users/manifests/sebastianschmieschek.pp
+++ b/modules/users/manifests/sebastianschmieschek.pp
@@ -1,6 +1,7 @@
 # Creates the user schmie for Sebastian Schmieschek
 class users::sebastianschmieschek {
   govuk_user { 'sebastianschmieschek':
+    ensure   => absent,
     fullname => 'Sebastian Schmieschek',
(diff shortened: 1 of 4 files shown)


modules/users/manifests/sebastianschmieschek.pp
```diff
@@ -1,9 +0,0 @@
-# Creates the user schmie for Sebastian Schmieschek
-class users::sebastianschmieschek {
-  govuk_user { 'sebastianschmieschek':
-    fullname => 'Sebastian Schmieschek',
```
issyl0: You'll need to `ensure => absent` for the user to be removed from Puppet along with their home directory, otherwise their files hang around forever. https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html

bevanloon: 🤦‍♂ I used to know that. Thanks Issy.

Changed in ed5779e

PR #10190, opened 2020-02-27
Remove the 'high disk time' alert.

sengi:
This is a [cause-based-alert] and has not proved useful. We have disk time
graphs on dashboards.

In the unlikely event that this leads to a gap in our alerting coverage,
that gap should be filled by adding an alert on the user-facing response
time rather than alerting on an internal detail like disk I/O time.

[cause-based-alert]: https://docs.google.com/document/d/199PqyG3UsyXlwieHaqbGiWVa8eMWi8zzAn0YfcApr8Q/preview#heading=h.6ammb5h32uqq

### DIFF
--- a/hieradata/class/api_mongo.yaml
+++ b/hieradata/class/api_mongo.yaml
@@ -51,8 +51,3 @@ mongodb::server::replicaset_members:
   'api-mongo-4':
     hidden: true
     priority: 0
-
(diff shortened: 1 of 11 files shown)


issyl0:
(cc @philandstuff - we cleaned up a long-standing alert here 💪)

PR #10191, opened 2020-02-27
Update Python version to 3.5 in govuk_python module

deborahchua:
This will install `Python 3.5.2` on the CI agent machines and others
that include the `govuk_python` modules (such as `mapit` and `ckan`).

This is currently blocking some of our branches from building
successfully as they require `Python 3.5 or later`.

Trello card: https://trello.com/c/lR6qMIm1/1780-5-upgrade-to-a-supported-python-version

### DIFF
--- a/modules/govuk_python/manifests/init.pp
+++ b/modules/govuk_python/manifests/init.pp
@@ -9,5 +9,5 @@ class govuk_python {
 
   Class['python::install'] -> Package <| provider == 'pip' and ensure != absent |>
 
-  ensure_packages(['python3', 'python3-dev'])
+  ensure_packages(['python3.5', 'python3.5-dev', 'python3-dev'])
 }


cbaines:
>  I assume this change means that `python3` now points to `python3.5` rather than `python3.4`.

I think you're imagining a world where we take software deployment seriously! I don't think Puppet will remove the `python3` or `python3-dev` packages, but there is the potential for them not to be installed on new machines, so this could cause additional inconsistencies between machines in AWS. Also, I don't think this will have any effect on update-alternatives, so `python3` will still point to `python3.4`, at least on the machines with `python3` still installed, for new machines it might be `python3.5`.

modules/govuk_python/manifests/init.pp
```diff
@@ -9,5 +9,5 @@
 
   Class['python::install'] -> Package <| provider == 'pip' and ensure != absent |>
 
-  ensure_packages(['python3', 'python3-dev'])
+  ensure_packages(['python3.5', 'python3.5-dev'])
```
cbaines: I think it's safest to keep installing `python3-dev`, as we're not managing this at a granular level, so we have no real idea of what uses Python 3.4.

thomasleese: That sounds sensible, one thing that also hadn't occurred to me is that there are probably other packages which depend on `python3` and so we're getting it anyway - but better to be explicit.

deborahchua: Thanks both! I've updated this now in https://github.com/alphagov/govuk-puppet/pull/10191/commits/76181452c6a73d71f171f81ab64a70bffaf95ba6

modules/govuk_python/manifests/init.pp
```diff
@@ -9,5 +9,5 @@
 
   Class['python::install'] -> Package <| provider == 'pip' and ensure != absent |>
 
-  ensure_packages(['python3', 'python3-dev'])
+  ensure_packages(['python3.5', 'python3-dev'])
```
cbaines: We should install `python3.5-dev` at least, as I imagine that things that Pip installs might depend on the headers included in that package to compile native extensions to Python.

PR #10199, opened 2020-03-01
Re-add Sebastian

bevanloon:
I deployed puppet on Friday evening to staging and it threw a lot of errors. I believe it's because secrets still has Seb in the hieradata but the manifest is gone. 
Re-adding the manifest should sort it out. We could alternatively merge https://github.com/alphagov/govuk-secrets/pull/915 which I think will solve the problem.

This PR is just in case it doesn't solve it.....


bevanloon:
I think alphagov/govuk-secrets#915 has done the trick so closing this.

PR #10202, opened 2020-03-02
Remove Rob Bond and Ruth Morris

bevanloon:
Rob no longer works on GOV.UK and so doesn't need to be in the email
whitelist

Ruth has moved onto bigger and better things.

### DIFF
--- a/hieradata/development.yaml
+++ b/hieradata/development.yaml
@@ -168,7 +168,6 @@ govuk::apps::email_alert_api::email_address_override_whitelist:
   - jos.koetsier@digital.cabinet-office.gov.uk
   - kevin.dew@digital.cabinet-office.gov.uk
   - leanne.cummings@digital.cabinet-office.gov.uk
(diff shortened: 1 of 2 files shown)


hieradata/development.yaml
```diff
@@ -168,7 +168,6 @@ govuk::apps::email_alert_api::email_address_override_whitelist:
   - jos.koetsier@digital.cabinet-office.gov.uk
   - kevin.dew@digital.cabinet-office.gov.uk
   - leanne.cummings@digital.cabinet-office.gov.uk
-  - robert.bond@digital.cabinet-office.gov.uk
   - ruth.morris@digital.cabinet-office.gov.uk
```
issyl0: Neither does Ruth...

bevanloon: 👍 Ruth will be back. I'm sure of it. 
Updated in https://github.com/alphagov/govuk-puppet/pull/10202/commits/762b5e24a89c09b22bd3ff9a480aae44fa3953ba

PR #10208, opened 2020-03-04
Rename access limited filenames to redacted.pdf

thomasleese:
This matches [the behaviour of Whitehall](https://github.com/alphagov/whitehall/blob/a7c839bd80ac0dc535c979f89dd3657eff0c510b/script/scrub-database#L94) and means we can avoid unnecessary Sentry errors when Whitehall tries to communicate with Asset Manager with a different filename to the one known by Asset Manager.

[Trello Card](https://trello.com/c/33ZuT6D4/885-5-handle-access-limited-assets-in-asset-manager) ・ [Sentry issue](https://sentry.io/organizations/govuk/issues/617165767/)

### DIFF
--- a/modules/govuk_env_sync/files/govuk_env_sync.sh
+++ b/modules/govuk_env_sync/files/govuk_env_sync.sh
@@ -582,7 +582,25 @@ function postprocess_router {
     spotlight_proxy_domain="${unmigrated_source_domain}"
   fi
   mongo_backend_domain_manipulator "spotlight-proxy" "${spotlight_proxy_domain}"
+}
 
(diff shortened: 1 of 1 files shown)


thomasleese:
I think they fixed it today. 😄

sengi:
> I think they fixed it today. 😄

Ah that was the sync of the actual assets to S3 (which was either broken since before the migration or working the whole time, I'm not sure because afaict it doesn't log anything and appears to return instantly if the Jenkins logs are to be believed 😅 )

The govuk_env_sync job here has been in service since a day or two after the migration.

Thanks for sorting out the redaction stuff!

PR #10209, opened 2020-03-04
Fix check_puppetdb_ssh_host_keys script

bilbof:
Need to import sys as per error in alert logs:
NameError: global name 'sys' is not defined


bilbof:
@issyl0 it's all good

PR #10221, opened 2020-03-05
Disable enhanced ecommerce data load

bilbof:
This isn't needed on a schedule right now so we can remove the cron schedule. It'll be needed in future so I've kept the code here.

### DIFF
--- a/hieradata_aws/production.yaml
+++ b/hieradata_aws/production.yaml
@@ -193,8 +193,6 @@ govuk::apps::travel_advice_publisher::enable_email_alerts: true
 govuk_search::monitoring::es_port: '80'
 govuk_search::monitoring::es_host: 'vpc-blue-elasticsearch6-domain-2hnib7uydv3tgebhabx74gdelq.eu-west-1.es.amazonaws.com'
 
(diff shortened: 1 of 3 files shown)


hieradata_aws/production.yaml
```diff
@@ -193,8 +193,6 @@ govuk::apps::travel_advice_publisher::enable_email_alerts: true
 govuk_search::monitoring::es_port: '80'
 govuk_search::monitoring::es_host: 'vpc-blue-elasticsearch6-domain-2hnib7uydv3tgebhabx74gdelq.eu-west-1.es.amazonaws.com'
 
-govuk_jenkins::jobs::enhanced_ecommerce_search_api::cron_schedule: '0 6 * * 1-5' # Every weekday at 6am
```
barrucadu: Is https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/templates/jobs/enhanced_ecommerce.yaml.erb going to work with this being `undef`?

bilbof: Maybe not - I've put in a fix for that

PR #10224, opened 2020-03-05
Make "puppet last run errors" alert show as Critical.

sengi:
We keep disabling Puppet agent on machines and then forgetting about it.
This has led to several mini-incidents over the last few months and has
caused problems when migrating apps to AWS. The warning alerts tend to
be ignored (understandably) and so mistakes are left unchecked.

On one occasion a Puppet agent was left disabled for several months on a
Jenkins box and this led to CI breaking horribly when Puppet was
re-enabled, because in the meantime someone had merged a Puppet PR
containing a mistake which removed all permissions from the Jenkins
config. The PR appeared fine when it was tested, because the machine
which it affected never ran the Puppet agent.

It therefore appears that we've been ignoring this alert, but it does
actually require action. Let's see if making it show up under the
Critical alerts section helps.

### DIFF
--- a/modules/puppet/manifests/monitoring.pp
+++ b/modules/puppet/manifests/monitoring.pp
@@ -23,9 +23,10 @@ class puppet::monitoring (
   $app_domain = hiera('app_domain')
 
   @@icinga::passive_check { "check_puppet_${::hostname}":
-    service_description => 'puppet last run errors',
-    host_name           => $::fqdn,
-    freshness_threshold => 7200,
-    notes_url           => monitoring_docs_url(puppet-last-run-errors),
+    service_description   => 'puppet last run errors',
+    host_name             => $::fqdn,
+    freshness_threshold   => 7200,
+    freshness_alert_level => 'critical',
+    notes_url             => monitoring_docs_url(puppet-last-run-errors),
   }
 }


sengi:
> The motivation for this seems sound. However, I think we should take a step back and clarify how we're interpreting criticals vs. warnings. For me, a critical should be something that's actively degrading user experience, and a warning should be something that might if left unattended. With those definitions, this alert should continue to be a warning.
> 
> I'm hoping there's another way we can address this, as there is an issue there. Perhaps more of a process issue, though?

Did you have a particular alternative in mind? I can think of several but none that take minimal effort to implement.

The fact remains that we've been ignoring the warning alerts and that Puppet failures require action, so we need to do something. I don't mind whether it's this or something else, as long as it fixes the problem.

sengi:
@benthorner I basically agree with all your points there. This change is an imperfect and unfair compromise within an already far-from-ideal system, making it slightly worse in some ways in order to maybe address a real problem.

I'd much prefer a solution which doesn't add slightly to the already noisy alerts which 2nd-line has to deal with. Having a time-limit on disabling Puppet agent seems attractive, but would really just swap one footshooter for another (and would be nontrivial to implement and test).

Having the alert nag the person who disabled the Puppet agent would probably be the next best thing, but that requires a lot of stuff that we don't currently have.

So I'm left with this rather unsatisfactory PR that isn't fair on 2nd-line and works (if it works) by causing the culprits to be nagged by an actual person who's annoyed that they unfairly had to look at yet another alert because someone didn't clean up after themselves.

I'm kinda torn between a) merging this, b) putting a big nannyish warning on `govuk_puppet --disable` reminding people to remind themselves to re-enable Puppet agent and c) both of the above 😕

benthorner:
@sengi thanks for explaining some of the alternatives. I think this is fine to go in - I suspect it would be such a big deal in practice, and may have the positive outcome of motivating us to improve the alerts, due to them being more visible.

issyl0:
Can we do something with `govuk_puppet --disable` to enforce that people put notes as to _why_ it's being disabled (and also their name)?

sengi:
> Can we do something with `govuk_puppet --disable` to enforce that people put notes as to _why_ it's being disabled (and also their name)?

I think so. I don't know of any use case that would stop us from making `govuk_puppet --disable` require specifying the reason/note.

sengi:
We could also consider increasing the staleness threshold. It's currently 2 hours, which seems to me quite a short time. (Bear in mind this is only for the case where Puppet agent hasn't run at all in too long — it'll still alert right away if Puppet agent reports a failure. Indeed the alert was already a `critical` in that case.)

PR #10225, opened 2020-03-05
Limit access to staging and production Jenkins

bevanloon:
Members of the GOV.UK github team should have access to integration
Jenkins.

Members of the GOV.UK Production github team should have access to
staging and production Jenkins.

### DIFF
--- a/hieradata_aws/class/production/jenkins.yaml
+++ b/hieradata_aws/class/production/jenkins.yaml
@@ -1,4 +1,43 @@
 ---
+jenkins_admin_permission_list: &jenkins_admin_permission_list
+  - 'hudson.model.Hudson.Administer'
+  - 'hudson.model.Hudson.Read'
+  - 'hudson.model.Hudson.RunScripts'
+  - 'hudson.model.Item.Build'
+  - 'hudson.model.Item.Cancel'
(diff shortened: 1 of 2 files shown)


hieradata_aws/class/integration/jenkins.yaml
```diff
@@ -5,6 +5,24 @@ govuk_jenkins::config::theme_colour: '#005EA5'
...
+  -
+    user: 'anonymous'
+    permissions:
+      - 'hudson.model.Hudson.Read'
+      - 'hudson.model.Item.Discover'
+  -
+    user: 'github'
+    permissions:
+      - 'hudson.model.Item.Build'
+      - 'hudson.model.Item.Read'
```
cbaines: Is adding this here going to change anything? It looks to me that the same configuration is applied at the class level for the Jenkins machines in AWS: https://github.com/alphagov/govuk-puppet/blob/master/hieradata_aws/class/jenkins.yaml#L35-L51

cbaines: Ah, I see now, you're adding this here to keep the existing behaviour for Integration, while changing the behaviour for Staging and Production...

PR #10232, opened 2020-03-09
Set a template id and api key for short-url-manager to use govuk notify

benjamineskola:
cf. https://github.com/alphagov/short-url-manager/pull/452

https://trello.com/c/6FJKUWIs/1794-3-short-url-manager-use-notify-instead-of-amazon-ses


thomasleese:
I've rebased and squashed the commits, I'll merge and deploy this when the build has succeeded.

PR #10249, opened 2020-03-18
govuk_env_sync.sh improvements

cbaines:


modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -124,7 +124,22 @@ function nagios_passive {
   # We require to map the monitored services to the configuration files/govuk_env_sync::tasks
   if [ -n "${configfile:-""}" ]; then
     nagios_service_description="GOV.UK environment sync $(basename "${configfile%.cfg}")"
-    printf "%s\\t%s\\t%s\\t%s\\n" "${ip_address}" "${nagios_service_description}" "${nagios_code}" "${nagios_message}" | /usr/sbin/send_nsca -H alert.cluster >/dev/null
+
+    max_retries=4
```
sengi: optional: consider declaring these vars as local

cbaines: Good point, I've pushed another commit which makes all the variables in this function local: https://github.com/alphagov/govuk-puppet/pull/10249/commits/a806414b2d5b1482f94b65ad1dc03ace8151d287

PR #10250, opened 2020-03-18
Use different reboot locks for different tasks

cbaines:
Previously, the with_reboot_lock script would attempt to create
/etc/unattended-reboot/no-reboot/reboot_lock when it didn't exist. I
think the intention was to avoid the machine from rebooting, but this
also has the effect of serialising anything using this script.

I'm been looking at the use of this for the govuk_env_sync cron
jobs. They don't need to run one after the other, but have been doing
due to this script.

Rather than using a single lock file, have the with_reboot_lock script
take the filename as the first argument. This commit also updates
where this script is used in govuk-puppet.

### DIFF
--- a/modules/govuk_unattended_reboot/templates/usr/local/bin/with_reboot_lock.erb
+++ b/modules/govuk_unattended_reboot/templates/usr/local/bin/with_reboot_lock.erb
@@ -3,7 +3,7 @@
 set -eu
 
 lock_file_name="$1"
-lock_file="<%= @lock_dir %>/$lock_file_name"
+lock_file="<%= @lock_dir %>/reboot_lock-${lock_file_name}"
 
 function lock() {
   while [ -f $lock_file ]; do
--- a/modules/govuk_unattended_reboot/templates/usr/local/bin/with_reboot_lock.erb
+++ b/modules/govuk_unattended_reboot/templates/usr/local/bin/with_reboot_lock.erb
@@ -2,7 +2,8 @@
 
 set -eu
 
-lock_file="<%= @lock_dir %>/reboot_lock"
+lock_file_name="$1"
+lock_file="<%= @lock_dir %>/$lock_file_name"
 
 function lock() {
   while [ -f $lock_file ]; do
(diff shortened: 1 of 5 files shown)


modules/govuk_unattended_reboot/templates/usr/local/bin/with_reboot_lock.erb
```diff
@@ -2,7 +2,8 @@
 
 set -eu
 
-lock_file="<%= @lock_dir %>/reboot_lock"
+lock_file_name="$1"
+lock_file="<%= @lock_dir %>/$lock_file_name"
```
whi-tw: I'd quite like to have the file still have the same prefix, just in case we need to find these lock files to unlock later if something's become gummed up

```suggestion
lock_file="<%= @lock_dir %>/reboot_lock-${lock_file_name}"
```

modules/govuk_unattended_reboot/templates/usr/local/bin/with_reboot_lock.erb
```diff
@@ -19,4 +20,4 @@ function unlock() {
 lock
 trap "unlock" EXIT
 
-"${@}"
+"${@:222}"
```
whi-tw: Rather than pulling a substring out of the args, I'd just shift out the first argument and proceed as usual
```suggestion
shift
"${@}"
```

modules/govuk_unattended_reboot/templates/usr/local/bin/with_reboot_lock.erb
cbaines: That sounds fine to me.

modules/govuk_unattended_reboot/templates/usr/local/bin/with_reboot_lock.erb
cbaines: Haha, this was meant to be 2, rather than 222...

But yeah, I think this is fine as well.

cbaines: I've amended the original commit now to use shift rather than `"${@:2}"`

PR #10253, opened 2020-03-18
Use the standard HTTP response code when host is unknown.

sengi:
HTTP servers are supposed to send 400 Bad Request when the Host header has a bad field-value, not 500 Internal Server Error.

The practical reasons for this are:

 * A wrong host header field-value is an error on the part of the client, not the server, so it should be a 400 series response.
* A wrong host header is not a retryable error. The error code needs to indicate that the client should not resubmit the same request without modification.

https://tools.ietf.org/html/rfc7230#section-5.4

### DIFF
--- a/modules/govuk/manifests/node/s_backend.pp
+++ b/modules/govuk/manifests/node/s_backend.pp
@@ -32,7 +32,7 @@ class govuk::node::s_backend inherits govuk::node::s_base {
 
   include nginx
 
-  # The catchall vhost throws a 500, except for healthcheck requests.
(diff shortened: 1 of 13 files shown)


sengi:
Cheers for the review, @cbaines! I'll hold off on merging this for now if that's ok, just on the precautionary principle what with it being such a crucial time for GOV.UK.

It occurs to me that I need to verify that Fastly falls back to the S3/GCS mirrors if it receives a 400 response (in case we were to have an outage that resulted in vhost configs disappearing from the nginx configs, for example).

PR #10256, opened 2020-03-19
Set a template id and api key for service-manual-publisher to use gov.uk notify

benjamineskola:
Required for https://github.com/alphagov/service-manual-publisher/pull/775 (will also need the API key adding to the secrets).

https://trello.com/c/eraJ8WC2/1821-5-service-manual-publisher-use-notify-instead-of-ses


benjamineskola:
Messed this up by branching from the wrong place, trying again in https://github.com/alphagov/govuk-puppet/pull/10257

PR #10261, opened 2020-03-20
Drop the memory thresholds for the Search API procfile worker

cbaines:
This seems to be using up a significant portion of the 4GB of memory
which the machine has. As Sidekiq is in play here, restarting it will
probably free up memory, so have Icinga do that earlier.

### DIFF
--- a/modules/govuk/manifests/apps/search_api.pp
+++ b/modules/govuk/manifests/apps/search_api.pp
@@ -185,7 +185,9 @@ class govuk::apps::search_api(
   }
 
   govuk::procfile::worker { 'search-api':
-    enable_service => $enable_procfile_worker,
+    enable_service            => $enable_procfile_worker,
+    memory_warning_threshold  => 1250,
+    memory_critical_threshold => 2000,
   }
 
   govuk::procfile::worker { 'search-api-publishing-queue-listener':


cbaines:
> Hmm, does generating a critical Icinga alert restart the worker? If not, the commit message doesn't quite make sense to me confused

Well, it'll actually restart the service when the check hits the warning status. The event handler is configured here [1], and the script that actually does the restarting is here [2] (there's also some other bits in govuk-puppet to make this work).

1: https://github.com/alphagov/govuk-puppet/blob/drop-memory-thresholds-for-the-search-api-procfile-worker/modules/govuk/manifests/procfile/worker.pp#L128
2: https://github.com/alphagov/govuk-puppet/blob/drop-memory-thresholds-for-the-search-api-procfile-worker/modules/monitoring/files/usr/local/bin/event_handlers/govuk_procfile_worker_high_memory.sh#L32

> One other wondering: I take it the default values for warning and critical are 2000 MB and 4000 MB respectively. If a machine only has 4GB of memory in the first place, should the default critical be lowered?

Maybe. I think when I introduced the Icinga check, I just set the default thresholds to a bit above the procfile worker that was using the most memory (although I can't remember what this was).

Looking back at the search machines in Carrenza, they had 8GB of memory, which might explain why we didn't have this issue then.

PR #10263, opened 2020-03-20
Remove limit prom router discovery to integration

tlwr:
What
----

We only asked prometheus to discover router apps in integration

We think this is ready for prime time now, so we are removing the restriction

Now Prometheus will discover and scrape all router app instances

How to review
---

Code review

Read https://github.com/alphagov/router/pull/148/


issyl0:
Is this mergeable?

tlwr:
It is mergeable, but will not have anything to scrape until https://github.com/alphagov/router/pull/148 is merged and deployed

PR #10264, opened 2020-03-23
[DO NOT MERGE] Remove publishing apps from Carrenza Staging

sarahseewhy:
* Removes the relevant publishing app host entries from Carrenza Staging environment.
* This is part of the publishing apps migration.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -18,21 +18,8 @@ node_class: &node_class
   backend:
     apps:
       - canary-backend
-      - collections-publisher
-      - contacts
-      - content-publisher
-      - content-tagger
-      - hmrc-manuals-api
-      - manuals-publisher
-      - maslow
-      - publisher
-      - search-admin
-      - service-manual-publisher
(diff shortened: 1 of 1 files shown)


hieradata/common.yaml
```diff
@@ -18,21 +18,8 @@ node_class: &node_class
...
-      - manuals-publisher
-      - maslow
-      - publisher
-      - search-admin
-      - service-manual-publisher
-      - short-url-manager
       - sidekiq-monitoring
       - signon
-      - specialist-publisher
-      - travel-advice-publisher
```
cbaines: Note that this will just stop Puppet from caring about the apps.

The processes will continue to run, and the files on the machines will still be in place.

If you want Puppet to stop the apps, and remove the files, you'll need to keep these lines in place, and make the apps absent first.

camdesgov: Ah yes...thanks Chris! We will ensure the apps are absent before we merge this post migration.

PR #10267, opened 2020-03-23
Improve how the deploy nofile limit is increased for cache machines

cbaines:
Rather than having two configuration files with different values for the same parameter, they'll just be one file with the intended value.

### DIFF
--- a/hieradata_aws/class/cache.yaml
+++ b/hieradata_aws/class/cache.yaml
@@ -18,4 +18,4 @@ varnish::environment_ip_prefix: "%{hiera('environment_ip_prefix')}"
 
 # Increase the maximum number of file descriptors usable by the router
 # A file descriptor is used per connection
-govuk::deploy::config: 65536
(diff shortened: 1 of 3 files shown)


hieradata_aws/class/cache.yaml
```diff
@@ -15,3 +15,7 @@ govuk::apps::router::mongodb_nodes:
   - 'router-backend-3'
 
 varnish::environment_ip_prefix: "%{hiera('environment_ip_prefix')}"
+
+# Increase the maximum number of file descriptors usable by the router
+# A file descriptor is used per connection
+govuk::deploy::config: 65536
```
thomasleese: Does this need to specify the parameter?

```suggestion
govuk::deploy::config::nofile_limit: 65536
```

cbaines: Indeed, good spot :+1:

PR #10269, opened 2020-03-23
Raise router nofile limit

cbaines:
The current approaches to this seem to be insufficient, `cat
/proc/.../limits` seems to report 1024.

Increasing the limit is important as the router uses files for TCP
connections, and we want it to be able to handle lots of connections
at once. I copied the value from the s_cache.pp file.


modules/govuk/files/usr/local/bin/govuk_spinup
```diff
@@ -85,6 +85,10 @@ echo "GOVUK_APP_ROOT=${GOVUK_APP_ROOT}"
 echo "GOVUK_APP_RUN=${GOVUK_APP_RUN}"
 echo "GOVUK_APP_LOGROOT=${GOVUK_APP_LOGROOT}"
 
+if [ -n "$GOVUK_APP_ULIMIT_NOFILE" ]; then
```
barrucadu: This should be `if [ -n "${GOVUK_APP_ULIMIT_NOFILE:-}" ]; then` because of the `set -u` further up

PR #10271, opened 2020-03-24
Only run mongodb env_sync tasks on mongo master

whi-tw:
Utilise the same "is_writable_mongo" check for pushes as is currently
used for pulls. This check is scoped only if \$dbms is "mongo.

This will ensure that this task does not run on every one of the mongo
nodes, only on the master.

### DIFF
--- a/modules/govuk_env_sync/files/govuk_env_sync.sh
+++ b/modules/govuk_env_sync/files/govuk_env_sync.sh
@@ -682,6 +682,11 @@ nagios_code=2
 
 case ${action} in
   push)
+    if [ "${dbms}" == "mongo" ] && [ "$(is_writable_mongo)" != "true" ]; then
+      trap - EXIT # remove the trap, as we do not want to send a nagios message in this case
+      log "This machine is not a mongo master. Skipping."
+      exit
+    fi
     create_tempdir
     create_timestamp
     set_filename


modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -682,6 +682,12 @@ nagios_code=2
 
 case ${action} in
   push)
+    if [ "${dbms}" == "mongo" ]; then
+      [ "$(is_writable_mongo)" != "true" ] \
+        && log "This machine is not a mongo master. Skipping." \
+        && trap - EXIT \
+        && exit
```
cbaines: It might be a bit easier to parse as:
```
if [ "${dbms}" == "mongo" ] && [ "$(is_writable_mongo)" != "true" ]; then
  log "This machine is not a mongo master. Skipping." \
  trap - EXIT
  exit
fi
```

whi-tw: Absolutely agree - that was a hideous way of writing things. I've been doing a bunch of docker stuff, so one-lining things is beneficial.

I've made some changes - hopefully that's better now!

cbaines: > Absolutely agree - that was a hideous way of writing things. I've been doing a bunch of docker stuff, so one-lining things is beneficial.

No problem, exposure to Docker can be very harmful!

> I've made some changes - hopefully that's better now!

Yep, :+1:

modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -682,6 +682,11 @@ nagios_code=2
 
 case ${action} in
   push)
+    if [ "${dbms}" == "mongo" ] && [ "$(is_writable_mongo)" != "true" ]; then
+      trap - EXIT # remove the trap, as we do not want to send a nagios message in this case
+      log "This machine is not a mongo master. Skipping."
+      exit
+    fi
```
cbaines: Having a comment about why this is necessary might be useful

whi-tw: I've got all the info in the commit message, and I wouldn't necessarily want to replicate `git blame` with super verbose comments in the code. Does that suffice?

cbaines: Having the motivation in the commit message is great. I guess the reasoning behind this is complicated enough that it's difficult to explain why this code is here concisely.

I'm fine with keeping things as they are.

PR #10272, opened 2020-03-24
Data sync complete staging fixes and improvements

cbaines:
This will hopefully fix the Data Sync Complete job on the Carrenza side of Staging, and start it doing things on the AWS side, which should resolve issues with Whitehall scheduled publications.

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/data_sync_complete_staging.pp
+++ b/modules/govuk_jenkins/manifests/jobs/data_sync_complete_staging.pp
@@ -18,13 +18,28 @@ class govuk_jenkins::jobs::data_sync_complete_staging (
 ) {
   if $::aws_migration {
     $aws = true
+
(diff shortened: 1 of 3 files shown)


modules/govuk_jenkins/manifests/jobs/data_sync_complete_staging.pp
```diff
@@ -16,9 +16,32 @@
   $auth_token = undef,
   $signon_domains_to_migrate = [],
 ) {
+  if $::aws_migration {
+    $aws = true
+    $carrenza = false
```
thomasleese: Do we need this variable?

cbaines: Ah, no, I was looking at making parts conditional to only run on Carrenza, but then decided against it. I've now removed the variable.

PR #10278, opened 2020-03-26
Rate limit specific high-impact Google Apps script

Tetrino:
We are seeing spikes of load on the search-api as a specific
GApps script makes up to multiple thousands of requests over a couple of
minutes. The script is using multiple IPs and thus bypassing our
conventional load limiting.

However they all share the same app type in the user agent header, which
allows us to track which requests are being made from this one source.

Unfortunately this event is regular enough that we have decided to take
direct action.

This pull request attempts to create a zone that handles
Google-Apps-Script user agents specifically, to avoid such high spikes
from third parties.

Trello card on search board: https://trello.com/c/4bjhI0QU/1456-investigate-rate-limiting-google-apps-scripts

### DIFF
--- a/modules/govuk/templates/node/s_backend_lb/rate-limiting.conf.erb
+++ b/modules/govuk/templates/node/s_backend_lb/rate-limiting.conf.erb
@@ -26,3 +26,14 @@ limit_req_zone $post_limit_req zone=contact:5m rate=6r/m;
 # Return 429 (Too Many Requests) instead of the default 503.
 limit_req_status 429;
 limit_conn_status 429;
+
+# Create a rate limiting zone that handles limiting excessive load from 
+# all Google-App-Script calls.
+map $http_user_agent $script_limit {
+  default "";
+  ~*(Google-Apps-Script) $http_user_agent;
+}
+
+limit_req_zone $script_limit zone=rate:30m rate=10r/s;
+limit_req_zone $script_limit zone=performance:5m rate=5r/s;
+limit_conn_zone $script_limit zone=connections:10m;
--- a/modules/router/templates/rate-limiting.conf.erb
+++ b/modules/router/templates/rate-limiting.conf.erb
@@ -26,3 +26,14 @@ limit_req_zone $post_limit_req zone=contact:5m rate=6r/m;
 # Return 429 (Too Many Requests) instead of the default 503.
 limit_req_status 429;
 limit_conn_status 429;
+
+# Create a rate limiting zone that handles limiting excessive load from
+# all Google-App-Script calls.
+map $http_user_agent $script_limit {
+  default "";
+  ~*(Google-Apps-Script) $http_user_agent;
+}
+
+limit_req_zone $script_limit zone=rate:30m rate=10r/s;
+limit_req_zone $script_limit zone=performance:5m rate=5r/s;
+limit_conn_zone $script_limit zone=connections:10m;


modules/govuk/templates/node/s_backend_lb/rate-limiting.conf.erb
```diff
@@ -26,3 +26,15 @@ limit_req_zone $post_limit_req zone=contact:5m rate=6r/m;
 # Return 429 (Too Many Requests) instead of the default 503.
 limit_req_status 429;
 limit_conn_status 429;
+
+# Create a rate limiting zone that handles limiting excessive load from
+# a specific high-load google apps script by ID. This may need to be changed to limit 
+# all Google-App-Script calls.
+map $http_user_agent $script_limit {
+  default "";
+  ~*(UAEmdDd-Tj3yuNDB-5kGWRZCbhvY_Im35Vdc) $binary_remote_addr;
```
barrucadu: I don't think we should limit by this specific ID.  If a script with ID "foo" makes a million requests in a second, we should limit that too.

I'd instead like to see something which extracts the ID from the user agent and uses that.

Tetrino: So you're thinking we store it somewhere? One option I do have is to limit it based on type e.g. `google-apps-script`

Tetrino: I wasn't sure if that was the best approach.

barrucadu: So, we currently have rate limiting based on IPs, right?  If the same IP sends too many requests in a short period of time, we serve a 429 response.

It would be good if we could do the same thing here, but based on the user-agent, where it's a google apps scripts user-agent.

modules/govuk/templates/node/s_backend_lb/rate-limiting.conf.erb
```diff
@@ -26,3 +26,15 @@ limit_req_zone $post_limit_req zone=contact:5m rate=6r/m;
...
+# a specific high-load google apps script by ID. This may need to be changed to limit 
+# all Google-App-Script calls.
+map $http_user_agent $script_limit {
+  default "";
+  ~*(Google-Apps-Script) $binary_remote_addr;
+}
+
+limit_req_zone $limit_req zone=rate:30m rate=10r/s;
+limit_req_zone $limit_req zone=performance:5m rate=5r/s;
+limit_conn_zone $limit_req zone=connections:10m;
```
barrucadu: Should these 3 lines (and in the other file) use `$script_limit` instead of `$limit_req`?

modules/govuk/templates/node/s_backend_lb/rate-limiting.conf.erb
```diff
@@ -26,3 +26,15 @@ limit_req_zone $post_limit_req zone=contact:5m rate=6r/m;
 # Return 429 (Too Many Requests) instead of the default 503.
 limit_req_status 429;
 limit_conn_status 429;
+
+# Create a rate limiting zone that handles limiting excessive load from
+# a specific high-load google apps script by ID. This may need to be changed to limit 
+# all Google-App-Script calls.
```
barrucadu: This comment is no longer accurate as it's limiting all google apps scripts

modules/govuk/templates/node/s_backend_lb/rate-limiting.conf.erb
```diff
@@ -26,3 +26,15 @@ limit_req_zone $post_limit_req zone=contact:5m rate=6r/m;
...
+map $http_user_agent $script_limit {
+  default "";
+  ~*(Google-Apps-Script) $binary_remote_addr;
```
barrucadu: I think with `$binary_remote_addr` here we're just limiting by IP address, so this is adding nothing over existing configuration, I think we want `$http_user_agent `

PR #10287, opened 2020-03-31
Revert "Rate limit specific high-impact Google Apps script"

barrucadu:
```
Notice: /Stage[main]/Nginx::Config/Exec[nginx_configtest]/returns: nginx: [emerg] limit_req_zone "rate" is already bound to key "$limit_req" in /etc/nginx/conf.d/rate-limiting.conf:43
Notice: /Stage[main]/Nginx::Config/Exec[nginx_configtest]/returns: nginx: configuration file /etc/nginx/nginx.conf test failed
```

I'm not too sure what the needed change is here - is it just changing the "zone=30m" to something else like "ua_zone=30m"?  Reverting for now.

Reverts alphagov/govuk-puppet#10278


barrucadu:
Also, Chrome is [planning to deprecate the user-agent header](https://9to5google.com/2020/01/14/google-deprecate-chrome-user-agent-string-privacy/), so we may not be able to trust it for other Google things in the future.  This approach only works because we're assuming each script (or at least each user) has a unique user-agent.

PR #10291, opened 2020-04-02
Manage aptly GPG key through puppet

barrucadu:
The key was, as far as I can tell, manually added to the Carrenza apt-1 machine years ago.  It's not there in AWS.  So, we need to start managing this through puppet.

This PR has two changes:

1. Have s_apt take the key as hiera, and manage the keyring.
2. Change everything which hard-codes the fingerprint to use the one from hiera.

The current private key will need putting into secrets too.

---

[Trello card](https://trello.com/c/Y1Yg59D4/1891-get-aptly-gpg-key-working-in-aws)

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -68,6 +68,7 @@ deployable_applications: &deployable_applications
   travel-advice-publisher: {}
 
 apt_mirror_hostname: 'apt.publishing.service.gov.uk'
+apt_mirror_fingerprint: '3803E444EB0235822AA36A66EC5FE1A937E3ACBB'
 
 apt::apt_update_frequency: 'daily'
 apt::purge_preferences_d: true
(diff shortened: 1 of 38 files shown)


hieradata/common.yaml
```diff
@@ -68,6 +68,7 @@ deployable_applications: &deployable_applications
   travel-advice-publisher: {}
 
 apt_mirror_hostname: 'apt.publishing.service.gov.uk'
+apt_mirror_fingerprint: '3803E444EB0235822AA36A66EC5FE1A937E3ACBB'
```
cbaines, line 71: I was wondering what the a fingerprint meant in the context of serving Debian packages, then I realised it's the fingerprint of the GPG key used to sign the packages. I'm not saying you should change it, but `apt_mirror_gpg_key_fingerprint` might be clearer.

barrucadu: I'll rename it

PR #10299, opened 2020-04-15
Remove "AWS SES quota" alert

ChrisBAshton:
Removes AWS SES quota alert and associated config.

Whilst we're dealing with legacy SES debt, I've removed some Licensify code that I don't think is used anymore.

The only remaining SES config in govuk-puppet are the various [postfix::smarthost configs](https://github.com/alphagov/govuk-puppet/blob/648ca88938edef69743a0b7b916360928dd51eb4/hieradata/production.yaml#L370-L372), which I think are still required until all publishing apps' emails have moved to Notify.

Sister PR: https://github.com/alphagov/govuk-secrets/pull/950
Trello: https://trello.com/c/j5np2Vlm/1912-3-remove-aws-ses-quota-usage-higher-than-expected-alert


thomasleese:
There might be some secrets in `govuk-secrets` we can remove as well, probably the ones that relate to the unused Licensify SES keys.

ChrisBAshton:
Thanks @thomasleese - I've now raised https://github.com/alphagov/govuk-secrets/pull/950.

PR #10311, opened 2020-04-22
Adding Govuk_envsync Cron Job On All Carrenza Production Mongo Nodes

camdesgov:
This job only runs from the master/primary Mongo node, but the primary
node changes periodically. This means the sync job fails when Mongo-2 becomes
a secondary node. This PR corrects that behaviour by ensuring the sync cron jobs
are present on all nodes in the cluster.


issyl0:
What's mongo-4? https://github.com/alphagov/govuk-puppet/blob/d267d9943dd4c93121992cab5ede244099a93ae2/hieradata/class/mongo.yaml#L43 suggests that it might be for DR? When did we last failover to it, and do we want this?

cbaines:
> This job only runs from the master/primary Mongo node, but the primary
> node changes periodically. This means the sync job fails when Mongo-2 becomes
> a secondary node. This PR corrects that behaviour by ensuring the sync cron jobs
> are present on all nodes in the cluster.

I don't quite follow the motivation here. These are all "push" jobs, so they could run on either the primary or secondary instances?

camdesgov:
> I don't quite follow the motivation here. These are all "push" jobs, so they could run on either the primary or secondary instances?

There is a line in the govuk env sync script that is preventing the push from happening unless it's the primary node:
https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_env_sync/files/govuk_env_sync.sh#L686
I'm not too sure exactly why this was originally added, but I didn't feel confident just removing it.... Maybe it was added to guard against replication lag on the secondaries, and to ensure the backup always contained the most up to date data... 

We're primarily trying to make the mongo dump/backup process a bit simpler for when we come to do the publishing apps migration.  After the remaining publishing apps are migrated we'll be able to shutdown the Carrenza mongo nodes.

camdesgov:
> What's mongo-4?
> 
> https://github.com/alphagov/govuk-puppet/blob/d267d9943dd4c93121992cab5ede244099a93ae2/hieradata/class/mongo.yaml#L43-L45
> 
> suggests that it might be for DR? When did we last failover to it, and do we want this?

Hmm...yeah perhaps mongo 4 is surplus to requirements.... but as we will be shutting down all four of the Carrenza mongo nodes after the remaining Publishing apps migration (before the end of this quarter) perhaps we should leave mongo 4 as it is for now, just in case?

cbaines:
> > I don't quite follow the motivation here. These are all "push" jobs, so they could run on either the primary or secondary instances?
> 
> There is a line in the govuk env sync script that is preventing the push from happening unless it's the primary node:
> https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_env_sync/files/govuk_env_sync.sh#L686
> I'm not too sure exactly why this was originally added, but I didn't feel confident just removing it.... Maybe it was added to guard against replication lag on the secondaries, and to ensure the backup always contained the most up to date data...

While the commit message doesn't quite say it [1], I believe this was about stopping data from being pushed from each of the machines in AWS.

1: https://github.com/alphagov/govuk-puppet/pull/10271/commits/d3c61f5ce9552511e5ff5655c9a2ac5bda2e5590

I think the approach here is OK, but then we should do the same as is done in AWS with the hieradata, and specify all the jobs in the file for the class + environment [1].

1: https://github.com/alphagov/govuk-puppet/blob/master/hieradata_aws/class/staging/mongo.yaml

Another approach would be to change that new condition to just apply on AWS.

camdesgov:
> I think the approach here is OK, but then we should do the same as is done in AWS with the 
> hieradata, and specify all the jobs in the file for the class + environment [1].

👍 I've updated the PR with this suggestion

PR #10315, opened 2020-04-24
Remove rubocop-govuk from Jenkins CI

benthorner:
This is now built with GitHub Actions.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -524,9 +524,7 @@ govuk_ci::master::pipeline_jobs:
   rack-logstasher: {}
   rails_translation_manager: {}
   router-data: {}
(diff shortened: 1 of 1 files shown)


hieradata/common.yaml
```diff
@@ -524,7 +524,6 @@ govuk_ci::master::pipeline_jobs:
   rack-logstasher: {}
   rails_translation_manager: {}
   router-data: {}
-  rubocop-govuk: {}
   scss-lint-govuk: {}
   seal: {}
```
issyl0, line 528:
```suggestion

```

The Seal was switched to GitHub Actions in https://github.com/alphagov/seal/pull/162.

PR #10316, opened 2020-04-24
Use live URL when draft is missing

beccapearce:
`draft-content-store` sets `PLEK_HOSTNAME_PREFIX` to prepend `draft-` to the
URIs of its dependencies, so that it uses the special "draft" versions
of those dependencies. Two of these dependencies draft-info-frontend and
draft-feedback, have been causing errors since their draft version is
is missing.

To prevent the errors we overridden the URIs to their live
versions in draft-content-store.

Trello card: https://trello.com/c/LNbGiOcM/1898-5-investigate-an-interesting-sentry-error-relating-to-the-router

### DIFF
--- a/modules/govuk/manifests/node/s_draft_content_store.pp
+++ b/modules/govuk/manifests/node/s_draft_content_store.pp
@@ -10,9 +10,12 @@ class govuk::node::s_draft_content_store() inherits govuk::node::s_base {
   include govuk::node::s_content_store
 
   $app_domain = hiera('app_domain')
+  $app_domain_internal = hiera('app_domain_internal')
 
   govuk_envvar {
     'PLEK_HOSTNAME_PREFIX': value => 'draft-';
     'PLEK_SERVICE_SIGNON_URI': value => "https://signon.${app_domain}";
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/node/s_draft_frontend.pp
```diff
@@ -28,5 +28,7 @@
     'PLEK_SERVICE_LICENSIFY_URI': value => "https://licensify.${app_domain_internal}";
     'PLEK_SERVICE_MAPIT_URI': value => "https://mapit.${app_domain_internal}";
     'PLEK_SERVICE_SEARCH_URI': value => "https://search.${app_domain_internal}";
+    'PLEK_SERVICE_FEEDBACK_URI': value => "https://feedback.${app_domain_internal}";
+    'PLEK_SERVICE_INFO_FRONTEND_URI': value => "https://info-frontend.${app_domain_internal}";
```
cbaines, line 32: To have an effect on the routes in the Router API/Router, this change will need to affect the draft content store, and I'm not sure the draft content store is on this machine?

PR #10321, opened 2020-04-29
Enable continuous deployment to Staging

benthorner:
Depends on: https://github.com/alphagov/govuk-jenkinslib/pull/71
Depends on: **TODO - PR to add necessary govuk-secrets**

    Previously we had an 'integration-app-deploy' job that was triggered by
    a successful build of the master branch for each app on CI. This renames
    the job to 'Deploy_App_Downstream' and modifies the 'Deploy_App' job to
    call it after a successful run of Smokey i.e. Continuous Deployment.
    
    Continuous Deployment (CD) was originally proposed in RFC 70 [1]. Manual
    deployments to staging and then production take up a lot of time. Although
    we have Smokey tests, we still rely on manual testing for many deployments,
    which means 'full' CD would be quite a challenge. Enabling CD to Staging
    is a step in that direction, which will have a few benefits:
    
    - It reduces the amount of time spent deploying things
    - It encourages us to improve our Smokey tests
    - It encourages us to not ignore our Smokey tests
    
    While there is a risk of 'broken' code being deployed to Staging, this
    is also a risk now: sometimes we don't see the warnings [2]. In general,
    we shouldn't be deploying 'broken' code; and if that does happen, we
    should have automated (Smokey) tests to protect us from ourselves,
    instead of relying on (people remembering to do) manual testing.
    
    [1]: https://github.com/alphagov/govuk-rfcs/blob/master/rfc-070-path-towards-continuous-deployment-cd.md
    [2]: https://github.com/alphagov/smokey/pull/633

Post-merge things to address:

- [ ] Update the dev docs to explain the new process (see [here](https://docs.publishing.service.gov.uk/manual/deploying.html#1-test-on-integration))
- [ ] Decide if we want to [exclude staging deploys from #govuk-deploy](https://github.com/alphagov/govuk-puppet/pull/10321#issuecomment-621244804)


modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
```diff
@@ -0,0 +1,53 @@
...
+      - shell: |
+        if ! [ "$TAG" = "release_*" ]; then
+          echo "Skipping downstream deploy for non-release tag $TAG"
+          exit 0
+        fi
+
+        <% if @deploy_url %>
+        JSON="{\"parameter\": [{\"name\": \"TARGET_APPLICATION\", \"value\": \"$TARGET_APPLICATION\"}, {\"name\": \"TAG\", \"value\": \"$TAG\"}, {\"name\": \"DEPLOY_TASK\", \"value\": \"$DEPLOY_TASK\"}, {\"name\": \"NOTIFY_RELEASE_APP\", \"value\": \"true\"}, {\"name\": \"SLACK_NOTIFICATIONS\", \"value\": \"true\"}], \"\": \"\"}"
+
+        # Deploy to AWS Integration environment
```
richardTowers, line 21: This comment isn't correct any more, as it could be going to staging, right?

issyl0:
We currently don't surface integration app deploys in the #govuk-deploy channel on Slack. Given the new Staging ones will all be automated too, do we want to exclude those as well?

modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
benthorner: Oops - fixed 👍. 

(The original author was really keen on the word 'Integration'!)

benthorner:
@issyl0 good point, although I don't really look at that channel, so I have no opinion. I've added it to a list of things to think about.

kevindew:
Thanks for pushing this to attention, it's good to have the discussions around this and I like the angle of giving our Smoke tests more love. Thoughts that could to my mind based on this are:

- Is this the first step in a longer piece of work? In isolation it makes me question the purpose of staging, but if we're going to drop a prod equivalent soon after this seems less of a pain point
- Sounds like if this is merged we need to get devs into a master-is-always-deployable mindset (which, whatever happens, with this would be a good thing in my opinion) since we still have a lot of "is this ok to deploy" checks
- I think there's some apps (Content Publisher comes to mind) that have no smoke tests
- I assume we'd need full Smokey suite to pass, which could make it quite disruptive if anyone is testing out a branch on integration that breaks a smoke test (on this point I somewhat wonder if it'd better if we deployed both staging and integration deploy automatically on a master merge and just go to production if staging smoke tests pass, do we really it need it to pass two environments worth of smoke tests?)
- The pain point I anticipate this giving me is automatic DB migrations. I seem to find myself every few weeks or so having a longer migration to run and doing it outside a deployment since long migrations crash deployments and can leave you an inconsistent state. Personally I'd be for deploys to never run a migration but I don't think that'd be popular. However similarly I don't want to have to run around trying to stop deploys from starting. Maybe rake tasks for slower things is a way to go? but yeah this point is my bug bear.

On the PR:

- Does integration know which apps are AWS or not? From reading it looks like this job will try both Carrenza and AWS, but I'm probably missing a nuance
- It might be good to explicitly name Carrenza parameters since AWS is now essentially the main environment
- Do we know that integration can talk to staging? I'm not sure if it's firewalled at all.

benthorner:
@kevindew thanks for commenting! I'm going to put the comments on the code itself to one side for a little while, while we discuss the overall issue. I'm not expecting this to work first time! One thing to clarify: this is intentionally dumb and tries to deploy to both staging providers, which is sometimes valid; hopefully it'll fail gracefully, or else we can build in some intelligence.

Now for the main points:

> I assume we'd need full Smokey suite to pass

Yes. I think this will motivate people to not break Smokey, and attend to it when it's broken. We can still use the old manual process as a fallback.

I also had an idea to start tagging Smokey tests by app, so we can limit the run to just relevant tests. This would also have the benefit of reducing the run time.

> I somewhat wonder if it'd better if we deployed both staging and integration deploy automatically

Currently, staging is something of a 'last resort' when we've been a bit careless and broken Integration. This change could/should have the effect that we treat Integration and Staging more interchangeably.  This will help us move away from the 'last resort' safety net, to the point where we can then ask: what extra value is a Staging environment giving us?

If we can get to a single integration/staging environment, I think we'll have an argument to doing CD all the way to production, since we'll have less tolerance for dodgy deployments that break the only non-production environment we have to test things in.

> some apps (Content Publisher comes to mind) that have no smoke tests

Indeed. Currently there's little/no motivation to write any. The potential to break more stuff should motivate us to write some, as they become necessary.

As a starting point, we could have a generic smoke test that pings the `/healthcheck` endpoint of each app, much like the Icinga alert. This should provide some base level coverage.

> we need to get devs into a master-is-always-deployable mindset

We certainly do. Maybe we're there already? Or is this hinting at doing more manual testing before we hit "merge"?

Whether our deployments are manual or automatic, there's always a risk of breaking something. Manual deployments offer more time to manually spot issues, if someone actually remembers to look for them. Automatic deployments drive us to detect issues reliably/automatically, and also allows us to deploy those fixes faster.

> Is this the first step in a longer piece of work?

I hope so, but then I'm not in charge exactly. I think it will take time for people to adjust to the new workflow, and iron out the kinks: more Smokey tests, more careful checking of PRs. Beyond that I can see us moving to CD in production on a per-app basis.

> The pain point I anticipate this giving me is automatic DB migrations.

This one seems tangential to me, since we already do automatic deploys to Integration. Can you help me understand why CD to Staging would worsen the situation?

kevindew:
> @kevindew thanks for commenting! I'm going to put the comments on the code itself to one side for a little while, while we discuss the overall issue. I'm not expecting this to work first time! One thing to clarify: this is intentionally dumb and tries to deploy to both staging providers, which is sometimes valid; hopefully it'll fail gracefully, or else we can build in some intelligence.

Sure, trying it out you'll get a 500 from Jenkins if you try (java.lang.IllegalArgumentException: Illegal choice for parameter TARGET_APPLICATION: frontend) which seems a better outcome than a failed or attempted build. 

> I also had an idea to start tagging Smokey tests by app, so we can limit the run to just relevant tests. This would also have the benefit of reducing the run time.

There is a somewhat crude approach to this already based on the naming of the cucumber tests. There's is this rather daunting [python script](https://github.com/alphagov/smokey/blob/master/nagios_check_cache.py) that works out which features to report to different icinga checks.

> Currently, staging is something of a 'last resort' when we've been a bit careless and broken Integration. This change could/should have the effect that we treat Integration and Staging more interchangeably. This will help us move away from the 'last resort' safety net, to the point where we can then ask: what extra value is a Staging environment giving us?
> 
> If we can get to a single integration/staging environment, I think we'll have an argument to doing CD all the way to production, since we'll have less tolerance for dodgy deployments that break the only non-production environment we have to test things in.

It wouldn't really describe it as a last resort I've always understood it to be an environment that is intentionally near identical to production for a realistic experience of what running something in production will be. Whereas integration has:

- Different hardware
- Different data (filtered and different users)
- Different access controls
- A higher tolerance for experimental testing/breaking the environment

I do agree it seems a pain to have both and somewhat superfluous. Does feel like some of the access / date issues would be a pain to killing one though.

I didn't really gather where you were for or against the suggestion to deploy to staging immediately rather than waiting for Smokey integration to pass?

> > we need to get devs into a master-is-always-deployable mindset
> 
> We certainly do. Maybe we're there already? Or is this hinting at doing more manual testing before we hit "merge"?
> 
> Whether our deployments are manual or automatic, there's always a risk of breaking something. Manual deployments offer more time to manually spot issues, if someone actually remembers to look for them. Automatic deployments drive us to detect issues reliably/automatically, and also allows us to deploy those fixes faster.

I have the impression there's still often a culture of merge into master and then test the feature after merge or even on staging. Evidenced that it's not totally unusual for someone to say something is not ready to deploy. Though I imagine the Coronavirus service has helped improve this.

> > Is this the first step in a longer piece of work?
> 
> I hope so, but then I'm not in charge exactly. I think it will take time for people to adjust to the new workflow, and iron out the kinks: more Smokey tests, more careful checking of PRs. Beyond that I can see us moving to CD in production on a per-app basis.

It'd be good to have some sort of view forward as I see this change as introducing a new ambiguity to the staging environments purpose and I wouldn't want that for more than a short-term period.

> > The pain point I anticipate this giving me is automatic DB migrations.
> 
> This one seems tangential to me, since we already do automatic deploys to Integration. Can you help me understand why CD to Staging would worsen the situation?

I find this a pain point with integration too. Normally it's softened because I'll have deployed the branch to integration manually and run the migration first then when the integration deploy kicks in the migration is already run. This wouldn't be a great workflow all the way to production though.

benthorner:
Closing, as we will be revisiting this work in stages, as part of a beta for Continuous Deployment.

PR #10329, opened 2020-05-04
[WIP] Rename `email_address_override_whitelist`

chao-xian:
…allowlist`

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -211,6 +211,8 @@ task :check_consistency_between_aws_and_carrenza do
     govuk::apps::email_alert_api::email_address_override
     govuk::apps::email_alert_api::email_address_override_whitelist
     govuk::apps::email_alert_api::email_address_override_whitelist_only
+    govuk::apps::email_alert_api::email_address_override_allowlist
+    govuk::apps::email_alert_api::email_address_override_allowlist_only
     govuk::apps::email_alert_api::email_archive_s3_bucket
     govuk::apps::email_alert_api::email_archive_s3_enabled
     govuk::apps::email_alert_api::enabled
--- a/modules/govuk/manifests/apps/email_alert_api.pp
+++ b/modules/govuk/manifests/apps/email_alert_api.pp
@@ -62,11 +62,19 @@
 # [*email_address_override*]
 #   Configure all emails to be sent to this address instead.
 #
+# TO BE DELETED:
 # [*email_address_override_whitelist*]
 #   Allow a whitelist of email addresses that ignore the override option.
 #
+# TO BE DELETED:
 # [*email_address_override_whitelist_only*]
 #   Specify that only emails that are whitelisted should be sent out.
+#
+# [*email_address_override_allowlist*]
+#   Allow a list of email addresses that ignore the override option.
+#
+# [*email_address_override_allowlist_only*]
+#   Specify that only emails that are allowed should be sent out.
 #
 # [*delivery_request_threshold*]
 #   Configure the number of requests that the rate limit will allow to be made in one minute.
@@ -132,8 +140,10 @@ class govuk::apps::email_alert_api(
   $email_alert_auth_token = undef,
   $email_service_provider = 'NOTIFY',
   $email_address_override = undef,
-  $email_address_override_whitelist = undef,
-  $email_address_override_whitelist_only = false,
+  $email_address_override_allowlist = undef,
+  $email_address_override_allowlist_only = false,
+  $email_address_override_whitelist = $email_address_override_allowlist,
+  $email_address_override_whitelist_only = $email_address_override_allowlist_only,
   $delivery_request_threshold = undef,
   $db_username = 'email-alert-api',
   $db_password = undef,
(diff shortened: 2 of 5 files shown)


hieradata/development.yaml
```diff
@@ -151,7 +151,7 @@ govuk::apps::email_alert_api::db_password: "%{hiera('govuk::apps::email_alert_ap
 govuk::apps::email_alert_api::db_port: 6432
 govuk::apps::email_alert_api::db_allow_prepared_statements: false
 # This list should be kept in sync with https://www.notifications.service.gov.uk/services/ecfc7e2f-5145-45a6-9413-5d9b6e813ea9/users
-govuk::apps::email_alert_api::email_address_override_whitelist:
+govuk::apps::email_alert_api::email_address_override_allowlist:
```
cbaines, line 154: At least to me, `allowlist` doesn't make as much sense. I'm fine with changing this, but how about `allowed_email_addresses_override`?

cbaines: Or maybe just `allowed_email_addresses`, I'm not sure what the override bit means.

chao-xian: I like your suggestion!

Rakefile
```diff
@@ -209,8 +209,8 @@ task :check_consistency_between_aws_and_carrenza do
     govuk::apps::email_alert_api::db_password
     govuk::apps::email_alert_api::delivery_request_threshold
     govuk::apps::email_alert_api::email_address_override
-    govuk::apps::email_alert_api::email_address_override_whitelist
-    govuk::apps::email_alert_api::email_address_override_whitelist_only
+    govuk::apps::email_alert_api::allowed_email_addresses
+    govuk::apps::email_alert_api::allowed_email_addresses_only
```
benthorner, line 213: This is one confusing boolean (the original, that is). Looking in [the code](https://github.com/alphagov/email-alert-api/blob/4890e19da81d19a91fdafed783a0ffcd9f907b3d/app/services/delivery_request_service/email_address_overrider.rb), it seems its purpose is to:

- If `true`, only send email to the configured addresses
- If `false`, send all other email to an "override address"

I think we could just scrap this variable, and just rely on the presence of the override address if one is set.

modules/govuk/manifests/apps/email_alert_api.pp
```diff
@@ -261,16 +261,16 @@
       value   => $publishing_api_bearer_token;
   }
 
-  if $email_address_override_whitelist_only {
-    govuk::app::envvar { "${title}-EMAIL_ADDRESS_OVERRIDE_WHITELIST_ONLY":
-      varname => 'EMAIL_ADDRESS_OVERRIDE_WHITELIST_ONLY',
+  if $allowed_email_addresses_only {
+    govuk::app::envvar { "${title}-allowed_email_addresses_ONLY":
+      varname => 'allowed_email_addresses_ONLY',
```
benthorner, line 266: We seem to have lost the original casing here.

kevindew:
In order to allow migrating to the changed env vars it's probably worth having a period of time where there are both the old whitelist env vars and the new allowed ones so that we don't need to sync app deploys with puppet deploys. Then once the app is switched over remove the whitelist ones.

Rakefile
chao-xian: It appears to be used for logic [here](https://github.com/alphagov/govuk-puppet/pull/10329/files/674788995ba1e2abf377923d15041e23e1b935c5#diff-b1c1d8bf6c96caf2e78a19d9bdbbf71bR264)

chao-xian:
I've reworked this PR to take on board Kevin's suggestion to support both variables temporarily as we switch over.

modules/govuk/manifests/apps/email_alert_api.pp
```diff
@@ -132,8 +140,10 @@
   $email_alert_auth_token = undef,
   $email_service_provider = 'NOTIFY',
   $email_address_override = undef,
-  $email_address_override_whitelist = undef,
-  $email_address_override_whitelist_only = false,
+  $email_address_override_allowlist = undef,
+  $email_address_override_allowlist_only = false,
+  $email_address_override_whitelist = $email_address_override_allowlist,
+  $email_address_override_whitelist_only = $email_address_override_allowlist_only,
```
kevindew, line 146: I'd suggest you can delete the parameters for the whitelist options now (and the associated hiera config)

modules/govuk/manifests/apps/email_alert_api.pp
```diff
@@ -261,6 +271,7 @@
       value   => $publishing_api_bearer_token;
   }
 
+  # TO BE DELETED:
```
kevindew, line 274: It'd be good to say when this can be deleted so you know when/if you can do the deletion.

modules/govuk/manifests/apps/email_alert_api.pp
```diff
@@ -261,6 +271,7 @@
...
   if $email_address_override_whitelist_only {
```
kevindew, line 275: This could be `$email_address_override_allowlist_only` with a similar approach taken for the actual list itself.

chao-xian:
Having a stale PR clear out... closing 😓

chao-xian:
Thank you to everyone for their feedback!

ChrisBAshton:
Looks like Kelvin intended to close this - so I'll do that now.

PR #10340, opened 2020-05-06
adjust govuk_env_sync integration

fredericfran-gds:
Actions:
1. push local links manager backup from staging to S3
2. pull local links manager backup from staging S3 to integration


hieradata_aws/class/staging/db_admin.yaml
```diff
@@ -43,6 +43,17 @@ govuk_env_sync::tasks:
     temppath: "/tmp/local-links-manager_production"
     url: "govuk-production-database-backups"
     path: "postgresql-backend"
+  "push_local-links-manager_staging_daily":
+    ensure: "present"
+    hour: "4"
+    minute: "30"
+    action: "pull"
```
cbaines, line 50: There's an inconsistency here between the action and the name above "push_local-links-manager_staging_daily"

fredericfran-gds: will be fix here: https://github.com/alphagov/govuk-puppet/pull/10342 thanks!

PR #10353, opened 2020-05-14
Upgrade Mongo On CI Agents 1 - 3

camdesgov:
Upgrade Mongo to 2.6 from 2.4.9 on ci_agent machines 1 - 3. This
is a necessary step towards migrating the remaining publishing apps
currently using Mongo databases from Carrenza to AWS. We will be updating the apps
to use DocumentDB which is incompatible with versions of Mongo below 3.6. We 
are upgrading the CI Agent Machines to bring the versions inline with the versions
used in all other environments. For more information please see this trello card:
https://trello.com/c/YXXu4s5c/3422-fix-mongo-documentdb-errors

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1039,7 +1039,7 @@ limits::entries:
 
 mongodb::repository::apt_mirror_hostname: "%{hiera('apt_mirror_hostname')}"
 mongodb::repository::apt_mirror_gpg_key_fingerprint: "%{hiera('apt_mirror_fingerprint')}"
-mongodb::server::version: '2.4.9'
+mongodb::server::version: '2.6.12'
 
 monitoring::checks::aws_origin_domain: 'dev.govuk.digital'
 monitoring::checks::http_username: "%{hiera('http_username')}"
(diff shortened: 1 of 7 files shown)


cbaines:
I get from the commit message that this relates to upgrading Mongo, but what I'm missing from the commit message is why Mongo is being upgraded on these specific machines, and what's significant about the version "2.4.9"?

camdesgov:
> I get from the commit message that this relates to upgrading Mongo, but what I'm missing from the commit message is why Mongo is being upgraded on these specific machines, and what's significant about the version "2.4.9"?

Updated the commit message to hopefully add a better explanation. Not sure if you remember, but some time ago, the publishing apps with Mongo databases were showing errors when we migrated them to AWS staging. This was down to a version of the Mongoid gem being incompatible with DocumentDB. We could fix this by upgrading the mongoid gem to 6.3.0 and the mongo gem to 2.5.0. .....but unfortunately we couldn't just upgrade the gem as the mongo 2.5.0 gem is incompatible with the lower version of MongoDB we had running on our servers( 2.4.9) : https://docs.mongodb.com/ruby-driver/master/reference/driver-compatibility/.  The CI Agent machines have their versions pinned at 2.4.9 (I believe to match the version of mongo we used in prod ), so we're updating them inline with the other environments.

PR #10358, opened 2020-05-15
Fall back to rubygems if gem is not in gemstash

richardTowers:
... or if our gemstash goes down for some reason.

Pretty sure we're on > 1.12.x everywhere now, and if not this will force
us to fix it.

Note that the name of this config setting is slightly different to the
one in the comment - `/__FALLBACK...` instead of `.FALLBACK`. This is
what I get in my `~/.bundle/config` if I run `bundle config mirror.https://rubygems.org.fallback_timeout true`,
which is what the docs say I should do to set this flag.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -121,5 +121,5 @@ DEPENDENCIES
   sshkey (= 1.9.0)
   webmock (~> 1.20.0)
 
-RUBY VERSION
-   ruby 1.9.3p550
+BUNDLED WITH
+   1.17.3
--- a/modules/govuk_bundler/templates/bundle_config.erb
+++ b/modules/govuk_bundler/templates/bundle_config.erb
@@ -1,4 +1,3 @@
 ---
-# FIXME: Once all bundler versions are > 1.12.x i.e. ruby > 2.x
-# BUNDLE_MIRROR__HTTPS://RUBYGEMS.ORG.FALLBACK_TIMEOUT: 'true'
+BUNDLE_MIRROR__HTTPS://RUBYGEMS__ORG/__FALLBACK_TIMEOUT: 'true'
 BUNDLE_MIRROR__HTTPS://RUBYGEMS.ORG: <%= @server %>/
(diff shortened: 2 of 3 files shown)


sihugh:
Worth a shot 👍

kevindew:
Unfortunately we're still on Ruby 1.9.3 for this repo (and potentially a couple of other places) where we need Ruby to run before puppet has run and we're dealing with Trusty's Ruby version.

richardTowers:
😭 

> Pretty sure we're on > 1.12.x everywhere now, and if not this will force
us to fix it.

Why am I so naive 😭

kevindew:
You might actually be in luck about the bundler version. Having a quick look on a ci-agent machine  has the bundler version for Ruby 1.9.3-p550 being 1.14.5.

richardTowers:
Looks like it's a bit older on the jenkins master:

```
richardtowers@ci-integration-jenkins-1:~$ rbenv shell 1.9.3-p550
richardtowers@ci-integration-jenkins-1:~$ bundle --version
Bundler version 1.7.4
```

Is that going to be a problem?

kevindew:
> Looks like it's a bit older on the jenkins master:
> 
> ```
> richardtowers@ci-integration-jenkins-1:~$ rbenv shell 1.9.3-p550
> richardtowers@ci-integration-jenkins-1:~$ bundle --version
> Bundler version 1.7.4
> ```
> 
> Is that going to be a problem?

No idea, quite how fatal this version restriction would be. Given at least one CI agent has a newer version (see below) it at least means it can be updated. Which I guess can be achieved rather simply with some yolo commands on the machines and much harder by tweaking the provisioning image.

```
kevindew@ci-integration-ci-agent-1:~$ rbenv shell 1.9.3-p550
kevindew@ci-integration-ci-agent-1:~$ bundler --version
Bundler version 1.14.5
```

Could just give it a try and see what happens? You can do that by deploying puppet to carrenza infrastructure at https://ci-deploy.integration.publishing.service.gov.uk/ then run puppet on the Jenkins for Deploying (jenkins-1.ci) and Jenkins for CI (ci-master-1.ci) and see if the latter can deploy puppet and the latter can run a puppet PR build.

PR #10360, opened 2020-05-18
Foundations to swap asset host for government-frontend and static

kevindew:
Trello: https://trello.com/c/oNEtjIVp/99-serve-static-assets-from-www-hostname

This is the first step to implementing one of the key components of [RFC 115](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-115-enabling-http2-on-govuk.md) where we switch the serving application assets from the `assets.publishing.service.gov.uk` hostname to the `www.gov.uk` origin hostname.

As I plan to link to this to PR in a few places it seems worth summarising the what and why of changes:

- Application assets (CSS/JS/Images) will no longer be served from the `assets.publishing.service.gov.uk` hostname and instead be served off the `www.gov.uk` hostname. This allows all assets and page to be served from the same TCP connection and allow Fastly to perform HTTP2 Prioritisation across the assets (which should produce a faster initial usable page rendering).
- Asset manager assets will remain served from the `assets.publishing.service.gov.uk` hostname (there is some discussion in the RFC that these should also be moved to the `www.gov.uk` hostname however this is a bigger challenge)
- Static templates will continue to be served off the assets hostname publicly (internally apps communicate via a static hostname). The fact we serve the static app from the assets hostname doesn't make a huge amount of sense to me since this is a route to an application, given we plan to eventually remove static I hope this can be deferred rather than needing a migration to a different public hostname.

This PR itself sets up the means to host proxy assets to apps through the www hostname. The expectation is that this is done under a path prefixed `assets` so to not conflict with any pages. This proxying is set-up to mirror the connection rules provided on the `assets.publishing.service.gov.uk` hostname (although I am skeptical that these numbers are particularly tuned and I think they are just values that have stuck). It then changes the basic app template so that apps can have multiple asset paths defined (this allows for a seamless migration).

I've applied this configuration initially to government-frontend and static as the first apps to try out this changeover. Government-frontend was chosen as it's one of the main frontend apps for GOV.UK and thus hopefully quite representative of the challenges to resolve this. Static was chosen because that's a somewhat special app with quite particular needs - for instance developers use the live version of static to insert templates into either local development apps or Heroku preview apps. In order to support these use cases static is set up with a hostname for assets so that it always returns absolute URLs to assets and fonts are set-up with cross origin access so that they can be used when accessing the asset from a different hostname (e.g. `*.dev.gov.uk` using static on `www.gov.uk`).


modules/router/templates/router_include.conf.erb
```diff
@@ -33,6 +33,17 @@ location ~ ^\/[A-Z]+[A-Z\W\d]+$ {
   rewrite ^(.*)$ https://$host$uri_lowercase permanent;
 }
 
+<%- @app_specific_static_asset_routes.each do |alias_path, vhost_name| -%>
+# Static asset serving for <%= vhost_name %>
+set $upstream_<%= vhost_name.delete "-" %> https://<%= vhost_name %>.<%= @app_domain_internal %>;
```
thomasleese, line 38: I'd be tempted to pull this line out into some variables, as I can see they're also used below in the block itself. Although I'm not sure how conventional it is to define variables in ERB files.

kevindew: Sure, I've done that now.

kevindew:
@thomasleese I also snuck the configurations for frontend and collections as I've almost finished setting them up.

PR #10369, opened 2020-05-20
Add check for deploy freeze when triggering downstream deploy

theseanything:
This adds a step to the Deploy_App_Downstream job to query the Release app and prevent deployments if the deploy freeze attribute is true.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
@@ -9,6 +9,19 @@
           days-to-keep: 30
           artifact-num-to-keep: 5
     builders:
+      - shell: |
+          # Check for deploy freeze set in Release App
+          APPLICATION_METADATA=$(curl -s \
+            -H "Accept: application/json" \
+            -H "Authorization: Bearer $RELEASE_APP_NOTIFICATION_BEARER_TOKEN" \
+            "https://release.publishing.service.gov.uk/applications/$TARGET_APPLICATION")
+
+          DEPLOY_FREEZE=$(echo "$APPLICATION_METADATA" | jq .deploy_freeze)
+
+          if $DEPLOY_FREEZE; then
+            echo "This application is under a deploy freeze in Release app. Aborting"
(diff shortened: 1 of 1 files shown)


theseanything:
> Script looks good 👍. Can we move it to its own step, before the sleep/Smokey step? I think it makes sense to abort early for this check; we benefit from doing the commit checks a bit later as they also debounce deployments for a string of PRs merged around the same time.

Sure thing!

modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
```diff
@@ -9,15 +9,6 @@
...
-          # Wait for any app restarts to complete before running Smokey
-          sleep 60
-      - trigger-builds:
-          - project: Smokey
-            current-parameters: true
-            block: true
-      <% end %>
       - shell: |
           # Workaround for our inconsistent repo vs. deployment naming
           case "$TARGET_APPLICATION" in
```
benthorner, line 14: I think this bit needs to stay where it was. Fortunately, it's only used there. It's probably worth doing a test deployment of this on Integration, just in case it doesn't work.

theseanything: `$TARGET_APPLICATION` is needed for the call to the release API too. I assumed environment variables are rolling over to subsequent shell scripts, because it's used again in line 65.

theseanything: But I'll test in integration and write back with the results!

benthorner: > $TARGET_APPLICATION is needed for the call to the release API too.

Yup. I'm referring to the whole case statement, which sets a `REPO` variable that's only needed in the final step.

theseanything: Gottcha - soo my brain was thinking this was actually setting the TARGET_APPLICATION variable 🤦  - I'll shift it down

PR #10377, opened 2020-05-22
Replace 'integration-*-deploy' jobs

benthorner:
https://trello.com/c/fq2LN9bl/153-rewrite-the-deploy-to-integration-job-so-it-can-be-used-for-cd

Depends on: https://github.com/alphagov/govuk-secrets/pull/980
Depends on: https://github.com/alphagov/licensify/pull/487

This does a couple of things:

- Renames 'integration-puppet-deploy' to 'Deploy_Puppet_Downstream'
- Removes 'integration-licensify-deploy'


modules/govuk_jenkins/templates/jobs/deploy_puppet_downstream.yaml.erb
```diff
@@ -0,0 +1,35 @@
...
+    builders:
+      - shell: |
+          JSON="{\"parameter\": [{\"name\": \"TAG\", \"value\": \"$TAG\"}], \"\": \"\"}"
+
+          # Deploy to Integration environment
+          CRUMB=$(curl https://<%= @jenkins_downstream_api_user %>:<%= @jenkins_downstream_api_password %>@<%= @deploy_url %>/crumbIssuer/api/json | jq --raw-output '. | .crumb')
+
+          curl -f -H "Jenkins-Crumb:$CRUMB" -XPOST https://<%= @jenkins_downstream_api_user %>:<%= @jenkins_downstream_api_password %>@<%= @deploy_url %>/job/Deploy_Puppet/build -d token=<%= @puppet_auth_token %> --data-urlencode json="$JSON"
+
+          # Deploy to AWS Integration environment
```
theseanything, line 19:
```suggestion
          # Deploy to AWS environment
```
or remove these comments? They make me think this only applies to the integration environment?

modules/govuk_jenkins/templates/jobs/deploy_puppet_downstream.yaml.erb
```diff
@@ -0,0 +1,35 @@
+---
+- job:
+    name: Deploy_Puppet_Downstream
+    project-type: freestyle
+    description: "Kicks off a Puppet deploy in a downstream environment"
```
theseanything, line 5: I'm finding the name and description quite confusing - am I right in thinking this job is to trigger the "Deploy puppet" jenkins job on the "downstream" jenkins? From the current name, I'd think that is job is actually doing the work to "deploy puppet". 

Can we be more specific in the name and description? e.g. Trigger_Deploy_Puppet_Downstream

benthorner: Hmm...this would be quite time consuming to change, as we should make it consistent with the existing `Deploy_App_Downstream` job. I think it's helpful that they both begin with `Deploy_X`, so that these related jobs are grouped together in the UI.

If we do change the name, I'd like to stick with a `Deploy_App` prefix. What about `Deploy_Puppet_Downstream_Trigger`?

modules/govuk_jenkins/templates/jobs/deploy_puppet_downstream.yaml.erb
benthorner: I've updated to say "downstream" (just to be clear).

PR #10387, opened 2020-05-29
Support running only app-specific scenarios

benthorner:
https://trello.com/c/VdbPVqto/154-support-running-only-smokey-tests-that-are-relevant-to-an-app

Depends on: https://github.com/alphagov/smokey/pull/668

This adds an optional string parameter that will be used by the
'jenkins.sh' script in Smokey to only run scenarios with a tag that
matches, noting that many scenarios are not relevant to a given app.
We need this feature to support Continuous Deployment, so that deploys
are not blocked by unrelated failures.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/smokey.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/smokey.yaml.erb
@@ -29,6 +29,10 @@
             predefined-parameters: |
                 NSCA_CHECK_DESCRIPTION=<%= @service_description %>
                 NSCA_OUTPUT=<%= @service_description %> failed
+    parameters:
+      - string:
+          name: TARGET_APPLICATION
+          description: Application to target (optional).
     builders:
       - shell: |
           set +x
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/templates/jobs/smokey.yaml.erb
```diff
@@ -29,13 +29,16 @@
...
+          name: TARGET_APPLICATION
+          description: Application to target (optional).
     builders:
       - shell: |
           set +x
           export TARGET_PLATFORM=<%= @environment %>
-          export MYTASK=<%= @smokey_task %>
-          set -x
           ./jenkins.sh
+          set -x
```
theseanything, line 41: How come `set -x` is moved after executing `./jenkins.sh`?

benthorner: Ah, that was a quick change to see if we needed it in `jenkins.sh` as well (we do, as it doesn't persist between processes). I left it here, as it was weird to only be echoing some of the commands, although I realise I didn't explain why 😬. Shall I move it back?

theseanything: No problem! Yeah I think move it back because its a change not really relevant to the scope of this PR.

benthorner: Sure, updated.

benthorner:
That was tempting. When I tried it, I realised I kept them separate, as they don't rely on each other. I'll leave them as-is, to avoid fiddling with the commit messages.

PR #10392, opened 2020-06-03
Sync and restyle Sidekiq job count

benthorner:
https://trello.com/c/epDHdjIe/311-test-increasing-database-connections-and-workers

Previously the line showing the count of processed Sidekiq jobs used
a different interval to the other lines on the graph, which lead to
confusion since the points on the graph were being aggregated over
different time intervals. This fixes that.

After fixing the time interval, it's hard to distinguish the job count
line from the lines showing the job times. This restyles the line so
it's clear that it represents a different kind of series.

## Before

Screenshot 2020-06-03 at 11 12 58

## After

Screenshot 2020-06-03 at 11 13 27

### DIFF
--- a/modules/grafana/files/dashboards/sidekiq_job_run_time.json
+++ b/modules/grafana/files/dashboards/sidekiq_job_run_time.json
@@ -43,7 +43,9 @@
           "seriesOverrides": [
             {
               "alias": "Count",
+              "dashes": true,
               "fill": 0,
+              "linewidth": 2,
               "yaxis": 2
             }
           ],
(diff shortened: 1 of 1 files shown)


cbaines:
Interesting, can you provide screenshots to show before and after this change.

benthorner:
@cbaines sure - I meant to and forgot 😬. Screenshots are now in the description.

(I did also spend some time trying to replace the second graph with a heatmap. The problem we found is that it's easy to look at the scatter of points and see it as representing volume, but each point is actually an aggregation. So it would be more useful to have a heatmap that tells you "at this time, lots of jobs took x time, some took y time, a couple took z time, etc. Unfortunately, it doesn't seem to be possible to get Grafana's heatmap panel to use a custom bucket count.)

PR #10424, opened 2020-06-12
Fix misleading 'check_proc_running' alert messages

ChrisBAshton:
Some of our alerts were down as 'X not running', as opposed to
'X running'. When the alert happens, the 'Service' message we
see in Icinga/Nagstamon is the 'X running'/'X not running'
message. It's a little confusing to see 'X running' as an alert
when the issue is that it ISN'T running.

### DIFF
--- a/modules/collectd/manifests/service.pp
+++ b/modules/collectd/manifests/service.pp
@@ -10,7 +10,7 @@ class collectd::service {
 
   @@icinga::check { "check_collectd_running_${::hostname}":
     check_command       => 'check_nrpe!check_proc_running!collectd',
-    service_description => 'collectd system statistics daemon',
+    service_description => 'collectd system statistics daemon not running',
     host_name           => $::fqdn,
     notes_url           => monitoring_docs_url(check-process-running),
   }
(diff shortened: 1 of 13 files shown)


cbaines:
I think the way these Icinga services are meant to be used is that they describe something working. Like a process running. That makes sense with the OK status. If you have a check which says "process X is running" and the status is OK, then all is good, the process is running.

If you have a check which says "process Y is running" and the status is WARNING/CRITICAL, then something is probably up, maybe the process isn't running.

ChrisBAshton:
I agree that's how it was probably intended to be used, but most of our other alerts are in the negative:

* "puppet last run errors"
* "finder-frontend nginx 5xx rate too high"
* "collections-publisher upstart not up"
* "email-alert-api: unprocessed content changes"
* "high zombie procs"
* "high load on"

It saves a bit of confusion to make the nature of these statuses consistent.

Admittedly there are a couple of other alerts which should also be looked at in that regard:

* "collections-publisher app healthcheck"
* "Smokey"
* "Runs a rake task on Search API that generates the sitemap files. It is safe to re-run in-hours."

PR #10427, opened 2020-06-12
Add splunk module to frontend

MahmudH:
### How to add govuk_splunk module to a machine
1. Include the module into the box you want to monitor the logs for. In this instance, we are going to send logs from the frontend node to Splunk. 
2. Create the inputs file in the govuk_splunk module.
3. Add the path of the inputs.conf file and make sure it is present. 

Trello: https://trello.com/c/OsPfVq9q/1943-5-spike-into-getting-logs-into-splunk-pipeline-%F0%9F%8D%90

### DIFF
--- a/modules/govuk/manifests/node/s_frontend.pp
+++ b/modules/govuk/manifests/node/s_frontend.pp
@@ -9,6 +9,7 @@ class govuk::node::s_frontend inherits govuk::node::s_base {
 
   include nginx
   include govuk_awscloudwatch
+  include govuk_splunk
 
   # The catchall vhost throws a 500, except for healthcheck requests.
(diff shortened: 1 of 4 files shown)


cbaines:
So, the frontend machines run a number of apps, and it's not clear to me from these changes which logs from the machine/apps that these changes are aiming to collect, can you clarify?

MahmudH:
The frontend machine logs that Cyber will see in Splunk are what is present in the frontend box through nginx logs (/var/log/nginx). 

These are: `email-alert-frontend`, `frontend`, `static`, `government-frontend`, `feedback`, `collections`, `info-frontend`, ` service-manual-frontend`.

PR #10437, opened 2020-06-16
Add rerun populate_missing_data_etl

beccapearce:
his adds a job to repopulate the data that has been collected by the
[content data api etl job](https://deploy.blue.production.govuk.digital/job/content_data_api_import_etl_master_process/).

The data collected by this initial etl job was giving inaccurate results
we traced this down to a [delay in results showing up in Google
analytics](https://support.google.com/analytics/answer/1070983?hl=en#:~:text=Data%20processing%20latency,for%20up%20to%20two%20days),
results can take between 24-48 hours to appear in GA, the initial etl
is run at 7am leaving only 7 hours for the data to appear in GA.

The newly added job will collect the data after 2 days leaving time for
for the data to appear correctly in GA.

We have added this new job instead of moving the current job back since
there are many references in the code of content-data to data collected
yesterday. The date ranges are also done from data received yesterday.
Changing this would be a big job, so initially we use the potentially
inaccurate data and then correct it after 2 days.

Trello card: https://trello.com/c/ODGWWnXt/2005-5-content-data-incorrect-users-who-found-this-useful-data-for-march

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -879,6 +879,7 @@ govuk_jenkins::packages::govuk_python::apt_mirror_gpg_key_fingerprint: "%{hiera(
 
 govuk_jenkins::jobs::deploy_app::graphite_host: "graphite.%{hiera('app_domain_internal')}"
 govuk_jenkins::jobs::deploy_app::graphite_port: '443'
+govuk_jenkins::jobs::content_data_api_re_run::re_run_rake_etl_master_process_cron_schedule: '0 3 * * *'
 
 govuk_jenkins::deploy_all_apps::deploy_environment: "%{hiera('govuk_jenkins::job_builder::environment')}"
 
--- a/Rakefile
+++ b/Rakefile
@@ -423,6 +423,7 @@ task :check_consistency_between_aws_and_carrenza do
     govuk_jenkins::deploy_all_apps::apps_on_nodes
     govuk_jenkins::deploy_all_apps::deploy_environment
     govuk_jenkins::jobs::content_data_api::rake_etl_master_process_cron_schedule
+    govuk_jenkins::jobs::content_data_api_re_run::re_run_rake_etl_master_process_cron_schedule
     govuk_jenkins::jobs::deploy_app::graphite_host
     govuk_jenkins::jobs::deploy_app::graphite_port
     govuk_jenkins::jobs::deploy_emergency_banner::clear_cdn_cache
--- a/hieradata_aws/class/integration/jenkins.yaml
+++ b/hieradata_aws/class/integration/jenkins.yaml
@@ -24,6 +24,7 @@ govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::jobs::govuk_taxonomy_supervised_learning
   - govuk_jenkins::jobs::monitor_taxonomy_health
   - govuk_jenkins::jobs::passive_checks
+  - govuk_jenkins::jobs::content_data_api_re_run
   - govuk_jenkins::jobs::publication_delay_report
   - govuk_jenkins::jobs::publish_special_routes
   - govuk_jenkins::jobs::record_taxonomy_metrics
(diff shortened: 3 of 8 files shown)


modules/govuk_jenkins/templates/jobs/populate_missing_data_etl.yaml.erb
```diff
@@ -0,0 +1,23 @@
...
+          - project: run-rake-task
+            block: true
+            predefined-parameters: |
+              TARGET_APPLICATION=content-data-api
+              MACHINE_CLASS=backend
+              RAKE_TASK=etl:rerun_master[<% "#{(Time.now-2*24*60*60).strftime("%Y-%m-%d")}, #{(Time.now-2*24*60*60).strftime("%Y-%m-%d")}"%>]
+    wrappers:
+      - ansicolor:
+          colormap: xterm
+  <% if @rerun_rake_etl_master_process_cron_schedule %>
```
thomasleese, line 18: Do we need this if it's always set? Since it's in `common.yaml`.

modules/govuk_jenkins/manifests/jobs/populate_missing_data_etl.pp
```diff
@@ -0,0 +1,22 @@
...
+# - rake etl:rerun_master
+#
+# === Parameters:
+#
+# [*rerun_rake_etl_master_process_cron_schedule *]
+#   The cron timings for the etl:master process
+#   Default: undef
+#
+class govuk_jenkins::jobs::populate_missing_data_etl (
+  $rerun_rake_etl_master_process_cron_schedule = undef,
```
thomasleese, line 13:
```suggestion
  $rerun_rake_etl_master_process_cron_schedule,
```

Since it's always set we could make it a required argument.

beccapearce:
@thomasleese should the job be added to all the Hieradata files or just some specific ones?

thomasleese:
@beccapearce I think it can be added to all the `production`, `staging` and `integration` ones, but it only needs to be in AWS I think. It doesn't need to be added to the `vagrant` or `ci_master` ones.

hieradata_aws/common.yaml
```diff
@@ -879,6 +879,7 @@ govuk_jenkins::packages::govuk_python::apt_mirror_gpg_key_fingerprint: "%{hiera(
 
 govuk_jenkins::jobs::deploy_app::graphite_host: "graphite.%{hiera('app_domain_internal')}"
 govuk_jenkins::jobs::deploy_app::graphite_port: '443'
+govuk_jenkins::jobs::populate_missing_data_etl::rerun_rake_etl_master_process_cron_schedule: '0 3 * * *'
```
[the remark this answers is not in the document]
beccapearce: Since there is already a file for a `content_data_api` job I've renamed it `content_data_api_re_run`

hieradata/common.yaml
```diff
@@ -681,6 +681,7 @@ govuk_jenkins::jobs::smokey::signon_password: "%{hiera('smokey_signon_password')
 
 govuk_jenkins::jobs::run_rake_task::applications: *deployable_applications
 govuk_jenkins::jobs::deploy_app::applications: *deployable_applications
+govuk_jenkins::jobs::populate_missing_data_etl::rerun_rake_etl_master_process_cron_schedule: '0 3 * * *'
```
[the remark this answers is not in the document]
beccapearce: This fails the linting task if it isn't in hieradata/common.yaml 
https://ci.integration.publishing.service.gov.uk/job/govuk-puppet/job/add-rerun-etl-job/8/console

modules/govuk_jenkins/manifests/jobs/populate_missing_data_etl.pp
beccapearce: I get some test fails when trying this: 
```
`Puppet::Error:
       Must pass re_run_rake_etl_master_process_cron_schedule to Class
```

https://ci.integration.publishing.service.gov.uk/job/govuk-puppet/job/add-rerun-etl-job/11/console

Would i need to change anything else to make it work?

thomasleese: Hmm, I'm not sure exactly how the testing works in this repo, but I think you might need to add a value to https://github.com/alphagov/govuk-puppet/blob/master/spec/fixtures/hieradata/common.yaml ?

hieradata/common.yaml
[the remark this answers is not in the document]
thomasleese: I think you can add this to the `AWS_ONLY_KEYS` section of the linting check to fix that: https://github.com/alphagov/govuk-puppet/blob/0507f34ae2d92efc6309c015e87892ee65ff2b4c/Rakefile#L137

PR #10447, opened 2020-06-18
Activate Continuous Deployment

benthorner:

This adds a conditional step to the Deploy_App job, which will trigger
a downstream deploy for a standard "deploy" build of a specific release
tag of a supported app. Adding all these conditionals means we avoid
triggering invalid builds of the Deploy_App_Downstream job, which would
look alarming if someone views the job logs.

Note that merging a PR always triggers a standard "deploy" build [1],
so we should avoid triggering a downstream deployment for other deploy
tasks; most apps will automatically run migrations as part of "deploy".

[1]: https://github.com/alphagov/govuk-jenkinslib/blob/b5af39d69a14c94bcecb770f7bba40b0ec672af7/vars/govuk.groovy#L134

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
@@ -42,6 +42,30 @@
 
             git clone ${APP_DEPLOYMENT_GIT_URL} --branch master --single-branch --depth 1 ./
             ./jenkins.sh
+        - conditional-step:
+            condition-kind: and
+            condition-operands:
+              # Only trigger downstream deploy for "deploy" tasks
+              - condition-kind: strings-match
+                condition-string1: "$DEPLOY_TASK"
+                condition-string2: "deploy"
+              # Only trigger downstream deploy for "release_123" tags
+              - condition-kind: regex-match
+                regex: "release_.*"
+                label: "$TAG"
+              # Only trigger downstream deploy for support apps
+              - condition-kind: or
+                condition-operands:
+                  <% @deploy_downstream_applications.keys.each do |app| %>
+                  - condition-kind: strings-match
+                    condition-string1: "$TARGET_APPLICATION"
+                    condition-string2: "<%= app %>"
+                  <% end %>
+                  - condition-kind: never # until we have multiple apps
+            steps:
+              - trigger-builds:
+                  - project: Deploy_App_Downstream
+                    current-parameters: true
     publishers:
         - trigger:
             project: Smokey
--- a/lib/puppet-lint/plugins/check_hiera.rb
+++ b/lib/puppet-lint/plugins/check_hiera.rb
@@ -58,6 +58,10 @@ PuppetLint.new_check(:hiera_explicit_lookup) do
     # nginx virtual host configuration
     'router::assets_origin::asset_manager_uploaded_assets_routes',
 
+    # Allow deploy_app.pp to share knowledge about the applications supported
+    # by deploy_app_downstream. This will become redundant once all apps support Continuous Deployment.
+    'govuk_jenkins::jobs::deploy_app_downstream::applications',
+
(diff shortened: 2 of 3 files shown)


benthorner:
We will set a time for when this PR should be merged, and coordinate this with freezing deploys on Smart Answers, to avoid ambiguity about when the pipeline will start being used.

benthorner:
Tested and works 🎉

modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
```diff
@@ -42,6 +42,27 @@
...
+                    condition-string1: "$TARGET_APPLICATION"
+                    condition-string2: "<%= app %>"
+                  <% end %>
+                  - condition-kind: never
+              - condition-kind: regex-match
+                regex: "release_.*"
+                label: "$TAG"
+              - condition-kind: strings-match
+                condition-string1: "$DEPLOY_TASK"
+                condition-string2: "deploy"
```
ChrisBAshton, line 61: For readability, could you move this condition to be first, then the regex one, then finally the more complicated 'matching an application' condition? The `condition-kind: and` shortly followed by the `condition-kind: or` messed with my head and I think it would be easier to parse if they were simply reordered 😁 

Would there be a tiny efficiency benefit too, since the regex/string conditions are quicker to evaluate?

modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
```diff
@@ -42,6 +42,27 @@
 
             git clone ${APP_DEPLOYMENT_GIT_URL} --branch master --single-branch --depth 1 ./
             ./jenkins.sh
+        - conditional-step:
```
ChrisBAshton, line 45: It might be kind to future developers to put a comment above this line along the lines of

```
        # Only deploy if this is a "deploy" task, release tag and a supported app
        # https://docs.openstack.org/infra/jenkins-job-builder/builders.html#builders.conditional-step
        - conditional-step:
```

modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
```diff
@@ -42,6 +42,27 @@
...
+              - condition-kind: regex-match
+                regex: "release_.*"
+                label: "$TAG"
+              - condition-kind: strings-match
+                condition-string1: "$DEPLOY_TASK"
+                condition-string2: "deploy"
+            steps:
+              - trigger-builds:
+                  - project: Deploy_App_Downstream
+                    current-parameters: true
```
ChrisBAshton, line 65: Perhaps worth clarifying this with a comment too:

```yaml
                    current-parameters: true # pass all of this job's build parameters to the downstream build
```

I just don't find the jenkins-job-builder config options particularly self-documenting!

benthorner: This parameter is pretty standard for the plugin, so I'm not going to comment here; it's covered by the documentation.

modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
benthorner: Sure, I've added comments for each of the conditions.

modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
benthorner: Sure, I've reordered to have the simpler conditions first. I'm not sure the runtime of them is significant, though!

benthorner:
@ChrisBAshton thanks for your comments! I think I've addressed them all - could you take another look? 

> are we sure we have the Conditional BuildStep plugin installed?

We do - this was covered by my manual testing of this PR in Integration.

PR #10449, opened 2020-06-19
Fix missing release app token in CI

benthorner:
https://trello.com/c/JwvD6oaQ/167-add-deploy-freeze-check-to-deploy-downstream-job

Depends on: https://github.com/alphagov/govuk-secrets/pull/998

Previously we added an HTTP API call to the production release app
to abort downstream deploys if they are meant to be frozen. These
jobs are failing in CI because the environment variable used to get
the bearer token is missing. It's unclear where it comes from.

This ensures the variable is explicitly passed to the job, and moves
it and the other GitHub secret into a wrapper, so they are not exposed
in the console output of the job.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
@@ -13,7 +13,7 @@
           # Check for deploy freeze set in Release App
           APPLICATION_METADATA=$(curl -s \
             -H "Accept: application/json" \
-            -H "Authorization: Bearer $RELEASE_APP_NOTIFICATION_BEARER_TOKEN" \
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
```diff
@@ -69,6 +69,16 @@
     wrappers:
         - ansicolor:
             colormap: xterm
+        - inject-passwords:
+            global: false
+            mask-password-params: true
+            job-passwords:
+              - name: GITHUB_API_TOKEN
+                password:
+                  '<%= @github_token %>'
```
kevindew, line 78: should this be `@github_api_token` ?

benthorner: Oops - thanks for spotting! Updated 👍

PR #10451, opened 2020-06-22
Fix metric namespaces for Email Alert API

benthorner:
Previously we updated the metric collection worker to use GovukStatsd
defaults to export metrics [1]. Unfortunately this caused an unintentional
change in the namespace of the metrics exported. The new per-machine
namespace is actually more consistent with every other app, so this
changes the alerts to use the new namespace.

Previously the machine-agnostic metrics were automatically averaged into
a single gauge. This replicates the previous averaging by using an
explicit Graphite function.

[1]: https://github.com/alphagov/email-alert-api/pull/1293/files#diff-045ba4e36dc7c541de20b27f5d9702b0L48


benthorner:
- [ ] deploy this to Integration and check each alert is associated with a live metric
- [ ] plan follow-up work to report when these metrics are dead, so we know in future

PR #10462, opened 2020-06-23
Rough sketch of solr6 installation

risicle:
Ceci n'est pas un pull request.

### DIFF
--- a/modules/govuk_solr6/manifests/install.pp
+++ b/modules/govuk_solr6/manifests/install.pp
@@ -1,0 +1,54 @@
+# == Class: govuk_solr::install
+#
+# Installs the solr package
+#
+class govuk_solr6::install (
+  $version  = '6.4.2',
+  $base_url = 'https://archive.apache.org/dist/lucene/solr/',
+  $solr_install_dir = undef,
+  $solr_home = undef,
+  $solr_user = undef,
+) {
+
+  $solr_install_dir_parent = dirname($solr_install_dir)
+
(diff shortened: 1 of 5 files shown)


modules/govuk_solr6/manifests/install.pp
```diff
@@ -0,0 +1,54 @@
+# == Class: govuk_solr::install
+#
+# Installs the solr package
+#
+class govuk_solr6::install (
+  $version  = '6.4.2',
+  $base_url = 'https://archive.apache.org/dist/lucene/solr/',
```
fredericfran-gds, line 7: I'm not an expert but maybe you need to have a copy of solr on our own apt server: https://docs.publishing.service.gov.uk/manual/debian-packaging.html

risicle: If only we could use a debian package... the debian and ubuntu packages of solr are 3.x, so this script has to do all the installation manually. It's essentially a highly modified copy of the existing `govuk_solr` module, copied because i think we're going to need to keep the old solr working on e.g. production for quite a while until we're ready to migrate...

risicle: So reading further, maybe I _should_ create a debian package from this and add it to https://github.com/alphagov/packager/tree/master/fpm/recipes

risicle:
Taking a different approach with this now (see https://github.com/alphagov/packager/pull/181)

PR #10464, opened 2020-06-24
Fix the fix (indentation error)

benthorner:

Reference: https://github.com/alphagov/govuk-puppet/pull/10463


benthorner:
(Tested and actually works this time.)

PR #10465, opened 2020-06-24
Don't trigger downstream deploys in production

benthorner:

Previously we tried to use the list of downstream applications as
a proxy for this behaviour, forgetting this is set commonly for all
environments, to avoid duplication. Since it would be surprising to
override it in production, this adds an explicit flag to disable
downstream deployment triggers in production.


benthorner:
Tested in Integration **and** production this time 😆.

PR #10470, opened 2020-06-25
Limit requests for new email subscriptions

benthorner:
This applies a rate limit to the specific request that causes
    subscription confirmation emails to be sent. At the moment, we
    are seeing a high number of bot-like requests to this endpoint.
    The limit is based on the one used for /contact/* paths, after
    realising we can only specify a per-minute granularity:
    
    - 6 requests per minute (360 per hour)
    - burst of 0, since we want to have a hard limit
    
    Note that the lack of shared knowledge between the nginx machines,
    coupled with the round robin load balancing between them means the
    actual rate limit will be slightly higher, based on the number of
    cache machines we have and the load balancing behaviour. Currently,
    the worst  case limit is 6 * 9 per minute (3240 per hour).


benthorner:
Tested and works on Integration.

PR #10479, opened 2020-06-29
Updates Email Alert Metrics dashboard

1pretz1:
Updates the dashboard with the following:

* fixes link to email-alert-api machines dashboard
* adds link to Notify status page
* adds link to Sentry for email-alert-api
* Updates 'Notify Emails Sent' dashboard name to be more representative of
  what the graph is actually showing & updates the rate limit to be more realistic.

Trello:
https://trello.com/c/U4IPak1a/357-consolidate-and-improve-grafana-dashboards

### DIFF
--- a/modules/grafana/files/dashboards/email_alert_api.json
+++ b/modules/grafana/files/dashboards/email_alert_api.json
@@ -174,7 +174,7 @@
             },
             {
               "refId": "B",
-              "target": "alias(constantLine(10800), 'per minute rate limit')"
+              "target": "alias(constantLine(10500), 'per minute rate limit')"
             },
             {
               "refId": "D",
--- a/modules/grafana/files/dashboards/email_alert_api.json
+++ b/modules/grafana/files/dashboards/email_alert_api.json
@@ -39,7 +39,7 @@
       "height": 196,
       "panels": [
         {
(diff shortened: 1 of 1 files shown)


benthorner:
> adds a graph from the Email Incident Debugging dashboard (Notify Throughput)

Is this one covered by the graph at the top of the dashboard? (Perhaps we could re-title it so that's clearer.)

1pretz1:
They differ slightly in that the one at the top of the page is concerned with both successful and failed requests whereas the one I've added is only concerned with successful requests. I've update the title according which should make the intention of the graph clearer. I also like that it's more granular than the one at the top of the page.

I have a Q though - is Notify's rate limit actually 350r/s? I know our code rate limits to 350 but I'm seeing 360 among certain graphs.

1pretz1:
Ah just read through https://trello.com/c/Hb7kL8pX/354-document-govuk-rate-limiting and answered my own question - I'm going to update some things to make them more accurate.

benthorner:
Thanks @1pretz1. I think it's confusing to have two graphs about the rate limit, and we should just stick with the graph at the top, although perhaps renaming it would be helpful. The graph at the top is more accurate because our rate limiter works on a minutely basis; graphing the per second rate is misleading because it looks like we're exceeding the limit when we're not. However, summarising in fixed/adjacent minutely buckets is also misleading, because the rate limiter works on a rolling window (a moving bucket). The graph at the top summarises in 30s buckets, as a compromise between these two. 

> the one I've added is only concerned with successful requests

That's a good point. It's unclear if we're penalised for failed requests on the Notify end. Since [we poke the rate limiter before we send the email](https://github.com/alphagov/email-alert-api/blob/master/app/workers/delivery_request_worker.rb#L19), we should include success and fail, since this is closer to the behaviour of our internal rate limiting.

> Updates the dashboard with the following

The diff looks funny because the data sources appear to have been abstracted. I think this happens when you save the dashboard to a file. When updating the dashboard in Puppet, it's clearer to do "Export < View JSON".

1pretz1:
Thanks for the review @benthorner - I've followed you're recommendations and this should now be good to go.

PR #10481, opened 2020-07-01
Add unprocessed emails count check

AlanGabbianelli:
This adds a check for the number of emails present in the fact-check
inbox. Those emails should be automatically processed and deleted.
We want to be notified if the inbox contains emails because that means
they were either not processed or not deleted and we want to investigate
the reason behind it.

I set the warning level to 1 email and the critical level to 2 emails.
This is completely arbitrary and we can change it in the future.

Related PR: https://github.com/alphagov/publisher/pull/1301

Trello card: https://trello.com/c/NbgFTMR4/2022-5-alert-on-fact-check-replies-that-werent-parsed

### DIFF
--- a/modules/govuk/manifests/apps/publisher/unprocessed_emails_count_check.pp
+++ b/modules/govuk/manifests/apps/publisher/unprocessed_emails_count_check.pp
@@ -1,0 +1,25 @@
+# == Define: govuk::apps::publisher::unprocessed_emails_count_check
+#
+# Creates an Icinga check which checks Graphite for the number of unprocessed
+# fact-check emails.
+#
+# === Parameters
+#
+# [*ensure*]
+#   Whether to enable the check.
+#   Default: present
+#
+class govuk::apps::publisher::unprocessed_emails_count_check(
+  $ensure = present,
+) {
+  @@icinga::check::graphite { 'check_publisher_unprocessed_fact_check_emails':
+    ensure    => $ensure,
+    host_name => $::fqdn,
+    target    => 'transformNull(averageSeries(keepLastValue(stats.gauges.govuk.app.publisher.*.unprocessed_emails.count)))',
+    warning   => '1',
+    critical  => '2',
+    from      => '1hour',
+    desc      => 'publisher - unprocessed fact-check emails are present',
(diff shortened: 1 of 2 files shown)


modules/monitoring/manifests/checks.pp
```diff
@@ -45,6 +45,10 @@
   include monitoring::checks::grafana_dashboards
   include monitoring::checks::cdn_logs
 
+  if $::aws_migration {
+    include govuk::apps::publisher::unprocessed_emails_count_check
+  }
+
   if $::aws_migration {
     include govuk::apps::email_alert_api::checks
   }
```
thomasleese, line 54: I think we can combine these two into one if statement:

```puppet
if $::aws_migration {
  include govuk::apps::publisher::unprocessed_emails_count_check
  include govuk::apps::email_alert_api::checks
}
```

modules/govuk/manifests/apps/publisher/unprocessed_emails_count_check.pp
```diff
@@ -0,0 +1,25 @@
...
+) {
+  @@icinga::check::graphite { 'check_publisher_unprocessed_fact_check_emails':
+    ensure    => $ensure,
+    host_name => $::fqdn,
+    target    => 'transformNull(averageSeries(keepLastValue(stats.gauges.govuk.app.publisher.*.unprocessed_emails.count)))',
+    warning   => '1',
+    critical  => '2',
+    from      => '1hour',
+    desc      => 'publisher - unprocessed fact-check emails are present',
+    notes_url => monitoring_docs_url(publisher-unprocessed-fact-check-emails),
```
thomasleese, line 23: I notice this page doesn't exist yet - but I'm guessing that's the next step?

modules/monitoring/manifests/checks.pp
AlanGabbianelli: Ah, yes, of course! 😅
I missed that for some reason! 🤦

modules/govuk/manifests/apps/publisher/unprocessed_emails_count_check.pp
AlanGabbianelli: Yes, that's the last bit left to do in the card. I'll add that page shortly.

PR #10487, opened 2020-07-02
govuk_solr: reverse ::install/::config dependencies in $remove mode

risicle:
Followup to https://github.com/alphagov/govuk-puppet/pull/10485

This is to prevent puppet trying to e.g. remove the solr user before stopping the service.

Might this work?

### DIFF
--- a/modules/govuk_solr/manifests/config.pp
+++ b/modules/govuk_solr/manifests/config.pp
@@ -32,32 +32,34 @@ class govuk_solr::config (
     true  => stopped,
   }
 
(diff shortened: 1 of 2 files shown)


risicle:
I tested this and it didn't - all it did was create a dependency cycle. Will continue tinkering.

risicle:
Trying another approach, skipping the config file resources completely in `$reverse` mode. Don't really care if they stay. They might get obliterated by removal of their parent dir anyway.

risicle:
Closed because that "other approach" didn't work either. See #10489

PR #10489, opened 2020-07-03
govuk_solr: add "disable" mode, which actually just stops & disables the jetty service

risicle:
A simpler alternative to https://github.com/alphagov/govuk-puppet/pull/10487

This just stops & disables the service, leaving the installation in place. It shouldn't interfere with the new, package-installed `solr`.


risicle:
Hooray this seems to work.

PR #10490, opened 2020-07-06
add govuk_solr6: add new module to install solr 6.x from custom debian package

risicle:
https://trello.com/c/nyR51ZeI

Designed to allow control of *actual* presence via `$present` parameter. Made `$present` on ci machines 7 and 8 to begin with, which have `govuk_solr` disabled.

~Requires openjdk8, which calls for a little dance with `govuk_java` to enforce that.~

### DIFF
--- a/hieradata/class/ci_agent.yaml
+++ b/hieradata/class/ci_agent.yaml
@@ -16,6 +16,8 @@ govuk_mysql::server::innodb_flush_log_at_trx_commit: 2
 govuk_mysql::server::innodb_buffer_pool_size_proportion: 0.05
 govuk_mysql::server::query_cache_size: 0
 
+govuk_solr6::present: false
+
 lv:
   data:
     pv: '/dev/sdb1'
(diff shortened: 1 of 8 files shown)


risicle:
Nope, on machines where we set it to `$present`,

```
puppet-agent[30894]: Could not retrieve catalog from remote server: Error 400 on SERVER: Duplicate declaration: Class[Govuk_java::Set_defaults] is already declared in file /usr/share/puppet/production/current/modules/govuk_ci/manifests/agent.pp:23; cannot redeclare at /usr/share/puppet/production/current/modules/govuk_solr6/manifests/init.pp:38 on node
```

On machines where we don't have it `$present`,

```
W: Failed to fetch http://******.gov.uk/solr/dists/trusty/stable/binary-amd64/Packages  404  Not Found
```

risicle:
The issue with the duplicate declaration is because clearly jenkins wants to set this too, I can live with this setting (I think jenkins also wants openjdk8), but I need to be able to enforce java8 for machines which _don't_ already have jenkins setting this for them...

risicle:
Right the above now _works_, but only because I omitted the openjdk8 requirement entirely. On machines with jenkins this will appear to work fine, but machines without jenkins won't be guaranteed to have openjdk8 selected as the current java.

risicle:
_Now_ I've also removed the attempt at ensuring the service is stopped before removal of the package, because puppet just doesn't seem to be set up to be able to cope with the idea that a service that you're trying to ensure is stopped may already not exist.

PR #10491, opened 2020-07-06
Remove redirect for leicestershire coronavirus

kevindew:
Trello: https://trello.com/c/IQ26eiaW/240-make-content-publisher-use-changing-auth-bypass-tokens

This redirect was added because the draft link was tweeted and was used
by many people. We've not had any requests for this link since June 30th
and it no longer returns a content page (it returns a redirect to
authenticate)


kevindew:
Replaced by https://github.com/alphagov/govuk-puppet/pull/10495 - there was a bit more configuration to remove.

PR #10493, opened 2020-07-07
Move 'Configure GitHub Repos' job to production

thomasleese:
It doesn't quite make sense for this job to run in integration as it performs tasks against our production repos.

Depends on https://github.com/alphagov/govuk-secrets/pull/1005.


issyl0:
Could we move this to Concourse instead? This came up when I started looking at what Jenkins/Concourse do for us as part of Replatforming strategy.

PR #10497, opened 2020-07-07
govuk_solr: attempt #2 to add "disable" mode, which actually just stops & disables the jetty service

risicle:
Let's try that again, shall we?

https://trello.com/c/zLLNaKtv

A re-spin of #10485 which moves the definitions of `$required_packages` and `$java_home` out of the conditional block, with the hopes that it will now be visible to `jetty.erb`.


risicle:
(I'm going to try applying this semi-manually to ci-integration first and watch its actions carefully)

risicle:
This does indeed appear to have no effect on nodes that don't have `$disable` set.

PR #10499, opened 2020-07-08
Simplify checks for Smokey loop features

benthorner:
Related to: https://github.com/alphagov/govuk-developer-docs/pull/2652

Previously we deprecated the concept of "priority" in Smokey [1].
This removes the redundant checks, and rephrases them to be more
clearly associated with Smokey.

[1]: https://github.com/alphagov/smokey/pull/693


benthorner:
- [ ] Test on Integration with the corresponding branch of Smokey

modules/icinga/manifests/check_feature.pp
```diff
@@ -0,0 +1,14 @@
+# FIXME: This class needs better documentation as per https://docs.puppetlabs.com/guides/style_guide.html#puppet-doc
+define icinga::check_feature(
+  $feature,
+  $ensure = present,
+  $notes_url = undef
+) {
+  icinga::check { "smokey_loop_for_${feature}":
```
ChrisBAshton, line 7: I was following along until this last commit, where the commit message doesn't really describe what the commit is doing.

I get that `modules/icinga/manifests/check_feature.pp` was once doing one thing, then was renamed and edited to do something else, and now we've re-created `modules/icinga/manifests/check_feature.pp` to do something again. It would be good to have a paragraph in this last commit to explain how/why we've come full circle.

Thanks!

benthorner: Oh dear, you're right to be confused - I got in a tangle with the renaming. The new commit is a one-liner; could you take another look?

PR #10520, opened 2020-07-17
Extend data sync end time and shift back Email Alert API db sync

kevindew:
Trello: https://trello.com/c/koo6gJov/355-increase-visibility-of-sentry-alerts-for-email-apps-and-services

This extends the time we consider a GOV.UK data sync to be in progress until 8:00 am UTC. This is to reflect that currently the data sync can go on past 7:00 and in that time produce a lot of errors for Sentry. 

To provide further buffer room this also moves the Email Alert API data syncs forward so that they risk less intersection with the deadline. From my investigation in AWS it doesn't look like the DB sync causes a noticeable performance impact in production when it is sending emails into the evening. 

When merged I plan to amend the developer docs.

### DIFF
--- a/hieradata_aws/class/integration/db_admin.yaml
+++ b/hieradata_aws/class/integration/db_admin.yaml
@@ -34,8 +34,8 @@ govuk_env_sync::tasks:
     path: "postgresql-backend"
   "pull_email_alert_api_production_daily":
     ensure: "present"
-    hour: "3"
-    minute: "10"
+    hour: "1"
+    minute: "30"
     action: "pull"
     dbms: "postgresql"
     storagebackend: "s3"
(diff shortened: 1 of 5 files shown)


1pretz1:
Makes sense to me, just one question:

Would moving the email-alert-api push job in production to before days end (23:00) lose us some data between then and midnight? I'm not too familiar with the data sync but I'd assume this would be caught in the next days sync.

kevindew:
> Makes sense to me, just one question:
> 
> Would moving the email-alert-api push job in production to before days end (23:00) lose us some data between then and midnight? I'm not too familiar with the data sync but I'd assume this would be caught in the next days sync.

Yup it would, though we don't make any assurance that it is before or after midnight. This is [already the case with CKAN](https://github.com/alphagov/govuk-puppet/blob/4bcf837e0dc4f48be59716921841307dfa291ebe/hieradata_aws/class/production/db_admin.yaml#L2-L12) for instance.

PR #10521, opened 2020-07-20
Bump yarn to version 1.22.4

kevindew:
The present keyword only installs the latest version when there isn't a
version already installed. This tells puppet to install the particular
latest version.

Installing this resolves a warning Yarn is raising since Node was
upgraded to version 12.

### DIFF
--- a/modules/yarn/manifests/init.pp
+++ b/modules/yarn/manifests/init.pp
@@ -5,10 +5,10 @@
 # === Parameters:
 #
 # [*version*]
-#   Version to install. Defaults to 'present' which will install latest version
+#   Version to install. Defaults to '1.22.4-1'
 #
 class yarn(
-  $version = 'present',
+  $version = '1.22.4-1',
 ) {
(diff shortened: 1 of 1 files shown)


modules/yarn/manifests/init.pp
```diff
@@ -5,10 +5,10 @@
 # === Parameters:
 #
 # [*version*]
-#   Version to install. Defaults to 'present' which will install latest version
+#   Version to install. Defaults to '1.22.4-1'
 #
 class yarn(
-  $version = 'present',
+  $version = '1.22.4-1',
```
issyl0, line 11: I'm concerned that (like everything else), we won't remember to update this.

kevindew: Is there an alternative approach you'd recommend instead?

As it's a mirrored repo we already have to go into apt and publish a new snapshot to update Yarn. So it's already a rather manual process.

PR #10523, opened 2020-07-20
govuk_solr6: add basic handling of solr configsets and cores

risicle:
https://trello.com/c/7iYmKqEe and https://trello.com/c/lkTmfhcY

Together, these should allow us to provision solr "cores" (analogous to postgres "databases") using our puppet configuration.

Unlike postgres databases, where each database is permitted to alter its own schema to its desire, solr (at least in the old fashioned way we're forced to use it with ckan) rather wants its schema to be given to it explicitly "up front". This is typically done by naming a `configSet` at core creation time which encompasses a number of settings for the core, including the schema. As of solr 6.x, these `configSet`s can't be created through any management interface and have to be placed in solr's data dir directly. So the `govuk_solr6::configset` resource allows us to ensure this is in place by simply providing a name and a `schema.xml`. Included is `ckan28`, with the `schema.xml` from ckan 2.8.3.

The `govuk_solr6::core` resource then allows us to provision cores. This is done through solr's http interface, and uses an `exec` resource to create the new core, but first checking to see whether a core with this name already exists. It won't ever try to do anything else as we don't want to risk puppet screwing around with potentially important data.

### DIFF
--- a/modules/govuk_solr6/files/ckan28.schema.xml
+++ b/modules/govuk_solr6/files/ckan28.schema.xml
@@ -1,0 +1,188 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one or more
+ contributor license agreements.  See the NOTICE file distributed with
+ this work for additional information regarding copyright ownership.
+ The ASF licenses this file to You under the Apache License, Version 2.0
+ (the "License"); you may not use this file except in compliance with
+ the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+<!--
+     NB Please copy changes to this file into the multilingual schema:
+        ckanext/multilingual/solr/schema.xml
+-->
+
+<!-- We update the version when there is a backward-incompatible change to this
+schema. In this case the version should be set to the next CKAN version number.
+(x.y but not x.y.z since it needs to be a float) -->
+<schema name="ckan" version="2.8">
+
+<types>
+    <fieldType name="string" class="solr.StrField" sortMissingLast="true" omitNorms="true"/>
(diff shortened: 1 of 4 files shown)


modules/govuk_solr6/manifests/core.pp
```diff
@@ -0,0 +1,32 @@
...
+  $solr_port = 8983,
+  $solr_host = 'localhost',
+) {
+  exec { "test \"$(curl -sL -w '%{http_code}' 'http://${solr_host}:${solr_port}/solr/admin/cores?action=CREATE&name=${name}&configSet=${configset}' -o /dev/null)\" = 200":
+    path     => '/usr/bin:/bin',
+    provider => shell,
+    unless   => "test \"$(curl -sL -w '%{http_code}' 'http://${solr_host}:${solr_port}/solr/${name}/schema' -o /dev/null)\" = 200",
+    require  => [
+      Configset[$configset],
+      Service['solr'],
```
ChrisBAshton, line 29: Mixture of ' and " in some of these files.

modules/govuk_solr6/files/ckan28.schema.xml
```diff
@@ -0,0 +1,188 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
ChrisBAshton, line 1: Perhaps this file could be added in a separate commit so that it's easier to see the significance of the other file changes in this commit.

modules/govuk_solr6/manifests/core.pp
risicle: :sob: I have to! The puppet linter complains when I use a `"` that has no substitutions and can't use `'` for ones that do! :sob:

modules/govuk_solr6/files/ckan28.schema.xml
risicle: Yeah ok I might do that...

risicle:
Thanks man, I'll add more comments from here - didn't realize I had so much to say when I started writing this PR desc.

modules/govuk_solr6/manifests/core.pp
ChrisBAshton: 😢 Understood!

risicle:
There is literally nobody with more context, the most I'm expecting is someone with more puppet experience than me.

ChrisBAshton:
I wouldn't claim to be that person - perhaps @benthorner or @cbaines ?

risicle:
It's mergey time!

PR #10524, opened 2020-07-22
Stop keeping Icinga log archive forever

benthorner:
Previously we placed Icinga logs in an unintuitive place (/var/lib)
and relied on Icinga's inbuilt log rotation. Since Icinga's inbuilt
log rotation appears to keep archived logs forever, we were seeing
the monitoring machine disk filling up over time. This disables log
rotation within Icinga, and adds configuration to do it the same way
as all our other apps.

### DIFF
--- a/modules/icinga/manifests/config.pp
+++ b/modules/icinga/manifests/config.pp
@@ -57,15 +57,12 @@ class icinga::config (
   }
 
   if $::aws_migration {
-
-    file { '/var/lib/icinga/log':
-      ensure => directory,
-      owner  => 'nagios',
-      group  => 'adm',
-      mode   => '2755',
+    @logrotate::conf { 'icinga':
+      matches => '/var/log/icinga/*.log',
+      maxsize => '100M',
     }
 
-    file { '/var/lib/icinga/log/archives':
+    file { '/var/log/icinga':
       ensure => directory,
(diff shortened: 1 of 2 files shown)


benthorner:
Tested on Integration: log files are there, and `logrotate -f /etc/logrotate.conf` runs without errors.

modules/icinga/manifests/config.pp
```diff
@@ -57,15 +57,12 @@
...
 
   if $::aws_migration {
-
-    file { '/var/lib/icinga/log':
-      ensure => directory,
-      owner  => 'nagios',
-      group  => 'adm',
-      mode   => '2755',
+    @logrotate::conf { 'icinga':
+      matches => "/var/log/${title}/*.log",
```
ChrisBAshton, line 61: Where does `title` come from? Other apps seem to [name the directory explicitly](https://github.com/alphagov/govuk-puppet/blob/30b6c7aed5bcd73b95ca9ff9b1828416999de549/modules/nginx/manifests/logging.pp#L13-L14) (probably to avoid someone asking where `title` comes from!).

benthorner: Sure, it's gone 👍. I thought I knew where it came from, but now I'm not so sure.

PR #10525, opened 2020-07-22
[DO NOT MERGE YET] add disabled signon database restore in AWS Production

fredericfran-gds:
This is required for the migration of signon to AWS from Carrenza


fredericfran-gds:
not needed yet

PR #10527, opened 2020-07-22
Remove unactionable high load alert

benthorner:
This alert does not signify a definite problem: it's OK for a machine
to be under high load for a period of time, if that happens to be when
we have a lot of work to do. If high load really is a problem, it should
manifest in other, objective alerts. This removes the alert, on the
basis that this data is already available in our monitoring dashboards.

### DIFF
--- a/modules/icinga/manifests/client/checks.pp
+++ b/modules/icinga/manifests/client/checks.pp
@@ -87,13 +87,6 @@ class icinga::client::checks (
     $grafana = "grafana.${app_domain}"
   }
 
-  @@icinga::check { "check_load_${::hostname}":
-    check_command       => 'check_nrpe_1arg!check_load',
-    service_description => 'high load on',
-    host_name           => $::fqdn,
-    action_url          => "https://${grafana}/dashboard/file/machine.json?refresh=1m&orgId=1&var-hostname=${::fqdn_metrics}",
-  }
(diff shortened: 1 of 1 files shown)


benthorner:
This is currently firing on Integration: `backend (10.1.5.22) |   | high load on`.

thomasleese:
I don't feel too strongly, but I feel like this is one of those alerts that's useful to have around for informational purposes. If we're seeing, for example, high 5xx errors, being able to see that there is _also_ a high load alert, it gives the person debugging something to investigate immediately, without having to find the appropriate machine graph in Grafana.

thomasleese:
Also if we are going to remove this, it might be good to remove the `check_load` plugin as well so we're not leaving bits behind (https://github.com/alphagov/govuk-puppet/blob/master/modules/icinga/files/etc/nagios/nrpe.d/check_load.cfg).

benthorner:
@thomasleese while it's tempting to agree, the high load alert may also be a distraction: the load may be unrelated to the 5xx. This is where I think it would be helpful to:

1. Not have the alert for a while
1. Add it back when we find we miss it (with docs!)

Thanks for pointing out the redundant plugin. I've raised https://github.com/alphagov/govuk-puppet/pull/10528 to remove it.

PR #10529, opened 2020-07-22
Chart count of emails marked as spam

laurentqro:
Adds two new rows on the email alert API [product dashboard](https://grafana.blue.production.govuk.digital/dashboard/file/email_alert_api_product.json?refresh=1m&orgId=1), with a
new panel to show the count of emails marked as spam, and another to show the count of subscriptions terminated due to an email being marked as spam.

Screenshot 2020-08-05 at 16 53 38

Related to https://github.com/alphagov/email-alert-api/pull/1336.

https://trello.com/c/alIJoEDk/408-chart-emails-marked-as-spam-and-subscriptions-ended-due-to-this

### DIFF
--- a/modules/grafana/files/dashboards/email_alert_api_product.json
+++ b/modules/grafana/files/dashboards/email_alert_api_product.json
@@ -271,6 +271,200 @@
       "showTitle": false,
       "title": "Dashboard Row",
       "titleSize": "h6"
+    },
+    {
+      "collapse": false,
+      "height": 295,
+      "panels": [
+        {
+          "aliasColors": {},
+          "bars": true,
+          "dashLength": 10,
+          "dashes": false,
+          "datasource": "Elasticsearch",
+          "fill": 1,
+          "id": 4,
+          "legend": {
+            "avg": false,
+            "current": false,
+            "max": false,
+            "min": false,
+            "show": true,
+            "total": false,
+            "values": false
+          },
+          "lines": false,
+          "linewidth": 1,
+          "links": [],
+          "nullPointMode": "null",
+          "percentage": false,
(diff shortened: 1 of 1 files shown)


modules/grafana/files/dashboards/email_alert_api_product.json
```diff
@@ -271,6 +271,200 @@
...
+              "metrics": [
+                {
+                  "field": "view",
+                  "id": "1",
+                  "meta": {},
+                  "settings": {},
+                  "type": "count"
+                }
+              ],
+              "query": "application: email-alert-api AND status: [200 TO 299] AND route: spam_reports*",
```
kevindew, line 334: It looks like this would catch a situation where we had multiple endpoints under the controller of spam_reports, but this would create a rather strange bar chart as they'd be indistinguishable. I'd switch this to be precise with `spam_reports#create`.

I don't think we gain anything from the status filter either so can probably drop that without consequence.

modules/grafana/files/dashboards/email_alert_api_product.json
```diff
@@ -271,6 +271,200 @@
...
+            }
+          ],
+          "spaceLength": 10,
+          "span": 6,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "refId": "A",
+              "target": "alias(summarize(groupByNode(consolidateBy(stats_counts.govuk.app.email-alert-api.*.unsubscribed.reason.marked_as_spam, 'sum'), 1, 'sum'), '1d', 'sum', false), 'daily volume')",
```
kevindew, line 416: should the alias for this be daily unsubscribes?

PR #10534, opened 2020-07-23
Remove unactionable Sidekiq queue size checks

benthorner:
https://trello.com/c/imXE6RlT/936-stop-alerting-for-email-queue-size

Related to: https://github.com/alphagov/govuk-developer-docs/pull/2672

Previously we would alert on the size of the Sidekiq queues, which
was a unique alert for Email Alert API. Since the size of the queue
simply represents the amount of work to do, it cannot be used to
detect a problem - latency is the critical metric here. This removes
all the alerts for the size of Sidekiq queues, which we can still
monitor with our existing Grafana dashboards.

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api/checks.pp
+++ b/modules/govuk/manifests/apps/email_alert_api/checks.pp
@@ -14,26 +14,18 @@ class govuk::apps::email_alert_api::checks(
       'email_generation_digest',
       'cleanup',
     ]:
-      size_warning     => '75000',
-      size_critical    => '100000',
       latency_warning  => '300', # 5 minutes
       latency_critical => '600'; # 10 minutes
 
     'delivery_immediate':
-      size_warning     => '150000',
-      size_critical    => '500000',
       latency_warning  => '1200', # 20 minutes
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/apps/email_alert_api/checks.pp
```diff
@@ -14,26 +14,18 @@
       'email_generation_digest',
```
1pretz1, line 14: Could we also increase the threshold for the `email_generation_digest` queue. We've seen it alert occasionally spiking around the 8 minute mark? I was thinking 60 / 90 minutes.

Also is the `email_generation_immediate` queue now redundant? I can't see it mentioned in the email-alert-api codebase.

benthorner: @1pretz1 I'm going to decline your request to keep the scope of this PR tight. When we do change the thresholds, there's a bit more planning/thought that needs to go into it:

- Set them so that the alert is definitely a problem.
- Explain the thresholds in the docs.
- Ensure the alert can be acted on.

> Also is the email_generation_immediate queue now redundant?

Probably - at least we won't alert on that one!

PR #10538, opened 2020-07-24
Add Dependabot config

ChrisBAshton:
See RFC 126:
https://github.com/alphagov/govuk-rfcs/blob/master/rfc-126-custom-configuration-for-dependabot.md

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: bundler
+    directory: /
+    schedule:
+      interval: daily
+    allow:
+      # Internal gems
+      - dependency-name: "govuk*"
+        dependency-type: direct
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
...
+  - package-ecosystem: bundler
+    directory: /
+    schedule:
+      interval: daily
+    allow:
+      # Internal gems
+      - dependency-name: "govuk*"
+        dependency-type: direct
+      # Framework gems
+      - dependency-name: puppet
```
thomasleese, line 12: Although it is a framework dependency, I don't see us upgrading to a newer version of Puppet any time soon and it would have to be a conscious effort rather than just merging in a Dependabot PR. I reckon to avoid the PRs, it's probably better to remove this from the list.

PR #10542, opened 2020-07-24
Fix false alerts about uninstalled packages

benthorner:
https://trello.com/c/5JpkmLwB/940-fix-not-alerting-about-outstanding-security-updates

Please see this commits for more details.


benthorner:
Thanks @ChrisBAshton! I've fixed that, and added another commit to a dev doc to try and help future people. Could you take another look at this and that?

PR #10543, opened 2020-07-24
Be more chill with machine reboot alerts

benthorner:
https://trello.com/c/XAkKo8Vi/943-reduce-reboot-alert-frequency-to-at-most-once-a-week

Currently we get an alert every time an updated package is installed,
which could be daily. This adds a delay to reboot alerts, so that we
don't get more than one a week - like Dependabot, we should limit the
amount of effort we expend on this kind of alert.

### DIFF
--- a/modules/monitoring/manifests/checks/reboots.pp
+++ b/modules/monitoring/manifests/checks/reboots.pp
@@ -14,12 +14,14 @@ class monitoring::checks::reboots {
   }
 
   icinga::check { 'check_reboots_required':
-      check_command       => 'check_reboots_required',
-      use                 => 'govuk_normal_priority',
-      host_name           => $::fqdn,
-      check_interval      => 60,
-      service_description => 'At least 1 machine requires reboots due to apt updates',
-      notes_url           => monitoring_docs_url(rebooting-machines),
-      require             => Icinga::Check_config['check_reboots_required'],
+    check_command              => 'check_reboots_required',
+    use                        => 'govuk_normal_priority',
+    host_name                  => $::fqdn,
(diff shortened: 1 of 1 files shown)


richardTowers:
Am I right in thinking that one effect of this would be that we'd be admitting to ourselves that OS level security patches can take ~1 week to roll out?

(Of course, I appreciate we're probably not doing much better than that at the moment, but _admitting_ it is scary)

benthorner:
@richardTowers that's a good point, and _may_ be true for those updates that _require_ a reboot (many don't). It's a bit of a smell that the current alert is for "generic reboots required", so we're highly likely to ignore it and not know it's (potentially) urgent. So perhaps we should leave it as-is, and make the actual check better.

What do you think?

benthorner:
> If we're ignoring these alerts for multiple weeks anyway, and this is about reducing the number of alerts so we do action them when they fire then I'm more in favour of it

@richardTowers that's about the shape of it. We also seem to have a problem with the auto-reboot side of these alerts, as well as which machines should/n't auto-reboot (doesn't make sense to me, as it is).

We've had the current alert in production for 18 days now. No one seems to be in a rush to do anything about it. We've been hoping to get to it on 2ndline, but it's been busy.

Perhaps a helpful way to think about this: if there really were a significant security vulnerability that we needed to merge in right away, we'll be doing it manually, irrespective of the alerts.

benthorner:
Dah, unfortunately I neglected this alert when [fixing the approach to delay in others](https://github.com/alphagov/govuk-puppet/pull/10555). I've had to tweak the config to actually achieve the stated effect.

Could you take another look?

PR #10545, opened 2020-07-27
Ignore transient 5xx spikes for apps

benthorner:
https://trello.com/c/ciBPNeOg/945-stop-alerting-for-app-5xx-spikes

### DIFF
--- a/modules/nginx/manifests/config/vhost/proxy.pp
+++ b/modules/nginx/manifests/config/vhost/proxy.pp
@@ -154,9 +154,9 @@ define nginx::config::vhost::proxy(
     warning                  => $alert_5xx_warning_rate,
     critical                 => $alert_5xx_critical_rate,
     from                     => '30minutes',
-    desc                     => "${title} nginx 5xx rate too high",
+    desc                     => "${title} high nginx 5xx rate",
     host_name                => $::fqdn,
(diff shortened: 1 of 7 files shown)


modules/nginx/manifests/config/vhost/proxy.pp
```diff
@@ -149,14 +149,15 @@
...
-    from      => '3minutes',
-    desc      => "${title} nginx 5xx rate too high",
-    host_name => $::fqdn,
-    notes_url => monitoring_docs_url(nginx-5xx-rate-too-high-for-many-apps-boxes),
+    ensure                   => $ensure,
+    target                   => "transformNull(stats.${counter_basename}.http_5xx,0)",
+    warning                  => $alert_5xx_warning_rate,
+    critical                 => $alert_5xx_critical_rate,
+    from                     => '30minutes',
+    desc                     => "${title} nginx 5xx rate too high",
```
ChrisBAshton, line 157: Minor: Can we make the `desc` consistent between these two files? (Especially if they're pointing to the same monitoring docs URL).

benthorner: Sure, I've gone with the licensify one, and updated the docs to match: https://github.com/alphagov/govuk-developer-docs/pull/2681

PR #10546, opened 2020-07-27
Tweak process alerts so they detect real problems

benthorner:
https://trello.com/c/ihy7rIkl/946-stop-alerting-for-small-quantities-of-defunct-zombie-processes


benthorner:
Example of a "normal" number of processes (on monitoring.integration): 231

benthorner:
> so I'd personally opt for a warning level of say 15000

That's what I chose originally 😆. On reflection, if we get up to 15000, then that's a good indicator of an escalating issue, so worth alerting at that halfway point. I've tweaked the thresholds to that.

PR #10547, opened 2020-07-27
Switch ckan integration to govuk_solr6

risicle:
https://trello.com/c/OxoHouSW

This:
 - gives the `ckan` app the ability to work with multicore solr and provision a solr core in such cases.
 - adds the `govuk_solr6` module to `s_ckan` instance classes, though doesn't _enable_ it by default.
 - goes on to enable `govuk_solr6` and disable `govuk_solr` for the integration `ckan`.

This needs to be released hand-in-hand with an app release of the `ckanext-datagovuk` `master-2.8` branch, because the 2.7 branch hasn't been patched to work with multicore solr.

### DIFF
--- a/hieradata_aws/class/ckan.yaml
+++ b/hieradata_aws/class/ckan.yaml
@@ -5,3 +5,5 @@ govuk::apps::ckan::priority_worker_processes: '1'
 govuk::apps::ckan::bulk_worker_processes: '4'
 govuk::apps::ckan::enable_harvester_fetch: true
 govuk::apps::ckan::enable_harvester_gather: true
+
+govuk_solr6::present: false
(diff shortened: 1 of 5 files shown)


risicle:
Note: if this PR gets applied then **reverted**, the daemons might need a manual kicking because the new `solr` _package_ won't get _un_-installed and will still be running as a daemon occupying the port the old solr (run as the `jetty` service`) wants to use. Stateful management is great.

risicle:
FWIW I've tested this (on integration) with the "`govuk_solr` disabled/`govuk_solr6` enabled" switch _not_ flipped too, which should be the state that gets pushed out to staging and production following the merge of this. It safely continues to work the old way.

PR #10550, opened 2020-07-28
Remove unactionable 429 alerts

benthorner:
https://trello.com/c/0KX2kT9o/950-remove-unactionable-429-alerts

Rate limiting is an intentional feature of GOV.UK. We do not need
to alert when the feature is being used, which is all the docs say.

This was appearing on Integration.


benthorner:
(Interestingly, this was originally [an urgent priority alert](https://github.com/alphagov/govuk-puppet/pull/6317/files), but there's no explanation as to why, and we've since downgraded it.)

PR #10551, opened 2020-07-28
Caveat some auto-reboots may not work

benthorner:
https://trello.com/c/XAkKo8Vi/943-investigate-why-auto-reboots-arent-working

### DIFF
--- a/modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
+++ b/modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
@@ -66,7 +66,7 @@ else
   end
   if safe_machines.length > 0
     puts "\n"
-    puts 'These hosts should reboot automatically overnight:'
+    puts 'These hosts should reboot automatically overnight (unless they happen to be a primary MongoDB machine):'
     safe_machines.sort.each do |hostname|
       hydrated_name = hostname_with_node_name(hostname, machines_names_hash)
       puts "- #{hydrated_name}"


modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
```diff
@@ -66,7 +66,7 @@ else
   end
   if safe_machines.length > 0
     puts "\n"
-    puts 'These hosts should reboot automatically overnight:'
+    puts 'These hosts should reboot automatically overnight (unless they happen to be a primary MongoDB machine):'
```
ChrisBAshton, line 69: Could we do this in code instead?

```ruby
safe_machines = safe_machines - ['mongodb']
```

benthorner: Ah, so MongoDB machines _are_ safe, except the one that's the primary. Since the primary is determined dynamically, we can't say which one it is upfront. It does seem better to auto-reboot where we can, so I thought this was a good compromise.

thomasleese: We should probably also add something that tells people they can step down the primaries to allow them to reboot overnight. I think the wording at the moment suggests there's nothing to be done.

https://docs.publishing.service.gov.uk/manual/alerts/rebooting-machines.html#rebooting-mongodb-machines

benthorner: @thomasleese I'm not following, sorry. Which wording suggests there's nothing to be done?

thomasleese: I guess for me the phrase "unless they happen to be a primary MongoDB machine" doesn't tell the person who sees this alert _what_ to do, more than there _might_ be something to do. I can imagine the alert could continue to be ignored without a specific action to take. Whereas, there is an action they can take, which is to step down any of the primaries in the list so that they will reboot automatically over night.

benthorner: Sure - anything to get people actually fixing the alert!

How about this? https://github.com/alphagov/govuk-puppet/pull/10572

PR #10554, opened 2020-07-28
Complete Staging sync after it's actually finished

benthorner:
https://trello.com/c/EtiHksj2/952-fix-whitehall-healthcheck-alerts-on-staging-due-to-scheduled-docs

Related to: 367c6fb5d684c228ca7bcb96a8a45397d58c48a4

Previously we were seeing warnings in Staging about scheduled docs
not being published, which are fixed by simply running this job
manually. This changes the trigger time so that the job runs when
the synced data to 'complete' is more likely to be present.


benthorner:
@ChrisBAshton sure, I've extended the commit message. (I was just copying from Integration.)

PR #10557, opened 2020-07-29
Change feedback notify api key

edwardkerry:
The Feedback application sends e-mails to users who sign up to GOV.UK surveys via Notify.

We are changing the Notify service. 

As part of this we are changing the variable name for the Notify API key to be in line with other applications.

As the previous implementation had only one service across environments the email template ID was hard-coded into the Feedback application. The new implementation has a distinct service for each environment, so the email template ID is added as an environment variable.

When this change has been deployed, a subsequent PR will remove the old value of `survey_notify_service_api_key`.

[trello](https://trello.com/c/zs2mHfAM/2061-3-feedback-app-uses-its-own-notify-account)

### DIFF
--- a/modules/govuk/manifests/apps/feedback.pp
+++ b/modules/govuk/manifests/apps/feedback.pp
@@ -20,7 +20,7 @@
 # [*support_api_bearer_token*]
 #   The bearer token to allow this app to submit anonymous feedback to the Support API.
 #
-# [*survey_notify_service_api_key*]
+# [*govuk_notify_api_key*]
 #   The API key to allow this app to talk to GOV.UK Notify and send emails
 #   to people who want to sign up to take a survey
 #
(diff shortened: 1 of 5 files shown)


modules/govuk/manifests/apps/feedback.pp
```diff
@@ -99,10 +103,17 @@
...
-      value   => $survey_notify_service_api_key,
+  if $govuk_notify_api_key != undef {
+    govuk::app::envvar { "${title}-GOVUK_NOTIFY_API_KEY":
+      varname => 'GOVUK_NOTIFY_API_KEY',
+      value   => $govuk_notify_api_key,
+    }
+  }
+
+  if $govuk_notify_template_id != undef {
+    govuk::app::envvar {  "${title}-GOVUK_NOTIFY_TEMPLATE_ID":
```
thomasleese, line 114:
```suggestion
    govuk::app::envvar { "${title}-GOVUK_NOTIFY_TEMPLATE_ID":
```

edwardkerry: good spot!

thomasleese: 👀 🔍

PR #10561, opened 2020-07-29
Remove setup_actionmail_ses_config parameter

benjamineskola:
Remove the file installed by symlink_mailer_config, since it will no longer be referenced anywhere now that SES is not being used, cf. alphagov/govuk-app-deployment#403

(depends on alphagov/signon#1476 being deployed first)


modules/govuk/manifests/deploy.pp
```diff
@@ -11,19 +10,17 @@
...
-    actionmailer_enable_delivery  => $actionmailer_enable_delivery,
-    setup_actionmailer_ses_config => $setup_actionmailer_ses_config,
-    aws_ses_smtp_host             => $aws_ses_smtp_host,
-    aws_ses_smtp_username         => $aws_ses_smtp_username,
-    aws_ses_smtp_password         => $aws_ses_smtp_password,
-    require                       => Anchor['govuk::deploy::begin'],
+    actionmailer_enable_delivery => $actionmailer_enable_delivery,
+    aws_ses_smtp_host            => $aws_ses_smtp_host,
+    aws_ses_smtp_username        => $aws_ses_smtp_username,
+    aws_ses_smtp_password        => $aws_ses_smtp_password,
```
thomasleese, line 22: Can we remove these too, as I think they were only used to generate the ActionMailer configuration file?

PR #10569, opened 2020-07-31
Add reply_to email IDs for Feedback's survey Notify integration

edwardkerry:
This adds a `govuk_notify_reply_to_id` variable to Feedback that corresponds with an e-mail address in the Notify service that handles survey e-mails. This will make the reply_to address for these emails `govuksurvey@digital.cabinet-office.gov.uk`

[Trello](https://trello.com/c/zs2mHfAM/2061-3-feedback-app-uses-its-own-notify-account)

### DIFF
--- a/modules/govuk/manifests/apps/feedback.pp
+++ b/modules/govuk/manifests/apps/feedback.pp
@@ -41,9 +41,9 @@ class govuk::apps::feedback(
   $google_client_email = undef,
   $google_private_key = undef,
   $survey_notify_service_api_key = undef,
-  $govuk_notify_api_key = undef,
-  $govuk_notify_template_id = undef,
-  $govuk_notify_reply_to_id = undef,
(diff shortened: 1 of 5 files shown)


modules/govuk/manifests/apps/feedback.pp
```diff
@@ -40,6 +43,7 @@
   $survey_notify_service_api_key = undef,
   $govuk_notify_api_key = undef,
   $govuk_notify_template_id = undef,
+  $govuk_notify_reply_to_id = undef,
```
thomasleese, line 46:
```suggestion
  $govuk_notify_reply_to_id,
```

I wonder if it's better to make this a required argument (and actually perhaps same for the ones above) because then we don't need the if conditional on the environment variable.

edwardkerry: Sounds sensible 👍

PR #10570, opened 2020-07-31
integration db_admin govuk_env_sync::tasks: add pull_ckan_production task

risicle:
"disabled" so it can only be run manually. Integration ckan's database hasn't been refreshed in quite a while it seems.


risicle:
Wheeee

PR #10571, opened 2020-07-31
Remove redudant alerts

1pretz1:
We decided in [this doc](https://docs.google.com/document/d/1fTwyoSD7iWKJYrVk8-yhRSyYBRgZG1jePuW-LzNI74I/edit#heading=h.jvdcecid8r7s) that there were some alerts that didn't offer us much
value. These included the checks around a high sidekiq retry set size
and the total email send requests to Notify. Both of these are purely
informational and don't actually indicate that anything is wrong so we
shouldn't be alerting on them.

Trello:
https://trello.com/c/TbOhx0B6/438-remove-redundant-alerts


1pretz1:
Thanks for the review Kev - I've added to the commit message regarding the daily limit.

PR #10573, opened 2020-08-03
Only allow deploying CDN or updating CDN directories for applicable services

kevindew:
The GOV.UK CDN config can now update dictionaries in www, assets, performanceplatform and mirror services. To reflect this change we need to add more services to the select field for updating CDN directories. When I looked at this I realised we were in a situation where `deploy_cdn` job allowed deploying to services that don't exist in the corresponding environment, whereas `update_cdn_dictionaries` only allowed deploying to applicable services. I've now configured them both to only have applicable services.

### DIFF
--- a/hieradata/class/production/jenkins.yaml
+++ b/hieradata/class/production/jenkins.yaml
@@ -32,6 +32,15 @@ govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::jobs::validate_published_dns
 
 govuk_jenkins::jobs::deploy_cdn::enable_slack_notifications: true
+govuk_jenkins::jobs::deploy_cdn::services:
+  - apt
+  - assets
+  - mirror
+  - performanceplatform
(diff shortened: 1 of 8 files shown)


karlbaker02:
As part of this, perhaps we should consider updating the [dev docs](https://github.com/alphagov/govuk-cdn-config#deploy-dictionaries) to explicitly list the Fastly services you can deploy to (as in the previous section _Deploy Service_)?

Changes here look sensible 👍

kevindew:
> As part of this, perhaps we should consider updating the [dev docs](https://github.com/alphagov/govuk-cdn-config#deploy-dictionaries) to explicitly list the Fastly services you can deploy to (as in the previous section _Deploy Service_)?
> 
> Changes here look sensible 👍

That sounds a good idea - do you want to open a quick PR? I do worry a bit about the amount of places duplicating information but it's not like we change CDN services a lot.

PR #10580, opened 2020-08-05
Remove warning alerts 4 Messages & ContentChanges

1pretz1:
Removes the warning alerts as they're superfluous and don't actually
give us any useful information.

This is especially true now we've merged in this [retry behaviour] PR.
We only want to raise a critical when the retry logic has run and
failed.

We should deploy this at the same time (just after) as https://github.com/alphagov/email-alert-api/pull/1358.

More info in the Trello card:
https://trello.com/c/BT5x4iAU/457-update-warning-critical-alerts-for-messages-and-content-changes

[retry behaviour]: https://github.com/alphagov/email-alert-api/pull/1318

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api/checks.pp
+++ b/modules/govuk/manifests/apps/email_alert_api/checks.pp
@@ -46,10 +46,10 @@ class govuk::apps::email_alert_api::checks(
     notes_url => monitoring_docs_url(email-alert-api-incomplete-digest-runs),
   }
 
-  @@icinga::check::graphite { 'email-alert-api-critical-content-changes':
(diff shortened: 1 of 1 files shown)


1pretz1:
Thanks for the review @kevindew !

Oh aye, that PR is [here](https://github.com/alphagov/email-alert-api/pull/1358).

Good point on the naming though, you're right in that it is a bit smelly! I've renamed the metric name in email-alert-api in the PR above and made the related changes for the check in an extra commit on this PR. 

Could you give it another look?

kevindew:
Also, only just realised you went 90s and used the number 4 instead of "for" 😱

1pretz1:
Haha, you have to creative when given a char limit 😄

PR #10581, opened 2020-08-05
Split up out of hours and in hours machines

thomasleese:
Some machines need to be rebooted manually, but within that some machines can be rebooted in hours and some have to be rebooted out of hours.

At the moment the alert doesn't provide that distinction so this makes it easier to see what needs to be done.

### DIFF
--- a/modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
+++ b/modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
@@ -65,10 +65,16 @@ if safe_machines.length + unsafe_machines.length == 0
 else
   puts 'Some hosts need to be rebooted to apply updates'
 
-  if unsafe_machines.length > 0
+  if out_of_hours_unsafe_machines.length > 0
     puts "\n"
-    puts 'These hosts need to be rebooted manually:'
-    print_hostnames(unsafe_machines, machines_names_hash)
+    puts 'These hosts need to be rebooted manually out of hours:'
+    print_hostnames(out_of_hours_unsafe_machines, machines_names_hash)
+  end
+
+  if in_hours_unsafe_machines.length > 0
+    puts "\n"
+    puts 'These hosts need to be rebooted manually in hours:'
+    print_hostnames(in_hours_unsafe_machines, machines_names_hash)
   end
 
   if safe_machines.length > 0
--- a/modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
+++ b/modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
@@ -21,7 +21,13 @@ def hostname_with_node_name(hostname, machines_names_hash)
   else
     hostname
   end
+end
 
+def print_hostnames(hostnames, machines_names_hash)
+  hostnames.sort.each do |hostname|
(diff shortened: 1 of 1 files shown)


modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
```diff
@@ -21,12 +21,20 @@ def hostname_with_node_name(hostname, machines_names_hash)
...
+    puts "- #{hydrated_name}"
+  end
 end
 
 all_machines = `/usr/local/bin/govuk_node_list`.split("\n")
 
-unsafe_machines = `/usr/local/bin/govuk_node_list --puppet-class govuk_safe_to_reboot::careful,govuk_safe_to_reboot::no`.split("\n")
+out_of_hours_unsafe_machines = `/usr/local/bin/govuk_node_list --puppet-class govuk_safe_to_reboot::no`.split("\n")
+in_hours_unsafe_machines = `/usr/local/bin/govuk_node_list --puppet-class govuk_safe_to_reboot::careful`.split("\n")
+unsafe_machines = out_of_hours_unsafe_machines + in_hours_unsafe_machines
```
benthorner, line 37: Minor: this seems to be used in a repetitive "+" everywhere. We could just define "machine_to_reboot" to DRY it up everywhere.

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
```diff
@@ -21,12 +21,20 @@ def hostname_with_node_name(hostname, machines_names_hash)
...
+  hostnames.sort.each do |hostname|
+    hydrated_name = hostname_with_node_name(hostname, machines_names_hash)
+    puts "- #{hydrated_name}"
+  end
 end
 
 all_machines = `/usr/local/bin/govuk_node_list`.split("\n")
 
-unsafe_machines = `/usr/local/bin/govuk_node_list --puppet-class govuk_safe_to_reboot::careful,govuk_safe_to_reboot::no`.split("\n")
+out_of_hours_unsafe_machines = `/usr/local/bin/govuk_node_list --puppet-class govuk_safe_to_reboot::no`.split("\n")
```
benthorner, line 35: Minor: the use of "no" contradicts the fact that we're suggesting to reboot it. Appreciate this is a historical issue with the choice of word. It would be clearer if we could rename the strings to "in-hours" and "out-of-hours" or similar.

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
thomasleese: Yep, I couldn't decide how far to refactor this script because to be honest it could do with completely re-writing... but anyway, I've fixed this in 71aaae77054faff900420d19b7f079bd686a0dab.

modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required
thomasleese: I think I'll leave this for now because I'm unsure what effect that would have, but I'll make a note to see if we can get RE or Platform Health to change it, because I agree the phrasing is confusing.

PR #10587, opened 2020-08-07
Fix low rate limit due to broken GitHub Auth

benthorner:
https://trello.com/c/9FQp21SD/185-investigate-authenticated-github-requests-in-deployappdownstream

Previously we used a shell variable to set the Authorization header
for the two GitHub requests we make to verify the integrity of a
requested tag, before deploying it downstream. Unfortunately, Jenkins
redacts the API token after the command is run, so the variable
literally contained "[******]". This inlines the use of the secret
token in the commands that require it.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
@@ -42,11 +42,10 @@
           esac
 
           # Check release to deploy is genuine and we're not going backwards
-          GITHUB_AUTH="-H 'Authorization: token $GITHUB_API_TOKEN'"
           GITHUB_API="https://api.github.com/repos/alphagov/$REPO"
 
-          LATEST_TAGS=$(curl $GITHUB_AUTH -s "$GITHUB_API/tags?per_page=1")
-          LATEST_MASTER=$(curl $GITHUB_AUTH -s "$GITHUB_API/commits?per_page=1")
+          LATEST_TAGS=$(curl -H "Authorization: token $GITHUB_API_TOKEN" -s "$GITHUB_API/tags?per_page=1")
+          LATEST_MASTER=$(curl -H "Authorization: token $GITHUB_API_TOKEN" -s "$GITHUB_API/commits?per_page=1")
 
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
```diff
@@ -45,8 +45,8 @@
           GITHUB_AUTH="-H 'Authorization: token $GITHUB_API_TOKEN'"
           GITHUB_API="https://api.github.com/repos/alphagov/$REPO"
 
-          LATEST_TAGS=$(curl $GITHUB_AUTH -s "$GITHUB_API/tags?per_page=1")
-          LATEST_MASTER=$(curl $GITHUB_AUTH -s "$GITHUB_API/commits?per_page=1")
+          LATEST_TAGS=$(curl -H "Authorization: token $GITHUB_API_TOKEN" -s "$GITHUB_API/tags?per_page=1")
+          LATEST_MASTER=$(curl "Authorization: token $GITHUB_API_TOKEN" -s "$GITHUB_API/commits?per_page=1")
```
issyl0, line 49:
```suggestion
          LATEST_MASTER=$(curl -H "Authorization: token $GITHUB_API_TOKEN" -s "$GITHUB_API/commits?per_page=1")
```

modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
```diff
@@ -45,8 +45,8 @@
           GITHUB_AUTH="-H 'Authorization: token $GITHUB_API_TOKEN'"
```
issyl0, line 45: You can delete this line too.

benthorner: Oops, I changed approach halfway through. Done.

modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
benthorner: Good spot - fixed 👍

PR #10597, opened 2020-08-11
Trial servicegroups in Icinga

benthorner:
https://trello.com/c/V1xnRofx/178-use-icinga-servicegroups-to-replace-smokey-healthcheck-scenarios

WIP


benthorner:
Spoke about this briefly with @seanrankine2. We agreed that, since this is proving particularly hard to setup, we will wait until we see another major issue that slips through, where an alert may have fired. When this happens, we'll make a point of documenting the scenario in detail, so we can plan work that will definitely prevent it recurring.

PR #10600, opened 2020-08-13
Revert "Only fail downstream deploy if action required"

benthorner:
https://trello.com/b/kmruvfMm/govuk-developer-tools

Reverts alphagov/govuk-puppet#10588

Unfortunately Jenkins continues executing later job stages 
when an earlier stage returns an exit status of "0", so this is 
not a valid strategy for gracefully aborting a job. We need to 
rethink how we can maintain the integrity checks of the CD 
pipeline, while also only alerting if action is required.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
@@ -20,7 +20,7 @@
 
           if $DEPLOY_FREEZE; then
             echo "This application is under a deploy freeze in Release app. Aborting"
-            exit 0 # Don't alert about predictable failures
+            exit 1
           fi
       <% if @smokey_pre_check %>
       - shell: |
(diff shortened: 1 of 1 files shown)


kevindew:
You could consider using the unstable status as a means to have alerts on failures and not on purposeful aborts. We use an approach for this on deploy jobs:

https://github.com/alphagov/govuk-puppet/blob/6c1c9cab61645ebacacf2725d52c62a50143856b/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb#L75-L78

These do end up confusing people but do offer a middle ground between failure and success states.

benthorner:
Ah that's interesting. Thanks for pointing it out. Unfortunately, I don't think we can use it, because the "shell" builders we have echo all the commands they run, which would trip the finder.

Happily, when checking to see if we could turn of the builder echoing commands, I found something that looks like it does what we want directly: https://github.com/alphagov/govuk-puppet/pull/10601

PR #10601, opened 2020-08-13
Only fail downstream deploys if action required

benthorner:
https://trello.com/c/Xa6d8HuJ/179-add-slack-govuk-developers-for-first-failures-of-the-deployappdownstream-job

    Previously we attempted to use a "0" exit status to silently abort
    this job when there are predictable failures. However, having the
    exit statements in separate shell steps from the final trigger means
    means the trigger (and other steps) are still executed [1]. This
    combines the shell steps together, so that we can gracefully fail
    the job (no Slack message) without triggering a deploy at the end.
    
    [1]: https://github.com/alphagov/govuk-puppet/pull/10600

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
@@ -29,7 +29,7 @@
 
           if $DEPLOY_FREEZE; then
             echo "This application is under a deploy freeze in Release app. Aborting"
-            exit 1
+            exit 0 # Don't alert about predictable failures
           fi
 
           # Workaround for our inconsistent repo vs. deployment naming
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
```diff
@@ -53,14 +53,14 @@
...
-            exit 1
+            exit 0 # Don't alert about predictable failures
           fi
 
           if [ "$LATEST_TAG_SHA" != "$LATEST_MASTER_SHA" ]; then
             echo "The TAG to deploy is supserseded, or not on master. Aborting."
-            exit 1
+            exit 0 # Don't alert about predictable failures
           fi
-      - shell: |
```
thomasleese, line 63: It might be worth splitting up the commit because there's lots of little changes in this PR, but the crucial one is this line here, which I completely missed the first time and didn't understand how this fixed the problem.

benthorner: Sure, it's a fair point - I was in a rush when I did this. I've split the change into 3 commits. Could you take another look?

PR #10617, opened 2020-08-19
ci_master in AWS integration listens on 80

fredericfran-gds:
This is because the tls is terminated at the load balancer level.


rtrinque (approved):
Ok

PR #10627, opened 2020-08-20
govuk::apps::ckan: add a maintenance_mode

risicle:
https://trello.com/c/WjOK9SKl

The fetching of the static error page is heavily cribbed from how the loadbalancer does this, and the actual redirection mechanism is depressingly complex due to how the nginx conf *outside* of govuk::apps:ckan's control is organized. Hopefully I've commented it enough to make some sense though.

### DIFF
--- a/modules/govuk/templates/ckan/nginx.conf.erb
+++ b/modules/govuk/templates/ckan/nginx.conf.erb
@@ -17,7 +17,7 @@ location = /scheduled-maintenance {
 # that this config is injected into specifies a number of `location`s itself that
 # we wouldn't otherwise be able to override.
 absolute_redirect off;
-rewrite ^/(?!(scheduled-maintenance))$ /scheduled-maintenance redirect;
+rewrite ^/(?!(scheduled-maintenance)).*$ /scheduled-maintenance redirect;
 <%- end -%>
 
 # Guard against open redirects from CKAN for requests with paths like
--- a/modules/govuk/templates/ckan/nginx.conf.erb
+++ b/modules/govuk/templates/ckan/nginx.conf.erb
@@ -1,3 +1,25 @@
+
+<%- if @maintenance_mode -%>
+error_page 503 /scheduled_maintenance.html;
+
+location = /scheduled_maintenance.html {
+  root /usr/share/nginx/www;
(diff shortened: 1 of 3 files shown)


modules/govuk/templates/ckan/nginx.conf.erb
```diff
@@ -1,3 +1,25 @@
...
+
+# doing an internal redirect from within a `server` block like this will cause
+# `error_page` to be ignored and an ugly error page to be returned. to get
+# around this, we have to do an *external* redirect to the error page (for
+# any request that *isn't* already requesting the error page, obviously).
+# we *have* to do the redirect from the `server` block because the base template
+# that this config is injected into specifies a number of `location`s itself that
+# we wouldn't otherwise be able to override.
+absolute_redirect off;
+rewrite ^/(?!(scheduled-maintenance))$ /scheduled-maintenance redirect;
```
richardTowers, line 20: Is this intended to redirect _every_ path, except `/scheduled-maintenance`?

It looks like this would only match `/` at the moment. Perhaps you want:

```suggestion
rewrite ^/(?!(scheduled-maintenance)).*$ /scheduled-maintenance redirect;
```

?

risicle: That's _exactly_ what it's doing. For the path `/scheduled-maintenance` we want to serve a 503, but for every other path, we want to redirect it to `/scheduled-maintenance`, which gets served by the above `location =`. I know, this is far from ideal but other solutions were worse.

risicle: Hmmm yes I might have to re-experiment to check this...

risicle: Yup you're absolutely right.

risicle: Though entertainingly enough adding the `.*` sends it into an infinite redirect, so I'm going to have to play with this some more...

risicle:
Descent into nginx hell complete. Might have to close this PR and do a blanket redirect to an external error page, running out of time to tinker with this nonsense.

PR #10630, opened 2020-08-21
Remove postfix

thomasleese:
Postfix is an SMTP server that allows processes on the machine to send emails and they get relayed to an external SMTP server. It was added in ff718d1 and further updated in acb7fe8 to use SES.

It seems to be using old credentials in the HMGGDS AWS account which we want to switch off. Rather than getting new credentials, I'm thinking it might be easier to remove `postfix` from our machines.

Since adding Postfix in 2013, our applications have changed quite a bit and they're now using GOV.UK Notify directly to send emails via an HTTP API. I'm not sure what else we could be sending through Postfix, perhaps some system emails, but as far as I can tell they're not being read by anybody and we don't seem to have any documentation that explains how we use Postfix.

## Alternatives

- Remove just the "Smarthost" part of Postfix. This seems to be the part which relays messages to SES, so perhaps we could just remove this part.
- Switch Postfix to use SES credentials not in HMGGDS.

### DIFF
--- a/modules/postfix/manifests/config.pp
+++ b/modules/postfix/manifests/config.pp
@@ -21,6 +21,7 @@
 #   The mail user/list name to use in the rewrite rules
 #
 class postfix::config(
+  $ensure = present,
   $smarthost,
   $smarthost_user,
   $smarthost_pass,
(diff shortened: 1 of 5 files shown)


theseanything:
Doesn't matter anymore - retiring EC2 soon anyways.

PR #10634, opened 2020-08-25
govuk::apps::ckan: add $blanket_redirect_url option

risicle:
https://trello.com/c/WjOK9SKl/

As an alternative to https://github.com/alphagov/govuk-puppet/pull/10627, we'll just redirect everything to a maintenance page hosted elsewhere.

### DIFF
--- a/modules/govuk/manifests/apps/ckan.pp
+++ b/modules/govuk/manifests/apps/ckan.pp
@@ -11,6 +11,10 @@
 # [*pycsw_port*]
 #   What port should the PyCSW companion app run on?
 #
+# [*blanket_redirect_url*]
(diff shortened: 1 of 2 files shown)


issyl0 (commented):
Out of interest, what will the URL be? The [status page](https://status.publishing.service.gov.uk), which has as section for DGU in it?

richardTowers (approved):
Judging from the conversation on slack, https://staging.data.gov.uk/ckan_maintenance (or its prod equivalent).

That's not a CKAN page, so there's no infinite redirect issue.

This seems fine to me.

risicle:
Yeah, I didn't know about the status page. Maybe we should do something about that...

PR #10635, opened 2020-08-25
staging govuk::apps::ckan: (temporarily) disable interfering processes

risicle:
https://trello.com/c/ssYoWQnK

There are a few things we want to turn off while we reindex ckan during the 2.8 migration, and we'd rather puppet wasn't always trying to turn them back on.


risicle:
I meant to make this a draft. Damn.

thomasleese (approved):

PR #10638, opened 2020-08-26
govuk::apps::ckan: add $blanket_redirect_skip_key argument

risicle:
https://trello.com/c/WjOK9SKl

Ken wants a way to access the site during maintenance, so honor the magic header `x-ckan-skip-blanket-redirect-key: ` (dat name!) if it matches a certain value. Value should be provided through govuk-secrets.

This might not be the best way of doing this, but it's probably the simplest as it doesn't involve lists of trusted IPs, `x-forwarded-for` complications, basic auth dialogs popping up for normal visitors etc...


rtrinque (approved):
Looks good

risicle:
Thanks!

PR #10643, opened 2020-08-27
production govuk::apps::ckan: set blanket_redirect_url to put in maintenance mode

risicle:
Don't want too many db modifications creeping in since last snapshot was taken.


rtrinque (approved):
Looks good

PR #10649, opened 2020-08-28
Migrate jenkins jobs from carrenza to AWS

rtrinque:
Carrenza staging is being decommissioned.


seb-szy (approved):
This is in line with a config we have in Carrenza. All good.

PR #10652, opened 2020-08-28
Enable solr 6 to run on all ci agents except 4

kentsanggds:
## What 

To allow CKAN to run tests on CI machines we need to setup solr 6 on those machines.


risicle:
I think you should probably set these on `govuk-puppet/hieradata/class/ci_agent.yaml` instead and add an exception on `ci-agent-4`.

risicle:
In fact you might not even need the exception. It's not like other things like mongodb make the effort to disable themselves for agent 4, just as long as we don't schedule jobs on it.

risicle (approved):

PR #10655, opened 2020-08-28
fix icinga monitoring about AWS ci agents alive/connected

fredericfran-gds:
In AWS, the ssl termination is at the load balancer and hence, the icinga plugin needs to use port 80 with localhost rather than 443.


rtrinque (approved):
Ok

PR #10656, opened 2020-08-28
disable backup of mongodb to s3 on ci agents

fredericfran-gds:
This is not done in carrenza and hence, should not be done in AWS


rtrinque (approved):
Ok

PR #10663, opened 2020-09-02
Increase the Icinga thread check threshold for the Content Store

cbaines:
As it has more Unicorn workers now, so there are more than 100
threads (which is the default).

### DIFF
--- a/modules/govuk/manifests/apps/content_store.pp
+++ b/modules/govuk/manifests/apps/content_store.pp
@@ -96,6 +96,7 @@ class govuk::apps::content_store(
     nagios_memory_warning       => $nagios_memory_warning,
     nagios_memory_critical      => $nagios_memory_critical,
     unicorn_worker_processes    => $unicorn_worker_processes,
+    alert_when_threads_exceed   => 155,
     create_default_nginx_config => $create_default_nginx_config,
   }
 


brucebolt (approved):

modules/govuk/manifests/apps/content_store.pp
```diff
@@ -96,6 +96,7 @@
     nagios_memory_warning       => $nagios_memory_warning,
     nagios_memory_critical      => $nagios_memory_critical,
     unicorn_worker_processes    => $unicorn_worker_processes,
+    alert_when_threads_exceed   => 155,
```
ChrisBAshton, line 99: I wonder if this can be dynamically calculated? Could we make use of the `$unicorn_worker_processes` above it?

cbaines: That could be attempted, but I trust my squinting at the data more than my ability to come up with a formula.

There are actually only 109 threads per machine at the moment, which is `(6 * $unicorn_worker_processes) + 1` (although that's just me guessing). I picked a value of 155 because that's a little more than I saw the thread count peak at during the last deploy, because during normal deployments, you have a mixture of old threads and new threads. Now the configuration of the Icinga check might avoid the check failing during deployments, but I used the larger value just to be safe.

PR #10666, opened 2020-09-03
Report HTTP errors as critical for healthchecks

benthorner:
https://trello.com/c/DyNP0Env/496-update-the-healthcheck-icinga-alert-timeout-status-to-be-critical-rather-than-unknown

Previously we missed a healthcheck alert for Email Alert API due to the app
being down, which manifested as an unknown "timeout" error. This raises the
profile of HTTP errors to critical, since these also reflect the health of
the app being checked. Please see the commits for more details.

## Before

Screenshot 2020-09-03 at 16 37 36

## After

Screenshot 2020-09-03 at 16 44 58


1pretz1 (approved):
🦜

PR #10670, opened 2020-09-04
Change bouncer traffic replay

cbaines:
Target the application load balancer in Staging, rather than the
classic load balancer, as this is more consistent with how requests
are routed in Production.


fredericfran-gds (approved):
looks good

PR #10671, opened 2020-09-04
Make AWS LB checks comprehensive

benthorner:
https://trello.com/c/U25eoIJN/497-add-a-aws-lb-check-specific-for-emailalertapi

Previously we had a small number of checks for the health of AWS
load balancers, which were incorrectly excluding some hosts, and
had no consistency in choice between internal/external/public

This replaces the current config to represent all of the services
that are load balanced. While we should see other alerts when
individual apps/machines are down, these alerts have the benefit
of also testing the network between the machines. Recently, we
were not alerted in good time when Email Alert API failed to start,
and this should go some way to addressing that, while making the
alerting less arbitrary for all other services.

Where multiple LBs exist for a service, I've used "internal" ones
for consistency, and because the "external" vs. "public" naming
is unclear, and some "-public" LBs appear to be redundant.

Where an Application LB (ALB) is checked, some of the individual
target groups also appear to be redundant. I've excluded these with
a comment to avoid someone thinking this is "correct".

Some LBs are intentionally only associated with a single instance.
In this case, we will only see a critical alert.

### DIFF
--- a/hieradata_aws/integration.yaml
+++ b/hieradata_aws/integration.yaml
@@ -351,22 +351,62 @@ monitoring::checks::aws_iam_key::region: 'eu-west-1'
 monitoring::checks::aws_iam_key::max_aws_iam_key_age: 320
 
 monitoring::checks::lb::region: 'eu-west-1'
-monitoring::checks::lb::loadbalancers:
+monitoring::checks::lb::loadbalancers: # prefer "internal"
+  blue-apt-internal:
+    healthyhosts_warning: 0
+  blue-backend-internal:
+    healthyhosts_warning: 0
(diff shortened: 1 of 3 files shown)


benthorner:
It's also worth noting that I've ignored a couple of seemingly-distinct LBs:

- govuk-prometheus-public (no idea what this is)
- govuk-feedback-public (seems redundant - no healthy instances in prod)
- govuk-search-api-public (assume this is the public version of `blue-search`)

benthorner:
Tested on Integration, with no unexpected alerts 💯

1pretz1 (approved):
Looks good, nice work! 🦡

PR #10672, opened 2020-09-04
remove carrenza production db sync tasks

fredericfran-gds:
This is because carrenza is being shutdown with migration to AWS


deanwilson (approved):
Merge away

PR #10673, opened 2020-09-07
remove carrenza production mysql backup

fredericfran-gds:
Carrenza production mysql db cluster has been shut down last week


deanwilson (approved):
Merge away

PR #10683, opened 2020-09-08
Improve alerts for when an app is down

benthorner:

This makes a few changes to alerts we see when an app is down:

- Only alert on processes if healthcheck not supported
- Show a critical processes alert instead of a warning


1pretz1 (approved):
Looks good to me, I'm glad we're raising 0 processes as critical's now instead of potentially ignored warnings!

We'll have to see if this creates any excess noise (I assume that's why they were only raised as warnings before?)

benthorner:
> I assume that's why they were only raised as warnings before?

There doesn't seem to be any explanation for it - original commit [here](https://github.com/alphagov/govuk-puppet/commit/cd199c36ee642b2b02310259da949ff33ea3a6cc). Now that I'm reading it, I have even less confidence that it's a necessary alert, noting that we 5/6 others to cover the main scenario of an app being down.

On second thoughts: let's ditch it. Closing in favour of https://github.com/alphagov/govuk-puppet/pull/10685.

PR #10684, opened 2020-09-08
make govuk_bundler in carrenza production use aws gemstash

fredericfran-gds:


sarahseewhy (approved):
I think this looks fine.

PR #10685, opened 2020-09-09
Remove redundant check for no processes

benthorner:

This check was introduced to check we were collecting process metrics
for each app [1]. We don't universally check the metrics we collect,
and we have no evidence to suggest this particular metric is a problem.
Note that we have an equivalent check for upstart, which also links
to the same documentation as this alert.

[1]: cd199c36ee642b2b02310259da949ff33ea3a6cc

### DIFF
--- a/modules/govuk/manifests/app/config.pp
+++ b/modules/govuk/manifests/app/config.pp
@@ -325,16 +325,6 @@ define govuk::app::config (
       ensure => $ensure,
       regex  => pick($collectd_process_regex, $default_collectd_process_regex),
     }
-    @@icinga::check::graphite { "check_${title_underscore}_app_process_count_${::hostname}":
-      ensure       => $ensure,
-      target       => "${::fqdn_metrics}.processes-app-${title_underscore}.ps_count.processes",
-      warning      => '@0', # WARN if there are 0 processes
-      critical     => '@-1', # Don't use the CRITICAL status for now
-      check_period => $check_period,
-      desc         => "No processes found for ${title_underscore}",
-      host_name    => $::fqdn,
-      from         => '30seconds',
-    }
     @@icinga::check::graphite { "check_${title}_app_cpu_usage${::hostname}":
       ensure         => $ensure,
       target         => "scale(sumSeries(${::fqdn_metrics}.processes-app-${title_underscore}.ps_cputime.*),0.0001)",


cbaines:
So, when I introduced this check, there were some misconfigured collectd plugins, which weren't monitoring any processes.

I think having the check is still valuable, as it should prevent getting in to that situation again in the future.

benthorner:
@cbaines I'm keen that we don't have redundant alerts, so I'd like to explore alternatives. We currently have 5-6 alerts that all fire when an app is down or sputtering, which is unhelpful noise. Some ideas:

- If this alert is meant to monitor collectd, we should rename and re-document it as such. Note that the only explanation of its purpose is currently in the original commit message.

- Is there another way we could protect against this? Since it doesn't affect the running platform, it's not ideal for it to be an alert. Could we test the configuration in Puppet, perhaps?

- What's the worst case scenario if this alert didn't exist? I imagine it will still get spotted from people looking at deployment dashboards and the like, so we'd still be motivated to fix it.

What do you think?

cbaines:
> @cbaines I'm keen that we don't have redundant alerts, so I'd like to explore alternatives. We currently have 5-6 alerts that all fire when an app is down or sputtering, which is unhelpful noise. Some ideas:
> 
>   * If this alert is meant to monitor collectd, we should rename and re-document it as such. Note that the only explanation of its purpose is currently in the original commit message.
> 
>   * Is there another way we could protect against this? Since it doesn't affect the running platform, it's not ideal for it to be an alert. Could we test the configuration in Puppet, perhaps?
> 
>   * What's the worst case scenario if this alert didn't exist? I imagine it will still get spotted from people looking at deployment dashboards and the like, so we'd still be motivated to fix it.
> 
> 
> What do you think?

I'm still unsure this is redundant, is there something else in Icinga which will spot issues with the collectd process monitoring?

I think it's fine to mention that this is looking at the collectd metrics in the Icinga check description.

I'm not sure this could be tested in Puppet, because correctly configuring collectd requires writing a regex that matches the right processes by name, and we don't specify process names for all the apps in govuk-puppet.

I've had my fun fixing the broken configuration, so if we remove the check and things break again, I don't think it matters much in the grand scheme of things. I just wanted to get everything working, and keep it working, so that we'd have a more complete set of metrics.

benthorner:
> is there something else in Icinga which will spot issues with the collectd process monitoring?

I don't think so. On balance, I'd like to remove it, but we could re-add if the config breaks again.

Would that be acceptable?

kevindew (approved):
I'm approving since it appears this discussion reached a conclusion where it was ok for this change to go in, can always we revert if we miss it.

PR #10688, opened 2020-09-09
configure gemstash for ci agents in AWS integration

fredericfran-gds:


sarahseewhy (approved):
I don't have much context on this but it looks fine.

PR #10676, opened 2020-09-07
Mark Issy's user as absent

issyl0:
- I'm leaving on Friday 11th.

https://trello.com/c/asODc0Zw/968-leaver-issy-long

PR #10689, opened 2020-09-10
Fully remove Issy's user from Puppet

issyl0:
- As per https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html.
- Once #10676 is deployed to Production.

https://trello.com/c/asODc0Zw/968-leaver-issy-long

Bye, folks! It's been fun! Keep doing excellent work! ❤️


deanwilson (approved):
Merge away and be free!

PR #10691, opened 2020-09-10
Move the user monitor Jenkins job to run and alert in AWS

issyl0:
- This was only configured to run in AWS Integration and Staging. The Production one - which we take as the source of truth - was only running in Carrenza. Since Carrenza is soon being turned off, I thought I'd save someone else some cleanup by moving it myself, because I got confused by not seeing it in AWS Production Jenkins and had to resort to running it locally manually to see if my changes worked.
- For something that ensures access to things for the right people, it's nice for this to be in the place that people look at regularly.

### DIFF
--- a/hieradata/class/staging/jenkins.yaml
+++ b/hieradata/class/staging/jenkins.yaml
@@ -13,7 +13,6 @@ govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::jobs::smokey
   - govuk_jenkins::jobs::smokey_deploy
   - govuk_jenkins::jobs::update_cdn_dictionaries
-  - govuk_jenkins::jobs::user_monitor
   - govuk_jenkins::jobs::validate_published_dns
 
 govuk_jenkins::jobs::deploy_cdn::enable_slack_notifications: true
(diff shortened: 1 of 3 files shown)


hieradata/class/staging/jenkins.yaml
```diff
@@ -13,7 +13,6 @@ govuk_jenkins::job_builder::jobs:
   - govuk_jenkins::jobs::smokey
   - govuk_jenkins::jobs::smokey_deploy
   - govuk_jenkins::jobs::update_cdn_dictionaries
-  - govuk_jenkins::jobs::user_monitor
```
sengi, line 16: It looks like you're not adding it back into staging on AWS - is that intended?

issyl0: It's [already there](https://github.com/alphagov/govuk-puppet/blob/f8b97320c9101ba25de39242f6b68aa7f95a7b7a/hieradata_aws/class/staging/jenkins.yaml#L82).

sengi: Oh, silly me 🤦  Sorry for the noise.

sengi (approved):

PR #10692, opened 2020-09-10
Fix link to the user monitor Jenkins job (now in AWS) from its alert

issyl0:
- This was using `app_domain`, but that has been [set to publishing.service.gov.uk in the Production AWS hieradata](https://github.com/alphagov/govuk-puppet/blob/05ae00cf0ae0d211c8b0c091692be3d0afff366d/hieradata_aws/production.yaml#L2) after everything was assumed migrated. However, `deploy.publishing.service.gov.uk` is the (still live) Carrenza Production Jenkins, and the user monitor job there doesn't exist.
- The `deploy.production.govuk.digital` URL needs the stackname as well because otherwise GitHub authentication callbacks don't work and you have to click the Login button twice.


sengi (approved):
LGTM. I think hardcoding the `govuk.digital` domain is the lesser of evils here. It's not hard to find should we ever decide to tackle consolidating on `publishing.service.gov.uk`.

PR #10700, opened 2020-09-14
Revert "swap name of lvm in ci_agents"

fredericfran-gds:
Reverts alphagov/govuk-puppet#10699


sarahseewhy (approved):
Out of curiosity, why did we have to revert the change?

fredericfran-gds:
I have some issue with ordering of ebs volumes, I thought this would fixed it but it didn't.

PR #10701, opened 2020-09-15
postgresql ci-agent-4

fredericfran-gds:


fredericfran-gds:
the hiera.yml or hiera_aws.yml file is only loaded on `service puppetserver restart`, not on jenkins deploy.

PR #10705, opened 2020-09-16
Correct healthcheck endpoints for 2 apps

benthorner:

This is consistent with the JSON endpoints that exist [1] [2].

[1]: https://github.com/alphagov/release/blob/master/app/controllers/application_controller.rb#L20
[2]: https://github.com/alphagov/asset-manager/blob/master/app/controllers/healthcheck_controller.rb#L5


kevindew (approved):
Looks good.

The change did make me wonder, at least in the case of release, whether this was done due to a need or just catching consistency?

benthorner:
@kevindew sorry, I forgot to specify: it was a surprise to me that so many apps don't have a JSON healthcheck. I started on auditing them, but it was taking a fair amount of effort. These are the two potential corrections I found.

PR #10712, opened 2020-09-18
archive monitoring icinga data

fredericfran-gds:
just use the carrenza icinga config that is working.
we rotate logs every day and then uses a cronjob to delete old files.


camdesgov (approved):
Seems sensible - approved.

PR #10714, opened 2020-09-21
Specify http_host for CKAN

thomasleese:
This should fix a recent incident where there's an open redirect vulnerability in CKAN core. Unfortunately we can't upgrade to the version of CKAN where this is fixed, so the alternative is to require that the `Host` header cannot be modified.

The original fix in CKAN is https://github.com/ckan/ckan/commit/0a34afecc76b9cc8030098371b09fdfcd053356a.

[Trello Card](https://trello.com/c/oRVFNc2T/2107-5-investigate-fix-open-redirect-by-host-header-vulnerability-using-nginx) &middot; [Incident Report](https://docs.google.com/document/d/1w1Gp4wOFwSd5WanVS9jzUuyHuwSD3AmTY6qfN5DC9Xc/edit#)

### DIFF
--- a/modules/govuk/manifests/app/nginx_vhost.pp
+++ b/modules/govuk/manifests/app/nginx_vhost.pp
@@ -62,6 +62,9 @@
 #   Boolean, whether to enable HTTP/1.1 for proxying from the Nginx vhost
 #   to the app server.
 #
+# [*http_host*]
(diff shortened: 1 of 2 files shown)


cbaines (approved):
Sounds good to me :+1:

PR #10720, opened 2020-09-23
Add ensure => absent to Felisia's manifest

bevanloon:
Felisia has left GDS. This is the [first step to removing Felisia from puppet](https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html)


karlbaker02 (approved):
👍

PR #10722, opened 2020-09-23
Configure Jenkins GitHub servers via puppet

kevindew:
This configuration was previously managed by the UI which meant that it
broke when CI was moved to another location. This is a configuration
file that we only want on CI master instance - as this is the only box
with the credential - therefore I configured this in govuk_ci rather
than govuk_jenkins.

The credential this relies on, github-token-govuk-ci, doesn't seem to be
managed by Puppet. However I can see that this
is already a convention for this project with similar tokens referenced
in modules/govuk_ci/templates/application_build_job.xml.erb and
modules/govuk_jenkins/templates/config/org.jenkinsci.plugins.workflow.libs.GlobalLibraries.xml


cbaines (approved):
Sounds good to me

PR #10724, opened 2020-09-23
Remove Signon Plek Override In AWS

camdesgov:
This override was added during the migration from AWS to Carrenza, to ensure
that apps hosted by different providers would still be able to communicate with
Signon.

Now that all apps are in AWS, we want them to use the internal load balancer to
connect to Signon. To do this, we must ensure they have
signon.production.govuk-internal.digital set as their Signon URL.

This is an incident action item, you can find more information about this here:
https://docs.google.com/document/d/1npjn9yQ0z_p9ie2RBikt-VpigJWoAz76pgHwNOMlZKM/edit#

### DIFF
--- a/modules/govuk/manifests/deploy/config.pp
+++ b/modules/govuk/manifests/deploy/config.pp
@@ -126,19 +126,6 @@ class govuk::deploy::config(
       'GOVUK_APP_DOMAIN_EXTERNAL':  value => $app_domain;
     }
 
-    # 1. Signon is still in Carrenza Staging and Production.
-    if ($::aws_environment == 'staging') or ($::aws_environment == 'production') {
-      # draft_content_store overrides PLEK_SERVICE_SIGNON_URI itself because it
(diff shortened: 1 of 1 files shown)


fredericfran-gds (changes requested):
I don't think this is correct. Right now all the Plek entries (except the draft-content-store ones) are overridden, which we don't want. I would suggest removing the whole If block, including the unless part, etc.

camdesgov:
> I don't think this is correct. Right now all the Plek entries (except the draft-content-store ones) are overridden, which we don't want. I would suggest removing the whole If block, including the unless part, etc.

Amended as discussed, could you take another peek?

fredericfran-gds (approved):
looks good, please test.

camdesgov:
> looks good, please test.

Arg! Tested the branch again just now in staging and loads of errors come up on the cache machines... I don't think it's just coincidence.....I'm closing this PR for now and I've updated card.

PR #10727, opened 2020-09-24
Support the transformation_sql_file for mysql in govuk_env_sync

cbaines:
So this can maybe be used to apply specific changes to Staging.

### DIFF
--- a/modules/govuk_env_sync/files/govuk_env_sync.sh
+++ b/modules/govuk_env_sync/files/govuk_env_sync.sh
@@ -42,10 +42,11 @@ set -o errtrace
 #     Path to use on storage backend, prefix in case of S3.
 #
 #   s) transformation_sql_file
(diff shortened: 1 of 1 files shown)


bevanloon (approved):
Looks sensible to me. Is it worth updating the [comment](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_env_sync/files/govuk_env_sync.sh#L46) and removing the explicit reference to Postgres?

cbaines:
> Looks sensible to me. Is it worth updating the [comment](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_env_sync/files/govuk_env_sync.sh#L46) and removing the explicit reference to Postgres?

I've updated it to include MySQ, just to be extra specific that this doesn't apply to MongoDB and Elasticsearch...

PR #10737, opened 2020-09-25
Allow cookies for coronavirus find support

gclssvglx:
Add inbound and outbound exception for the smart answers coronavirus find support flow as it uses cookie-based sessions to manage state.


theseanything (changes requested):
Could we update the commit description explaining this an exception to smart answers as it uses cookie based sessions to manage state?

reggieb (commented):
Looks good. It looks like Varnish doesn't offer a lot of options so chaining `&&` is the way to go

gclssvglx:
> Could we update the commit description explaining this an exception to smart answers as it uses cookie based sessions to manage state?

Sure thing - all done!

theseanything (approved):

PR #10739, opened 2020-09-28
Fix Puppet error on Integration

benthorner:
https://trello.com/c/sc2PLcNR/984-whos-fixing-me-ckan-upstart-not-up-in-integration

Error: /Stage[main]/Govuk_solr::Config/Service[jetty]: Could not evaluate: Could not find init script or upstart conf file for 'jetty'

Previously we still defined a 'service' resource even when the
associated init file was absent, which lead to an error. This
avoids defining the service unless its config is present, the
only downside being an attempt to remove disable jetty may leave
an old process running, which seems like a niche scenario.

### DIFF
--- a/modules/govuk_solr/manifests/config.pp
+++ b/modules/govuk_solr/manifests/config.pp
@@ -23,14 +23,14 @@ class govuk_solr::config (
   unless $disable {
     file{"${solr_home}/current/conf":
       ensure  => directory,
-    } ->
+    }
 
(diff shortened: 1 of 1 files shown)


risicle (approved):
Yeah I think I kinda realized this would be a possibility when it came to the actual removal and require some ~hack~ tweak like this.

risicle:
Ah but I don't have approval permission anymore

kentsanggds (approved):

PR #10741, opened 2020-09-29
Stop sync between Production and Staging CKAN

kentsanggds:
## What

Whilst pentesting is happening stop the DB sync between production and staging as it could wipe out the pentest users on Staging.

### DIFF
--- a/hieradata_aws/class/staging/db_admin.yaml
+++ b/hieradata_aws/class/staging/db_admin.yaml
@@ -1,6 +1,6 @@
 govuk_env_sync::tasks:
   "pull_ckan_production_daily":
-    ensure: "present"
+    ensure: "disabled"
     hour: "3"
     minute: "15"
     action: "pull"


hieradata_aws/class/staging/db_admin.yaml
```diff
@@ -1,6 +1,6 @@
 govuk_env_sync::tasks:
   "pull_ckan_production_daily":
-    ensure: "present"
+    ensure: "disabled"
```
richardTowers, line 3: Should it be "absent"? Or is "disabled" equivalent? Looking at the history, it seems we've got examples of both uses.

richardTowers: I should have grepped before asking:

https://github.com/alphagov/govuk-puppet/blob/101e060c2dc2ea62decd31f7911a8c296fbbd8a9/modules/govuk_env_sync/manifests/task.pp#L38-L40

richardTowers (approved):

kentsanggds:
I've seen both, disabled seems to be more prevalent so I've gone for that this time.

PR #10743, opened 2020-10-01
Disable configurations via the Config API in Solr

deborahchua:
We have discovered a vulnerability in the version of Solr (6) we
are using for CKAN.

Adding the new `solr.init.sh` startup script allows us to override
the default script (`/etc/default/solr.in.sh`), specifically to disallow
any changes made to configurations via the Config API. This
also means we cannot use the edit capabilities of the Config API
until further fixes are in place.

According to https://access.redhat.com/security/vulnerabilities/CVE-2017-12629
this should help mitigate the issue, but doesn't remove the XXE
in the first place which we will look into fixing next.

Trello card: https://trello.com/c/7m2n9oy1/2158-5-stop-ckan-solr-rce-vulnerability

### DIFF
--- a/modules/govuk_solr6/files/solr.init.sh
+++ b/modules/govuk_solr6/files/solr.init.sh
@@ -1,0 +1,127 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


richardTowers (approved):
This looks good to me.

Where did you get `solr.init.sh` from? Did you copy it from one of our CKAN servers?

(I'm just checking whether I need to check if it changes anything else accidentally)

kentsanggds:
That was copied from the default one on Integration, I don't think it's any different on Staging or Prod but can check

risicle (approved):
Broadly LGTM but you'll never know till you try it :+1:

deborahchua:
Sorry @richardTowers I explained that incorrectly(!) Yep the file exists in both staging and production so shouldn't raise any issues 🤞

PR #10747, opened 2020-10-05
Fix EBS mountpoint for Icinga Logs

smford:
The logs and archive should it on their own storage, this change mounts
the storage in to the correct place of /var/log/icinga.

Logs and archives are not stored within /var/lib/icinga and having it
mounted there is incorrect.


fredericfran-gds (approved):
cool

PR #10753, opened 2020-10-08
Update email alert api token for pentesting on staging

bevanloon:
Pentesters need a bearer token for email-alert-api to carry out their work.

We've create a user and associated bearer token in Signon and subsequently suspended the user and revoked the token.

These will be persisted to staging during the normal data sync, but to minimise manual intervention, we also want to unsuspend the api user and reinstate the bearer token.

Add the pentest user to the existing SQL script to unsuspend the user and also add an extra SQL clause to reinstate the bearer token.


karlbaker02 (approved):
All looks sensible 👍

PR #10754, opened 2020-10-08
Add a jenkins task to publish a single special route

leenagupte:
Trello: https://trello.com/c/EUGVdXmo/834-publish-local-lockdown-finder-content-item
Related to: https://github.com/alphagov/special-route-publisher/pull/117

# What?
Special-route-publisher only exists to be run as a jenkins task, so it doesn't need deploying after changes are merged.
This PR adds a task to allow people to publish a specific route, rather than all of them.

# Note to reviewer

This job needs to allow users to pass a base path as a parameter. I _think_ this is how to do it.


thomasleese (approved):
This makes sense to me, you can deploy the branch to integration to test it out before merging.

PR #10758, opened 2020-10-13
Update local links manager to use defined memory thresholds

karlbaker02:
This PR updates local links manager to use defined memory thresholds from `hieradata_aws/common.yaml`, in addition to updating the values for these thresholds based on what was previously in `local_links_manager.pp` as well as from looking at the live memory the app is using.

### DIFF
--- a/modules/govuk/manifests/apps/local_links_manager.pp
+++ b/modules/govuk/manifests/apps/local_links_manager.pp
@@ -98,6 +98,11 @@
 #   The external subdomain used by the app
 #   Default: undef
 #
+# [*nagios_memory_warning*]
+#   The threshold that memory must be reached to be triggered as a warning
+# [*nagios_memory_critical*]
+#   The threshold that memory must be reached to be triggered as a critical issue
+#
 class govuk::apps::local_links_manager(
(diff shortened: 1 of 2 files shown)


modules/govuk/manifests/apps/local_links_manager.pp
```diff
@@ -98,6 +98,11 @@
 #   The external subdomain used by the app
 #   Default: undef
 #
+# [*nagios_memory_warning*]
+#   The threshold that memory must be reached to be triggered as a warning
```
hannako, line 102: Nitpick alternative comments! 

When an app's memory use reaches this value, a Nagios warning alert is triggered.
When an app's memory use reaches this value, a Nagios critical alert is triggered.

hannako (approved):

modules/govuk/manifests/apps/local_links_manager.pp
karlbaker02: Thanks, I took the comments from the asset_manager.pp manifest (they're also the same in the cache_clearing_service.pp manifest)

PR #10759, opened 2020-10-13
Stop suppressing non-JSON logs

benthorner:
https://trello.com/c/F8uizllI/537-fix-kibana-not-ingesting-email-alert-service-logs

## Testing

Screenshot 2020-10-14 at 13 41 13

## Before

Screenshot 2020-10-14 at 13 39 39

## After

Screenshot 2020-10-14 at 13 39 57

### DIFF
--- a/modules/filebeat/manifests/prospector.pp
+++ b/modules/filebeat/manifests/prospector.pp
@@ -30,14 +30,14 @@ define filebeat::prospector (
   $exclude_lines         = [],
   $max_bytes             = '10485760',
   $multiline             = {},
-  $json                  = {},
+  $json                  = false,
   $tags                  = [],
   $symlinks              = false,
 ) {
 
-  validate_hash($fields, $multiline, $json)
+  validate_hash($fields, $multiline)
   validate_array($paths, $exclude_files, $include_lines, $exclude_lines, $tags)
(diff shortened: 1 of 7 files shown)


modules/filebeat/manifests/prospector.pp
```diff
@@ -30,12 +30,12 @@
...
-  $json                  = {},
+  $json                  = false,
   $tags                  = [],
   $symlinks              = false,
 ) {
 
-  validate_hash($fields, $multiline, $json)
+  validate_hash($fields, $multiline)
   validate_array($paths, $exclude_files, $include_lines, $exclude_lines, $tags)
   validate_bool($tail_files, $close_renamed, $close_removed, $close_eof, $clean_removed, $symlinks)
```
kevindew, line 40: For consistency with existing validating should $json be in this validate_bool grouping? (does seem somewhat inconsistent with most of our puppet to be validiating bools though)

kevindew (approved):
Nice one, this looks a great improvement. 👍

modules/filebeat/manifests/prospector.pp
benthorner: Oh yes, I totally missed that! May as well validate, since it's there...

benthorner: ...and it turns out I forgot to change a couple of the (truthy) hashes, so it was useful in this case!

PR #10760, opened 2020-10-13
Remove Chris

hannako:
Removing Chris Baines from Puppet as he's moved on. 
https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html


karlbaker02 (approved):
👍

PR #10770, opened 2020-10-16
Expose GOVUK_DATA_SYNC_PERIOD ENV variable for govuk_app_config

ChrisBAshton:
This is to enable https://github.com/alphagov/govuk_app_config/pull/160,
which ignores certain exceptions if they occur during the data
sync period. We want to set this via Puppet so that it is DRY and
configured in only one place.

Trello: https://trello.com/c/rAlvGlSp/2123-5-ignore-data-sync-errors-in-govukappconfig

### DIFF
--- a/modules/govuk/lib/puppet/parser/functions/data_sync_times.rb
+++ b/modules/govuk/lib/puppet/parser/functions/data_sync_times.rb
@@ -1,0 +1,38 @@
+module Puppet::Parser::Functions
+  newfunction(:data_sync_times, :type => :rvalue, :doc => <<-EOS
+Return information about the data sync times, depending on the argument passed
+    EOS
+  ) do |args|
+    start_hour = 22
+    start_minute = 0
+    finish_hour = 8
+	finish_minute = 0
+	expected_args = 1
+
+    if args.size != expected_args
+      raise ArgumentError, "data_sync_times: Wrong number of arguments " +
+      "(given #{args.size}, expected #{expected_args})"
+    end
+
+    type = args[0]
+
+    case type
+    when 'time_range'
+      "#{start_hour}:#{start_minute}-#{finish_hour}:#{finish_minute}"
+    when 'start_time'
+      "#{start_hour}:#{start_minute}"
+    when 'start_hour'
+      start_hour
+    when 'start_minute'
+      start_minute
+    when 'finish_time'
+      "#{finish_hour}:#{finish_minute}"
+    when 'finish_hour'
+      finish_hour
+    when 'finish_minute'
+      finish_minute
+    else
+      raise ArgumentError, "data_sync_times: Unknown argument #{type}"
+    end
+  end
+end
--- a/modules/govuk/manifests/deploy/config.pp
+++ b/modules/govuk/manifests/deploy/config.pp
@@ -129,7 +129,7 @@ class govuk::deploy::config(
 
   if $::aws_migration and ($::aws_environment != 'production') {
     govuk_envvar {
-      'GOVUK_DATA_SYNC_PERIOD': value => '22:00-8:00'; # This should match up with the govuk_data_sync_in_progress class
+      'GOVUK_DATA_SYNC_PERIOD': value => data_sync_times('time_range');
     }
   }
 
--- a/modules/govuk/manifests/deploy/config.pp
+++ b/modules/govuk/manifests/deploy/config.pp
@@ -127,6 +127,12 @@ class govuk::deploy::config(
     'GOVUK_CSP_REPORT_URI': value      => $csp_report_uri;
   }
 
+  if $::aws_migration and ($::aws_environment != 'production') {
+    govuk_envvar {
+      'GOVUK_DATA_SYNC_PERIOD': value => '22:00-8:00'; # This should match up with the govuk_data_sync_in_progress class
+    }
+  }
+
   if $::aws_migration {
     $app_domain_internal = hiera('app_domain_internal')
 
--- a/modules/govuk_data_sync_in_progress/manifests/init.pp
+++ b/modules/govuk_data_sync_in_progress/manifests/init.pp
@@ -17,12 +17,6 @@ define govuk_data_sync_in_progress(
 ) {
   $fact_path = '/etc/govuk/env.d/FACTER_data_sync_in_progress'
 
-  $start_hour = 22
-  $start_minute = 0
-
-  $finish_hour = 8
-  $finish_minute = 00
-
   if !defined(File[$fact_path]) {
     file { $fact_path:
       ensure  => present,
@@ -35,8 +29,8 @@ define govuk_data_sync_in_progress(
     cron { 'data_sync_started':
       command => "echo 'true' > ${fact_path}",
       user    => 'deploy',
-      hour    => $start_hour,
-      minute  => $start_minute,
+      hour    => data_sync_times('start_hour'),
+      minute  => data_sync_times('start_minute'),
     }
   }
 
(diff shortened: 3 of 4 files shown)


modules/govuk/manifests/deploy/config.pp
```diff
@@ -121,6 +121,7 @@
 
     'ERRBIT_ENVIRONMENT_NAME': value   => $errbit_environment_name;
     'SENTRY_CURRENT_ENV': value        => $errbit_environment_name;
+    'GOVUK_DATA_SYNC_PERIOD': value    => '22:00-8:00'; # This should match up with the govuk_data_sync_in_progress class
```
ChrisBAshton: Note to self - need to only set this variable in Integration and Staging.

modules/govuk/lib/puppet/parser/functions/data_sync_times.rb
```diff
@@ -0,0 +1,38 @@
...
+    EOS
+  ) do |args|
+    start_hour = 22
+    start_minute = 0
+    finish_hour = 8
+    finish_minute = 0
+
+    if args.size != 1
+      raise(ArgumentError, "data_sync_times: Wrong number of arguments " +
+      "(given #{args.size})")
```
kevindew, line 13:
```suggestion
      raise ArgumentError, 
            "data_sync_times: Wrong number of arguments " + "(given #{args.size})"
```

It seems a bit unusual to have braces on a raise in Ruby since it's treated more as a keyword than a method. I've no idea if puppet does anything special here.

modules/govuk/lib/puppet/parser/functions/data_sync_times.rb
```diff
@@ -0,0 +1,38 @@
+module Puppet::Parser::Functions
+  newfunction(:data_sync_times, :type => :rvalue, :doc => <<-EOS
+Return information about the data sync times, depending on the argument passed
+    EOS
```
kevindew, line 4: Ooh this is exotic, I've never seen a puppet function definition before. It seems a neat solution to the problem 👍

modules/govuk/lib/puppet/parser/functions/data_sync_times.rb
```diff
@@ -0,0 +1,38 @@
+module Puppet::Parser::Functions
+  newfunction(:data_sync_times, :type => :rvalue, :doc => <<-EOS
```
kevindew, line 2: I wonder if you want to name this as `format_data_sync_time` since the primary role of the function is the formatting?

modules/govuk/lib/puppet/parser/functions/data_sync_times.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    if args.size != 1
+      raise(ArgumentError, "data_sync_times: Wrong number of arguments " +
+      "(given #{args.size})")
+    end
+
+    type = args[0]
+
+    case type
+    when 'TIME_RANGE'
```
kevindew, line 19: Is there a convention for using screaming snake case here? If so I wonder if it's a bit misleading as it looks like it's a constant but it's not. I'd guess in Ruby world a symbol would be most appropriate but I guess that's not available in puppet, though we do use lowercase strings e.g. `"absent"`

kevindew (commented):
This looks neat Chris. I've a few minor comments on the function but approach seems sound.

modules/govuk/lib/puppet/parser/functions/data_sync_times.rb
ChrisBAshton: I can see how it could misleadingly look like a constant. I've now switched to lowercase.

modules/govuk/lib/puppet/parser/functions/data_sync_times.rb
ChrisBAshton: I took inspiration from [other functions we have in govuk-puppet](https://github.com/alphagov/govuk-puppet/blob/f9f16fc8756e3965af6b697175f71d6c94343179/modules/govuk/lib/puppet/parser/functions/monitoring_docs_url.rb), after initially trying (and failing) to define the start/finish times in YAML. Could have pursued YAML further but as we sometimes need the raw values and sometimes need a representation of them, seemed just as clean to put them in here. 👍

modules/govuk/lib/puppet/parser/functions/data_sync_times.rb
AlanGabbianelli: Since the function returns values and doesn't change anything, I'd probably avoid the use of a verb like "format" in its name. Calling it `format_data_sync_time` makes it sound like it's either editing something somewhere or returning edited versions of passed attributes. I think `data_sync_times` is accurate for what it does: returning appropriate data sync time values based on the argument passed.

modules/govuk/lib/puppet/parser/functions/data_sync_times.rb
```diff
@@ -0,0 +1,38 @@
...
+  ) do |args|
+    start_hour = 22
+    start_minute = 0
+    finish_hour = 8
+    finish_minute = 0
+
+    if args.size != 1
+      raise ArgumentError, "data_sync_times: Wrong number of arguments " +
+      "(given #{args.size})"
+    end
```
AlanGabbianelli, line 14: If we specify the number of given arguments, it would probably be useful to also suggest the expected number. Something like:
```suggestion
	finish_minute = 0
	expected_args = 1

    if args.size != expected_args
      raise ArgumentError, "data_sync_times: Wrong number of arguments " +
      "(given #{args.size}, expected #{expected_args})"
    end
```

(with or without parentheses, I don't mind them with `raise`)

AlanGabbianelli (approved):
Looks good, I just left a small suggestion (not a blocker!)

modules/govuk/lib/puppet/parser/functions/data_sync_times.rb
ChrisBAshton: I think I agree with Alan here 👍

modules/govuk/manifests/deploy/config.pp
```diff
@@ -127,6 +127,12 @@
     'GOVUK_CSP_REPORT_URI': value      => $csp_report_uri;
   }
 
+  if $::aws_migration and ($::aws_environment != 'production') {
+    govuk_envvar {
+      'GOVUK_DATA_SYNC_PERIOD': value => data_sync_times('time_range');
```
ChrisBAshton, line 132: So this does seem to be setting it (don't worry about the lack of leading zeros - `Time.parse` still parses them fine):

```bash
$ more /etc/govuk/env.d/GOVUK_DATA_SYNC_PERIOD
22:0-8:0
```

But if I `echo $GOVUK_DATA_SYNC_PERIOD` I get nothing. Do we know how/whether these variables get exported as traditional ENV vars @kevindew ?

kevindew: I'm not sure exactly how they work - I think the govuk_setenv command sets them:

```
kevindew@ec2-integration-blue-email_alert_api-ip-10-1-6-234:~$ govuk_setenv email-alert-api env | grep GOVUK_DATA_SYNC_PERIOD
GOVUK_DATA_SYNC_PERIOD=22:0-8:0
```

You can see the env var if you go into an app console. For example Email Alert API on integration: 

```
irb(main):002:0> ENV["GOVUK_DATA_SYNC_PERIOD"]
=> "22:0-8:0"
```

ChrisBAshton: That's a relief - thanks for confirming!

PR #10774, opened 2020-10-16
Remove sleep before running smoke tests

kevindew:
Trello: https://trello.com/c/oPNhsGl3/536-investigate-delaying-an-app-deployment-until-the-app-has-restarted-to-avoid-false-smokey-builds

This is marked as a draft until https://github.com/alphagov/govuk-app-deployment/pull/409 is merged

With unicorn reloads now being supervised [1] we no longer need to sleep
and Smokey tests can run as soon as the deploy is successful.

[1]: https://github.com/alphagov/govuk-app-deployment/pull/409


benthorner (approved):
👍

PR #10776, opened 2020-10-19
Update CI agent Solr schema

kentsanggds:
## What

Update the schema to drop the element `defaultSearchField` as this is causing a problem with Solr 6.6.2


fredericfran-gds (approved):
looks good

PR #10777, opened 2020-10-19
Add environment variables for Transition Checker to use GOV.UK Accounts

brucebolt:
The Transition Checker will be the first live experiment for GOV.UK Accounts, therefore we need to set some environment variables to allow the checker (in `finder-frontend` to communicate with the account manager).

Note the feature flag is currently switched off in all environments.

Trello card: https://trello.com/c/CyOeHwIY

### DIFF
--- a/hieradata_aws/integration.yaml
+++ b/hieradata_aws/integration.yaml
@@ -41,6 +41,7 @@ govuk::apps::email_alert_frontend::subscription_management_enabled: true
 govuk::apps::feedback::govuk_notify_reply_to_id: 'fee22233-2f28-4b0b-8b6c-4410979f2275'
 govuk::apps::feedback::govuk_notify_template_id: 'eb9ba220-7d74-4aab-975a-bdbe718f69a3'
 govuk::apps::finder_frontend::feature_flag_accounts: false
+govuk::apps::finder_frontend::plek_account_manager_uri: 'https://www.account.staging.publishing.service.gov.uk'
 govuk::apps::frontend::govuk_notify_template_id: '1fc69d3a-09a2-40f9-852b-03f6fcef5340'
 govuk::apps::govuk_crawler_worker::enabled: false
 govuk::apps::govuk_crawler_worker::disable_during_data_sync: true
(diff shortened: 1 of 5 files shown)


hieradata_aws/integration.yaml
```diff
@@ -40,6 +40,7 @@ govuk::apps::email_alert_api::delivery_request_threshold: "3000"
 govuk::apps::email_alert_frontend::subscription_management_enabled: true
 govuk::apps::feedback::govuk_notify_reply_to_id: 'fee22233-2f28-4b0b-8b6c-4410979f2275'
 govuk::apps::feedback::govuk_notify_template_id: 'eb9ba220-7d74-4aab-975a-bdbe718f69a3'
+govuk::apps::finder_frontend::feature_flag_accounts: 'disabled'
```
barrucadu, line 43: Can this be a boolean, rather than a string which is either `enabled` or `disabled`?

modules/govuk/manifests/apps/finder_frontend.pp
```diff
@@ -62,6 +82,21 @@
...
+        value   => $account_oauth_client_secret;
+    "${title}-ACCOUNT-JWT-KEY-UUID":
+        varname => 'GOVUK_ACCOUNT_JWT_KEY_UUID',
+        value   => $account_jwt_key_uuid;
+    "${title}-ACCOUNT-JWT-KEY-PEM":
+        varname => 'GOVUK_ACCOUNT_JWT_KEY_PEM',
+        value   => $account_jwt_key_pem;
+    "${title}-FEATURE-FLAG-ACCOUNTS":
+        varname => 'FEATURE_FLAG_ACCOUNTS',
+        value   => $feature_flag_accounts;
```
barrucadu, line 99: We'll also need an appropriate Plek environment variable in here: https://github.com/alphagov/finder-frontend/blob/3535ad228a10ea1a269cadac0933103b5aee2a22/app/lib/services.rb#L35

hieradata_aws/integration.yaml
huwd: Could do, but then would need to be changed here and across all the docker images too.

https://github.com/alphagov/finder-frontend/blob/master/config/application.rb#L65

At the time the thinking was that this was more descriptive, and would also make for less ambiguity about a pulling in a truthy value from ENV?

huwd (approved):

hieradata_aws/integration.yaml
barrucadu: No, I meant can this hieradata field be a bool, which sets the env var to `feature_flag_accounts ? "enabled" : undef`

barrucadu: Environment variables contain strings, there's no getting around that

modules/govuk/manifests/apps/finder_frontend.pp
brucebolt: Ah, yes, forgot about this one.  Have added it now.

hieradata_aws/integration.yaml
brucebolt: Done

barrucadu (approved):

PR #10779, opened 2020-10-19
Update solr files structure, add custom solrconfig

kentsanggds:
## What

Add a custom `solrconfig.xml` file which contains defaults for `defaultSearchField` and `defaultOperator` which are no longer supported in the `schema.xml` file.

https://lucene.apache.org/solr/guide/6_6/other-schema-elements.html#OtherSchemaElements-DefaultSearchField_QueryOperator


fredericfran-gds (approved):
not an expert but I guess without testing we won't know.

PR #10783, opened 2020-10-20
Remove variable reassignment in Finder Frontend manifest

brucebolt:
This was causing issues with the Puppet being deployed.  Have renamed the variable to something else.


fredericfran-gds (approved):
looks good, thanks for doing that.

PR #10784, opened 2020-10-20
[Merge after thorough checks] Remove Carrenza redundant hiera

fredericfran-gds:
With the decommissioning of Carrenza, there are many redundant hiera
still present. This removes the hiera mostly related to Carrenza
Production that can be removed.

We know that this hiera are not needed by running in Carrenza Production:
```
govuk_node_list --classes
api
api_lb
backup
docker_management
graphite
jenkins
jumpbox
monitoring
performance_mongo
puppetmaster
rabbitmq
```

and

```
govuk_node_list
api-1.api.publishing.service.gov.uk
api-2.api.publishing.service.gov.uk
api-lb-1.api.publishing.service.gov.uk
api-lb-2.api.publishing.service.gov.uk
backup-1.management.publishing.service.gov.uk
docker-management-1.management.publishing.service.gov.uk
graphite-1.management.publishing.service.gov.uk
jenkins-1.management.publishing.service.gov.uk
jumpbox-1.management.publishing.service.gov.uk
jumpbox-2.management.publishing.service.gov.uk
monitoring-1.management.publishing.service.gov.uk
performance-mongo-1.api.publishing.service.gov.uk
performance-mongo-2.api.publishing.service.gov.uk
performance-mongo-3.api.publishing.service.gov.uk
performance-mongo-4.api.publishing.service.gov.uk
puppetmaster-2.management.publishing.service.gov.uk
rabbitmq-1.backend.publishing.service.gov.uk
rabbitmq-2.backend.publishing.service.gov.uk
rabbitmq-3.backend.publishing.service.gov.uk
```

Any classes or nodes not in the list above are removed.

Tests
------

Tests are fixed by:
1. adding/removing node classes to be tested in AWS and Carrenza to be
tested.  E.g. we remove node classes from Carrenza as we deleted the
associated hiera and hence some tests will not pass. We add some node
classes in AWS where these classes now exist there.

2. Because ci_agent and ci_master AWS hiera are under class/integration
rather than under class directly, we had to add some default hiera so
that the tests pass.

Refs:
-----
1. #10766


modules/govuk/spec/classes/govuk_nodes_spec_optional.rb
```diff
@@ -53,16 +54,41 @@
...
+            '/var/lib/docker' => {
+              'disk' => '/dev/mapper/data-docker',
+              'govuk_lvm' => 'docker',
+              'mountoptions' => 'defaults',
+            },
+          })
+          create_resources('govuk_mount', $mount)
+        EOT
+      end
+    else
```
sengi, line 80: The copy/paste makes it quite hard to see what this code is doing at a glance. Can we factor out the part that's the same in both cases of the conditional?

The [docs](https://rspec-puppet.com/documentation/classes/#specifying-code-to-include-before) say that pre_condition can be an array of strings, so if I were writing this I would assign the common stuff (i.e. everything except `$mount = ...`) to one string, and the `$mount = ...` part (that depends on whether it's a CI node) to another string and then write something like `let(:pre_condition) { [$lvm_precondition, $mount_precondition] }`.

I'd also add a comment to say why the extra stuff is needed for the CI node classes (as it's completely opaque to the reader otherwise).

fredericfran-gds: Thanks. We could simplify this a bit rather than this quick stab at things, let me have a look.

modules/govuk/spec/classes/govuk_nodes_spec_optional.rb
```diff
@@ -53,6 +54,30 @@
 
     let(:hiera_config) { temporary_hiera_file.path }
 
+
+    # The tests for ci_agent class requires the mount variable to be set.
+    # Before migration to AWS, ci_agent hiera was accessible under hiera/class
+    # but now resides under hiera_aws/class/integration and is not accessible since
+    # the tests are using environment vagrant rather than integration.
+    if class_name != "ci_agent"
+      default_mount = {}
```
sengi, line 63: nit: Should this be a string `"{}"`? Seems surprising that the type of the variable is different between the two branches of the conditional.

fredericfran-gds: it is a hash actually. Sorry I have been experimenting. I will ping you once it is ready, maybe I should I have put it into draft.

fredericfran-gds: Fix now, if you could have another look @sengi . thanks!

sengi (approved):
👍

PR #10786, opened 2020-10-20
Switch transition checker accounts experiment on in staging

barrucadu:
We're turning this on in staging ahead of the public launch for the
ITHC.

---

[Trello card](https://trello.com/c/kRU1953p/360-switch-accounts-on-in-staging)


brucebolt (approved):
🎉

PR #10798, opened 2020-10-27
Add an accounts feature flag to collections

barrucadu:
We want to display a different header on /transition and
/transition-check when accounts are enabled.  /transition is rendered
by collections.

---

[Trello card](https://trello.com/c/vdZrgBRy/363-make-changes-to-transition-for-trial-launch)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -151,6 +151,7 @@ task :check_consistency_between_aws_and_carrenza do
     govuk::apps::ckan::redis_port
     govuk::apps::ckan::s3_aws_access_key_id
     govuk::apps::ckan::s3_aws_secret_access_key
+    govuk::apps::collections::feature_flag_accounts
     govuk::apps::collections::unicorn_worker_processes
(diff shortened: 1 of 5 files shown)


Rakefile
```diff
@@ -151,6 +151,7 @@ task :check_consistency_between_aws_and_carrenza do
     govuk::apps::ckan::redis_port
     govuk::apps::ckan::s3_aws_access_key_id
     govuk::apps::ckan::s3_aws_secret_access_key
+    govuk::apps::collections::feature_flag_accounts
```
huwd, line 154: Does / should this file still do anything? Aren't we now fully off carrenza?

brucebolt (approved):

huwd (approved):

Rakefile
barrucadu: Some Carrenza still exists (eg, deploy.publishing.service.gov.uk), hopefully it'll go away soon and huge swathes of puppet can be simplified.

PR #10799, opened 2020-10-28
Remove check for old AWS IAM keys

benthorner:
https://trello.com/c/MHb7TZix/1017-remove-icinga-alert-at-least-1-aws-iam-key-is-older-than-320-days-and-needs-rotating

Previously we added this check based on a recommendation in AWS
Trusted Advisor. We later decided that this alert has a higher
cost than the value we get from actioning it:

- The number of keys to rotate is high, and the effort required to
rotate a single key is also high.

- Historically (over a year) 2ndline has not had the time to action
the majority of the keys highlighted by this alert.

- We do not believe there is a significant security risk to justify
the effort required in having this alert.

- This alert represents only one of the recommendations in AWS
Trusted Advisor.

In general, rotating keys after 90 days still means there are 90
days where a leaked key may be exploited. We have many issues with
the way we manage secrets, and we should think about resolving them
comphrehensively, rather than with individual alerts.


richardTowers (approved):
👍  - we discussed this in the last tech debt session and agreed that our handling of secrets needs more comprehensive thought than an alert that asks engineers to rotate them manually.

We expect the replatforming team will make significant progress on this in the course of their work.

In the meantime, we agreed that this alert was not valuable. Happy for it to be removed.

PR #10800, opened 2020-10-28
Enable accounts in all environments

barrucadu:
Merge this when we're going live.

See also https://github.com/alphagov/govuk-account-manager-prototype/pull/318


brucebolt (approved):
Approving in advance on going live

PR #10806, opened 2020-10-29
Add alert to detect dead nodes in a cluster

benthorner:
https://trello.com/c/Di1wXCVM/1025-fix-not-being-able-to-reboot-rabbitmq-machines

Related to: https://github.com/alphagov/govuk-developer-docs/pull/2841

This ensures we become aware when a node exists in the RabbitMQ
cluster, but the underlying machine is gone, which can happen
when AWS automatically regenerates our VMs. While this state is
not an immediate issue, it will prevent safe reboots.

### DIFF
--- a/modules/govuk_rabbitmq/manifests/monitoring.pp
+++ b/modules/govuk_rabbitmq/manifests/monitoring.pp
@@ -12,6 +12,11 @@ class govuk_rabbitmq::monitoring (
     require => Icinga::Plugin['check_http_timeout_noncrit'],
   }
 
+  @icinga::nrpe_config { 'check_rabbitmq_dead_nodes':
+    content => template('govuk_rabbitmq/check_rabbitmq_dead_nodes.cfg.erb'),
+    require => Icinga::Plugin['check_http_timeout_noncrit'],
+  }
+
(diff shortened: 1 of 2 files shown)


benthorner:
Tested and works in Production, where we experienced this issue ✅

deborahchua (approved):
🐰

sengi:
For posterity - this is because we (the late RE GOV.UK team) never got around to fixing https://trello.com/c/BChTMR4b/227-make-rabbitmq-cluster-resilient-to-a-single-instance-going-away (it's essentially the same underlying issue, just a slightly different manifestation of it)

I expect we'll just have to live with this extra toil until we (Replatforming) switch to Amazon MQ or whatever. Fixing the underlying issue (which is that Rabbit needs a way to tell the difference between a failed node which still exists and a node whose EC2 instance has gone away and been replaced by a new one) requires a newer version of Rabbit than we currently have (and it's unlikely to be worth the pain of upgrading Rabbit for that).

hashtag sad panda 😢

benthorner:
Thanks for context @sengi! I did wonder why the cluster didn't self-reconfigure. Fixing that seemed a world away from at least knowing about the issue. It's reassuring to know this alert is definitely for a known issue 💯.

PR #10807, opened 2020-10-30
Prevent auto-reboot for monitoring machines

benthorner:
https://trello.com/c/loHkCXV8/1024-stop-monitoring-machine-from-rebooting-overnight-manual-only

Previously we risked losing alerts if the auto-reboot failed overnight,
so we wouldn't get notified of critical issues with the platform. In
addition, the auto-reboot (at midnight) coincides with the other jobs
like data sync, so we get a load of alerts the next day because the
job was unable to post its status to Icinga. This changes the reboot
settings so that they must be done manually.

### DIFF
--- a/hieradata_aws/class/monitoring.yaml
+++ b/hieradata_aws/class/monitoring.yaml
@@ -1,5 +1,8 @@
 ---
 
+govuk_safe_to_reboot::can_reboot: 'careful'
+govuk_safe_to_reboot::reason: 'We lose alerting when this machine is down'
+
 govuk_cdnlogs::log_dir: '/var/log/cdn'
 
 icinga::client::checks::disk_space_warn: 12


sengi:
This seems like it creates extra toil, which is a shame. Is it feasible to make it reboot at a time that doesn't clash with overnight periodic jobs? Or is the time of the auto-reboot hardcoded at the moment and a pain to change?

benthorner:
Yeah, it does. I also just found out the same thing happened to the mirrorer machine i.e. anything that does a scheduled job around the time of the reboot, which appears to be 0-5AM.

It doesn't look like we're managing the file with Puppet. On the other hand, I'm not sure where it's coming from. Do you think it would be reasonable for us to start managing it?

sengi:
For scheduled jobs which run on the machine which is being rebooted, these should take a lock by being run via `/usr/local/bin/with_reboot_lock` (i.e. `/usr/local/bin/with_reboot_lock foo_command` instead of running `foo_command` directly). ([some examples](https://github.com/alphagov/govuk-puppet/commit/a215ef01d81d6f1e4ae4ff912ae79fb6fc110a95)) That doesn't solve the problem for the monitoring machine though of course, but it sounds like it might be relevant to mirrorer perhaps.

I think the actual reboot is done by the Ubuntu `unattended-upgrades` package when it runs. It sounds like it's possible to configure it to schedule the reboot instead of doing it as soon as it finishes upgrading, by setting `Unattended-Upgrade::Automatic-Reboot-Time` in `/etc/apt/apt.conf.d/*` (there's a commented-out example in `/etc/apt/apt.conf.d/99auto-upgrades` which shows the syntax).

Anyway either of those sound like they'd be a non-trivial amount of faff to test and roll out (even assuming I've not barked up the wrong tree), so don't feel obliged to investigate - I just thought I'd mention what I found when I went looking out of curiosity (motivated by an SRE's healthy disdain for toil 😁)

hieradata_aws/class/monitoring.yaml
```diff
@@ -1,5 +1,8 @@
 ---
 
+govuk_safe_to_reboot::can_reboot: 'careful'
+govuk_safe_to_reboot::reason: 'We lose alerting when this machine is down'
```
sengi, line 4: Would a more specific statement fit or would this be too long? Seems like missing those job status reports is the key thing that folks didn't realise here.

```suggestion
govuk_safe_to_reboot::reason: 'We lose monitoring and can miss passive alerts while this machine is down because jobs cannot report their status'
```

sengi (approved):

hieradata_aws/class/monitoring.yaml
sengi: I think that's about the same as the original for me. It's not a big deal - it's certainly true that we lose alerting while the machine is down so let's go with that.

benthorner:
Thanks for investigating @sengi! I'm not sure the auto-reboot is done by `unattended-upgrades`, since part of its config has `Upgrade::Automatic-Reboot: "false"`. Either way, I agree it would be a fair bit of effort to find, implement and test an in-hours reboot solution, so I'm going to merge this in - at least it fixes the problem 🤞.

PR #10810, opened 2020-11-02
Reconfigure alerts for 5xx to avoid spikes

benthorner:

This PR does three, potentially controversial things:

- Increases the averaging windows for 5xx alerts to 10 minutes
- Introduces new 5xx warning alerts with less averaging on them
- Removes the double standard we had for out-of-hours 5xx alerts

All of this is done with the aim of reducing the number of false
alarms we get on 2ndline, while also ensuring we have some early
warnings of potential issues with GOV.UK. CRITICAL alerts should
be reserved for things that are *definitely broken*.

### DIFF
--- a/modules/router/manifests/nginx.pp
+++ b/modules/router/manifests/nginx.pp
@@ -138,14 +138,26 @@ class router::nginx (
     require => File['/usr/share/nginx/www'],
   }
 
-  $graphite_5xx_target = "movingMedian(transformNull(stats.${::fqdn_metrics}.nginx_logs.www-origin.http_5xx,0),\"5min\")"
+  $graphite_5xx_warning_target = "movingMedian(transformNull(stats.${::fqdn_metrics}.nginx_logs.www-origin.http_5xx,0),\"5min\")"
+  $graphite_5xx_critical_target = "movingMedian(transformNull(stats.${::fqdn_metrics}.nginx_logs.www-origin.http_5xx,0),\"10min\")"
 
-  @@icinga::check::graphite { "check_nginx_5xx_on_${::hostname}":
-    target              => $graphite_5xx_target,
+  @@icinga::check::graphite { "check_nginx_5xx_on_${::hostname}_warning":
+    target              => $graphite_5xx_target_warning,
     warning             => 0.3,
+    critical            => 9999999, # the metric for this alert is only used for warnings
+    from                => '8minutes', # average over three 5 min windows
+    desc                => '5xx rate for www-origin [in office hours]',
+    host_name           => $::fqdn,
+    notes_url           => monitoring_docs_url(high-nginx-5xx-rate),
+    notification_period => 'inoffice',
+  }
+
+  @@icinga::check::graphite { "check_nginx_5xx_on_${::hostname}_critical":
+    target              => $graphite_5xx_target_critical,
+    warning             => 0.6, # the metric for this alert is only used for criticals
     critical            => 0.6,
     use                 => 'govuk_urgent_priority',
-    from                => '8minutes',
+    from                => '13minutes', # average over three 10 min windows
     desc                => '5xx rate for www-origin [in office hours]',
     host_name           => $::fqdn,
     notes_url           => monitoring_docs_url(high-nginx-5xx-rate),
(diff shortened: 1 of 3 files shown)


modules/router/manifests/nginx.pp
```diff
@@ -138,26 +138,38 @@
     require => File['/usr/share/nginx/www'],
   }
 
-  $graphite_5xx_target = "movingMedian(transformNull(stats.${::fqdn_metrics}.nginx_logs.www-origin.http_5xx,0),\"5min\")"
+  $graphite_5xx_warning_target = "movingMedian(transformNull(stats.${::fqdn_metrics}.nginx_logs.www-origin.http_5xx,0),\"5min\")"
+  $graphite_5xx_critical_target = "movingMedian(transformNull(stats.${::fqdn_metrics}.nginx_logs.www-origin.http_5xx,0),\"10min\")"
 
-  @@icinga::check::graphite { "check_nginx_5xx_on_${::hostname}":
-    target              => $graphite_5xx_target,
+  @@icinga::check::graphite { "check_nginx_5xx_on_${::hostname}_warning":
```
bilbof, line 141: Thanks for tuning this 👍 

Just want to make sure I completely understood the reason for separating the warning and critical checks. Is the intent of having `from` be '13minutes' for critical alerts and '8minutes' for the warning, that spikes will continue to appear as warnings while criticals will only appear for longer running, less spiky problems?

It might be confusing to have a warning and a critical for the same thing occurring at the same time. Perhaps rather than calling the alert `check_nginx_5xx_on_${::hostname}_warning`, we could call it `_spiking` (or something more descriptive).

Alternatively, could the same be achieved by lowering the warning threshold instead of creating two warnings?

modules/router/manifests/nginx.pp
```diff
@@ -153,27 +150,14 @@
...
-    notes_url           => monitoring_docs_url(high-nginx-5xx-rate),
-    notification_period => 'oncall',
+    target    => "movingMedian(transformNull(stats.${::fqdn_metrics}.nginx_logs.www-origin.http_5xx,0),\"10min\")",
+    warning   => 0.9, # the metric for this alert is only used for criticals
+    critical  => 0.9,
+    use       => 'govuk_urgent_priority',
+    from      => '13minutes', # average over three 10 min windows
+    desc      => '5xx rate for www-origin',
+    host_name => $::fqdn,
+    notes_url => monitoring_docs_url(high-nginx-5xx-rate),
```
bilbof, line 160: Second commit is an excellent change 👍

bilbof (approved):
LGTM, particularly the second commit.

I have a couple of questions inline: 1. will this continue to catch issues 2. an ignorable nit about whether we can combine the warning and critical alerts.

If you're happy this will continue to catch alerts, then :shipit:

modules/router/manifests/nginx.pp
benthorner: Sort of: a measure persistent is definitely what I'm trying to achieve here. The problem is that Icinga only supports one view of a metric per alert, and the mean averaging it applies due to `from` is naturally vulnerable to spikes. So, to get different measures of persistence, we need two alerts:

- The warning alert uses a metric that has a 5 min moving median average on it.
- The critical alert uses a metric that has a 10 min moving median average on it.
- The `from` value must therefore be at least the length of the window.
- All data points in the `from` window are averaged, so e.g. 13 is a "best of 3".

The thinking behind the config is:

- The median window is our _signal_ that something might be wrong.
- The `from` value means we are watching for a pattern in the signal (i.e. more than 2 signals).

Adding three minutes isn't strictly necessary; it has the minor advantage of making the graph linked from Icinga a bit clearer, since otherwise it'll just have one datapoint on it.

> Alternatively, could the same be achieved by lowering the warning threshold instead of creating two checks?

Previously we did have this compromise: a 5 min moving median window with a single alert. This still lead to pointless alerts, since it only took three high readings to trigger. An alternative we could try, instead of splitting the alert, would be to keep the 5 min moving median, but increase the `from` time to e.g. 15 mins.

How about that?

bilbof: Thanks for explaining, that makes sense.

> Adding three minutes isn't strictly necessary; it has the minor advantage of making the graph linked from Icinga a bit clearer, since otherwise it'll just have one datapoint on it.

To fix this I think we can provide an `action_url` to the check, which would be a link to the same Grafana chart with a longer time period.

> An alternative we could try, instead of splitting the alert, would be to keep the 5 min moving median, but increase the from time to e.g. 15 mins. How about that?

That sounds OK. We'd probably want to check in Graphite to ensure that this would catch persistent issues that we've seen in the past.

benthorner:
Thanks for reviewing @bilbof - it's complex stuff, and your questions have helped avoid it getting even more complicated, without trying something simpler first. Following our discussion on https://github.com/alphagov/govuk-puppet/pull/10810#discussion_r516032919, I've decided to split this PR in two, so that we can discuss the changes separately:

- https://github.com/alphagov/govuk-puppet/pull/10817 (averaging)
- https://github.com/alphagov/govuk-puppet/pull/10818 (DRY-up 5xx alert)

Could you take a look when you have some time (non-urgent)?

PR #10811, opened 2020-11-02
Limit collection of interface stats

benthorner:
https://trello.com/c/OUzLQlrp/1029-fix-integration-graphite-filling-up

Previously we collected stats for all interfaces, which lead to a
massive increase in data stored for our CI agents, due to their
extensive use of (Docker) veth interfaces. Since we only really
need stats about the physical interfaces, this applies filtering
to the plugin config [1] to restrict the data collected.

[1]: https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_interface

### DIFF
--- a/modules/collectd/files/etc/collectd/conf.d/default.conf
+++ b/modules/collectd/files/etc/collectd/conf.d/default.conf
@@ -4,11 +4,19 @@ LoadPlugin syslog
 	LogLevel info
 </Plugin>
 
+LoadPlugin interface
+
+<Plugin Interface>
+ Interface "lo"
+ Interface "/^veth/"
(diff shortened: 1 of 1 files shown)


benthorner:
Demo of how much of a problem this is for CI agents:
```
root@ip-10-1-4-56:/opt/graphite/storage/whisper# ls ci_agent-ip-10-1-* | grep interface | wc -l
22675
```

MahmudH (approved):
I've read this [interface-plugin](https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_interface) and looks good to me 👍

PR #10814, opened 2020-11-03
Remove Ovas, Samuel and Tirath

erino:
Ovas, Samuel and Tirath no longer need access to GOV.UK


bevanloon (changes requested):
I think we need to [add `ensure   => absent,`](https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html#header) to the manifests and let puppet run before we remove the manifests completely.

erino:
Good point, I often forget how Puppet works. Updated version here: https://github.com/alphagov/govuk-puppet/pull/10816

PR #10817, opened 2020-11-03
Only alert for persistent 5xx errors

benthorner:

Previously we applied a moving median average to various 5xx metrics,
which helps to dampen spikes over a short period. Icinga also does its
own (mean) averaging with the "from" param e.g. a "from" value of 8
means we average over three 5 min windows.

However, we can still get false alarms if we have a short period of high
5xx spikes. Increasing the window size is not ideal because this further
masks the data for the purpose of a warning. A compromise is to increase
the "from" value, so that we require more persistence for 5xx errors to
trigger a warning or a critical alert.

This increases the "from" param to 15 minutes for all 5xx alerts, so
that the windowed/de-spiked metric must either:

- Be higher over a shorter period of time; or
- Be over the threshold for a longer period of time

### DIFF
--- a/modules/licensify/manifests/apps/licensify.pp
+++ b/modules/licensify/manifests/apps/licensify.pp
@@ -134,7 +134,7 @@ class licensify::apps::licensify (
     target    => "movingMedian(transformNull(stats.${counter_basename}.http_5xx,0),\"5min\")",
     warning   => 0.05,
     critical  => 0.1,
-    from      => '5minutes',
+    from      => '15minutes',
     desc      => "${vhost_name} high nginx 5xx rate",
     host_name => $::fqdn,
     notes_url => monitoring_docs_url(high-nginx-5xx-rate),
--- a/modules/monitoring/manifests/edge.pp
+++ b/modules/monitoring/manifests/edge.pp
@@ -19,7 +19,7 @@ class monitoring::edge (
       warning   => 0.05,
       critical  => 0.1,
       host_name => $::fqdn,
-      from      => '5minutes',
+      from      => '15minutes',
       notes_url => monitoring_docs_url(fastly-error-rate),
     }
   }
--- a/modules/nginx/manifests/config/vhost/proxy.pp
+++ b/modules/nginx/manifests/config/vhost/proxy.pp
@@ -153,7 +153,7 @@ define nginx::config::vhost::proxy(
     target    => "movingMedian(transformNull(stats.${counter_basename}.http_5xx,0),\"5min\")",
     warning   => $alert_5xx_warning_rate,
     critical  => $alert_5xx_critical_rate,
-    from      => '5minutes',
+    from      => '15minutes',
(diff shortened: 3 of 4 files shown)


benthorner:
I've looked over 5xx incident reports for 2019. Some of them are false alarms, which this should help to prevent (in addition to the median averaging). All the other incident reports (with graphs) show that the 5xx errors persisted for a significant period, which matches the intent of these changes: the thresholds are the same, but we wait a little longer before we start panicking about errors with the site.

bilbof (approved):
Thanks for tuning these, this is a good change 👍 

The most impactful part of this PR is the change to the Nginx checks in `modules/router/manifests/nginx.pp`.

The alert "5xx rate for www-origin" is by far the most frequent page we see (judging from [pagerduty history](https://governmentdigitalservice.pagerduty.com/incidents?status=acknowledged%2Ctriggered%2Cresolved)). Part of that is because we get an alert per nginx instance, and we don't group the alerts.

E.g. most lines in the chart below ([link to chart](https://graphite.blue.production.govuk.digital/render?from=17:56_20201012&until=18:11_20201012&target=movingMedian(transformNull(stats.cache-ip-*.nginx_logs.www-origin.http_5xx,0),%225mins%22))) resulted in a separate page:

dupe-nginx

From some manual checks in graphite, you have already made a significant dent in the "Nginx pages us too much for blips" problem with https://github.com/alphagov/govuk-puppet/pull/10802, and we've not had Icinga page us for this check since.

I double checked that this won't prevent us from hearing about issues we'll want to respond to. I did a quick check of the October 12 incident with the various parameters for the "5xx rate for www-origin" alert (results below). It looks like this change would delay us hearing about issues like this by ~1 minute, which I think is acceptable. We can tune it further if we find we want to hear about issues sooner.

```
8mins period, 5mins window (current), 18:09: 0.7145833333333333
15mins period, 5mins window, 18:09: 0.3811111111111111
15mins period, 3mins window, 18:09: 1.2966666666666669
8mins period, 5mins window (current), 18:10: 1.39375
15mins period, 5mins window, 18:10: 0.7433333333333334
15mins period, 3mins window, 18:10: 1.658888888888889
15mins period, 5mins window, 18:11: 1.1055555555555556
```

I also checked the Fastly change and found we'll still see alerts quickly.

I have only compared the checks for router's Nginx and Fastly, so I'm not able to say whether the other changes are justified.

benthorner:
Thanks @bilbof! 5xx is a hard alert to try and optimise, since it's fundamentally arbitrary ("how many 5xx errors means broken?"). I did some visual checking of this, but nothing as thorough as the above - great stuff!

From all these manual checks and my own mental model of how the alert works, I'm confident we will continue to get timely errors for more definite issues with our apps, just as much as we will for Nginx (origin) and Fastly (edge).

To close, it's worth noting that we will always miss some alerts, as some pages just don't generate enough (error) traffic to trigger this. I've come to think of the alert as a fallback: we should have more specific checks if possible.

PR #10818, opened 2020-11-03
Stop having a double standard for origin 5xx

benthorner:

Previously we had two alerts for origin 5xx errors, with different
thresholds for in-hours vs. on-call. The importance of the site
being broken should not depend on the time of day, so this merges
the alerts together and thus unifies the thresholds:

- We take the lower threshold for the warning, to ensure we don't
lose visibility we have now. This won't apge anyone.

- We take the higher threshold for the critical, which is better
indicator that something is *definitely* broken.


bilbof (approved):
Great change. This makes sense, might be worth getting a wider consensus on the in-hours threshold change (from 0.6 -> 0.9) with another review - your call.

deborahchua (approved):
I think this also makes sense, and I can't think of a reason why we should have two separate alerts

PR #10822, opened 2020-11-04
Wait even longer to alert about established conns

benthorner:
https://trello.com/c/7QA5eAoC/483-do-an-audit-to-check-for-any-more-work-to-do-on-alerts-epic

Previously we applied a moving median average to the metric for this
alert, in order to filter out spikes [1]. However, it's still possible
to get false alarms due to temporary load that lasts a few minutes,
but doesn't represent an issue: the app recovers when as the load dies
down. This extends the data window for this alert from 5 minutes (the
default) to 15 minutes, reduced using a mean average [2].

Mean averaging is preferable because it's more responsive to increases
in the underlying metric. This means we get warnings earlier than if we
instead extended the existing median window to 15 minutes. Using the
combination of averages seems to be the best compromise:

- A 5 minute median average to remove one-off spikes
- A 15 minute mean average of that, to dampen short bursts

[1]: https://github.com/alphagov/govuk-puppet/pull/10804
[2]: https://github.com/pyr/check-graphite#how-it-works

## Before

Screenshot 2020-11-04 at 12 28 42

## After

Note how the time to reach a "warning" level is similar, but the overall max value is dampened.

Screenshot 2020-11-04 at 12 29 13

## Raw

Screenshot 2020-11-04 at 12 29 43

## Extended Median (alt)

Note how the overall values are much lower than with a mean average (above). Another perspective: it takes longer to reach a threshold value.

Screenshot 2020-11-04 at 12 30 12

### DIFF
--- a/modules/govuk/manifests/app/config.pp
+++ b/modules/govuk/manifests/app/config.pp
@@ -416,6 +416,7 @@ define govuk::app::config (
       critical       => $local_tcpconns_critical,
       desc           => "Established connections for ${title_underscore} exceeds ${local_tcpconns_warning}",
       host_name      => $::fqdn,
+      from           => '15minutes',
       notes_url      => monitoring_docs_url(established-connections-exceed),
       contact_groups => $additional_check_contact_groups,
     }


1pretz1 (approved):
Looks good, thanks for adding screenshots!

benthorner:
Note: we've also reached agreement on [a comparable PR to this one](https://github.com/alphagov/govuk-puppet/pull/10817#pullrequestreview-523598548), so doubly confident this change can go in 👍.

PR #10829, opened 2020-11-09
Aggregate Icinga alert counts for faster reporting

benthorner:
This introduces two new metric namespaces so that we can easily see
the total number of alerts overall and per-machine. Currently, the
number of metrics is too great to compute this.


benthorner:
Note I followed the example here: https://github.com/graphite-project/carbon/blob/master/conf/aggregation-rules.conf.example

benthorner:
Note I followed the example here: https://github.com/graphite-project/carbon/blob/master/conf/aggregation-rules.conf.example

deborahchua (approved):

PR #10831, opened 2020-11-09
Fix new aggregations for Icinga alerts

benthorner:
Previously the config was the wrong way round :-|.


benthorner:
Tested and works on Integration ✅

deborahchua (approved):

PR #10832, opened 2020-11-10
Duplicate OAUTH_* env vars with GDS_SSO prefixed equivalents

kevindew:
This change is made to allow the transition of GOV.UK apps from OAUTH_* env vars to GDS_SSO_OAUTH_* ones where their purpose is clearer. This adds duplicate ones prefixed with GDS_SSO_* so that apps can seamlessly transition to the prefixed equivalent. Once all apps are switched over I can then delete the ones without a prefix.

This PR also cleans up the env var setting for Authenticating Proxy and Search API, which were a bit unconventional. It's easier to review this commit by commit.

This is a precursor for: https://github.com/alphagov/gds-sso/pull/242

### DIFF
--- a/modules/govuk/manifests/apps/search_api.pp
+++ b/modules/govuk/manifests/apps/search_api.pp
@@ -219,21 +219,12 @@ class govuk::apps::search_api(
     "${title}-UNICORN_TIMEOUT":
       varname => 'UNICORN_TIMEOUT',
       value   => $unicorn_timeout;
-  }
-
-  govuk::app::envvar {
     "${title}-PUBLISHING_API_BEARER_TOKEN":
       varname => 'PUBLISHING_API_BEARER_TOKEN',
       value   => $publishing_api_bearer_token;
-  }
-
-  govuk::app::envvar {
     "${title}-EMAIL_ALERT_API_BEARER_TOKEN":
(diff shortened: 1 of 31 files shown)


modules/govuk/manifests/apps/search_api.pp
```diff
@@ -306,4 +277,11 @@
       varname => 'TENSORFLOW_SAGEMAKER_VARIANTS',
       value   => $tensorflow_sagemaker_variants;
   }
+
+  if $elasticsearch_b_uri != 'null' {
```
benthorner, line 287: Out-of-scope: it looks like this [was removed](https://github.com/alphagov/search-api/commit/aef1da207bc6183e1681c405b8883f29a2d6fe56). The variable is no longer present in production, so we could just remove it here.

benthorner (approved):
Nice cleanup for those random apps. That'll make Replatforming just that little bit easier!

modules/govuk/manifests/apps/search_api.pp
kevindew: Oh great - nice catch.

kevindew: I've pushed a change up for this, if you could take another look

benthorner (approved):

PR #10835, opened 2020-11-11
Try alternative fix for Slack notifications

benthorner:
Previously we attempted to fix Slack notifications by upgrading the
Slack plugin [1], but for some reason the job builder still won't
generate the necessary config to make use of it. Since we can still
achieve the desired affect with the legacy config, this switches to
using that form, so we get notified about every failure.

[1]: 7b457741b


benthorner:
Closing in favour of https://github.com/alphagov/govuk-puppet/pull/10836

PR #10836, opened 2020-11-11
Upgrade Job Builder to support newer plugins

benthorner:
Previously we upgraded the Slack plugin to utilise a new config
feature [1], but the job builder also needs to be updated so that
we generate the new form of config for the plugin. The docs for
the builder aren't clear about what changes were made in each
release, so I'm specifying the latest one [2].

[1]: 21263c5a1
[2]: https://pypi.org/project/jenkins-job-builder/#history


benthorner:
Tested and works on Integration ✅

theseanything (approved):

PR #10839, opened 2020-11-11
Fix broken provisioning for Jenkins

benthorner:
Previously we applied this same change for CI Jenkins [1].

[1]: https://github.com/alphagov/govuk-puppet/commit/31ff8bfab00bc1ad160e9ef29254943cb575557b#diff-b37901140d98466ceb6ea10b5a77a2b5a29ef97829d59c8f050fd24792490dceR78


fredericfran-gds (approved):
looks good

PR #10847, opened 2020-11-13
Remove unprefixed OAuth environment variables

kevindew:
In https://github.com/alphagov/govuk-puppet/pull/10832 we duplicated the OAuth environment variables to transition to a [renaming](https://github.com/alphagov/gds-sso/pull/242) without downtime. All these apps have now been deployed to use the new environment variable naming so the old environment variables can be removed. This PR removes them all.

It also does some clean-up on the documentation of the oauth_id and oauth_secret variables for consistency, with a couple of tangential fixes thrown in (sorry).

More info in the commits.


benthorner (approved):
🏴󠁧󠁢󠁥󠁮󠁧󠁿 

The other cleanups make sense to me - for secrets, I can see we didn't comment about the secret nature of "db_password", for example, so there was nothing consistent about this.

PR #10852, opened 2020-11-20
Expose arbitrarily hidden healthchecks

benthorner:
https://trello.com/c/ILpSOFhq/201-enable-cd-for-quick-win-and-possible-quick-win-apps

Previously we noted that the majority of our apps expose their
healthcheck endpoint [1], with the original concern for not
doing so being vague and not representative of the defaults in
place. Of the three apps that still had their healthcheck route
hidden, none of the commits make any mention of this or suggest
why the endpoint should be treated specially. In addition, two
of the three apps aren't even resolvable externally. This removes
the functionality to hide the healthcheck route and thus exposes
it for three inconsistent apps.

[1]: a91df2aa30d20a2a099b9bf843d1c8c159dfe4d2

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -64,16 +64,6 @@
 #   By default the alert doc will be "app-healthcheck-not-ok".
 #   Set to true to change this to "<app>-app-healthcheck-not-ok".
 #
-# [*expose_health_check*]
-# whether to expose the health check to the proxy.
-#
-# In some apps, such as bouncer, this must be exposed so load balancers know
-# whether to send requests to a particular server; in other apps, such as
-# publisher, this must not be exposed, because then it would provide an
-# uncached, unauthenticated, publicly accessible point of access into the
-# state of our admin systems.
-#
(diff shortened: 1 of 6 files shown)


thomasleese (approved):
Thanks, this also resolves this card! https://trello.com/c/6Pi4obSe/1125-review-exposed-and-unexposed-healthchecks-of-apps

benthorner:
Oh thanks for the reference @thomasleese 💯. I've had a quick look at the card you linked to, and realised Whitehall (Admin) had one extra endpoint blocked using `hidden_paths` directly. I've added another commit to remove that.

> Exposing the healthcheck can be a mild security risk for DDOS attacks and, for admin systems, provides an uncached, unauthenticated, publicly accessible point of access.

I noted this on one of the commits to add this functionality. I maintain it's unfounded, though, as many of the routes aren't actually exposed, and those that are are for publishing apps, which are uncached anyway.

benthorner:
Could you take another look?

thomasleese (approved):

PR #10853, opened 2020-11-20
Update email override list

sihugh:
Adding me so I can receive emails in non-production.

Also removing Steve and Vanita as they don't need this any more.


sihugh:
Has been replaced by a sensible way of doing things

PR #10859, opened 2020-11-25
Set up a GOVUK_NOTIFY_RECIPIENTS env var for Email Alert API

kevindew:
Trello: https://trello.com/c/G35ncFfw/645-simplify-configuration-for-email-sending

This first reduces the number of people who are configured to receive integration/staging email notifications to just developers in the GOV.UK Notifications and Transitions teams.

It then adds an environment variable `GOVUK_NOTIFY_RECIPIENTS` which is intended to help Email Alert API transition from a rather complicated configuration system (provider, whitelist, whitelist_only etc) to a much simpler one. It also offers the first step towards removing the whitelist terminology.

Once Email Alert API has been updated to use this new environment variable I'll come back and clean up. I didn't see much value in leaving comments in the code to explain what is legacy or not as there would be quite a lot to comment on and I expect it won't be long until they're removed.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -202,6 +202,7 @@ task :check_consistency_between_aws_and_carrenza do
     govuk::apps::email_alert_api::email_archive_s3_bucket
     govuk::apps::email_alert_api::email_archive_s3_enabled
     govuk::apps::email_alert_api::enabled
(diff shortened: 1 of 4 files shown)


benthorner (approved):
🇦🇶

PR #10861, opened 2020-11-26
Reduce collections response timeout from 15 seconds to 1 second

bilbof:
Nginx proxy_read_timeout for collections will be reduced to 1s.

The intent is to bail out on slow requests. Currently 99th percentile for collections is >1s.


thomasleese (approved):
Makes sense to me, although perhaps we don't need this anymore..

bilbof:
I won't merge it for now. Perhaps one for the incident review.

bilbof:
Closing as not pursued.

PR #10864, opened 2020-12-03
Remove Tom W from GOV.UK ssh authorised-keys.

sengi:
Tom W is no longer on RE GOV.UK and went to PaaS team and not Replatforming. He's also no longer on the oncall rota, so no longer needs access to GOV.UK AWS.

Rollout plan: the usual: merge, deploy Puppet in integration+staging, wait half an hour and see if it doesn't break, roll to prod.

See also alphagov/govuk-aws-data#813 and alphagov/govuk-user-reviewer#484.

### DIFF
--- a/hieradata_aws/integration.yaml
+++ b/hieradata_aws/integration.yaml
@@ -533,7 +533,6 @@ users::usernames:
   - suganyasivaskantharajah
   - thomasleese
   - timblair
-  - tomwhitwell
   - vitaliemogoreanu
   - williamfranklin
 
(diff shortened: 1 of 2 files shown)


richardTowers (approved):

sengi:
Aha there's a [playbook for this](https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html). Didn't find it at first. Says to roll out the `ensure: absent` first before removing from the list of users in integration, but seems to have worked anyway (puppet-agent really did the user from ssh_authorized_keys - I spot-checked the logs on one machine).

PR #10866, opened 2020-12-03
Switch Email Alert API Notify recipients in integration and staging

kevindew:
Trello: https://trello.com/c/l4ahPmzS/644-switch-integration-staging-to-send-emails-to-a-google-group

This changes the addresses that can receive Notify emails to be Google
group addresses rather than individuals, this has been done to simplify
the management of recipients. Previously when someone wanted to test
Email Alert API email sending they would need to add their email into
Notify and into puppet before testing, with this change they can test
immediately by viewing the group and using the group email address for
their test mails.


benthorner (approved):
🇲🇾

PR #10870, opened 2020-12-04
Disable pager-duty notification in carrenza production

rtrinque:
The only thing still running in carrenza production is performance
 platform. This service will not be monitored out of hours.


sengi (approved):
SGTM - probably also worth a review from 2nd-line for visibility as this affects their team more than ours (?)

PR #10873, opened 2020-12-08
Enable Continuous Deployment for HMRC Manuals API

MuriloDalRi:
Dependent on:

- https://github.com/alphagov/smokey/pull/759
- https://github.com/alphagov/govuk-saas-config/pull/74
- https://github.com/alphagov/hmrc-manuals-api/pull/544

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -1121,6 +1121,7 @@ govuk_jenkins::jobs::deploy_app_downstream::applications:
   content-publisher: {}
   finder-frontend: {}
   government-frontend: {}
+  hmrc-manuals-api: {}
   info-frontend: {}
   manuals-frontend: {}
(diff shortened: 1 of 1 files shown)


hieradata_aws/common.yaml
```diff
@@ -1122,6 +1122,7 @@ govuk_jenkins::jobs::deploy_app_downstream::applications:
   finder-frontend: {}
   government-frontend: {}
   info-frontend: {}
+  hmrc-manuals-api: {}
```
ChrisBAshton, line 1125: This should come before `info-frontend` so we can keep it alphabetical 👍

MuriloDalRi: 🤦 of course

ChrisBAshton (approved):

PR #10881, opened 2020-12-14
Double the number of unicorn workers for collections

huwd:
We estimate that each unicorn worker is using around 315MiB, given the
number of resources that we currently allocate to collections we should
be able to safely scale unicorn workers without going over 50% of the
boxes allocated RAM.

We'll also double the Icinga warning and critical levels so we do not
make 2ndline sad.

Co-authored-by: Jon Hallam <jonathan.hallam@digital.cabinet-office.gov.uk>


bilbof (approved):
LGTM! 👍

kevindew:
Great stuff - nice working out. Did you check the memory situation was the same in each environment? I can't remember if integration is under provisioned for this instance.

PR #10884, opened 2020-12-17
Tweak email send attempts dashboard graph

benthorner:
Previously this had an additional metric to cope with sent email
stats being a combination of those successfully sent, and those
that were rate limited; this is no longer the case [1].

This also adds metrics for pseudo-sent emails, which are in the
majority in Integration and Staging, as well as:

- Removing the misleading graph title (the 30s is just a way to
approximate the per minute rate).

- Using a "max" consolidation, so the graph does not appear
alarming if the viewer zooms out the time range. Using "max" still
means we surface any worrying values.

[1]: https://github.com/alphagov/email-alert-api/commit/294ce78dd36b95347db551a580d69eeda932b229#diff-2117167ba6330613e0c2bfe4df9385ddfa9d6a2bdf363f4e72cc91222ff67d3aL47

### DIFF
--- a/modules/grafana/files/dashboards/email_alert_api_technical.json
+++ b/modules/grafana/files/dashboards/email_alert_api_technical.json
@@ -161,31 +161,26 @@
           "steppedLine": false,
           "targets": [
             {
-              "hide": true,
+              "hide": false,
               "refId": "A",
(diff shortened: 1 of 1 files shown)


1pretz1 (approved):
Nice! +1 for the Pseudo metrics

PR #10888, opened 2020-12-18
Update Redis cache CPU alert

1pretz1:
Previously the ElastiCache CPU Utilization alert wasn't really useful as
it would never fire due to the thresholds being too high (80% and 90%).

The reason being is that 'CPUUtilization' would return how much CPU each
Redis process is using as a percentage versus whats on the machine. As
each Redis process (replica and primary) can only use one CPU at a time
then the 'CPUUtilization' metric for each node will never cross 50% as
each machine has two CPUs.

The 'EngineCPUUtilization' metric however measures how much CPU the
Redis process is using out of what it actually has available instead of
whats on the machine. This gives a more realistic measurement on the CPU
usage of Redis.

The new thresholds seems reasonable - 90% warn and 95% critical
(averaged over 5 mins) as we've only briefly crossed the 90% threshold
(90.7% for the primary) once in the last 4 weeks under normal operating 
conditions in production.

These thresholds would've alerted us of the system struggling during a recent 
load test for Email Alert API as the system was using consistently 98%/99% 
of the 'engine' CPU for hours which went unnoticed.

## Production Redis Engine CPU usage over the last 4 weeks

Screenshot 2020-12-18 at 10 58 33

## Under a load test in Staging causing degraded performance 

Screenshot 2020-12-18 at 12 27 30

### DIFF
--- a/modules/monitoring/files/usr/lib/nagios/plugins/check_aws_cache_cpu
+++ b/modules/monitoring/files/usr/lib/nagios/plugins/check_aws_cache_cpu
@@ -1,8 +1,8 @@
 #!/opt/python2.7/bin/python
 
 ##########################################################
-# This script will cpture and report on AWS ElastiCache #
-# CPUUtilization         	                         #
+# This script will capture and report on AWS ElastiCache #
+# EngineCPUUtilization         	                         #
 ##########################################################
 
 ###########################################################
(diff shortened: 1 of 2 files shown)


thomasleese (approved):
Good catch!

benthorner (approved):
Looks good to me too - graphs were useful 💯.

PR #10892, opened 2020-12-22
Temporarily re-add the 'marksutton' user

ChrisBAshton:
Mark got removed in #10878, a little too quickly: we hadn't yet
applied an `ensure => absent` to their class as per the 'Remove
a user from Puppet' instructions:
https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html

I can see that they still have a home directory on at least one
backend class machine on Integration, so we need to do a little
housekeeping. We'll temporarily add the user back to Puppet with
an `ensure => absent` instruction. The other properties don't seem
to be required:
https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_user/manifests/init.pp

When this is merged and Puppet is deployed, we should see the home
directory removed, and we can then revert this commit.

While I'm here, I spotted two other orphaned home directories on
Integration, so am applying the same fix to those users too.

Trello: https://trello.com/c/7ySrimBF/1058-leaver-mark-sutton

### DIFF
--- a/modules/users/manifests/chrisfarmiloe.pp
+++ b/modules/users/manifests/chrisfarmiloe.pp
@@ -1,0 +1,6 @@
+# Creates the chrisfarmiloe user - temporarily - so we can remove their home dir
+class users::chrisfarmiloe {
+  govuk_user { 'chrisfarmiloe':
+    ensure   => absent,
+  }
+}
--- a/modules/users/manifests/isabelllong.pp
+++ b/modules/users/manifests/isabelllong.pp
@@ -1,0 +1,6 @@
+# Creates the isabelllong user - temporarily - so we can remove their home dir
+class users::isabelllong {
+  govuk_user { 'isabelllong':
+    ensure   => absent,
+  }
+}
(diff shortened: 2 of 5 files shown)


kevindew (approved):
Looks cool - guess you'll find out with a whirl on integration if there are any problems.

ChrisBAshton:
Force pushed to rebase against `master`, to incorporate Mapit changes before deploying my branch to Integration.

ChrisBAshton:
Tried this and it didn't work. Home dirs still persisted.

Confirmed that adding an `ensure => absent` to an existing user _does_ remove the home dir, so it's only retrospectively applying the argument that doesn't work.

> Our `govuk_user/manifest/init.pp` doesn't look like it deletes the home directory when the user is removed: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_user/manifests/init.pp#L93

ChrisBAshton:
Closed in favour of https://github.com/alphagov/govuk-developer-docs/pull/2909.

PR #10894, opened 2020-12-22
Change mapit pv to /dev/xvdf in integration

barrucadu:
I've recycled the mapit machines in integration, and they all have a /dev/xvdf but no /dev/nvme1n1 ([like in production](https://github.com/alphagov/govuk-puppet/blob/master/hieradata_aws/class/mapit.yaml)).


richardTowers (approved):
This was so annoying when we spun up the test environment.

PR #10896, opened 2020-12-22
Frontend shared cache

richardTowers:
We use a gem called Dalli as a memcached client. This gem takes a
MEMCACHE_SERVERS environment variable, which it will use to work out
which memcached server to talk to.

Currently we're not setting this variable, so we're getting its default
value (127.0.0.1:11211).

To make it easy to override, this makes puppet always set the variable.
By default, we use the same default as Dalli, which should not change
the existing behaviour. In environments where we're using a shared
elasticache memcached instance, we can set it to the DNS name (and port)
of the cluster.

In GOV.UK's current infrastructure, there's a Search path set in
/etc/resolv.conf, so `frontend-memcached` will result in a lookup to
`frontend-memcached.production.govuk-internal.digital`, which will be
CNAMED to `frontend-memcached.blue.production.govuk-internal.digital`.

For now, we're going to experiment with a shared cache in integration
and staging. If it works well, we'll roll it out to production.

See alphagov/govuk-aws#1394 for the
infrastructure side to this.

### DIFF
--- a/modules/govuk/manifests/apps/collections.pp
+++ b/modules/govuk/manifests/apps/collections.pp
@@ -38,10 +38,7 @@
 #   Feature flag to switch GOV.UK Accounts on or off for the Transition Checker
 #
 # [*memcache_servers*]
-#   URL of a shared memcache cluster. Defaults to 127.0.0.1:11211 which is the
-#   default in "Dalli" our memcached client (
-#     https://github.com/petergoldstein/dalli/blob/1fbef3c/lib/dalli/client.rb#L35
-#   )
+#   URL of a shared memcache cluster.
 #
 class govuk::apps::collections(
   $vhost = 'collections',
@@ -53,7 +50,7 @@ class govuk::apps::collections(
   $nagios_memory_critical = undef,
   $email_alert_api_bearer_token = undef,
   $feature_flag_accounts = false,
-  $memcache_servers = '127.0.0.1:11211',
+  $memcache_servers = undef,
 ) {
   govuk::app { 'collections':
     app_type                 => 'rack',
--- a/modules/govuk/manifests/apps/collections.pp
+++ b/modules/govuk/manifests/apps/collections.pp
@@ -37,6 +37,12 @@
 # [*feature_flag_accounts*]
 #   Feature flag to switch GOV.UK Accounts on or off for the Transition Checker
 #
+# [*memcache_servers*]
+#   URL of a shared memcache cluster. Defaults to 127.0.0.1:11211 which is the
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/apps/collections.pp
```diff
@@ -82,5 +86,8 @@
     "${title}-FEATURE-FLAG-ACCOUNTS":
         varname => 'FEATURE_FLAG_ACCOUNTS',
         value   => $feature_flag_accounts_var;
+    "${title}-MEMCACHE_SERVERS":
+        varname => 'MEMCACHE_SERVERS',
```
barrucadu, line 90: Where is this variable used?  I don't see it referenced in collections.

richardTowers: It's used in Dalli (which is our memcached client):

https://github.com/petergoldstein/dalli/blob/1fbef3caf060d7d4e852c2a38e7c2987e208f0c7/lib/dalli/client.rb#L35

We should probably mention that in a comment though.

kevindew: It's the place the Ruby memcache driver looks for it: https://github.com/petergoldstein/dalli/blob/1fbef3caf060d7d4e852c2a38e7c2987e208f0c7/lib/rack/session/dalli.rb#L27

richardTowers: I've added a comment explaining this, which hopefully helps.

kevindew:
We've completed a round of testing this morning and are happy that this will benefit production. I've rebased this and another commit to apply this to production.

sihugh (approved):

hieradata_aws/class/draft_frontend.yaml
```diff
@@ -1,6 +1,8 @@
 ---
 
+govuk::apps::collections::memcache_servers: '127.0.0.1:11211'
```
sengi, line 3: What does it mean for this to be set to localhost on the draft app? Is it really just a dummy value? Do you intend for memcache to be disabled for draft?

Similarly below with the default on the arg to the class - I'm not clear as to the intent and the meaning.

kevindew: Good point. It refers to the instance of memcache on the local machine (one is installed by puppet). I wonder if it'll be cleaner if it's just undefined though and the machine picks it up automatically (as it does currently)

sengi: That seems much clearer to me, thanks!

sengi (approved):

PR #10898, opened 2020-12-22
Quadruple the unicorn workers for finder_frontend in staging.

sengi:
Also proportionally increase the thresholds for the Icinga memory "alerts", because these have some bizarre automation attached to them which helpfully kills the application process when the alert fires.

See alphagov/govuk-aws-data#827 for context.


sihugh (commented):
Might want to consider bumping the file handles for this one as well: https://github.com/alphagov/govuk-puppet/pull/10899/files?w=1

What @kevindew said about it being per-user and therefore more delicate on machines with multiple apps makes me reticent to bump this more widely.

sengi:
> Might want to consider bumping the file handles for this one as well: https://github.com/alphagov/govuk-puppet/pull/10899/files?w=1
> 
> What @kevindew said about it being per-user and therefore more delicate on machines with multiple apps makes me reluctant to bump this more widely.

I think it's a bad alert, but I agree let's just bump the threshold for this app for now. Done.

sihugh (approved):

PR #10901, opened 2020-12-23
Declare second Sidekiq worker (solely for sending emails)

ChrisBAshton:
Depends on https://github.com/alphagov/email-alert-api/pull/1534.

This `govuk::procfile::worker` declaration refers to the new
sidekiq_send_email.yml config in Email Alert API, introduced in
https://github.com/alphagov/email-alert-api/pull/1534

This is largely a copy and paste of the existing declaration above it,
but with a new name, which becomes the `$title` used throughout the
`govuk::procfile::worker` class (and also lends itself to the service
name we use in commands like `service {SERVICE_NAME} start`):
https://github.com/alphagov/govuk-puppet/blob/fbbecba00f5e4a06bed1c48b34445d7e4235c0fd/modules%2Fgovuk%2Fmanifests%2Fprocfile%2Fworker.pp

I also needed to pass `process_type => 'send_email_worker'` so that it knows
which worker to instantiate. I took inspiration from prior art in govuk-puppet
([1]) for Content Data API, whose process is called 'default-worker'
rather than 'worker' ([2]). We also used to have multiple workers in
the (now retired) Content Audit Tool ([3] and [4]), for which I remember
having to fix the regex on my first 2nd line shift ([5]). I don't think
we need to specify a custom process regex for this addition.

[1]: https://github.com/alphagov/govuk-puppet/blob/7e6a90da615e1094df305a367418139a76de0f8d/modules%2Fgovuk%2Fmanifests%2Fapps%2Fcontent_data_api.pp#L147-L152
[2]: https://github.com/alphagov/content-data-api/blob/2df29c27f2e715b07f5b8758519bb0d01ebdf6ff/Procfile#L1
[3]: https://github.com/alphagov/content-audit-tool/commit/6edc4e488dd8d7347517f5818ad977d9755fba01#diff-0a99231995da379e7aebabe76c9d849a23737a42c3b3a8994043e2aa80958424
[4]: https://github.com/alphagov/govuk-puppet/blob/e335930e21814e7aa27efa7e8f4e5360796091e9/modules/govuk/manifests/apps/content_audit_tool.pp#L96-L106
[5]: https://github.com/alphagov/content-audit-tool/pull/306

Trello: https://trello.com/c/52YiPQ3y/2273-investigate-splitting-queues-to-send-email-onto-a-different-set-of-workers

### DIFF
--- a/modules/govuk/manifests/apps/email_alert_api.pp
+++ b/modules/govuk/manifests/apps/email_alert_api.pp
@@ -151,6 +151,15 @@ class govuk::apps::email_alert_api(
     memory_critical_threshold => 10000,
   }
 
+  govuk::procfile::worker {'email-alert-api-send-email':
+    ensure                    => $ensure,
+    enable_service            => $enable_procfile_worker,
+    process_type              => 'send_email_worker',
+    alert_when_threads_exceed => 100,
+    memory_warning_threshold  => 8000,
+    memory_critical_threshold => 10000,
+  }
+
   Govuk::App::Envvar {
     ensure          => $ensure,
     app             => 'email-alert-api',


kevindew:
Awesome stuff working on this Chris. I was wondering if we wanted to this second worker to be focused only on fast tasks rather than just emails? as we've got tasks like MetricsCollectionWorker that can get caught behind slow processing of content changes. Just wondered if that was considered (could end up with 3 processes even)

ChrisBAshton:
Thanks @kevindew. It's not something that occurred to me, as I don't have a concrete idea of how long each class of worker takes to complete. It also isn't on the card, so it perhaps hasn't been considered at all.

I quite like the current idea of a parent worker that only deals with email sending, for simplicity and for a clear 1:1 mapping between jobs processed & emails sent. A third parent worker could follow a similar principle and deal only with email creation, leaving the original parent worker only to deal with auxiliary matters such as metrics collection and DB cleanup. I wonder if three workers might be over-engineering things a bit, as we don't seem to ever have had more than two in our apps.

Alternatively, as you suggest, we could create a 'fast jobs worker' and a 'slow jobs worker', and divvy out the job queues based on their speed rather than their purpose. I quite like that idea too - it gives Email Alert API more freedom to define new job classes without having to go and update govuk-puppet. But it might be initially confusing to conflate fast/slow parent workers with high/low priority 'send email' queues. I assume we'd have _all_ email sending queues in the 'fast jobs' worker, even the ones not in 'immediate' or 'immediate high'?

ChrisBAshton:
I've just had a look at the [MetricsCollectionWorker](https://github.com/alphagov/email-alert-api/blob/bae78a3cb7e970d290d4a95613149c1d5d34e4f1/app/workers/metrics_collection_worker.rb) and it looks pretty important. Without it, we have no visibility of how well Email Alert API is working (e.g. no way of [measuring unprocessed content changes](https://github.com/alphagov/email-alert-api/blob/21f768a9cfa6c5f16ba6eb9a13c26faa5424e802/app/workers/metrics_collection_worker/content_change_exporter.rb#L2-L4)). So this does seem like something high priority that shouldn't be left to process at the bottom of the queue.

The only situation I can think of where that might happen is if metrics collection is given a low priority of 1, and all of the workers are tied up with a big backlog of email generation. We could potentially be in a situation where metrics aren't reported for many minutes (even an hour?), which would be terrible in an incident. Could we not mitigate this by leaving it in the 'slow' worker queue, but giving it the highest priority?

If so, that seems a good approach to me.

If not, then I think you're right that we should not leave metrics collection in the slow job worker. It either ought to be in the fast jobs worker - which would then need a rename from `send_email_worker` - or given its own dedicated parent worker. I don't know enough about all the different workers and their speed and importance (we haven't touched on RecoverLostJobsWorker, EmailDeletionWorker, HistoricalDataDeletionWorker), so my inclination is to keep things simple:

- `fast-jobs-worker` for email sending and metrics collection
- `slow-jobs-worker` for everything else

This still doesn't feel quite right. I suspect for example that `EmailDeletionWorker` is actually a fast job, but that it's also a very low priority one, designed to prevent disk space clogging up over time. So perhaps `slow-or-low-priority-jobs-worker` would be a better name for that one.

Any thoughts, @kevindew? We're wrapping up this preliminary investigation work and would be good to have decided on a worker division strategy in principle, ready for [phase 2](https://trello.com/c/hnpIufBV/1989-load-test-email-alert-api-with-different-queues-and-concurrency-values) (though I don't think it's hugely high on the priority list this quarter).

PR #10907, opened 2020-12-24
Remove Healthchecks annotation from Grafana dashboards

bilbof:
This annotation doesn't seem to work; it's not a useful query. Removing it should reduce the number of requests made to Graphite.


richardTowers (approved):
Seems reasonable to me, if they don't work.

huwd (approved):

PR #10911, opened 2021-01-08
Add python3.6 and install on CKAN machines and CI agents

kentsanggds:
## What

Make python 3.6.12 package available on CKAN and CI agents so that CKAN 2.9 can run on python 3.6.12

## Reference 

https://trello.com/c/VrPqMpYM/2286-enable-the-ci-test-pipeline-to-run-against-ckan-29-on-python-3-on-a-particular-agent

### DIFF
--- a/modules/govuk_jenkins/manifests/packages/govuk_python.pp
+++ b/modules/govuk_jenkins/manifests/packages/govuk_python.pp
@@ -25,10 +25,13 @@ class govuk_jenkins::packages::govuk_python (
     key          => $apt_mirror_gpg_key_fingerprint,
   }
 
-  package { 'govuk-python':
-    ensure  => $govuk_python_version,
-    require => Apt::Source['govuk-python'],
-  }
+  ensure_packages(
+    ['govuk-python'],
+    {
+      ensure  => $govuk_python_version,
+      require => Apt::Source['govuk-python'],
+    }
+  )
 
   package { 'govuk-python-setuptools':
     ensure  => $govuk_python_setuptools_version,
--- a/modules/govuk_python3/manifests/init.pp
+++ b/modules/govuk_python3/manifests/init.pp
@@ -1,0 +1,26 @@
+# == Class: govuk_python3
+#
+# Install python3 package
+#
+class govuk_python3 (
+  $govuk_python_version = '3.6.12',
+  $apt_mirror_hostname,
+  $apt_mirror_gpg_key_fingerprint,
+) {
+
+  apt::source { 'govuk-python3':
+    location     => "http://${apt_mirror_hostname}/govuk-python",
+    release      => 'stable',
+    architecture => $::architecture,
+    repos        => 'main',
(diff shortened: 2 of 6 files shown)


modules/govuk_python3/manifests/init.pp
```diff
@@ -0,0 +1,23 @@
...
+    location     => "http://${apt_mirror_hostname}/govuk-python",
+    release      => 'stable',
+    architecture => $::architecture,
+    repos        => 'main',
+    key          => $apt_mirror_gpg_key_fingerprint,
+  }
+
+  package { 'govuk-python3':
+    ensure  => $govuk_python_version,
+    require => Apt::Source['govuk-python'],
```
richardTowers, line 21: Should this be:

```suggestion
    require => Apt::Source['govuk-python3'],
```

?

kentsanggds: No, the source is still govuk-python, its just the version number thats different

kentsanggds: hmmm, actually, looks like the branch is failing now, I'll try with the change

fredericfran-gds (commented):
is there a reason why you are not bumping: https://github.com/alphagov/govuk-puppet/blob/8cca9aad2b68d6cb396a135f47524fafeca1c947/modules/govuk_python/manifests/init.pp#L12

modules/govuk_python/manifests/init.pp:12
  ensure_packages(['python3.5', 'python3.5-dev', 'python3-dev'])
<https://github.com/alphagov/govuk-puppet|alphagov/govuk-puppet>alphagov/govuk-puppet | Added by GitHub

kentsanggds:
Not sure what impact bumping that up would be on apps which rely on python 3.5, when I initially used 3.6 it was on minor version 8, I then did a bump up to version 12 for security patches and that caused some other issues which I fixed. It might be the same case with python 3.5

fredericfran-gds:
what kind of issues? Maybe we should try because there is both python 3.5 and 3.6.12 now on ci-agents as you include govuk_python too

modules/govuk_jenkins/manifests/packages/govuk_python.pp
```diff
@@ -25,10 +25,13 @@
     key          => $apt_mirror_gpg_key_fingerprint,
   }
 
-  package { 'govuk-python':
-    ensure  => $govuk_python_version,
-    require => Apt::Source['govuk-python'],
-  }
+  ensure_packages(
+    ['govuk-python'],
```
fredericfran-gds, line 29: did you mean govuk-python3 ?

kentsanggds: nope, it's on out aptly server under govuk-python, tested it out on Integration and it works.

fredericfran-gds: ok cool

fredericfran-gds (approved):
looks good and was tested

PR #10912, opened 2021-01-08
Use govuk-python::app-source class

kentsanggds:
## What

This fixes the installation of python3.6 on ci agents and the duplication of the `govuk-python` apt-source.


fredericfran-gds (approved):
looks good

PR #10913, opened 2021-01-08
add govuk-python

fredericfran-gds:


kentsanggds (approved):
I was hoping that https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_ci/manifests/agent.pp#L45 would ensure that it would installed on the CI agents, but although python 3.6 is available its not installed. 

I did some manual testing on the CI machines and it seems like python 3.5 is available, so it should be possible to also install python 3.6 side by side.

PR #10914, opened 2021-01-11
Add prometheus data source to Grafana

richardTowers:
It looks like this can work the same way as graphite does - with
https://prometheus as the URL. This works because /etc/resolve.conf sets
a search path of integration.govuk-internal.digital.
prometheus.integration.govuk-internal.digital is a CNAME to
prometheus.blue.integration.govuk-internal.digital, which points to
Prometheus' internal load balancer.

I'm not sure how Grafana is happy making an HTTPS connection to this
URL, becauce the certificate is definitely not valid for
https://prometheus. But I've tested it and it seems happy (just as it is
with graphite), and not having to pass the URL in as a variable makes
our lives significantly simpler.

### DIFF
--- a/modules/grafana/manifests/datasources.pp
+++ b/modules/grafana/manifests/datasources.pp
@@ -61,6 +61,19 @@ class grafana::datasources(
     '2>/dev/null','|','grep',shellquote('{"message":"Data source not found"}'),
   ], ' ')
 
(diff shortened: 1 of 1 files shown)


sengi (approved):
Fits with what's already there 🤷😅

PR #10915, opened 2021-01-12
install python 3.6.12

fredericfran-gds:
I tested on both ci_agents and ckan and it seems to work


benthorner (approved):
I'm not entirely sure what this PR does, but I trust that it's been tested 👍. Having a single Python module looks sensible.

PR #10916, opened 2021-01-12
use default python for icinga alert scripts

fredericfran-gds:
This is a side effect of govuk-python being used to install python 3.6. See #10915

### DIFF
--- a/modules/monitoring/manifests/checks.pp
+++ b/modules/monitoring/manifests/checks.pp
@@ -24,13 +24,13 @@ class monitoring::checks (
 
   exec { 'install_boto':
         path    => ['/opt/python2.7/bin', '/usr/bin', '/usr/sbin'],
-        command => '/opt/python2.7/bin/pip install boto',
+        command => 'pip install boto',
         require => Class['::govuk_jenkins::packages::govuk_python'],
   }
 
   exec { 'install_boto3':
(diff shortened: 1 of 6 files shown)


richardTowers:
This change looks okay to me, but I don't really understand the cause.

We've installed a new version of python, and now the old version of python isn't in the same place anymore? Why?

modules/monitoring/manifests/checks.pp
```diff
@@ -24,13 +24,13 @@
 
   exec { 'install_boto':
         path    => ['/opt/python2.7/bin', '/usr/bin', '/usr/sbin'],
-        command => '/opt/python2.7/bin/pip install boto',
+        command => 'pip install boto',
         require => Class['::govuk_jenkins::packages::govuk_python'],
   }
 
   exec { 'install_boto3':
         path    => ['/opt/python2.7/bin', '/usr/bin', '/usr/sbin'],
```
benthorner, line 32: Does this need to change as well?

fredericfran-gds:
I'm going to close this PR in favour of #10917. 

This whole issue came into about when python 3.6.12 was added here: https://apt.publishing.service.gov.uk/govuk-python/dists/stable/main/binary-amd64/Packages

Should have created a new package govuk-python3

PR #10917, opened 2021-01-12
fix python 3.6 install

fredericfran-gds:
govuk_python needs to be set only on ci_agents and ckan to 3.6. In other places, we keep it on 2.7.14.

### DIFF
--- a/hieradata_aws/class/ckan.yaml
+++ b/hieradata_aws/class/ckan.yaml
@@ -7,3 +7,5 @@ govuk::apps::ckan::enable_harvester_fetch: true
 govuk::apps::ckan::enable_harvester_gather: true
 
 govuk_solr6::present: false
+
+govuk_python::govuk_python_version: '3.6.12'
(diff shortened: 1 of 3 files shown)


benthorner (commented):
Will we still have issues on e.g. the CKAN machines? (Looking at [the previous PR](https://github.com/alphagov/govuk-puppet/pull/10916/files#diff-97461aaa9723d1d8c38bce4d3815f5bffe09357fa3e0abb471c4a486edf1fe45R1).)

fredericfran-gds:
> Will we still have issues on e.g. the CKAN machines? (Looking at [the previous PR](https://github.com/alphagov/govuk-puppet/pull/10916/files#diff-97461aaa9723d1d8c38bce4d3815f5bffe09357fa3e0abb471c4a486edf1fe45R1).)

according to @kentsanggds no, we want 3.6.12 on CKAN.

benthorner:
We still have alerts on CKAN machines, though. I notice the previous PR was trying to fix alerts to cope with Python 3. Is there something special about CKAN / CI machines that means this isn't an issue?

benthorner (approved):
Looking again at the previous PR, the [files affected](https://github.com/alphagov/govuk-puppet/pull/10916/files) are specific to the monitoring machines, which (with this PR) will still be running on Python 2.x. As long as there's nothing on the CKAN / CI agents that requires legacy Python, I agree we should switch to the newer version.

PR #10919, opened 2021-01-14
Removes Sentry error-reporting for failed Puppet builds

ChrisBAshton:
Depends on #10920.

The govuk-puppet [project in Sentry](https://sentry.io/organizations/govuk/issues/?project=125252) has consistently had > 100 errors a day, for a very long time now. It is not actively monitored, as we already have alerting in Icinga.

Moreover, Puppet is on Ruby version 1.9.3, and [uses an old sentry-raven gem version](https://github.com/alphagov/govuk-puppet/commit/5f699843de071d39ec6147684fa0ce65aa033957). It is therefore a blocker for us [upgrading sentry-raven to sentry-ruby](https://trello.com/c/1zVPYfTR/1979-replace-sentry-raven-with-sentry-ruby), which has dropped support for Ruby 2.3 and below.

This commit carefully unpicks the initial work added in 95f86d28a76c7b1b6a7ec8e5b2dc9ef89912e5c4, and subsequent refactor in https://github.com/alphagov/govuk-puppet/commit/c182a41e2c544b618768f0e05bb1be5d485dc0aa#diff-3ca7c12e32772363a8eaa238f62dd4cf2b354b93fb78f4531e9e357403531536.

Note that there are still plenty of occurrences of "sentry" throughout govuk-puppet. This is because we use govuk-puppet
(alongside govuk-secrets) to provide the unique Sentry DSN for every app on GOV.UK. This creates a `SENTRY_DSN` ENV var which is picked up by the application, which has its own version of sentry-raven running (via its dependency on govuk_app_config), so that it knows where to send its Sentry errors. I've had to be careful not to remove that.

(Here's an example of where SENTRY_DSN was added, and the kind of thing I've had to avoid removing: https://github.com/alphagov/govuk-puppet/pull/6237

...and here is its govuk-secrets counterpart: https://github.com/alphagov/govuk-secrets/pull/19/files#diff-648cfbf5de807e8057060d24c0756e5fdc26ca2541404a1933ef530565e0874bR740-R741)

NB, once this PR is merged, we can probably archive this repo: https://github.com/alphagov/puppet-reporter-sentry

Trello: https://trello.com/c/1pAh0QpK/2294-3-remove-govuk-puppet-project-from-sentry

**Once this PR is sense-checked, I’ll manually deploy on Integration and monitor our Sentry events for a while before merging.** Want to ensure it doesn't break the error-reporting for our apps.


modules/puppet/manifests/puppetserver.pp
```diff
@@ -40,13 +32,6 @@
   Class['::puppet::puppetserver::generate_cert']
     ~> Class['::puppet::puppetserver::nginx']
 
-  if $sentry_dsn {
-    class { '::puppet::puppetserver::sentry':
-      require => Class['::puppet::puppetserver::package'],
-      notify  => Class['::puppet::puppetserver::service'],
-    }
-  }
```
thomasleese, line 48: Unfortunately, just removing this code won't actually remove the configuration for it (https://serverfault.com/questions/455657/how-to-revert-a-configuration-with-puppet). Instead you'll need to first merge and deploy a change with `ensure => absent` and once that's run on all our machines, you can then merge this in.

The added complication here is that we wrote the `puppet::puppetserver::sentry` class and it doesn't accept an `ensure` parameter. So instead I'd suggest changing this line https://github.com/alphagov/govuk-puppet/blob/a1cd0a425e6d9c1cf9f321e91f39bf3cd87db39a/modules/puppet/manifests/puppetserver/sentry.pp#L12 to be `ensure => absent` and removing this https://github.com/alphagov/govuk-puppet/blob/a1cd0a425e6d9c1cf9f321e91f39bf3cd87db39a/modules/puppet/manifests/puppetserver/sentry.pp#L7-L9 so it doesn't run next time.

modules/puppet/manifests/master/package.pp
```diff
@@ -34,11 +30,6 @@
     provider => system_gem,
   }
 
-  package { 'sentry-raven':
-    ensure   => '2.2.0',
-    provider => 'system_gem',
-  }
```
thomasleese, line 40: Similar to my other comment, this will first need to be `ensure => absent` to make sure it gets uninstalled.

thomasleese (commented):
Definitely like this, I've got some comments regarding how to remove things from Puppet (it's slightly more complicated than it should be!)

modules/puppet/manifests/puppetserver.pp
ChrisBAshton: D'oh, I'd forgotten that! 🤦 Good catch. I've raised #10920.

ChrisBAshton:
Tested on integration, Sentry errors for other apps still come through, so we haven't broken anything: https://sentry.io/organizations/govuk/issues/2152180977/?project=202220

Without merging & deploying all the way through, it's hard to say if this stops the Puppet Sentry errors, as we don't seem to have a concept of environments in the govuk-puppet project on Sentry: https://sentry.io/organizations/govuk/issues/?groupStatsPeriod=auto&project=125252&query=is%3Aunresolved&statsPeriod=24h

It does _look_ as though the number of logged errors has gone down since I deployed this branch to integration though. 🎉

thomasleese (approved):

ChrisBAshton:
I've just archived https://github.com/alphagov/puppet-reporter-sentry, as this PR removes that dependency.

PR #10920, opened 2021-01-14
Ensure Sentry Puppet resources are removed prior to deletion

ChrisBAshton:
In #10919, we're removing Sentry error-reporting for failed Puppet builds,
but all this does is tell Puppet not to manage those resources
anymore. It doesn't actually remove them.

We need to tell Puppet to remove the resources before we can
safely remove the resource configurations.

See https://serverfault.com/questions/455657/how-to-revert-a-configuration-with-puppet

Trello: https://trello.com/c/1pAh0QpK/2294-3-remove-govuk-puppet-project-from-sentry

Co-authored-by: Thomas Leese <thomas.leese@digital.cabinet-office.gov.uk>

### DIFF
--- a/modules/puppet/manifests/master/package.pp
+++ b/modules/puppet/manifests/master/package.pp
@@ -35,7 +35,7 @@ class puppet::master::package(
   }
 
   package { 'sentry-raven':
-    ensure   => '2.2.0',
+    ensure   => absent,
     provider => 'system_gem',
   }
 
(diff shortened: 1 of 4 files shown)


thomasleese (commented):
I reckon we'll want this change https://github.com/alphagov/govuk-puppet/pull/10919/files#diff-e907b2599685443a359e9365bce1885fab56e5f67cfeaf0047d6bd47917e22c2 in this PR as well otherwise I think Puppet will try and load a configuration file which no longer exists, other than that I think it's good!

thomasleese (approved):

ChrisBAshton:
Manually tested sending an error via the console, after deploying this change to staging. It went through OK:
https://sentry.io/organizations/govuk/issues/2152180977/?project=202220

So I'm confident this change hasn't broken our Sentry integration. I'll do the same test in #10919 post-merge.

PR #10922, opened 2021-01-14
Enable Slack notifications to #govuk-deploy when Puppet deployed

ChrisBAshton:
This used to be enabled. We believe the notifications were
disabled on AWS when we had Carrenza running, as this would have
meant two notifications each time and got noisy.

### DIFF
--- a/hieradata_aws/class/production/jenkins.yaml
+++ b/hieradata_aws/class/production/jenkins.yaml
@@ -104,7 +104,7 @@ govuk_jenkins::jobs::update_cdn_dictionaries::services:
   - performanceplatform
   - www
 
-govuk_jenkins::jobs::deploy_puppet::enable_slack_notifications: false
+govuk_jenkins::jobs::deploy_puppet::enable_slack_notifications: true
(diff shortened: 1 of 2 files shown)


thomasleese (approved):
Since `enable_slack_notifications` is `true` everywhere now, I'd suggest that we can probably just remove it as an option - means the next person to look at this code doesn't need to go and check if it's `true` or `false`.

Regardless, this works, so happy to approve!

ChrisBAshton:
For the record, this change didn't re-enable Slack messages on Puppet deploys 😞 

Initially we had error messages along the lines of "invalid token", so reverted https://github.com/alphagov/govuk-secrets/pull/1104. That got us closer, but now we're seeing this in the output ([example](https://deploy.blue.staging.govuk.digital/job/Deploy_Puppet/1522/console)):

```
17:32:26 [Slack Notifications] found #1521 as previous completed, non-aborted build
17:32:26 [Slack Notifications] will send OnSuccessNotification because build matches and user preferences allow it
17:32:26 Triggering a new build of Smokey
17:32:26 Finished: SUCCESS
```

It thinks it's successfully sending the Slack message, but nothing is appearing in Slack. Needs further investigation.

PR #10924, opened 2021-01-15
Remove REDIS_HOST and REDIS_PORT env vars

kevindew:
Trello: https://trello.com/c/dXud4WIt/206-inconsistent-redisurl-config

No GOV.UK apps need these anymore as they all operate with REDIS_URL
values.


thomasleese (approved):
Looks good, we should be aware that this is going to restart all those apps when Puppet deploys.

kevindew:
> Looks good, we should be aware that this is going to restart all those apps when Puppet deploys.

Oh yes, good point 😬

PR #10925, opened 2021-01-15
Revert "Remove REDIS_HOST and REDIS_PORT env vars"

kevindew:
Reverts alphagov/govuk-puppet#10924

I'm reverting this as there are some problems with Search API and Publishing E2E tests that could be related to do this and I plan to revert Search API to still use these variables.


karlbaker02 (approved):
Looks good 👍  Hopefully we'll get to the bottom of why these changes are causing the issues we're seeing!

PR #10928, opened 2021-01-18
Pull main branch from Smokey repo

bilbof:
Smokey is changing to use 'main' rather than 'master' as its default branch.

There are no difference between the branches: https://github.com/alphagov/smokey/compare/master...main

Once this puppet change has gone out I'll change the default branch for Smokey to be 'main' in the GitHub UI.


richardTowers (approved):
Nice, go for it!

PR #10930, opened 2021-01-19
Fix race condition in updating router_nodes_file.

sengi:
The cron job which updates the file of Router nodes for Router API to contact when a route changes was un-atomically clobbering the file, fetching the new list of nodes and then writing the new list to the file. This means there's typically a half-second window every 5 minutes where Router API will see an empty or incomplete node list.

Instead, have it create a new file and atomically move it into place.

While we're there, make it check for an empty list of nodes. It turns out govuk_node_list returns 0 (success) when querying for a nonexistent node class or a node class which contains no nodes. It's probably safer not to update the file if this happens.

This should only be a short-term fix until we implement a more sensible (and simpler) config reload mechanism for Router, likely based on a database watch / pubsub / change notification type of arrangement.

Tested: ran the command locally with the variables set in the shell, verified that it creates/updates the file with the expected output when router_nodes_class is set to an actual node class and that the file doesn't get updated when router_nodes_class is set to a nonexistent node class.

I plan to test further in integration after merge, by checking that the cron job still updates the file (and I'll roll back if it's broken of course).

### DIFF
--- a/modules/govuk/manifests/apps/router_api.pp
+++ b/modules/govuk/manifests/apps/router_api.pp
@@ -105,7 +105,7 @@ class govuk::apps::router_api(
     $router_nodes_file = '/etc/router_nodes'
 
     cron::crondotdee { 'update-router-nodes':
-      command => "/usr/local/bin/govuk_node_list -c ${router_nodes_class} | sed 's/$/:${router_port}/g' > ${router_nodes_file}",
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
Looks good to me.

In case anyone's wondering what `[ -s ... ]` does, `man [` has this to say:

```
     -s file       True if file exists and has a size greater than zero.
```

... which is what we want here.

kevindew:
Amazing 👍  Thanks @sengi and @richardTowers for resolving this

sengi:
It's probably using the builtin `[`, but it'll behave the same. I guess I could have used `[[` if we know the shell is bash, but this is slightly more portable 🤷

sengi:
Works in integration with the new cronjob. File is still getting updated every 5 minutes and has the expected contents and the same mode and owner/group.

PR #10934, opened 2021-01-26
Install tfenv on AWS Integration Jenkins

fredericfran-gds:
We install tfenv on AWS Integration Jenkins so that Terraform 0.11.14 (current) and 0.13.6 (new) can co-exist on the same system.

We need Terraform 0.13.6 because some of its new features is needed for adding new functionality in govuk-aws infra-security project.

Ref:
1. [trello card](https://trello.com/c/kkFNCkop/335-give-2ndline-iam-admin-access-to-aws)

Co-Authored-By: Karl Baker <karlbaker02@users.noreply.github.com>

### DIFF
--- a/modules/govuk_jenkins/manifests/packages/tfenv.pp
+++ b/modules/govuk_jenkins/manifests/packages/tfenv.pp
@@ -1,0 +1,19 @@
+# == Class: govuk_jenkins::packages::tfenv
+#
+# Installs tfenv https://github.com/tfutils/tfenv
+# to manage multiple versions of terraform installed concurrently on the system
+#
+# === Parameters
+#
+# [*terraform_versions*]
+#   The list of terraform versions to be installed
+#
+#
+class govuk_jenkins::packages::tfenv (
+  $terraform_versions = ['0.11.14', '0.13.6'],
+){
+
+  include ::tfenv
+  ::tfenv::terraform { $terraform_versions: }
+
+}
--- a/modules/tfenv/manifests/init.pp
+++ b/modules/tfenv/manifests/init.pp
@@ -1,0 +1,44 @@
+# == Class: tfenv
+#
+class tfenv(
+  $install_dir          = '/opt/tfenv',
+  $tfenv_git_repo       = 'https://github.com/tfutils/tfenv',
+  $tfenv_revision       = 'v2.0.0',
+) {
+
+  $packages = [
+    git,
+    unzip,
+  ]
+
+  ensure_packages($packages, {'ensure' => 'present'})
+
+  file { $install_dir:
+    ensure  => directory,
+    owner   => 'jenkins',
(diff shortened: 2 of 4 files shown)


modules/tfenv/manifests/init.pp
```diff
@@ -0,0 +1,65 @@
+# == Class: tfenv
+#
+class tfenv(
+  $install_dir          = '/opt/tfenv',
+  $tfenv_git_repo       = 'https://github.com/kamatama41/tfenv',
```
richardTowers, line 5: Should this be
```suggestion
  $tfenv_git_repo       = 'https://github.com/tfutils/tfenv',
```

? Looks like it redirects.

modules/govuk_jenkins/manifests/packages/tfenv.pp
```diff
@@ -0,0 +1,26 @@
...
+#
+class govuk_jenkins::packages::tfenv (
+  $terraform_versions = ['0.11.14', '0.13.6'],
+){
+
+  class { '::tfenv':
+    manage_user    => false,
+    manage_group   => false,
+    tfenv_user     => 'jenkins',
+    tfenv_group    => 'jenkins',
```
richardTowers, line 20: Feels like we could just hardcode these in the class. It's unlikely we're going to want to run tfenv as anything other than the jenkins:jenkins. And we could remove the manage_user / manage_group bits, since they're always false in practice.

Puppetfile
```diff
@@ -27,6 +27,7 @@ mod 'saz/sudo', '3.1.0'
 mod 'saz/timezone'
 mod 'covermymeds-pgbouncer', '0.3.3'
 
+
```
richardTowers, line 30: Accidentally committed whitespace?

modules/tfenv/manifests/init.pp
fredericfran-gds: Yep, can fix that, we copied across the code contents of the puppet module: https://forge.puppet.com/modules/sergk/tfenv because of some dependencies conflict

modules/govuk_jenkins/manifests/packages/tfenv.pp
fredericfran-gds: Yep, can fix that, we copied across the code contents of the puppet module: https://forge.puppet.com/modules/sergk/tfenv because of some dependencies conflict

Puppetfile
fredericfran-gds: yep, we will fix that, it was leftover from trying to include the puppet module directly but failed due to dependencies conflicts.

richardTowers (approved):
Thanks, looks good to me.

fredericfran-gds:
had to change the origin of the git repo manually on Jenkins but now puppet is running successfully there.

PR #10938, opened 2021-02-03
Add Nginx cache for Mapit

theseanything:
This PR adds the ability to set a Ngnix cache for GOV.UK apps and enables it for the Mapit application.

Mapit as a backing service currently has no caching available and its underlying data is infrequently updated. Therefore subsequent same requests create unnecessary database look ups and overhead, which can add latency and lower capacity to user facing services such as the COVID Tiers postcode checker and the local transactions pages.

Initial investigation was done to see if cache could be done at an application level (Django) - however this involves a codebase change and is still pending with the original developers.

The query param `?no cache=true` can added to request to bypass the cache and used to make sure health checks are monitoring the underlying application rather than the cache.

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -143,6 +143,14 @@
 # possible number of "special cases". As such, use of this option is
 # discouraged, and it is included for backwards compatibility purposes only.
 #
+# [*nginx_cache*]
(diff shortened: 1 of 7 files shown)


gclssvglx (approved):
I don't fully understand all of this to be honest - but it makes sense in terms of setting the config vars.

PR #10941, opened 2021-02-08
Set golang global_version as latest version

bilbof:
Also re-adds 1.12.1 as this is used by the CSP Forwarder
https://github.com/alphagov/govuk-csp-forwarder

### DIFF
--- a/modules/golang/manifests/init.pp
+++ b/modules/golang/manifests/init.pp
@@ -11,7 +11,12 @@ class golang {
     require        => Class['govuk_ppa'],
   }
 
-  goenv::version { ['1.11.13', '1.12.1', '1.15.8']: }
+  goenv::version { [
+    '1.7.1',   # Used by alphagov/govuk_crawler_worker
+    '1.11.13', # Used by alphagov/router
+    '1.12.1',  # Used by alphagov/govuk-csp-forwarder
+    '1.15.8',  # Used by alphagov/router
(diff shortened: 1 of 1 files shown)


modules/golang/manifests/init.pp
```diff
@@ -7,11 +7,11 @@
   include govuk_ppa
 
   class { 'goenv':
-    global_version => '1.9.1',
+    global_version => '1.15.8',
```
karlbaker02, line 10: Will setting the global version to the latest have any impact on jobs etc where we're not on the latest Go?

bilbof: Yes, if those repositories don't have a `.go-version` file set. I think all repos should explicitly set a go version that they use. I looked through the govuk repos that use go, and found all set a `.go-version` file.

If there is an app that doesn't have this version installed, the reason should be obvious, they'll get an error like `goenv: version '1.15.8' is not installed`.

karlbaker02: Fair enough, sounds like a reasonable approach. For the recent work installing Terraform we opted to keep the default to what all the projects use and only set the newer version explicitly in the one project which uses it. I think here though each project is using a different version so probably better to let the user know that it fails telling them explicitly which version isn't installed 👍

karlbaker02 (approved):
Looks good

PR #10947, opened 2021-02-11
Update Mapit to use Python 3.6

brucebolt:
We [upgraded the Python used in CI to be Python 3.6](https://github.com/alphagov/govuk-puppet/commit/64fe47f435a049ac348a9ec6b296b29baa5cbe55#diff-c4a319376e182c98cdaf96574876b36ea7a793955d0fdc2499bcb72340def948R17), which is required for Mapit to run Django 3. However we didn't do this for the Mapit machines themselves.

Trello card: https://trello.com/c/rm2WTeqk


richardTowers (approved):
oops!

PR #10954, opened 2021-02-16
Fix Static ALB healthcheck

bilbof:
The Application Load Balancer health check uses
this path for static's healthcheck.

This path has been removed so we can replace it
with its replacement in the corresponding PR:
https://github.com/alphagov/static/pull/2412


beccapearce (approved):
Makes sense to me!

PR #10956, opened 2021-02-18
Add GCP icinga alert

karlbaker02:
This commit adds an Icinga check that will check daily whether the GCP
mirror sync has completed succesfully.

This check only applies to the production AWS account.

Trello: https://trello.com/c/Dr8OMcJg

### DIFF
--- a/modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
+++ b/modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+set -eou pipefail
+
+STATE_UNKNOWN=3
+
+function usage()
+{
+cat << EOF
+usage: $0 [options]
+
+This script checks the status of the latest transfer job operation which mirrors content from AWS to GCP. 
+It relies on the google-cloud-sdk to query the Google Storage Transfer Service API. This script is meant 
+for Nagios.
+
+OPTIONS:
+    -h or --help           Show this message
+
+    --google_application_credentials_file_path=x    Required: Enter the absolute file path to the Google application credentials file. This should a JSON file
+                                                    containing the private key and other credentials that can be used to assume a service account and are available
+                                                    to download upon the creation of a key pair. See https://cloud.google.com/docs/authentication/getting-started 
+                                                    for more details.
+
+    --gcp_project_id=x                              Required: Enter the GCP project id for the project which owns the Transfer Job specified in the transfer_job_name
+                                                    parameter. 
+
+    --transfer_job_name=x                           Required: Enter the Transfer Job name for the job which syncs mirror content to GCP. This can be found in the 
+                                                    configuration for the job. If the name starts with transferJobs/, omit this when entering this parameter.
+
+EOF
+}
+
+function error()
+{
+  echo -e "${1}"
+}
+
+
+#
+# Awesome parameter parsing, see https://stackoverflow.com/a/14203146
+#
+for i in "$@"; do
+case ${i} in
+  --google_application_credentials_file_path=* )
+    GOOGLE_APPLICATION_CREDENTIALS_FILE_PATH="${i#*=}"
+    shift
+    ;;
+
+  --gcp_project_id=* )
+    GCP_PROJECT_ID="${i#*=}"
+    shift
+    ;;
+
+  --transfer_job_name=* )
+    TRANSFER_JOB_NAME="${i#*=}"
+    shift
+    ;;
+
+  help | --help | -h)
+    usage
+    exit ${STATE_UNKNOWN}
+    ;;
+
+  *)
+    error "Error, unknown parameter \"${i}\" given!"
+    echo
+    usage
+    exit ${STATE_UNKNOWN}
+    ;;
+  esac
+done
+
+#
+# Validation
+#
+
+if [[ "${GOOGLE_APPLICATION_CREDENTIALS_FILE_PATH}" == "" ]]; then
+    error "You have to supply the absolute file path for the Google credentials file"
+    usage
+    exit ${STATE_UNKNOWN}
+fi
+
+if [[ "${GCP_PROJECT_ID}" == "" ]]; then
+    error "You have to supply a Google project id"
+    usage
+    exit ${STATE_UNKNOWN}
+fi
+
+if [[ "${TRANSFER_JOB_NAME}" == "" ]]; then
+    error "You have to supply a transfer job name"
+    usage
+    exit ${STATE_UNKNOWN}
+fi
+
+export GOOGLE_APPLICATION_CREDENTIALS="$GOOGLE_APPLICATION_CREDENTIALS_FILE_PATH"
+
+GCP_TRANSFER_SERVICE_TOKEN=$(gcloud auth application-default print-access-token)
+
+latest_operation_name=$(curl --header "Content-Type: application/json" \
+  --header "Authorization: Bearer $GCP_TRANSFER_SERVICE_TOKEN" \
+  --request GET "https://storagetransfer.googleapis.com/v1/transferJobs/$TRANSFER_JOB_NAME?projectId=$GCP_PROJECT_ID" \
+  | jq -r '.latestOperationName')
+
+latest_operation_details=$(curl --header "Content-Type: application/json" \
+  --header "Authorization: Bearer $GCP_TRANSFER_SERVICE_TOKEN" \
+  --request GET "https://storagetransfer.googleapis.com/v1/$latest_operation_name")
+
+status=$(echo $latest_operation_details | jq -r '.metadata.status')
+
+if [[ "$status" == "SUCCESS" ]]; then
+  finish_datetime=$(echo "$latest_operation_details" | jq -r '.metadata.endTime')
+  echo "Mirror files successfully synced to GCP at $finish_datetime"
+  exit 0
+fi
+
+echo "A non-success result was returned - see GCP Storage Transfer Service console for more information."
+echo "Failed operation: $latest_operation_name"
+exit 1
--- a/modules/monitoring/manifests/checks/mirror.pp
+++ b/modules/monitoring/manifests/checks/mirror.pp
@@ -8,14 +8,56 @@
 #   Can be used to disable checks/alerts for a given environment.
 #   Default: false
 #
+# [*gcp_mirror_sync_project_id*]
+#   Sets the GCP project id to use for the transfer job.
+#   Required if enabled=true.
+#
+# [*gcp_mirror_sync_transfer_job_auth_json*]
+#   Sets the auth JSON required by the Google SDK.
+#   Required if enabled=true.
+#
+# [*gcp_mirror_sync_transfer_job_name*]
+#   Sets the transfer job name which controls the mirror sync.
+#   Required if enabled=true.
+#
 class monitoring::checks::mirror (
   $enabled = false,
+  $gcp_mirror_sync_project_id = undef,
+  $gcp_mirror_sync_transfer_job_auth_json = undef,
+  $gcp_mirror_sync_transfer_job_name = undef,
 ) {
   icinga::check_config { 'mirror_age':
     source => 'puppet:///modules/monitoring/etc/nagios3/conf.d/check_mirror_age.cfg',
   }
 
   if $enabled {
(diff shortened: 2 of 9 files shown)


modules/monitoring/manifests/checks/mirror.pp
```diff
@@ -8,13 +8,57 @@
 #   Can be used to disable checks/alerts for a given environment.
 #   Default: false
 #
+# [*gcp_mirror_sync_project_id*]
+#   Sets the GCP project id to use for the transfer job.
+#   Default: ''
```
sengi, line 13: This contradicts the actual code. You've got these defaulting to `undef` (not empty string), which makes sense because you presumably want them to be optional so that they can be omitted when `enabled = false` but still raise an error if they're not set when they're needed.
```suggestion
#   Required if enabled=true.
```

(similarly for the others)

modules/monitoring/manifests/checks/mirror.pp
```diff
@@ -8,13 +8,57 @@
...
   $enabled = false,
+  $gcp_mirror_sync_project_id = undef,
+  $gcp_mirror_sync_transfer_job_auth_json = undef,
+  $gcp_mirror_sync_transfer_job_name = undef,
 ) {
   icinga::check_config { 'mirror_age':
     source => 'puppet:///modules/monitoring/etc/nagios3/conf.d/check_mirror_age.cfg',
   }
 
+  if ($::aws_migration and $::aws_environment == 'production') {
```
sengi, line 33: Does this conditional serve any purpose? Why not just use `$enabled` and merge it with the block below? Have I missed something?

modules/monitoring/manifests/checks/mirror.pp
```diff
@@ -8,13 +8,57 @@
...
+    }
+
+    icinga::check_config { 'mirror_file_sync':
+      source  => 'puppet:///modules/monitoring/etc/nagios3/conf.d/check_mirror_file_sync.cfg',
+      require => File[$google_application_credentials_file_path],
+    }
+
+    icinga::check { 'check_mirror_file_sync':
+      check_command       => "check_mirror_file_sync!${google_application_credentials_file_path}!${gcp_mirror_sync_project_id}!${gcp_mirror_sync_transfer_job_name}",
+      service_description => 'Latest GCP mirror file sync failed',
```
sengi, line 55: The service description shouldn't say that something's failed even when the check succeeds.
```suggestion
      service_description => 'Check status of latest GCP mirror sync job',
```

modules/monitoring/manifests/init.pp
```diff
@@ -31,6 +31,10 @@
     include monitoring::uptime_collector
   }
 
+  if $::aws_environment == 'production' {
+    include monitoring::gcloud
+  }
+
```
sengi, line 37: This conditional is unnecessary since you have `enabled` defaulting to false, so the presence/absence of the alert is already neatly controlled by setting `enabled` in Hiera in production.

sengi (approved):
Mostly just style nits, only one functional issue which is I think you can ditch the "if prod" conditionals and just use `enabled=true` in Hiera to say which environments you want to have the alert (please fix that one if you can, because it's potentially confusing).

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
```diff
@@ -0,0 +1,124 @@
...
+                                                    configuration for the job. If the name starts with transferJobs/, omit this when entering this parameter.
+
+EOF
+}
+
+function error()
+{
+  RED='\033[0;31m'
+  NC='\033[0m' # No Color
+  echo -e "${RED}${1}${NC}";
```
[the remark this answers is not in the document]
smford: Good point, I'll remove this pizazz because it's not neccessary.

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
```diff
@@ -0,0 +1,124 @@
...
+if [ $status != "SUCCESS" ]; then
+  echo "A non-success result was returned - see GCP Storage Transfer Service console for more information."
+  echo "Failed operation: $latest_operation_name"
+  exit 1
+else
+  finish_datetime=$(echo $latest_operation_details | jq -r '.metadata.endTime')
+
+  echo "Mirror files successfully synced to GCP at $finish_datetime"
+  exit 0
+fi
```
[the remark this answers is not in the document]
smford: Excellent point, thanks!

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
```diff
@@ -0,0 +1,124 @@
...
+    ;;
+  esac
+done
+
+#
+# Validation
+#
+
+if [[ "${GOOGLE_APPLICATION_CREDENTIALS_FILE_PATH}" == "" ]];
+then
```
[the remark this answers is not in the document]
smford: Thanks, made consistent.

modules/monitoring/manifests/checks/mirror.pp
smford: You are correct, moving it to use $enabled

modules/monitoring/manifests/checks/mirror.pp
smford: Cheers, updated.

modules/monitoring/manifests/init.pp
smford: My reasoning for doing it this way was to only allow the gcloud sdk to be installed on the machines that really need it, and reducing the size of the puppet catalog.  By having it here, without the conditional, it will install on all monitoring boxes.  I think an alternative could be moving the install of the packages to within the check, which would mean it gets captured by the $enable.

This gcloud sdk is also installed here: modules/govuk_ci/manifests/agent/gcloud.pp.  There is certainly some improvement that can be done around this and I shall ponder it over night.

sengi: Ohh sorry, I misread this in a very silly way (as if it said `monitoring::checks::mirror` rather than `monitoring::gcloud`) and mixed it up with the earlier comment about `enabled`. So I was talking complete nonsense here 🙈 Let's put it down to a year of isolation and hope I'm not just losing my marbles 😅 

No big deal then but I wonder if it's worth the small cost of installing gcloud in all the environments if only to keep non-prod similar to prod so there's a chance of catching any future breakage? I'm fine with whatever you think best on this.

smford: I agree with the small cost of installing on all environments.  Cheers for the feedback.

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env bash
+set -eou pipefail
+
+STATE_UNKNOWN = 3
```
sengi, line 4: from shellcheck
```suggestion
STATE_UNKNOWN=3
```

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
```diff
@@ -0,0 +1,118 @@
...
+latest_operation_name=$(curl --header "Content-Type: application/json" \
+  --header "Authorization: Bearer $GCP_TRANSFER_SERVICE_TOKEN" \
+  --request GET "https://storagetransfer.googleapis.com/v1/transferJobs/$TRANSFER_JOB_NAME?projectId=$GCP_PROJECT_ID" \
+  | jq -r '.latestOperationName')
+
+latest_operation_details=$(curl --header "Content-Type: application/json" \
+  --header "Authorization: Bearer $GCP_TRANSFER_SERVICE_TOKEN" \
+  --request GET "https://storagetransfer.googleapis.com/v1/$latest_operation_name")
+
+status=$(echo $latest_operation_details | jq -r '.metadata.status')
```
sengi, line 108: from shellcheck
```suggestion
status=$(echo "$latest_operation_details" | jq -r '.metadata.status')
```

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
```diff
@@ -0,0 +1,118 @@
...
+
+if [[ "$status" == "SUCCESS" ]]; then
+  finish_datetime=$(echo $latest_operation_details | jq -r '.metadata.endTime')
```
sengi, line 111:
```suggestion
  finish_datetime=$(echo "$latest_operation_details" | jq -r '.metadata.endTime')
```

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
```diff
@@ -0,0 +1,118 @@
...
+  help | --help | -h)
+    usage ;
+    exit 3;
+    ;;
+
+  *)
+      error "Error, unknown parameter \"${i}\" given!";
+      echo "";
+    usage ;
+    exit 3;
```
sengi, line 68: (remove superfluous semicolons, use STATE_UNKNOWN like elsewhere in the file, remove empty arg to echo, fix indentation on the `*` case.)

```suggestion
for i in "$@"; do
case ${i} in
  --google_application_credentials_file_path=* )
    GOOGLE_APPLICATION_CREDENTIALS_FILE_PATH="${i#*=}"
    shift
    ;;

  --gcp_project_id=* )
    GCP_PROJECT_ID="${i#*=}"
    shift
    ;;

  --transfer_job_name=* )
    TRANSFER_JOB_NAME="${i#*=}"
    shift
    ;;

  help | --help | -h)
    usage
    exit ${STATE_UNKNOWN}
    ;;

  *)
    error "Error, unknown parameter \"${i}\" given!"
    echo
    usage
    exit ${STATE_UNKNOWN}
```

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
```diff
@@ -0,0 +1,118 @@
...
+    error "You have to supply a Google project id";
+    usage;
+    exit ${STATE_UNKNOWN};
+fi;
+
+if [[ "${TRANSFER_JOB_NAME}" == "" ]]; then
+    error "You have to supply a transfer job name";
+    usage;
+    exit ${STATE_UNKNOWN};
+fi;
```
sengi, line 93: (remove unnecessary semicolons)
```suggestion
if [[ "${GOOGLE_APPLICATION_CREDENTIALS_FILE_PATH}" == "" ]]; then
    error "You have to supply the absolute file path for the Google credentials file"
    usage
    exit ${STATE_UNKNOWN}
fi

if [[ "${GCP_PROJECT_ID}" == "" ]]; then
    error "You have to supply a Google project id"
    usage
    exit ${STATE_UNKNOWN}
fi

if [[ "${TRANSFER_JOB_NAME}" == "" ]]; then
    error "You have to supply a transfer job name"
    usage
    exit ${STATE_UNKNOWN}
fi
```

sengi (approved):
Bit more syntax cleanup (sorry to be picky but let's set a good example!) Hopefully these are suitable for just straightforward applying.

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
smford: Thanks!

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
smford: Thanks!

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
smford: Thanks!

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
smford: Thanks!

modules/monitoring/files/usr/lib/nagios/plugins/check_mirror_file_sync
smford: Thanks!

PR #10957, opened 2021-02-18
Revert "Add Nginx cache for Mapit"

theseanything:
Nginx caching was originally added as quick way to cache Mapit, whilst work to [fix Mapit caching to memcached](https://github.com/mysociety/mapit/pull/373) was being introduced upstream. This fix has now been adopt and Mapit now caches responses in Memcached. This means the Nginx caches a redundant and can be removed to simply our caching set up. It is also preferable to use Memcached over Ngnix, as it allows us to use shared cache across all instance of Mapit. Reverts alphagov/govuk-puppet#10938

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -143,14 +143,6 @@
 # possible number of "special cases". As such, use of this option is
 # discouraged, and it is included for backwards compatibility purposes only.
 #
-# [*nginx_cache*]
-# configuration to enable caching within ngnix
-#
-# This parameter can be used to set up a nginx cache and takes a hash with
-# values for the following keys: `key_name`, `key_size`, `max_age` and
-# `max_size`. Use of this option is discouraged.
-#
-# Default: {}
 #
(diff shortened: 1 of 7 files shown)


kevindew (commented):
Nice one 👍 

I wonder if we can also remove https://github.com/alphagov/govuk-puppet/blob/8445b2f6c2e4c453cdcd6dd2ea9522a99ad2ee18/modules/govuk/manifests/node/s_mapit.pp#L23-L27 since we won't be using memcache on the machines anymore

kevindew (approved):

theseanything:
> Nice one 👍
> 
> I wonder if we can also remove
> 
> https://github.com/alphagov/govuk-puppet/blob/8445b2f6c2e4c453cdcd6dd2ea9522a99ad2ee18/modules/govuk/manifests/node/s_mapit.pp#L23-L27
> 
> since we won't be using memcache on the machines anymore

Yup - good shout! I'll do that in a separate PR, so this PR is just the revert.

PR #10959, opened 2021-02-18
Add shared elasticache location for Mapit

theseanything:
This adds a "MEMCACHE_SERVERS" environment variable to point Mapit it at a shared elasticached instance. This should double caching efficiency as we currently cache per instance. It will also allow us to pre-warm our cache more quickly without having to warm multiple caches.


kevindew (approved):
Nice one 👍

PR #10963, opened 2021-02-23
Add production check for Whitehall unenqueued scheduled documents

brucebolt:
In alphagov/whitehall#6001, we added an endpoint to check for the number of scheduled editions that have not been queued in Sidekiq.

We already have this data included in the Whitehall app check, but this is shown in Icinga for every host (rather than once  
suggests the entire app is critical, rather than one part. This will be removed later to ensure continuity.

Trello card: https://trello.com/c/UAbbL13x

### DIFF
--- a/modules/monitoring/manifests/checks.pp
+++ b/modules/monitoring/manifests/checks.pp
@@ -18,6 +18,7 @@ class monitoring::checks (
   $http_username                  = 'UNSET',
   $http_password                  = 'UNSET',
   $whitehall_overdue_check_period = undef,
+  $whitehall_scheduled_check_period = undef,
 ) {
 
   ensure_packages(['jq'])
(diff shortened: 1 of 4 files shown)


modules/monitoring/manifests/checks.pp
```diff
@@ -81,6 +83,26 @@
...
+    icinga::check { "check_whitehall_scheduled_from_${::hostname}":
+      check_command              => 'check_whitehall_scheduled',
+      service_description        => 'scheduled publications in Whitehall not queued',
+      use                        => 'govuk_urgent_priority',
+      host_name                  => $::fqdn,
+      notes_url                  => monitoring_docs_url(whitehall-scheduled-publishing),
+      action_url                 => "https://${whitehall_hostname}${whitehall_scheduled_url}",
+      check_period               => $whitehall_scheduled_check_period,
+      attempts_before_hard_state => 10,
+      retry_interval             => 1,
```
1pretz1, line 101: Is this interval in minutes or seconds?

1pretz1: Ah - I assume minutes:

```
Unless you've changed the interval_length directive from the default value of 60, this number will mean minutes.
```

brucebolt: I just copied it from the block above 😝.  Based on [this comment](https://github.com/alphagov/govuk-puppet/blob/b0bd7e2acbecfc51e12d0c7a547b7f8af03ca844/modules/monitoring/manifests/checks/reboots.pp#L21), I would assume it's in minutes.

1pretz1: Nice 👍 10 mins seems reasonable.

1pretz1 (approved):

PR #10965, opened 2021-03-01
Fix wrong MIME type being aplied to css files during mirroring

rtrinque:
This is a workaround for a bug in s3cmd :
https://github.com/s3tools/s3cmd/issues/643


fredericfran-gds (approved):
looks good

PR #10966, opened 2021-03-01
Clean up references to disused gds-operations GitHub org.

sengi:
- All the non-archived repos in github.com/gds-operations have been transferred to github.com/alphagov (including `puppet-aptly` and `ubuntu_unused_kernels`) so update refs to point at alphagov for those.
- The `vcloud-*` repos were archived a long time ago, so just remove refs to those.

[GitHub's transfer process](https://docs.github.com/en/github/administering-a-repository/transferring-a-repository) preserves pretty much everything and even redirects the old URLs, so this change is really just for tidiness rather than a necessity. Everything's already working fine with the redirects. I've checked that the committish refs here are still valid on the transferred repos.


bilbof (approved):
LGTM

PR #10971, opened 2021-03-04
Fix alerts for govuk_mirror_sync.

sengi:
* Remove alerts for the long-since-defunct Carrenza mirrors.
* Promote the alert which actually does tell us when there's a problem with the mirrors to `critical` severity, because it (entirely understandably) gets ignored as a warning but it definitely requires action.

See commit messages for more detail.


sengi:
Force-pushed to squash a fixup commit into 81f82ee to delete the tests which were made obsolete by removing the old alerts.

sengi:
Realised I can override the "Freshness threshold exceeded" message to say something more helpful, so I did.

richardTowers (approved):

PR #10972, opened 2021-03-04
Add healthcheck route for Sidekiq monitoring

1pretz1:
Related PR:
https://github.com/alphagov/sidekiq-monitoring/pull/71

Trello:
https://trello.com/c/C1cWKFAb/2392-5-enable-continuous-deployment-for-sidekiq-monitoring

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -321,6 +321,8 @@ task :check_consistency_between_aws_and_carrenza do
     govuk::apps::sidekiq_monitoring::link_checker_api_redis_port
     govuk::apps::sidekiq_monitoring::publishing_api_redis_host
     govuk::apps::sidekiq_monitoring::publishing_api_redis_port
+    govuk::apps::sidekiq_monitoring::redis_host
+    govuk::apps::sidekiq_monitoring::redis_port
     govuk::apps::sidekiq_monitoring::search_api_redis_host
     govuk::apps::sidekiq_monitoring::search_api_redis_port
     govuk::apps::sidekiq_monitoring::support_api_redis_host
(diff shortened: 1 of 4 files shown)


edwardkerry (approved):

brucebolt (approved):

Rakefile
```diff
@@ -321,6 +321,8 @@ task :check_consistency_between_aws_and_carrenza do
     govuk::apps::sidekiq_monitoring::link_checker_api_redis_port
     govuk::apps::sidekiq_monitoring::publishing_api_redis_host
     govuk::apps::sidekiq_monitoring::publishing_api_redis_port
+    govuk::apps::sidekiq_monitoring::redis_host
+    govuk::apps::sidekiq_monitoring::redis_port
```
ChrisBAshton, line 325: I've not looked into this much, but is this a legacy way of configuring Redis?

(Based on several Dependabot PRs that reference this commit: https://github.com/alphagov/govuk_sidekiq/commit/85c0cb87605d37ae94f7c98e254197c360f6ef53)

...TLDR should we be setting REDIS_URL instead?

1pretz1: Thanks for the comment @ChrisBAshton!

The end result is a `REDIS_URL` ENV variable being set, which is [prefixed](https://github.com/alphagov/sidekiq-monitoring/blob/master/config.ru#L7) for Sidekiq Monitroing instead of the app calling upon `REDIS_HOST` and `REDIS_PORT` which as you say, is legacy. The [spec](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/spec/defines/govuk__app__envvar__redis_spec.rb#L12) for the [Redis envar](https://github.com/alphagov/govuk-puppet/pull/10972/files#diff-db708426d4a7957a7fa0e18d1754a307c06decac126e898a1a7021b2e32caaabR245-R249) called probably illustrates how this is set so I think we're okay 👍 

There is a question on why we seem to have so many puppet variables, per app which set the Redis host and port individually when they all seem to use the same default values, however I thought this could be cleaned up later, as a separate PR.

ChrisBAshton: Makes sense, thanks.

ChrisBAshton (approved):

PR #10973, opened 2021-03-04
Make gcp mirror sync script resilient

smford:
This PR gets the GCP mirror transfer job name from the Storage Transfer API. The reason for this change is because when we change the Terraform for the job and redeploy it, a new Transfer Job is created by GCP, meaning we'd have to go and change secrets to the new job name; this isn't sustainable and instead a better way is to look up the job when we do the Icinga check.


karlbaker02 (approved):
Paired on this, looks good 👍

PR #10984, opened 2021-03-10
Remove Stephen H and Tim B as users

deborahchua:
They have left GOV.UK

Trello cards:
https://trello.com/c/djydeAa7/1099-leaver-tim-blair-%F0%9F%98%AD
https://trello.com/c/WyTZ6obE/1100-leaver-stephen-harker


richardTowers (approved):
Thank you for picking this up!

PR #10985, opened 2021-03-10
Add account-api

AlanGabbianelli:
This is to prepare the servers for account-api. I followed [govuk-puppet's documentation][0] as suggested in our [Set up a new Rails application][1].

[0]: https://github.com/alphagov/govuk-puppet/blob/master/docs/adding-a-new-app.md#including-the-app-on-machines
[1]: https://docs.publishing.service.gov.uk/manual/setting-up-new-rails-app.html#puppet-dns-sentry-and-beyond

Trello card: https://trello.com/c/ZeAKl2SM/651-set-up-a-new-app

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -12,7 +12,8 @@ backup::mysql::alert_hostname: 'alert'
 
 node_class: &node_class
   account:
-    apps: []
+    apps:
+      - account-api
   asset_master:
     apps:
       - asset_env_sync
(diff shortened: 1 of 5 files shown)


hieradata_aws/common.yaml
```diff
@@ -354,6 +356,7 @@ gdal::repo::apt_mirror_hostname: "%{hiera('apt_mirror_hostname')}"
 gdal::repo::apt_mirror_gpg_key_fingerprint: "%{hiera('apt_mirror_fingerprint')}"
 
 govuk::apps::publisher::port: 3000
+govuk::apps::account_api::port: 3001
```
barrucadu, line 359: I'm a bit suspicious of port 3001 being unused when 3000 and 3002 are both taken.  How about using the highest listed port + 1, which would be 3243?

One more thing to look forward to after replatforming: not needing to worry about unique ports.

AlanGabbianelli: It sounds sensible, I'll change that.
Although, more transparency on what ports we should and shouldn't use would be useful. 😄

barrucadu (approved):

barrucadu (approved):

PR #10993, opened 2021-03-12
Add account-manager env vars to account-api

barrucadu:
[Trello card](https://trello.com/c/ZeAKl2SM/651-set-up-a-new-app)

### DIFF
--- a/hieradata_aws/integration.yaml
+++ b/hieradata_aws/integration.yaml
@@ -10,6 +10,7 @@ cron::daily_hour: 6
 
 environment_ip_prefix: '10.1'
 
+govuk::apps::account_api::plek_account_manager_uri: 'https://www.account.staging.publishing.service.gov.uk'
 govuk::apps::asset_manager::aws_s3_bucket_name: 'govuk-assets-integration'
 govuk::apps::asset_manager::aws_region: 'eu-west-1'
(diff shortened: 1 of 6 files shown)


hieradata_aws/integration.yaml
```diff
@@ -10,6 +10,7 @@ cron::daily_hour: 6
 
 environment_ip_prefix: '10.1'
 
+govuk::apps::account_api::plek_account_manager_uri: 'https://www.account.staging.publishing.service.gov.uk'
```
huwd, line 13: Should this be staging or integration?

huwd (changes requested):

hieradata_aws/integration.yaml
huwd:
```suggestion
govuk::apps::account_api::plek_account_manager_uri: 'https://www.account.integration.publishing.service.gov.uk'
```

barrucadu: This should actually be staging, as we're not running a separate integration account manager - see also finder-frontend's hieradata

huwd: ok ace, will approve

huwd (approved):

PR #10994, opened 2021-03-12
Enable account-api in staging and production

barrucadu:
These two override the node classes, rather than inheriting from
common.yaml.  So in fact the "common" value is only used for
integration.

Well, this explains why we don't have an alerts for a non-running
account-api in those environments...

---

[Trello card](https://trello.com/c/ZeAKl2SM/651-set-up-a-new-app)


AlanGabbianelli (approved):
> These two override the node classes, rather than inheriting from
common.yaml. So in fact the "common" value is only used for
integration.

That's a bit confusing, thanks for making it clear in the commit message. ⭐

PR #11002, opened 2021-03-22
Stop setting the router_nodes_class variable

karlbaker02:
This PR stops the setting of the `router_nodes_class` variable for the `draft_cache` and `router_backend` machine classes. Router is being changed to self-update routes from MongoDB rather than having Router API inform it when to update routes; no longer setting these variables in Puppet is part of a wider set of changes to allow this to happen.

### DIFF
--- a/hieradata_aws/class/draft_cache.yaml
+++ b/hieradata_aws/class/draft_cache.yaml
@@ -20,8 +20,6 @@ govuk::apps::router::mongodb_nodes:
   - 'router-backend-2'
   - 'router-backend-3'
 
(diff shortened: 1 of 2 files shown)


sengi (approved):
LGTM - it looks like the only references to this var are in modules/govuk/manifests/apps/router_api.pp, where it defaults to `undef` and doesn't try to do anything with it unless it evaluates true, so this should have the desired effect 👍

karlbaker02:
The alternative option to setting the Router nodes `router_nodes` (specified in the same modules/govuk/manifests/apps/router_api.pp) looks like it's never set, so agree with your analysis 👍

PR #11004, opened 2021-03-23
Delete hieradata and replace it with hieradata_aws

rtrinque:
This is related to https://github.com/alphagov/govuk-secrets/pull/1123
Following the decommission of our Carrenza account we don't need
the discrimination between AWS and carrenza anymore.
This moves all references to hiera_aws to the generic hiera/hieradata.


sengi (approved):
Thanks for keeping the commits separate here - would be super hard to review otherwise. And thanks for taking on this cleanup. Great to see stuff being deleted from govuk-puppet!

I see you have a bit of a rollout plan in alphagov/govuk-secrets#1123 - makes sense. I guess there's not a lot else we can do in terms of testing this besides the usual check-it-doesn't-break-in-integration/staging?

PR #11005, opened 2021-03-24
Increase queue size for Carbon Cache/Relay/Aggregator

smford:
We are seeing warnings that say the carbon aggregator queue is hitting
its maximum size, initiating a backoff, and and causing stats to be
discarded from graphite.  This PR increases the queue size between for
cache, relay and aggregator whilst also changes the low water
threshold, which should prevent the backoff mechanism from initiating as
quickly.

It is unclear if these values are appropriate and they will likely need
tweaking as we get a better sense of volume of stats.

```
tail -n 2 /var/log/upstart/carbon-aggregator.log
23/03/2021 15:08:52 :: [clients] CarbonClientFactory(127.0.0.1:2024:None) send queue is full (10000 datapoints)
23/03/2021 15:08:52 :: [clients] CarbonClientProtocol(127.0.0.1:2024:None) send queue has space available
```


rtrinque (approved):
Looks good

PR #11007, opened 2021-03-24
Add cron jobs to run specific data.gov.uk harvesters

brucebolt:
These are not running correctly in the automated runs, so are currently being run manually every day/week. This removes the need to do this until we are able to fix the underlying problem.

Trello card: https://trello.com/c/HoGDPOpn

### DIFF
--- a/modules/govuk/manifests/apps/ckan/cronjobs.pp
+++ b/modules/govuk/manifests/apps/ckan/cronjobs.pp
@@ -30,6 +30,41 @@ class govuk::apps::ckan::cronjobs(
     false => 'absent',
   }
 
+  govuk::apps::ckan::paster_cronjob { 'Find a Tender harvester run':
+    ensure         => $ensure,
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/ckan/cronjobs.pp
```diff
@@ -30,6 +30,41 @@
     false => 'absent',
   }
 
+  govuk::apps::ckan::paster_cronjob { 'Find a Tender harvester run':
+    ensure         => $ensure,
+    paster_command => 'harvester run_test find-tender-data-harvest',
+    plugin         => 'ckanext-harvest',
+    hour           => '7',
```
deborahchua, line 37: Sorry @brucebolt I forgot to update the card, the user has specified they'd like this harvester to run daily at 2.30am

brucebolt: No problem, I've changed the time

deborahchua (approved):

PR #11010, opened 2021-03-25
Stop existing harvesters before starting new process

brucebolt:
In eead57eab4d784122e3087b8e0fe35aa16053a16, we added cron jobs to run specific data.gov.uk harvesters at regular intervals, whilst the underlying problem is investigated. These did not run because there was an existing stalled harvest run marked as running.

This will force the stalled harvest jobs to stop 5 minutes before the scheduled run.

Also adds a cron job for the Contracts Finder, which has been requested by Crown Commercial Service.

Trello card: https://trello.com/c/HoGDPOpn

### DIFF
--- a/modules/govuk/manifests/apps/ckan/cronjobs.pp
+++ b/modules/govuk/manifests/apps/ckan/cronjobs.pp
@@ -46,6 +46,22 @@ class govuk::apps::ckan::cronjobs(
     minute         => '30',
   }
 
+  govuk::apps::ckan::paster_cronjob { 'Contract Finder harvester stop existing processes':
+    ensure         => $ensure,
+    paster_command => 'harvester job_abort contracts-finder-ocds-data-harvest',
+    plugin         => 'ckanext-harvest',
+    hour           => '5',
+    minute         => '25',
+  }
+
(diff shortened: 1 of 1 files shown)


1pretz1 (approved):
Changes look good 👍 

As part of this could we also update the harvesters that are running via cron to be 'manually' updated from within CKAN. I noticed most of the harvesters concerned are all ready configured this way apart from the [Contracts finder ocds](https://ckan.publishing.service.gov.uk/harvest/edit/contracts-finder-ocds-data-harvest) and [Find tender data](https://ckan.publishing.service.gov.uk/harvest/edit/find-tender-data-harvest) which are configured to update daily which may explain the stalled jobs.

brucebolt:
> As part of this could we also update the harvesters that are running via cron to be 'manually' updated from within CKAN? I noticed most of the harvesters concerned are all ready configured this way, apart from the [Contracts finder ocds](https://ckan.publishing.service.gov.uk/harvest/edit/contracts-finder-ocds-data-harvest) and [Find tender data](https://ckan.publishing.service.gov.uk/harvest/edit/find-tender-data-harvest) which are configured to update daily which may explain the stalled jobs.

That's a good idea. I'll change these to manual.

PR #11016, opened 2021-03-30
Add search and metrics healthchecks for content-data-api

brucebolt:
In https://github.com/alphagov/content-data-api/pull/1595, we added `/healthcheck/search` and `/healthcheck/metrics` endpoints which will allow us to decouple application metrics from instance metrics.  As these metrics do not indicate a healthy instance, they should not be used in load balancing (which they are at present).

This adds these as alerts in Icinga, so we can soon remove the search and metrics from the main application healthcheck.

Trello card: https://trello.com/c/X2AKAUKn and https://trello.com/c/aDoOuYf5

### DIFF
--- a/modules/monitoring/manifests/checks.pp
+++ b/modules/monitoring/manifests/checks.pp
@@ -19,6 +19,7 @@ class monitoring::checks (
   $http_password                  = 'UNSET',
   $whitehall_overdue_check_period = undef,
   $whitehall_scheduled_check_period = undef,
+  $content_data_api_check_period  = undef,
 ) {
 
   ensure_packages(['jq'])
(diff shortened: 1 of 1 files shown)


modules/monitoring/manifests/checks.pp
```diff
@@ -58,6 +59,36 @@
 
   include icinga::plugin::check_http_timeout_noncrit
 
+  $content_data_api_hostname    = "content-data-api.${app_domain}"
+  $content_data_api_metrics_url = '/healthcheck/metrics'
+  $content_data_api_search_url = '/healthcheck/search'
```
ChrisBAshton, line 64: Is it worth renaming these from `_url` to `_path`?

modules/monitoring/manifests/checks.pp
```diff
@@ -58,6 +59,36 @@
...
+
+  if $::aws_migration {
+    include icinga::client::check_json_healthcheck
+
+    icinga::check { "content_data_api_search_${::hostname}":
+      check_command              => "check_app_health!check_json_healthcheck!443 ${content_data_api_search_url} ${content_data_api_hostname} https",
+      service_description        => 'search healthcheck for content-data-api',
+      use                        => 'govuk_urgent_priority',
+      host_name                  => $::fqdn,
+      notes_url                  => monitoring_docs_url(content-data-api-app-healthcheck-not-ok),
```
ChrisBAshton, line 74: I guess this makes sense for now: https://docs.publishing.service.gov.uk/manual/alerts/content-data-api-app-healthcheck-not-ok.html

Should we have a follow-up PR here and in the Dev Docs to split out into two notes_url pages?

ChrisBAshton (approved):
LGTM 👍 no blocking suggestions.

modules/monitoring/manifests/checks.pp
brucebolt: Yes, that seems like a good idea 👍

modules/monitoring/manifests/checks.pp
brucebolt: Indeed.  I'd planned to update the documentation as part of https://trello.com/c/Wn4w66dE, prior to removing the alerts from the instance healthcheck.

PR #11024, opened 2021-04-07
Enable continuous deployment for maslow

barrucadu:
Depends on https://github.com/alphagov/smokey/pull/810 https://github.com/alphagov/maslow/pull/689 https://github.com/alphagov/govuk-saas-config/pull/90


bilbof (approved):
LGTM

PR #11028, opened 2021-04-07
Remove db_allow_prepared_statements

barrucadu:
I think this was originally introduced for pgbouncer.  We're not using
pgbouncer in AWS, and so this parameter was unset for every app.  It's
a bunch of configuration doing nothing, so let's delete it.


kevindew (approved):
Nice one, great clean up.

PR #11034, opened 2021-04-12
Remove app memory alerts

MuriloDalRi:
We often see "high memory for * app" errors on 2nd line.
These alerts aren't actionable and don't usually indicate any user-facing
problems. So long as there is enough memory on the machine, it isn't an
issue if an app is using "lots" of it.

https://trello.com/c/nfzw7FvL/2458-remove-app-memory-alerts-2


ChrisBAshton (approved):
This looks great 👍

PR #11037, opened 2021-04-13
Delete hieradata directory

rtrinque:
Following the decommission of Carrenza we remove dead code


fredericfran-gds (approved):
looks good, there is a conflict for some reason with master, need rebase

PR #11038, opened 2021-04-13
Enable larger attachments for Specialist Publisher

bilbof:
In addition to https://github.com/alphagov/specialist-publisher/pull/1847 this will enable users to upload larger attachments, who are currently seeing Nginx timeouts.

Ticket: https://govuk.zendesk.com/agent/tickets/4515218

This should save us time [manually scp-ing large files](https://docs.publishing.service.gov.uk/manual/manage-assets.html#large-attachments) to production for users.

It takes about 24s for Specialist Publisher to process the 250mb file on my fast connection (2.2mins in total, since Nginx buffers the file before proxying the request). For a good user experience we should probably add a note to the app saying 250mb is the max permitted size, since we'll likely see timeouts on larger files.

### DIFF
--- a/modules/govuk/templates/specialist_publisher_nginx_extra.conf.erb
+++ b/modules/govuk/templates/specialist_publisher_nginx_extra.conf.erb
@@ -1,0 +1,7 @@
+client_max_body_size 500m;
+
+# Extended timeout to allow the processing of large attachments
+location ~* attachments {
+  proxy_read_timeout 30s;
+  proxy_pass http://specialist-publisher-proxy;
+}
--- a/modules/govuk/manifests/apps/specialist_publisher.pp
+++ b/modules/govuk/manifests/apps/specialist_publisher.pp
@@ -131,7 +131,7 @@ class govuk::apps::specialist_publisher(
     health_check_path  => '/healthcheck',
     json_health_check  => true,
     log_format_is_json => true,
-    nginx_extra_config => 'client_max_body_size 500m;',
+    nginx_extra_config => template('govuk/specialist_publisher_nginx_extra.conf.erb'),
     asset_pipeline     => true,
   }
 


modules/govuk/templates/specialist_publisher_nginx_extra.conf.erb
```diff
@@ -0,0 +1,8 @@
+client_max_body_size 500m;
+
+# Permit large attachments
+location ~* attachments {
+  proxy_read_timeout 5m;
+  client_body_timeout 5m;
```
kevindew, line 6: Do you know if these are actually needed?

I'm no expert but I have a recollection that nginx buffers the uploaded file before proxying to the app and you don't really need any changes to the proxying config for the file upload time. You instead just need to consider how long the server spends dealing with the file - to which 5 minutes is an eternity. 

For Content Publisher we just have: https://github.com/alphagov/govuk-puppet/blob/baf3cf67291cdc4527d90cffc1f4c3d640fd67e9/modules/govuk/manifests/apps/content_publisher.pp#L148

bilbof: Thanks for this, you're right, 5min is excessive! From Chrome devtools I'm seeing that uploading a 250mb attachment takes 2.2mins, of which 1.8mins is sending the request (which I suppose is Nginx buffering the file?), and then 23.39s is spent waiting (which is where the read timeout takes effect).

The `client_body_timeout` is by default 60s, so this doesn't need overriding here - I've removed that.

A longer `proxy_read_timeout` is necessary to resolve this, but 30s will do the trick rather than 5min 😬  I've amended the commit.

kevindew: Great that sounds reasonable, bit of a shame it can take so long to generate a response but that's a much broader concern.

Just a minor suggestion would be to amend the comment of `# Permit large attachments` to instead state we're extending the timeout to allow the processing of large attachments, since permit is probably the wrong word for the situation. I don't think you'll need another review on that though.

kevindew (approved):

PR #11044, opened 2021-04-19
Improve errors when the name of the app is wrong

richardTowers:
Error messages for this use case used to say:

```
cd: can't cd to /var/apps/content_store
```

Now they'll say:

```
Could not find application content_store. Are you sure you're looking at the right machine?
The applications on this machine are:

...
content-store
...
```


thomasleese (approved):
Nice improvement!

PR #11046, opened 2021-04-20
Add a check for publishing-api downstream_high latency

barrucadu:
This is to replace a check in the app:
https://github.com/alphagov/publishing-api/blob/1f595a4bc6693cc8c44b410bcddc2c1f07fb3f9a/app/models/healthcheck/queue_latency.rb

---

[Trello card](https://trello.com/c/tl6RV1el/723-improve-govuk-healthchecks)

### DIFF
--- a/modules/monitoring/manifests/checks/sidekiq.pp
+++ b/modules/monitoring/manifests/checks/sidekiq.pp
@@ -8,10 +8,14 @@
 # [*enable_support_check*]
 #   Enable monitoring for check_support_default_queue_size
 #
+# [*enable_publishing_api_check*]
+#   Enable monitoring for check_publishing_api_downstream_high_queue_latency
+#
 #
 class monitoring::checks::sidekiq (
   $enable_signon_check = true,
   $enable_support_check = true,
+  $enable_publishing_api_check = true,
 ) {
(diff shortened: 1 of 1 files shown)


modules/monitoring/manifests/checks/sidekiq.pp
```diff
@@ -48,4 +52,16 @@
       host_name => $::fqdn,
     }
   }
+
+  if $enable_publishing_api_check {
+    icinga::check::graphite { 'check_publishing_api_downstream_high_queue_latency':
+      target    => 'transformNull(averageSeries(keepLastValue(stats.gauges.govuk.app.publishing-api.*.workers.queues.downstream_high.enqueued)), 0)',
+      from      => '24hours',
+      args      => '--dropfirst -36',
```
AlanGabbianelli, line 60: What's this for?

barrucadu: Copied from the check above, which says

```
      # Take an average over the most recent 36 datapoints, which at 5
      # seconds per datapoint is the last 3 minutes
```

AlanGabbianelli (approved):

PR #11056, opened 2021-04-21
Add env vars for the Elections API

sihugh:
Overrides will be added as necessary for the URL. The API key will reside in secret hieradata.


hannako (approved):
👍

PR #11058, opened 2021-04-22
Use liveness and readiness healthchecks

barrucadu:
This PR makes three related changes for apps which have liveness and readiness healthchecks:

1. 6603b25e936682bbbef75ad6f0939f13ae014885 They're exposed through nginx as `/_healthcheck-live_app-name` and `/_healthcheck-ready_app-name` (the current healthcheck is exposed as `/_healthcheck_app-name`)
2. a1c5aa81557a504dfdc29b1361da32cb272b4bff The readiness healthcheck is preferred for the Icinga check over a custom healthcheck.
3. 50c30a600617df809238715cb6c398b60cdea2aa The liveness healthcheck is used for the Icinga check if there is neither a readiness healthcheck nor a custom one.

The rest of the commits just enable the new healthchecks for the apps which have them.

When this PR is merged, govuk-aws can be updated to use the new healthchecks.

---

[Trello card](https://trello.com/c/tl6RV1el/723-improve-govuk-healthchecks)


modules/govuk/manifests/app/config.pp
```diff
@@ -390,7 +390,29 @@
...
+  if $has_readiness_health_check {
+    include icinga::client::check_json_healthcheck
+
+    $healthcheck_desc      = "${title} app healthcheck not ok"
+
+    if $health_check_custom_doc {
+      $healthcheck_doc_slug = regsubst($healthcheck_desc, ' ', '-', 'G')
+    } else {
+      $healthcheck_doc_slug = 'app-healthcheck-not-ok'
+    }
```
AlanGabbianelli, line 402: Do we need this conditional?
Can't we simply have `$healthcheck_doc_slug = regsubst($healthcheck_desc, ' ', '-', 'G')`?

barrucadu: Sometimes the condition is true, sometimes it's false.  I think usually it's false, but there are a few apps with custom docs.

AlanGabbianelli: Yes, what I meant is if we only have the regexp replacement function, when there aren't custom docs, the slug would become "app-healthcheck-not-ok" because of the description. But I see now that it would become "#{title}-app-healthcheck-not-ok" instead, which I guess is not ok!

AlanGabbianelli (approved):

PR #11063, opened 2021-04-23
And instance_prefix for grafana to show up memcached stats

hannako:
- Looks like instance_prefix is required to get data to show up, as
  apps can live on instances that do not share their name.


sihugh (approved):
We can remove the integration entries now, but 👍

PR #11069, opened 2021-04-26
Add charts to Mongodb Grafana dashboard

bilbof:
This adds four charts to the Mongodb dashboard:
* Mongodb lock percent ratio VS content-store writes
* content-store write volume VS GET request percentiles
* content-store Nginx status codes
* content-store write request percentiles

These charts helped to diagnose problems affecting Content Store.

This commit is pretty hard to review. I'd suggest just looking at [the dashboard that I exported the JSON from](https://grafana.blue.production.govuk.digital/dashboard/db/temporary-mongodb-incident?refresh=1m&orgId=1).

### DIFF
--- a/modules/grafana/files/dashboards_aws/mongodb.json
+++ b/modules/grafana/files/dashboards_aws/mongodb.json
@@ -35,12 +35,166 @@
           "linewidth": 1,
           "links": [],
           "minSpan": 3,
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
Dashboard looks good to me. JSON looks like JSON.

PR #11071, opened 2021-04-27
Forbid access to internal Signon API

bilbof:
We will be adding an internal API to Signon at /api in this PR: alphagov/signon#1617.

This introduces an L7 firewall to restrict public access to this API.


richardTowers (approved):
Looks good to me. I checked out of band, and we've already handled this in govuk-infrastructure with an ALB rule.

PR #11073, opened 2021-04-28
Remove newline at the end of config.xml.erb

thomasleese:
It appears to be that when Jenkins reloads its own configuration, it writes the file out again but without the trailing new line. This means when Puppet next runs, it detects that the file is different, writes the file, and reloads the Jenkins config. The newline was added in 066b0694165954491cef2e36d8ff9d12b4c9eb7b and we have removed it before in 29acd89ae71e964a3c518997b4325767e1a40a60 and dae2ac8bf349ba4ed7beef02f11a59ba8d9644e0.

~I'm not sure how we can set this up so we don't put the new line back by accident again...~ I've added a GitHub Actions workflow for this check, which might be a little unusual/unnecessary.

Because of this, Jenkins is being reloaded approximately every 30 minutes, and I think this is causing it to lose track of any jobs which are dependant on another one finished. Most recently, we've been seeing this with the `Deploy_App_Downstream` job causing the continuous deployment pipeline to get stuck. Checking the Jenkins logs, I found a job which had been stuck, and coincidentally, the Jenkins config had been reloaded just a few minutes after the stuck job started.

See https://trello.com/c/hiZUg0gC/1142-smokey-failures-long-timeouts-blocking-continuous-deployment for more details on this failure.


benjamineskola:
> I'm not sure how we can set this up so we don't put the new line back by accident again...

Can we add a pre-commit hook for it?

thomasleese:
> Can we add a pre-commit hook for it?

That's probably worth considering, I might have a play around and see if there's something light-touch we can do (maybe a GitHub Action).

For what it's worth, I still think this PR is worth merging as it is without the long-term fix, to prevent our CD pipeline from getting stuck in the future.

benjamineskola (approved):

thomasleese:
> That's probably worth considering, I might have a play around and see if there's something light-touch we can do (maybe a GitHub Action).

I've added a GitHub Actions workflow for this check. It won't stop a PR from being merged, but it will show up as red if there is a trailing new line.

I'll wait for a re-review before merging, because I appreciate it's a little unorthodox and I'm happy to not have that workflow there.

kevindew:
It's a great idea to add a test for this, as this is so easy to miss or realise it's problematic ⭐ 

I feel the GitHub action approach is quite surprising though. I'm wondering that a slightly more conventional approach (though still a bit weird) would be to have this tested as part of puppets existing unit tests.

You could put something in https://github.com/alphagov/govuk-puppet/blob/93663a92bd6dcd5c8b0393068363734e56b19910/modules/govuk_jenkins/spec/classes/jenkins__config_spec.rb like: 

```
    context 'no new line' do
      let (:params) { default_params.merge({
        :manage_config => true,
      })}

      it do
        expect(subject).to contain_file('/var/lib/jenkins/config.xml').without_content(/\n\Z/),
          "Ensure modules/govuk_jenkins/templates/config/config.xml.erb has no new line. See https://github.com/alphagov/govuk-puppet/pull/11073 for more details."
      end
    end
```

(it took me way too long to work out how to do that)

thomasleese:
> I feel the GitHub action approach is quite surprising though. I'm wondering that a slightly more conventional approach (though still a bit weird) would be to have this tested as part of puppets existing unit tests.

That's fair, it actually hadn't occurred to me to do it that way, but I agree it's less surprising. I've amended the PR with a commit which adds this.

modules/govuk_jenkins/spec/classes/jenkins__config_spec.rb
```diff
@@ -1,39 +1,47 @@
...
+      it { is_expected.to contain_file("/var/lib/jenkins/config.xml").with_content(/<authorizationStrategy class="hudson.security.GlobalMatrixAuthorizationStrategy">/) }
+      it { is_expected.to contain_file("/var/lib/jenkins/config.xml").with_content(/<permission>jenkins.perm.admin:admin_user<\/permission>\n\s+<permission>jenkins.perm.edit:admin_user<\/permission>/) }
+    end
+
+    context "no new line" do
+      let (:params) { default_params.merge(manage_config: true) }
+
+      it do
+        expect(subject).to contain_file("/var/lib/jenkins/config.xml").without_content(/\n\Z/),
+          "Ensure modules/govuk_jenkins/templates/config/config.xml.erb has no new line. See https://github.com/alphagov/govuk-puppet/pull/11073 for more details."
```
kevindew, line 43:
```suggestion
          "Ensure modules/govuk_jenkins/templates/config/config.xml.erb does not end with a new line character. See https://github.com/alphagov/govuk-puppet/pull/11073 for more details."
```

I realised this message didn't make much sense as the file can have plenty of new line characters it's just the last one we care about. It seemed surprisingly hard to phrase what we actually mean though so feel free to improve on my suggestion if you have a better idea.

kevindew (commented):
Nice one Tom. I've just got a little stuck on the message but I think it's just about good to go.

modules/govuk_jenkins/spec/classes/jenkins__config_spec.rb
thomasleese: I've amended the commit with an improved message - what do you think?

kevindew: Looks a winner 🥇

kevindew (approved):

PR #11083, opened 2021-05-05
Fix setting cookies for /find-coronavirus-support

theseanything:
This service is provided by Smart Answers and require cookies to work. Previously routes to /find-coronavirus-support were prefixed with /s/, however this has since been changes to no longer require the prefix slug. This commit prevents cookies from being stripped from requests without the /s/ prefix.


kevindew (approved):
Oh gosh our cookie hacks 🙈

PR #11085, opened 2021-05-06
Update the CI agents to use python 2.7.14

kentsanggds:
## What

This is needed to deploy CKAN 2.8 changes, python 3 was set up to be used for deployment of CKAN 2.9 and will be implemented at a later date as it is still undergoing development.


thomasleese (commented):
The change itself seems fine, do we know if this will affect any of our non-CKAN Python repos? I suspect not since, I think we originally changed to 3.6 for CKAN?

kentsanggds:
I don't think it affects any other apps, the main reason I wanted to install python 3.6 was for CKAN 2.9 tests on the CI agents. I'll work on getting python3 setup on another PR as that requires a little bit of work.

thomasleese (approved):

PR #11086, opened 2021-05-06
Bump search-api unicorns to 48 per machine

sihugh:
I want to reduce the number of machines we have in production as it's costing us money. Staging has 36 unicorns running on machines with half the CPU and RAM of those in production and seems fine when dealing with replayed production traffic.

The machines are under low load, and there's plenty of headroom in terms of CPU and memory used. [Example machine on Grafana](https://grafana.production.govuk.digital/dashboard/file/machine.json?refresh=1m&orgId=1&var-hostname=search-ip-10-13-5-135_eu-west-1_compute_internal&var-cpmetrics=All&var-filesystem=All&var-disk=All&var-tcpconnslocal=All&var-tcpconnsremote=All&from=now-2d&to=now)

This app is not particularly CPU intensive as it offloads most things to Elasticsearch or SageMaker, and does some processing of results on the way back through.

I propose to ratchet this up in stages (reducing the number of machines as we go), to potentially halve the number of machines eventually while retaining our pool of workers.

### DIFF
--- a/hieradata_aws/production.yaml
+++ b/hieradata_aws/production.yaml
@@ -222,7 +222,7 @@ govuk::apps::search_api::relevancy_bucket_name: 'govuk-production-search-relevan
 govuk::apps::search_api::sitemaps_bucket_name: 'govuk-production-sitemaps'
 govuk::apps::search_api::tensorflow_sagemaker_endpoint: 'govuk-production-search-ltr-endpoint'
 govuk::apps::search_api::enable_learning_to_rank: true
-govuk::apps::search_api::unicorn_worker_processes: "36"
+govuk::apps::search_api::unicorn_worker_processes: "48"
 govuk::apps::service_manual_publisher::govuk_notify_template_id: "e00e89f5-b622-4dcb-8f30-e6c70231a940"
 govuk::apps::short_url_manager::govuk_notify_template_id: "e00e89f5-b622-4dcb-8f30-e6c70231a940"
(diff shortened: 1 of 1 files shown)


bilbof (approved):
Great change. Search API's work is basically entirely IO-bound (workers mostly sit idle waiting for requests to complete) so adding more workers seems reasonable. I understand that you want to scale up gradually - from the looks of things I reckon we'll be able to go a fair bit further.

Our [c5.4xlarge](https://aws.amazon.com/ec2/instance-types/c5/) instances have 16 vCPU and 32 GB RAM. How many workers do you think we could run? Judging by current memory usage, we could have ~80 workers, and be able to handle the same number of concurrent requests with half the machines.

Assuming requests are distributed equally, that's 3 concurrent requests per core (with 48 workers), which seems like we'll still be under-utilising them, since Search API is not CPU-bound (as the workers do very little actual work).

sihugh:
Yep, that was my plan exactly (to get up to ~80). I'll probably be bolder on the next jump 😬

sengi (approved):
SGTM.

If the app isn't CPU-intensive then it might be worth considering switching from C instances ("compute-optimized" i.e. high CPU to RAM ratio) to M instances ("general-purpose" i.e. typical ratio of CPU/RAM) or R instances ("memory-optimized" i.e. high RAM to CPU ratio).

RAM looks about 30% utilised on the couple of instances I spot-checked, and they're hardly tickling the CPU.

PR #11093, opened 2021-05-07
Add govuk-python3 package to CI agents and CKAN servers

kentsanggds:
## What

As its not possible to deploy python 2.7.14 and python 3.6.12 using the same package name govuk-python, a new package was created govuk-python3 to allow both python packages to be deployed on a machine.

### DIFF
--- a/hieradata_aws/class/ckan.yaml
+++ b/hieradata_aws/class/ckan.yaml
@@ -8,4 +8,5 @@ govuk::apps::ckan::enable_harvester_gather: true
 
 govuk_solr6::present: false
 
-govuk_python::govuk_python_version: '3.6.12'
+govuk_python::govuk_python_version: '2.7.14'
+govuk_python3::govuk_python_version: '3.6.12'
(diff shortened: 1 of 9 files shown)


thomasleese (commented):
Do we need to include `include govuk_puppet3` to places like this as well: https://github.com/alphagov/govuk-puppet/blob/57cc29472ed5429bda246dcde0b4c830c9dbc10c/modules/govuk/manifests/deploy.pp#L6 ?

thomasleese:
Okay, I wonder if we should be removing the old `python3` install from the `govuk_python` class? https://github.com/alphagov/govuk-puppet/blob/bd6da193bb1a282df8433ff0ec0cad2bfdfecd69/modules/govuk_python/manifests/init.pp#L16-L17

Because it means at the moment that including the `govuk_python` class or the `govuk_python3` class, you get `python-3.5` - I wonder if that's confusing behaviour?

If we do decide do to this though, it does mean that where we do `include govuk_puppet` at the moment, we'll probably need to add a `include govuk_puppet3` line as well.

kentsanggds:
I think for now, I'm going to leave it, as in the future as python 2 is EOL, python 3 should be the default so we can probably drop the govuk_python3 module if we are still using puppet. 

If I did drop python 3 from govuk_python then it would also mean updating some other apps which would then require testing to ensure that they still work, so will probably have to plan this with the teams affected.

PR #11110, opened 2021-05-19
Improve Whitehall Icinga checks

thomasleese:
This improves how we configure the Whitehall Icinga checks, see individual commits for more details.

Originally, this PR came about from an investigation into why the `never` Icinga time period didn't seem to be working. I'm unable to reproduce the problem, and as far as I can tell the check period is working. See [this screenshot from Integration](https://alert.integration.publishing.service.gov.uk/cgi-bin/icinga/extinfo.cgi?type=2&host=ip-10-1-4-181.eu-west-1.compute.internal&service=overdue+publications+in+Whitehall) showing that the check hasn't occurred recently:

Screenshot 2021-05-19 at 09 30 23
Screenshot 2021-05-19 at 09 30 19

If the problem does re-occur we can continue to investigate, my suggestion would be to have an `$enabled` parameter, and in Integration/Staging it's set to `false` and the checks aren't created in the first place.

[Trello Card](https://trello.com/c/Tu6CDw6t/2491-investigate-fix-or-remove-puppet-scheduled-check-never-option-15-days)

### DIFF
--- a/hieradata_aws/integration.yaml
+++ b/hieradata_aws/integration.yaml
@@ -349,8 +349,7 @@ monitoring::checks::rds::region: 'eu-west-1'
 monitoring::checks::aws_iam_key::enabled: true
 monitoring::checks::aws_iam_key::region: 'eu-west-1'
 monitoring::checks::aws_iam_key::max_aws_iam_key_age: 320
-monitoring::checks::whitehall::overdue_check_period: "never"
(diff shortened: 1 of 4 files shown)


1pretz1 (approved):
Looks good to me! 🍏

ChrisBAshton (commented):
I know this is already reviewed and merged, just wanted to add a ✅ as these changes look great!

PR #11111, opened 2021-05-19
Set up RabbitMQ for account-api

barrucadu:
We want account-api to listen to publishing updates from RabbitMQ so
that we can update a user's saved pages when content gets updated.

For example, if we want to show the titles of the pages a user has
saved, we could do that by:

1. Querying content-store every time the user requests the list
2. Doing (1), but caching the results for a while
3. Querying content-store when the page is first saved, and updating
   local state when updates occur.

We're going for (3).  (1) introduces O(n) extra network requests; (2)
does the same, unless we have a long cache time, in which case we risk
showing stale data.  (3) introduces one extra network request when we
save a page, but no additional requests when the list is retrieved
and, while at risk of stale state, we can bound that risk by adopting
an SLO if we observe problems.

This commit is based on the cache-clearing-service RabbitMQ config.

---

[Trello card](https://trello.com/c/8EadgviV/772-get-account-api-consuming-publishing-messages-from-rabbitmq)

### DIFF
--- a/modules/govuk/manifests/apps/account_api.pp
+++ b/modules/govuk/manifests/apps/account_api.pp
@@ -37,6 +37,22 @@
 # [*db_name*]
 #   The database name to use for the DATABASE_URL environment variable
 #
+# [*rabbitmq_hosts*]
+#   RabbitMQ hosts to connect to.
+#   Default ['localhost']
+#
+# [*rabbitmq_user*]
+#   RabbitMQ username.
+#   Default 'account_api'
+#
+# [*rabbitmq_password*]
+#   RabbitMQ password.
+#   Default 'account_api'
+#
+# [*enable_publishing_queue_listener*]
+#   Run the worker which processes publishing updates.
+#   Default: false
+#
 # [*account_oauth_client_id*]
 #   Client ID for the Transition Checker in GOV.UK Account Manager
 #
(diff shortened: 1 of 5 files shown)


modules/govuk/manifests/apps/account_api.pp
```diff
@@ -65,6 +81,10 @@
   $db_port = undef,
   $db_password = undef,
   $db_name = 'account-api_production',
+  $rabbitmq_hosts = ['localhost'],
+  $rabbitmq_user = 'account_api',
+  $rabbitmq_password = 'account_api',
```
huwd, line 86: Why don't we care about these credentials being public?
Isn't this the sort of thing we usually load with a variable / secrets?

huwd, line 86: Why don't we care about these credentials being public?
Isn't this the sort of thing we usually load with a variable / secrets?

barrucadu: The actual password is in govuk_secrets

huwd: so that overwrites these?
Is it that puppet does not accept a null/nil value here then?

Is it that this could just as easily be `""` but `'account_api` is more discriptive?

barrucadu: Yes, then value in the hieradata (whether that be in this repo or govuk-secrets) overwrites the default value.

Puppet accepts a missing value, but this seems to be the convention in other places:  https://github.com/alphagov/govuk-puppet/blob/756f34ff8d896c0eb8cd7bda0d7342fc5fe7b35b/modules/govuk/manifests/apps/cache_clearing_service.pp#L37-L39

huwd (approved):

PR #11123, opened 2021-05-27
fix the update of python2 to 2.7.14

fredericfran-gds:
This is the 2nd attempt (1st attempt in #11122) to install python 2.7.14
on all machines since the default Trusty version of pip2 is not SNI
compliant and no longer works.

Futher improvement on #11122 are:
1. installing govuk_python earlier in the process before govuk_scripts

2. merge `modules/govuk_jenkins/manifests/packages/govuk_python.pp` into
`modules/govuk_python/manifests/init.pp` to remove redundancy.

3. link `/usr/bin/pip` to `/opt/python2.7/bin/pip`

4. python3.6 is installed on mapit instances using `govuk-python3` repo
rather than mistakenly the `govuk-python` repo. This is required since
installation of Python 2.7.14 is via `govuk-python` repo and we can't
have 2 co-existing versions of the same apt package.

Co-Authored-By: Chris Ashton <ChrisBAshton@users.noreply.github.com>

Trello: https://trello.com/c/zGBit7Av/2542-pip-is-broken-in-trusty


ChrisBAshton (approved):

modules/govuk/manifests/node/s_mapit.pp
```diff
@@ -2,6 +2,7 @@
 #
 class govuk::node::s_mapit inherits govuk::node::s_base {
 
+  include govuk_python3
```
ChrisBAshton, line 5: All the changes LGTM, except perhaps this one - why is this not `govuk_python` rather than `govuk_python3`?

fredericfran-gds: I updated the PR description to include:
```
python3.6 is installed on mapit instances using govuk-python3 repo
rather than mistakenly the govuk-python repo. This is required since
installation of Python 2.7.14 is via govuk-python repo and we can't
have 2 co-existing versions of the same apt package.
```

richardTowers (approved):
As far as I can tell, this looks good.

We discussed this a bit [on slack](https://gds.slack.com/archives/CADKZN519/p1622194427139200) and we think that this will apply cleanly for non-mapit nodes, and mapit nodes which are freshly provisioned. There's an issue with conflicting python versions trying to apply this to mapit nodes which are already running.

I suggest we:
- merge this
- pause puppet on the mapit machines
- deploy puppet
- one-by-one remove the mapit machines from the ASG so that AWS will reprovision them. This can be flaky, but it's important that reprovisioning machines still works, so it's worth doing
- let puppet run automatically on the other nodes

... or something like that.

PR #11124, opened 2021-06-01
Fix Mapit not starting correctly when a new instance spuns up

fredericfran-gds:
This involves fixing the issue and workaround mentioned in the
[playbook](https://docs.publishing.service.gov.uk/manual/mapit-database-not-available.html)
so that Puppet can instantiate Mapit correctly on instance creation.

We included the database import [script](https://github.com/alphagov/mapit/blob/main/import-db-from-s3.sh)
into the `modules/govuk/files/etc/govuk/import_mapit_data.sh` so that
Puppet can prepare the instance fully before triggering Jenkins for
the final instantiation phase.

### DIFF
--- a/modules/govuk/files/etc/govuk/import_mapit_data.sh
+++ b/modules/govuk/files/etc/govuk/import_mapit_data.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -x
+
 function stop_mapit_services () {
   # Stop Mapit Services
   sudo service nginx stop
@@ -17,11 +19,20 @@ function restart_mapit_services () {
   sudo service nginx start
 }
 
-stop_mapit_services
-# This script needs to run as a user that can sudo to become the postgres
-# user.  It's easiest if this is passwordless, so we can't use the deploy user
-# who can't sudo without a password
-# Run 'govuk_setenv mapit ./import-db-from-s3.sh' from the '/var/apps/mapit' as
 
-cd /var/apps/mapit && govuk_setenv mapit ./import-db-from-s3.sh
+function import_db_from_s3 () {
+  # get the data from s3
+  aws s3 cp --no-sign-request s3://govuk-custom-formats-mapit-storage-production/source-data/2021-05/mapit-may-2021-update.sql.gz /tmp/mapit.sql.gz
+
+  # drop the old db and bring up a new empty one
+  sudo -u postgres dropdb --if-exists mapit
+  sudo -u postgres createdb --owner mapit mapit
+  sudo -u postgres psql -c "CREATE EXTENSION postgis; CREATE EXTENSION postgis_topology;" mapit
+
+  # load up the downloaded data into this new db
+  zcat /tmp/mapit.sql.gz | sudo -u postgres psql mapit
+}
+
+stop_mapit_services
+import_db_from_s3
 restart_mapit_services
--- a/modules/govuk/manifests/apps/mapit.pp
+++ b/modules/govuk/manifests/apps/mapit.pp
@@ -117,16 +117,19 @@ class govuk::apps::mapit (
         value   => $django_secret_key;
   }
 
-  if ($::mapit_data_present == '0') {
+  if ($::mapit_data_present != '1') {
 
 
     file { '/etc/govuk/import_mapit_data.sh':
-      ensure => file,
-      source => 'puppet:///modules/govuk/etc/govuk/import_mapit_data.sh',
-
+      ensure  => file,
+      mode    => '0755',
+      source  => 'puppet:///modules/govuk/etc/govuk/import_mapit_data.sh',
+      require => Govuk_postgresql::Db['mapit'],
     }
+
     exec { 'populate mapit database':
       command => '/bin/bash /etc/govuk/import_mapit_data.sh 2>&1 | logger',
+      require => File['/etc/govuk/import_mapit_data.sh'],
     }
   }
 


modules/govuk/files/etc/govuk/import_mapit_data.sh
```diff
@@ -17,11 +17,24 @@ function restart_mapit_services () {
...
+    exit 1
+  fi
+
+  # drop the old db and bring up a new empty one
+  sudo -u postgres dropdb --if-exists mapit
+  sudo -u postgres createdb --owner mapit mapit
+  sudo -u postgres psql -c "CREATE EXTENSION postgis; CREATE EXTENSION postgis_topology;" mapit
+
+  # load up the downloaded data into this new db
+  zcat -f /tmp/mapit.sql.gz | sudo -u postgres psql mapit
```
sengi, line 35: Can we remove the `-f`? I don't think there should be any need to use the "force" option here.

modules/govuk/files/etc/govuk/import_mapit_data.sh
```diff
@@ -17,11 +17,24 @@ function restart_mapit_services () {
...
-# This script needs to run as a user that can sudo to become the postgres
-# user.  It's easiest if this is passwordless, so we can't use the deploy user
-# who can't sudo without a password
-# Run 'govuk_setenv mapit ./import-db-from-s3.sh' from the '/var/apps/mapit' as
 
-cd /var/apps/mapit && govuk_setenv mapit ./import-db-from-s3.sh
+function import_db_from_s3 () {
+  # get the data from s3
+  curl 'https://s3-eu-west-1.amazonaws.com/govuk-custom-formats-mapit-storage-production/source-data/2020-11/mapit-november-2020-update.sql.gz' -o /tmp/mapit.sql.gz
+  if ! echo "e6d7d0049348744cf4c8c2f3f921adfcc0a3c76d /tmp/mapit.sql.gz" | sha1sum -c -; then
```
sihugh, line 24: this sha will be a pain to update when the data changes. It'd be great if we could figure out a way for this not to be reliant on a puppet deployment. (Appreciate that you're fixing a thing right now :-) )

modules/govuk/manifests/apps/mapit.pp
```diff
@@ -117,16 +117,21 @@
...
+      ensure  => file,
+      mode    => '0755',
+      source  => 'puppet:///modules/govuk/etc/govuk/import_mapit_data.sh',
+      require => Class['govuk_postgresql::server'],
     }
+
     exec { 'populate mapit database':
-      command => '/bin/bash /etc/govuk/import_mapit_data.sh 2>&1 | logger',
+      command     => '/bin/bash /etc/govuk/import_mapit_data.sh 2>&1 | logger',
+      subscribe   => File['/etc/govuk/import_mapit_data.sh'],
```
sengi, line 132: Are you sure you want the subscribe here? i.e. do we want Puppet to re-run the import script if the import script changes? Seems like that would have availability implications, and probably be surprising to someone in future.

modules/govuk/manifests/apps/mapit.pp
```diff
@@ -117,16 +117,21 @@
         value   => $django_secret_key;
   }
 
-  if ($::mapit_data_present == '0') {
+  if ($::mapit_data_present != '1') {
```
sengi, line 120: Ah, Puppet 3 :headdesk: 😅

sengi (approved):
Cool, thanks for taking this on. Looks good, just a couple of minor potential issues to check out.

modules/govuk/files/etc/govuk/import_mapit_data.sh
sengi: Probably just remove the checksum then? Not sure it's really adding any value.

fredericfran-gds: I guess there are 3 options for the checksum:
1. when someone wants to update the URL of the S3 object database, they update the checksum too or
2. we remove the checksum like Chris said
3. we retrieve the checksum from s3 and compare it with the locally computed one if we are afraid the s3 object is corrupted on download?

On the second issue of not being reliant on Puppet, this is the easiest method, we could transfer everything in the capistrano deployment but the `deploy` user there is limited.

modules/govuk/files/etc/govuk/import_mapit_data.sh
fredericfran-gds: I will try, to be honest, I copied the script from Mapit without looking into the specifics.

modules/govuk/manifests/apps/mapit.pp
fredericfran-gds: I will remove that, anyway, it was not going to run because of the conditional below as you rightly saw.

modules/govuk/files/etc/govuk/import_mapit_data.sh
```diff
@@ -1,27 +1,38 @@
...
-stop_mapit_services
-# This script needs to run as a user that can sudo to become the postgres
-# user.  It's easiest if this is passwordless, so we can't use the deploy user
-# who can't sudo without a password
-# Run 'govuk_setenv mapit ./import-db-from-s3.sh' from the '/var/apps/mapit' as
 
-cd /var/apps/mapit && govuk_setenv mapit ./import-db-from-s3.sh
+function import_db_from_s3 () {
+  # get the data from s3
+  aws s3 cp --no-sign-request s3://govuk-custom-formats-mapit-storage-production/source-data/2021-05/mapit-may-2021-update.sql.gz /tmp/mapit.sql.gz
```
sengi, line 25: Btw the use of a fixed location in /tmp isn't good practice from a security standpoint, but in this case I don't think it's a big deal because if a bad actor is on the box they're not gonna need any help from a tempfile TOCTOU in this script 😅 Just mentioning for future reference - it might matter in some other situation you encounter.

modules/govuk/files/etc/govuk/import_mapit_data.sh
sengi: The download is already over HTTPS from a bucket which only we can write to, and the chance of corruption is negligible because this will run maybe a few times a month at most. If you're worried about removing the checksum, don't be :)

PR #11129, opened 2021-06-03
use old mapit db

fredericfran-gds:
This is due to the fact that new one has a missing local authority in the database and we will be serving 404s.


fredericfran-gds:
no longer needed

PR #11131, opened 2021-06-04
add python 2.7.14 bin path to /etc/environment file

fredericfran-gds:
This is required following the installation of python 2.7.14 in #11123


sengi (approved):
Seems reasonable 👍

PR #11133, opened 2021-06-10
Modify the CDN cache clear job to programatically get most popular endpoints

alex-ju:
The existing endpoints are pretty old. They also used GA data over a year so should be pretty consistent. This takes a different approach and pulls the most popular pages from the last 2 weeks down before clearing the cache on those.

[Trello card](https://trello.com/c/zRoRyhqa)


nicholsj:
There are other consistently "popular" pages that won't be included in the search-api call - eg `/search/all` and `/` (homepage). So I assume these will still need to be hardcoded

PR #11134, opened 2021-06-10
remove hardcoded path from s3cmd

fredericfran-gds:
With the new version of python 2 installed, the python2 pip binaries are in located in `/opt/python2.7/bin` and therefore, we should not hardcode the `s3cmd` path.   `/opt/python2.7/bin` is already on the default PATH env variable.

Ref:
1. #11131
2. #11123


sengi (approved):
👍

PR #11136, opened 2021-06-10
fix pip installation via exec

fredericfran-gds:
With the new version of pip (#11123), the output of pip has changed. This PR fixes that even through it is brittle to grep on the output.


bilbof (approved):
LGTM

PR #11142, opened 2021-06-10
Modify mongo cluster to use FQDN for nodes in production

rtrinque:
We need this to run the new platform based on ECS as it is unable
to resolve shortened host names.


bilbof (changes requested):
It looks like you've re-used the `integration` hostnames.

fredericfran-gds (approved):
looks good.

PR #11144, opened 2021-06-11
fix docker collectd plugin following python upgrade

fredericfran-gds:
Collectd-core has been compiled against the original system Python 2.7.6 and will fail to load the current docker-collectd-plugin plugin version.

We fixed the docker-collectd-plugin in [here](https://github.com/alphagov/docker-collectd-plugin/commit/e3e520e665d56a3d0b1d3bb697c0504d4494012d)

This PR will upgrade the version of the docker-collectd plugin that we are using.


rtrinque (approved):
Looks good

PR #11145, opened 2021-06-14
Upgrade to govuk_seed_crawler 2.1.0

thomasleese:
This upgrades to a more recent version of govuk_seed_crawler which should fix an issue where the mirror machine doesn't come up correctly.


ChrisBAshton (approved):
Makes sense 👍 thanks!

PR #11147, opened 2021-06-15
Add unicorn_worker_processes variable to account-api

barrucadu:
This is not set yet, so it won't have any effect.  But when we do want
to scale up we can just set a number in hieradata now.


huwd (approved):
Nice!
Any idea how long it'd take us to scale up?
Can continue the chat on Slack?

barrucadu:
It's a puppet run, so half an hour after being deployed.

PR #11148, opened 2021-06-15
Slo alert

huwd:
[Trello](https://trello.com/c/Crj6Z5SW/841-add-slack-alerts-for-when-we-drop-to-50-and-25-remaining-error-budget)

### DIFF
--- a/modules/monitoring/manifests/checks/accounts_slos.pp
+++ b/modules/monitoring/manifests/checks/accounts_slos.pp
@@ -1,0 +1,50 @@
+# == Class: monitoring::checks::accounts_slos
+#
+# Nagios alerts for Account Team SLOs
+#
+# === Parameters
+#
+# [*enabled*]
+#   This variable enables to define whether to perform Account SLO checks. [This is set to 'true' by default.]
+class monitoring::checks::accounts_slos (
+  $enabled = true,
+) {
+  if $enabled {
+    $error_budget_warning = 50 # Warn when <50% of budget remains
+    $error_budget_critical = 25 # Declare critical when <25% of budget remains
+
+    $latency_bad_metric = 'transformNull(offset(scale(removeBelowValue(maxSeries(stats.timers.*.nginx_logs.account-api.time_request.upper_90),0.850),0),1))'
+    $latency_all_metric = 'offset(scale(transformNull(maxSeries(stats.timers.*.nginx_logs.account-api.time_request.upper_90)),0),1)'
+    $latency_error_budget = '0.01'
+
+    @@icinga::check::graphite { 'check_latency_error_budget_remaining':
+      target         => "scale(asPercent(scale(offset(divideSeries(integral(${latency_bad_metric}),integral(${latency_all_metric})),-${latency_error_budget}),-1),${latency_error_budget}),-1)",
+      warning        => "-${error_budget_warning}",
+      critical       => "-${error_budget_critical}",
+      desc           => 'Latency Error budget for GOV.UK Accounts has fallen below acceptable thresholds',
+      notes_url      => monitoring_docs_url(slo-error-budget-accounts),
+      # Get the data over a 28 day period. The graphite metrics record on this
+      # basis, if we go for a shorter window they tend to exaggerate recent small faults.
+      from      => '28days',
+      host_name      => $::fqdn,
+      contact_groups => ['slack-channel-accounts-tech'],
+    }
+
+    $http_bad_metric = 'sumSeries(transformNull(stats_counts.*.nginx_logs.account-api.http_{422,5*}))'
+    $http_all_metric = 'sumSeries(transformNull(stats_counts.*.nginx_logs.account-api.http_*))'
+    $http_error_budget = '0.01'
+
+    @@icinga::check::graphite { 'check_http_error_budget_remaining':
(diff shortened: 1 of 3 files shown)


modules/monitoring/manifests/checks/accounts_slos.pp
```diff
@@ -0,0 +1,31 @@
...
+    $latency_all_metric = 'offset(scale(transformNull(maxSeries(stats.timers.*.nginx_logs.account-api.time_request.upper_90)),0),1)'
+    $latency_error_budget = '0.01'
+
+    @@icinga::check::graphite { 'check_error_budget_remaining':
+      target         => "asPercent(scale(offset(divideSeries(integral(${latency_bad_metric}),integral(${latency_all_metric})),-${latency_error_budget}),-1),${latency_error_budget})",
+      warning        => $error_budget_warning,
+      critical       => $error_budget_critical,
+      desc           => 'Error budget for GOV.UK Accounts has fallen below acceptable thresholds',
+      notes_url      => monitoring_docs_url(account-error-budget),
+      contact_groups => 'slack-channel-accounts-tech',
```
barrucadu, line 28: Does this need to be an array?

```
contact_groups => ['slack-channel-accounts-tech'],
```

modules/monitoring/manifests/checks/accounts_slos.pp
```diff
@@ -0,0 +1,31 @@
+# == Class: monitoring::checks::accounts_slos
+#
+# Nagios alerts for Account Team SLOs
+#
+# === Parameters
+#
+# [*enabled*]
+#   This variable enables to define whether to perform Account SLO checks. [This is set to 'true' by default.]
+class monitoring::checks::accounts_slos (
```
barrucadu, line 9: We'll need to include this class here: https://github.com/alphagov/govuk-puppet/blob/main/modules/monitoring/manifests/checks.pp#L42

barrucadu (commented):
It would be great if we could deploy this on integration, then deploy a branch of account-api which responds to every request with a 500, make a few requests, and check that the alert appears in the channel at the right time.

modules/monitoring/manifests/checks/accounts_slos.pp
```diff
@@ -0,0 +1,32 @@
...
+
+    $latency_bad_metric = 'transformNull(offset(scale(removeBelowValue(maxSeries(stats.timers.*.nginx_logs.account-api.time_request.upper_90),0.850),0),1))'
+    $latency_all_metric = 'offset(scale(transformNull(maxSeries(stats.timers.*.nginx_logs.account-api.time_request.upper_90)),0),1)'
+    $latency_error_budget = '0.01'
+
+    @@icinga::check::graphite { 'check_error_budget_remaining':
+      target         => "asPercent(scale(offset(divideSeries(integral(${latency_bad_metric}),integral(${latency_all_metric})),-${latency_error_budget}),-1),${latency_error_budget})",
+      warning        => $error_budget_warning,
+      critical       => $error_budget_critical,
+      desc           => 'Error budget for GOV.UK Accounts has fallen below acceptable thresholds',
```
barrucadu, line 26: Call this "Latency error budget" or something

modules/monitoring/manifests/checks/accounts_slos.pp
```diff
@@ -0,0 +1,32 @@
...
+    # Warn when Accounts has less than 50% of our error budget remaining
+    $error_budget_warning = 50
+    # Declare critical when Accounts has less than 25% of our error budget remaining
+    $error_budget_critical = 25
+
+    $latency_bad_metric = 'transformNull(offset(scale(removeBelowValue(maxSeries(stats.timers.*.nginx_logs.account-api.time_request.upper_90),0.850),0),1))'
+    $latency_all_metric = 'offset(scale(transformNull(maxSeries(stats.timers.*.nginx_logs.account-api.time_request.upper_90)),0),1)'
+    $latency_error_budget = '0.01'
+
+    @@icinga::check::graphite { 'check_error_budget_remaining':
```
barrucadu, line 22: Add a second check for the availability SLO

modules/monitoring/manifests/checks/accounts_slos.pp
```diff
@@ -0,0 +1,32 @@
...
+      target         => "asPercent(scale(offset(divideSeries(integral(${latency_bad_metric}),integral(${latency_all_metric})),-${latency_error_budget}),-1),${latency_error_budget})",
+      warning        => $error_budget_warning,
+      critical       => $error_budget_critical,
```
barrucadu, line 25: I think the behaviour is "alert if target is over threshold", but we want the opposite.

So instead of having a value from 0 to 100, let's have a value from -100 to 0:

```
target   => "scale(..., -1)",
warning  => "-${error_budget_warning}",
critical => "-${error_budget_critical}",
```

modules/monitoring/manifests/checks/accounts_slos.pp
```diff
@@ -0,0 +1,48 @@
...
+#
+# [*enabled*]
+#   This variable enables to define whether to perform Account SLO checks. [This is set to 'true' by default.]
+class monitoring::checks::accounts_slos (
+  $enabled = true,
+) {
+  if $enabled {
+    target   => "scale(..., -1)",
+    warning  => "-${error_budget_warning}",
+    critical => "-${error_budget_critical}",
```
barrucadu, line 15: Remove these lines

huwd:
Ok we're going to have to take another tack here,
looks like there's a bug / flaw in our version of graphite that is stopping us getting good aggregate metrics.

PR #11150, opened 2021-06-16
Set up Smokey env vars for GOV.UK Account tests

barrucadu:
We're creating a test user for Smokey to log in with, to confirm that
deploys don't break the log in journey.

---

[Trello card](https://trello.com/c/a1mlNSkg/849-add-a-smoke-test-for-signing-in-from-the-checker-results-page)


huwd (approved):
Will go pull down the secret portion now and check I can decrypt

PR #11151, opened 2021-06-16
Set PYTHONPATH variable to new python version 2.7.14

rtrinque:
There is a case when we need to call the old 2.7.6 python because
some libraries were compiled against it.
See: https://github.com/alphagov/govuk-puppet/commit/8eeaddc6fe07b2305b24400d9934d752bcc4a78a

This creates a problem because modules are now installed in
/opt/python2.7/lib/python2.7/site-packages/, and so are missing
as viewed from 2.7.6.
This set a common path for python modules.


fredericfran-gds (approved):
looks good, lets test in integration

PR #11153, opened 2021-06-17
fix virtualenv factor

fredericfran-gds:
virualenv factor uses the output of `virtualenv --version`.

Before the update to new python/pip, the output was only a version number e.g. `10.0.1`, now it is a longer text: e.g. `20.4.7 from /opt/python2.7/lib/python2.7/site-packages/virtualenv/__init__.pyc`

which the virtualenv factor cannot regex correctly.

Therefore, we modify the virtualenv factor to handle both scenarios:
Original code:
```
# Make virtualenv version available as a fact

Facter.add("virtualenv_version") do
  setcode do
    if Facter::Util::Resolution.which('virtualenv')
      Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(/^(\d+\.\d+\.?\d*).*$/)[0]
    end
  end
end
```

to new code:
```
# Make virtualenv version available as a fact

Facter.add("virtualenv_version") do
  setcode do
    if Facter::Util::Resolution.which('virtualenv')
      match=Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(/^(\d+\.\d+\.?\d*).*$/)
      if match.nil?
        Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(/^virtualenv (\d+\.\d+\.?\d*).*$/)[1]
      else
        match[0]
      end
    end
  end
end
```


ChrisBAshton:
I think it should be possible to account for both variants with a single regex, using an optional matching group at the beginning: 
```
if Facter::Util::Resolution.which('virtualenv')
  Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(/^(?:virtualenv )(\d+\.\d+\.?\d*).*$/)[0]
end
```

fredericfran-gds:
> I think it should be possible to account for both variants with a single regex, using an optional matching group at the beginning:
> 
> ```
> if Facter::Util::Resolution.which('virtualenv')
>   Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(/^(?:virtualenv )(\d+\.\d+\.?\d*).*$/)[0]
> end
> ```

Doesn't work:
```
irb(main):001:0> "virtualenv 20.4.7 from /opt/python2.7/lib/python2.7/site-packages/virtualenv/__init__.pyc".match(/^(?:virtualenv )(\d+\.\d+\.?\d*).*$/)[0]
=> "virtualenv 20.4.7 from /opt/python2.7/lib/python2.7/site-packages/virtualenv/__init__.pyc"
```

bilbof (approved):
Contents of `stankevich-python-1.9.8/lib/facter/virtualenv_version.rb ` in tarball:

```
$ tar -xvzf vendor/puppet/cache/stankevich-python-1.9.8.tar.gz
$ cat stankevich-python-1.9.8/lib/facter/virtualenv_version.rb 
```

```ruby
# Make virtualenv version available as a fact

Facter.add("virtualenv_version") do
  setcode do
    if Facter::Util::Resolution.which('virtualenv')
      match=Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(/^(\d+\.\d+\.?\d*).*$/)
      if match.nil?
        Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(/^virtualenv (\d+\.\d+\.?\d*).*$/)[1]
      else
        match[0]
      end
    end
  end
end
```

I'd prefer something like:

```ruby
Facter.add("virtualenv_version") do
  setcode do
    if Facter::Util::Resolution.which('virtualenv')
      version_output = Facter::Util::Resolution.exec('virtualenv --version 2>&1')
      version_output.match(/\d+\.\d+\.?\d*/).to_s # semver version
    end
  end
end
```

However, since we'll be deleting this very soon, I'm fine if you want to merge this. I've tested that it should work locally.

PR #11155, opened 2021-06-17
override and fix virtualenv_version fact

fredericfran-gds:
`virualenv_version` factor uses the output of `virtualenv --version`.

Before the update to new python/pip, the output was only a version number e.g. `10.0.1`, now it is a longer text: e.g. `20.4.7 from /opt/python2.7/lib/python2.7/site-packages/virtualenv/__init__.pyc`

which the virtualenv_version factor cannot regex correctly.

Therefore, we modify the virtualenv factor to handle both scenarios.

A high weight is used for this version of `virtualenv_version` facter to take precedence over the standard broken default one.


rtrinque (approved):
Looks good

PR #11156, opened 2021-06-17
Revert "fix virtualenv factor"

fredericfran-gds:
Reverts alphagov/govuk-puppet#11153 since #11155 is a better implementation of the new `virtualenv_version` facter. Another issue with this PR is that it was changing the facter in `stankevich/puppet-python` but I believe the puppet module `voxpupuli/puppet-python` is taking precedence.


rtrinque (approved):
Ok

PR #11161, opened 2021-06-18
Correctly set Publisher report permissions

benjamineskola:
The reports for publisher are run by a user that does not have write access to the output directory, and it seems only to work for reports where there's an existing report file owned by the right user. This corrects that.

https://trello.com/c/fhtbsEgu/2528-create-a-mainstream-url-report-which-shows-all-urls


1pretz1 (approved):
LTGM! 👍

PR #11162, opened 2021-06-21
Adding custom PYTHONPATH var when running collectd

rtrinque:
collectd needs to run python 2.7.6 because it was compiled against
 this version of python.

To fix a problem with pip we are now using python 2.7.14, see :
https://github.com/alphagov/govuk-puppet/pull/11122

However this creates a problem for collectd when trying to load
python modules since it looks for them in the 2.7.6 directory and
not the 2.7.14 where they are not installed.

We modify PYTHONPATH for the benefit of collectd to fix this.


fredericfran-gds (approved):
looks good, let's test in integration

PR #11164, opened 2021-06-21
Add ckan29 integration

kentsanggds:
## What

Deploy CKAN 2.9 settings and necessary files onto Integration and startup worker processes.

The intention is to leave CKAN 2.9 running on Integration for a week to see if there are any issues running harvest jobs or with the GUI or managing users / datasets / harvest sources.

## Reference 

https://trello.com/c/GZP41t9N/2553-deploy-ckan-29-onto-integration

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -152,6 +152,8 @@ deployable_applications: &deployable_applications
   cache-clearing-service: {}
   ckan:
     repository: 'ckanext-datagovuk'
+  ckan29:
+    repository: 'ckanext-datagovuk'
   collections: {}
   collections-publisher: {}
   contacts:
--- a/modules/govuk/manifests/apps/ckan.pp
+++ b/modules/govuk/manifests/apps/ckan.pp
@@ -95,8 +95,27 @@ class govuk::apps::ckan (
   $sentry_dsn                     = undef,
 ) {
   $ckan_home = '/var/ckan'
-  $ckan_ini  = "${ckan_home}/ckan.ini"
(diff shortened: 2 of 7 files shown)


hieradata_aws/common.yaml
```diff
@@ -64,6 +64,9 @@ node_class: &node_class
   ckan:
     apps:
       - ckan
+  ckan29:
```
fredericfran-gds, line 67: I don't get this, this Hiera change is instructing that there will be a new class of machine `ckan29` and install `ckan`.

fredericfran-gds: Haven't read through the code, I don't think **any** change to **this** file is needed.

fredericfran-gds (commented):
I have made one comment & I guess you have done some extensive testing in integration so happy to approve if you fix my comment.

fredericfran-gds:
Another comment is if you could squash the 20 commits into a smaller number of commits, maybe 1?

hieradata_aws/common.yaml
kentsanggds: There is some logic in the `ckan` app which deploys files according to the environment so no need to create a `ckan29` app. I did try this route initially but hit a number of issues.

kentsanggds:
OK, I'll squash all the commits into 1, I left it separated to make it easier to review

hieradata_aws/common.yaml
fredericfran-gds: i don't see the need for this code. can you point to the logic/code which uses this hiera?

fredericfran-gds (approved):
looks good, this was extensively tested in integration.

PR #11166, opened 2021-06-22
Add a safety check to unattended reboots

rtrinque:
Wait for a machine to finish provisioning before allowing reboot.
We test the presence of /var/lib/instance/boot-finished as it is
created once has finished initializing and running the user-data
snippets.


fredericfran-gds (approved):
Good find!

ChrisBAshton (commented):
Great job!

PR #11168, opened 2021-06-22
Use OAuth access token instead of service account

richardTowers:
The GOOGLE_CREDENTIALS key we were using before was a long-lived
credential with access to administer DNS.

We try to avoid giving Jenkins long lived tokens with high levels of
privilege. There are a few reasons why it's bad:

- The token is set to expire in the year 9999. A disgruntled former
  employee could steal this token and use it later.
- Using the service account complicates auditing who made a change to
  DNS - it just shows up as Jenkins in the GCP audit logs.
- Jenkins itself could be compromised (e.g. due to a bad plugin). If it
  always has access to privileged tokens, that could lead to an
  escalation.

For AWS, we have a pattern of pasting short lived STS credentials into
Jenkins.

Now that we've updated the GCP terraform provider to a version which
supports GOOGLE_OAUTH_ACCESS_TOKEN we can do the same thing for Google.

This requires that everyone who needs to make changes to DNS has the
`gcloud` CLI installed.

If this is approved, further work will be needed to:

- [ ] Update the documentation on how to deploy DNS
- [ ] Update the on-call checklist to make it clear that you need to have `gcloud` set up
- [ ] Email all the engineers on 2ndline to make them aware of the requirement to have `gcloud` set up
- [x] Update the GDS CLI (which has a function where it can fill in the form in Jenkins for you)

https://trello.com/c/LasnNuz4/561-rationalise-use-of-gcp-credentials-eg-for-dns-deployments

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_dns.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_dns.yaml.erb
@@ -27,10 +27,6 @@
             colormap: xterm
         - build-name:
             name: '${ENV,var="PROVIDERS"} ${ENV,var="ZONE"} ${ENV,var="ACTION"}'
-        - credentials-binding:
-            - text:
-               credential-id: <%= @gce_credential_id %>
-               variable: GOOGLE_CREDENTIALS
     builders:
         - shell: |
             export GOOGLE_PROJECT=<%= @gce_project_id %>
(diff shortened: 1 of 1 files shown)


sengi (approved):
LGTM.

Took me a minute to figure out where GOOGLE_OAUTH_ACCESS_TOKEN is actually used; turns out it's a [Terraform thing](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference).

We'll definitely need to [update gds-cli](https://github.com/alphagov/gds-cli/blob/main/pkg/govuk/dns.go) in order to use this. Pretty sure nobody's copy-pasting tokens any more (at least I hope they're not!)

richardTowers:
> Pretty sure nobody's copy-pasting tokens any more (at least I hope they're not!)

image

... I'll update gds-cli.

richardTowers:
PR to update gds-cli: https://github.com/alphagov/gds-cli/pull/359

I'll raise a docs PR tomorrow before merging this.

PR #11173, opened 2021-06-24
Fix Account SLOs need a critical threshold

huwd:
In commit x looks like I made puppet really sad.

```
Error: Could not retrieve catalog from remote server: Error 400 on
SERVER: Must pass critical to
Icinga::Check::Graphite[check_slo_error_rate_10_min_accounts] at
/usr/share/puppet/production/current/modules/monitoring/manifests/checks/accounts_slos.pp:27

Warning: Not using cache on failed catalog Error: Could not retrieve
catalog; skipping run
```

I believe this is because i had the bright idea to only set this up as a
warning. Mostly because it's an alert as an SLO early warning, and
critical seemed a bit alarmist.

So I've added a critical status as if we exceed 10% of an error budget
in a rolling 10 minute window. That'd be 10 times worse than the
warning, so seems fair.

### DIFF
--- a/modules/monitoring/manifests/checks/accounts_slos.pp
+++ b/modules/monitoring/manifests/checks/accounts_slos.pp
@@ -12,6 +12,8 @@ class monitoring::checks::accounts_slos (
   if $enabled {
     # Warn when error rate exeeds 1% in any 10 minute rolling window
     $alert_warning_high_http_error_rate = 0.01
+    # Turns out puppet gets sad when we don't have a critical, so set it so something impossible.
+    $critical_warning_high_http_error_rate = 2
 
     $http_bad_metric = 'sumSeries(transformNull(stats_counts.*.nginx_logs.account-api.http_{422,5*}))'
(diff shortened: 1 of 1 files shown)


huwd:
I think this is the solution, will test on integration now.
If it is... it's a tad frustrating that linting doesn't offer an opinion on a missing "critical".

sihugh (approved):

barrucadu (commented):
You could set a critical threshold which will never be reached, like 2.

huwd:
yeah? I thought about that... but then 10% does seem pretty bad... perhaps we should get a nice red box for that?
... then again perhaps that's all too possible and we'll just mess up 2ndlines day 🤔

barrucadu (approved):
I think it doesn't matter too much, our error rate has been consistently very low, other than that one time

huwd:
Done!

huwd:
just checked this with the 10% value deploys fine on integration, so I'll send this down the pipe once you're 👍🏻  @barrucadu

barrucadu (approved):

PR #11179, opened 2021-06-28
fix logstream/logship

fredericfran-gds:
Since the upgrade to Python 2.7.14 (see #11122) where the pip module binaries are stored in /opt/python2.7/bin, we need to update the PATH env for logstream/logship scripts so that they know where to find logship

### DIFF
--- a/modules/govuk_logging/templates/logstream_metrics.erb
+++ b/modules/govuk_logging/templates/logstream_metrics.erb
@@ -9,6 +9,9 @@ respawn
 # and should be killed off.
 respawn limit 5 20
 
+# Add the path for Python modules after the upgrade to Python 2.7.14 since logship is a Python module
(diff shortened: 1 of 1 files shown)


modules/govuk_logging/templates/logstream_metrics.erb
```diff
@@ -9,6 +9,9 @@ respawn
 # and should be killed off.
 respawn limit 5 20
 
+# Add the path for Python modules after the upgrade to Python 2.7.14 since logship is a Python module
+env PATH=/opt/python2.7/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games
```
sihugh, line 13: do we need `/usr/games` :-)

sihugh (approved):

modules/govuk_logging/templates/logstream_metrics.erb
fredericfran-gds: lol, yeah, I send the logs of my gaming sessions to graphite ;-)

PR #11180, opened 2021-06-29
Rewrite publisher route to organization (Only merge in after CKAN 2.9 released)

kentsanggds:
## What

CKAN 2.9 doesn't have the publisher route, so rewrite any urls going to publisher to go to organization instead.

### DIFF
--- a/modules/govuk/templates/ckan/nginx.conf.erb
+++ b/modules/govuk/templates/ckan/nginx.conf.erb
@@ -71,6 +71,11 @@ location /api/3<%= path %> {
 }
 <% end %>
 
+# Rewrite publisher to organization for CKAN 2.9 in case it has been bookmarked
+location /publisher {
+  rewrite ^(.*)/publisher(/.*)?$ $1/organization$2 last;
+}
+
 location /csw {
   <%- if @protected -%>
   deny all;


modules/govuk/templates/ckan/nginx.conf.erb
```diff
@@ -71,6 +71,11 @@ location /api/3<%= path %> {
 }
 <% end %>
 
+# Rewrite publisher to organization for CKAN 2.9 in case it has been bookmarked
+location /publisher {
+  rewrite ^(.*)/publisher(/.*)?$ $1/organization$2 last;
```
sengi, line 76: Have you tested this? I think it might create a loop. If I've understood the docs correctly, `last` means "don't process any more rewrite rules for the current URL, instead go straight to processing `location` blocks for the rewritten URL". See the first rewrite example [here](http://nginx.org/en/docs/http/ngx_http_rewrite_module.html).

I think you may need to either change `last` to `break`, or not put the rewrite rule inside `location /publisher`.

I haven't actually tested it though, so I could be wrong :)

sengi, line 76: If this is for redirecting people who may have bookmarked the old location, consider using 301 Moved Permanently (`permanent`).

kentsanggds: I've tried it and it works, I don't mind changing it to `break` instead.

kentsanggds: I might give that a try

sengi: If it works then it's almost certainly fine - I must have got it wrong :)

sengi (approved):

kentsanggds:
Using redirect instead or rewrite

PR #11184, opened 2021-07-02
Add interest cohort permission policy

bevanloon:
Add header to our Nginx config to set the permission policy header such
that `interest-cohort=()`. This follows on from [RFC 142]((https://github.com/alphagov/govuk-rfcs/blob/main/rfc-142-add-interest-cohort-permssion-policy-header.md)

Co-authored-by: Richard Towers <richard.towers@digital.cabinet-office.gov.uk>
Co-authored-by: Himal Mandalia <himal.mandalia@digital.cabinet-office.gov.uk>

### DIFF
--- a/modules/nginx/files/etc/nginx/add-permissions-policy.conf
+++ b/modules/nginx/files/etc/nginx/add-permissions-policy.conf
@@ -1,0 +1,1 @@
+add_header Permissions-Policy interest-cohort=();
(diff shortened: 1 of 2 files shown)


sihugh (approved):
This is what I'd do

richardTowers (approved):
I paired on this, but 👍🏻  anyway.

We considered setting the permissions policy everywhere the `add-sts.conf` file is included (there are lots of other examples). But this works, and it feels cleaner just to set it once.

PR #11187, opened 2021-07-05
Map Prometheus healthcheck requests through nginx.

sengi:
Prometheus load balancer healthchecks are currently failing with a 500 response from nginx because they end up on the default vhost (because ALB/ELB can't be made to send the Host header on healthchecks).

This maps the standard Prometheus healthcheck paths through the nginx default vhost to Prometheus.

Access to Prometheus is controlled by security groups so this doesn't change anything from an access perspective.


bilbof (approved):
LGTM

PR #11188, opened 2021-07-06
Add new environment variable GOVUK_ENVIRONMENT_NAME

karlbaker02:
This commit adds a new environment variable `GOVUK_ENVIRONMENT_NAME` across all environments. This environment variable is intended to act as the canonical source of truth of the environment name for the purposes of setting the environment label in the layout headers of apps; currently this is done through a mixture of older apps using the deprecated `govuk_admin_template` gem and getting environment information set via an initializer file as part of a Jenkins deploy, and through abuse of the environment variable `ERRBIT_ENVIRONMENT_NAME` (we no longer use Errbit so this environment variable shouldn't even be around).

Once this new environment variable is added, we can transition apps to use it, starting out with all of those apps which are using `govuk_admin_template` (transitioning these apps to be 12 factor in the process) ([here's an example](https://github.com/alphagov/publisher/pull/1477)).


kevindew (approved):
Looks good 👍

PR #11190, opened 2021-07-06
Fix: Prometheus liveness check is /-/healthy not /-/live.

sengi:
I messed this up in 9e32a54. We're not actively using this at the moment but might as well get it right.


bilbof (approved):
Sorry I should have caught this in review. Thanks for the quick fix!

PR #11192, opened 2021-07-08
Redirect publisher route to organization (Only merge after CKAN 2.9 upgraded)

kentsanggds:
## What

CKAN 2.9 doesn't have the publisher route, so rewrite any urls going to publisher to go to organization instead

## Reference


smford (approved):
lgtm

PR #11200, opened 2021-07-14
Set up sidekiq for account-api

barrucadu:
We're setting up sidekiq for account-api, which means we need to set
up redis env vars for it first.  When this commit has been deployed,
we can deploy account-api with the new worker code, and then deploy a
second puppet change to enable the worker.

This commit also sets up sidekiq-monitoring and the application
dashboard.  Those won't have any metrics to show until the worker is
enabled though.

---

[Trello card](https://trello.com/c/uRW8OS61/907-move-expired-authrequest-deletion-to-sidekiq)

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -432,6 +432,7 @@ govuk::apps::content_data_api::port: 3235
 # search-admin sidekiq monitoring uses 3241
 # sidekiq monitoring uses 3242
 govuk::apps::account_api::port: 3243
+# account-api sidekiq monitoring uses 3244
 
 govuk::apps::asset_manager::jwt_auth_secret: "%{hiera('jwt_auth_secret')}"
 govuk::apps::asset_manager::mongodb_nodes:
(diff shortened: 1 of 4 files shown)


hieradata_aws/common.yaml
```diff
@@ -786,6 +789,8 @@ govuk::apps::account_api::db::lb_ip_range: "%{hiera('environment_ip_prefix')}.0.
 govuk::apps::account_api::enable_publishing_queue_listener: true
 govuk::apps::account_api::rabbitmq::enabled: true
 govuk::apps::account_api::rabbitmq_hosts: [rabbitmq]
+govuk::apps::content_data_admin::redis_host: "%{hiera('sidekiq_host')}"
+govuk::apps::content_data_admin::redis_port: "%{hiera('sidekiq_port')}"
```
bevanloon, line 793: /s/content_data_admin/account_api ?

barrucadu: 🤦

bevanloon (approved):

PR #11201, opened 2021-07-15
Enable account-api procfile worker

barrucadu:
The worker code has been deployed now, so it can be switched on.

---

[Trello card](https://trello.com/c/uRW8OS61/907-move-expired-authrequest-deletion-to-sidekiq)


bevanloon (approved):
🚀

PR #11203, opened 2021-07-16
Disable harvest workers and set ckan maintenance redirect (Only merge when initiating CKAN 2.9 upgrade)

kentsanggds:
## What

Disable CKAN harvest workers and set the site wide redirect to the CKAN maintenance page.

## Reference 


smford (approved):
lgtm

PR #11204, opened 2021-07-16
Upgrade ckan 2.9 (Only merge after CKAN redirection to maintenance page)

kentsanggds:
## What

Update config files for CKAN 2.9 and update cronjobs with new ccommands for Staging and Production stacks.

## Reference 

### DIFF
--- a/modules/govuk/manifests/apps/ckan.pp
+++ b/modules/govuk/manifests/apps/ckan.pp
@@ -170,7 +170,7 @@ class govuk::apps::ckan (
       process_regex  => 'pycsw\.wsgi',
     }
 
-    class { 'cronjobs29':
+    class { 'cronjobs':
       ckan_port    => $port,
       pycsw_config => $pycsw_config,
     }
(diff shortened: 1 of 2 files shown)


hieradata_aws/class/integration/ckan.yaml
```diff
@@ -3,5 +3,13 @@
 govuk_solr::disable: true
 govuk_solr6::present: true
 
+govuk::apps::ckan::blanket_redirect_url: https://find-data-beta-integration.cloudapps.digital/ckan_maintenance
```
smford, line 6: are you wanting a:

'govuk::apps::ckan::enabled: true'

somewhere here?

kentsanggds: It's currently working, not sure if we need to explicitly declare that

smford: ok cool!

smford (approved):
lgtm!

PR #11209, opened 2021-07-21
install docker before associating docker group to directory

fredericfran-gds:
Mirrorer is breaking in production with error:
```
Error: Could not find group docker
Error: /Stage[main]/Govuk_containers::Redis/File[/srv/redis]/group: change from root to docker failed: Could not find group docker
```

This fixes the issue by installing `govuk_docker` (hence have the `docker` group) before giving `docker` group ownership of `/srv/redis` directory

PR #11210, opened 2021-07-21
fix for #11209

fredericfran-gds:
fixes #11209 which is now locking when applying when the puppet config. Maybe a dependency cycle: https://tickets.puppetlabs.com/browse/PUP-5596


rtrinque (approved):
looks good

PR #11212, opened 2021-07-22
fix deploy uid/gid for deploy user in mirrorer machine

fredericfran-gds:
Do not set recurse on the mirror directory as it consists of a hundreds of GBs
of small files and will cause puppet to be take a long time to finish
running and may cause the Mirrorer instance to fail first instantiation.
Instead we set the uid and gid of `deploy` to be the same always via hiera.

### DIFF
--- a/hieradata_aws/class/mirrorer.yaml
+++ b/hieradata_aws/class/mirrorer.yaml
@@ -13,3 +13,6 @@ mount:
     disk: '/dev/mapper/crawler-worker'
     govuk_lvm: 'worker'
     mountoptions: 'defaults'
+
+govuk::deploy::setup::deploy_uid: 2899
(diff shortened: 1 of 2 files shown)


hieradata_aws/class/mirrorer.yaml
```diff
@@ -13,3 +13,6 @@ mount:
     disk: '/dev/mapper/crawler-worker'
     govuk_lvm: 'worker'
     mountoptions: 'defaults'
+
+govuk::deploy::setup::deploy_uid: 2899
+govuk::deploy::setup::deploy_gid: 2899
```
karlbaker02, line 18: Did you mean to set this twice?

karlbaker02 (approved):
Looks good 💯

PR #11214, opened 2021-07-22
Upgrade seed crawler gem to 3.0.0

bilbof:
This also installs the Ruby version 2.6.3 which is required by the gem.


fredericfran-gds (approved):
Thanks!

PR #11215, opened 2021-07-22
install govuk_seed_crawler using Ruby 2.6.3

fredericfran-gds:
The new version of govuk_seed_crawler (3.0.0) needs Ruby 2.6.3
and we need to use puppet Exec to install rather than Package
since Package will use the system Ruby which is set to 1.9.3.


karlbaker02 (approved):
Looks good 👍

PR #11218, opened 2021-07-27
Fix formatting of description of Deploy DNS job for GCP

barrucadu:
The description is displayed on a single line.  It currently looks like this:

Screenshot 2021-07-27 at 14 17 30

Not realising it was three separate commands at first, I copied the whole `gcloud` command, which didn't work.  So, change each line to be a sentence ending with a full stop, and separate the commands with semicolons, to make it easier to read.


richardTowers (approved):
Ah, thanks for this. I didn't realise Jenkins was going to render the description like that.

PR #11222, opened 2021-08-02
Ensure namevar for content-data-api rds_config is unique

benjamineskola:
Namevar for exported resources must be unique; until now there had never been more than one instance with this resource so it was never an issue, but this is a requirement if we ever want to scale this up.


hannako (approved):
👍

PR #11225, opened 2021-08-03
Set CheckHostIP to no in reboot-cache-instances script

edwardkerry:
When running this script checking the known_hosts file can
cause the SSH connection to be lost if the IP address has changed
since last connecting.

Users of this script would manually remove the problematic line
from the known_hosts file, and then have to either re-run the script
or more likely have to manually investigate the current state and
continue making changes in the AWS UI.

This commit removes this check allowing the script to complete.

https://man.openbsd.org/ssh_config#CheckHostIP

### DIFF
--- a/tools/reboot-cache-instances.sh
+++ b/tools/reboot-cache-instances.sh
@@ -60,7 +60,7 @@ echo "Waiting for requests to complete... 30 seconds"
 sleep 30
 
 echo "Rebooting instance via ssh"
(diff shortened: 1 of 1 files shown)


gclssvglx (approved):
Great work @edwardkerry - nice one 👍   Just one suggestion, might be worth adding a link to the man doc (https://man.openbsd.org/ssh_config#CheckHostIP) for reference.

edwardkerry:
I've only updated the commit message following the approving review

PR #11226, opened 2021-08-05
Add Plek env var for account manager to email-alert-frontend

danacotoran:
We are making the email management pages look like the account if the
user's signed in through the account. Therefore there will be links from
`email-alert-frontend` back to the account manager and account feedback
page which is what this env var is needed for.

----

https://trello.com/c/F0ZDY0RC


barrucadu (commented):
~~I don't think we need this, as apps communicate with the account manager through the account-api, which is already set up.~~

barrucadu (approved):

PR #11227, opened 2021-08-05
Re enable harvesting on integration and staging

kentsanggds:
## What

Re-enable the harvest workers on Integration and Staging so that they can be tested before deploying CKAN 2.9 upgrade to production.

## Reference

### DIFF
--- a/modules/govuk/manifests/apps/ckan/cronjobs.pp
+++ b/modules/govuk/manifests/apps/ckan/cronjobs.pp
@@ -59,7 +59,7 @@ class govuk::apps::ckan::cronjobs(
   }
 
   govuk::apps::ckan::ckan_cronjob { 'Environment Agency harvester stop existing processes':
-    ensure       => $ensure,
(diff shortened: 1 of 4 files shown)


modules/govuk/manifests/apps/ckan/cronjobs.pp
```diff
@@ -59,15 +59,15 @@
...
+    ensure       => 'absent',
     ckan_command => 'harvester job-abort environment-agency-data-sharing-platform',
     weekday      => '5',
     hour         => '15',
     minute       => '55',
   }
 
   govuk::apps::ckan::ckan_cronjob { 'Environment Agency harvester run':
-    ensure       => $ensure,
+    ensure       => 'absent',
```
sengi, line 70: I don't have context on this so I'm just double-checking that this commit is here intentionally, since you didn't mention it in the PR description.

sengi (approved):

PR #11231, opened 2021-08-09
Rewrite ckan healthcheck

kentsanggds:
## What

A healthcheck is failing as it is attempting to go to `_healthcheck` so add a rewrite in nginx to return the response from `healthcheck` endpoint.

## Reference 

https://trello.com/c/7HPu6jyN/2647-deployment-of-ckan-29-steps


karlbaker02 (approved):
Looks like a reasonable change 👍

PR #11234, opened 2021-08-11
Remove Splunk Log forwarding from GOV.UK

smford:
We no longer need splunk and this PR begins the removal process of
splunk from our systems.  Splunk was installed as a test and is no
longer needed.

This will be a two phased release:
Release 1: Release this PR which will remove all the splunk stuff from
the servers
Release 2: A second govuk-puppet PR which will delete this module, which
can then be deployed at the same time as the removal of the secrets in
this PR: https://github.com/alphagov/govuk-secrets/pull/1191


smford:
Abandoning - theres a better way, raising another pr.

Issue #1235, opened 2015-12-16
Fix check_unicorn_workers to warn when too few processes


alexmuller:
This pull request was [originally opened on GitHub Enterprise](https://github.gds/gds/puppet/pull/1235) on Tuesday 07 January 2014 at 15:55 GMT before GOV.UK's Puppet code was migrated to a public GitHub repository.

PR #11236, opened 2021-08-11
Remove Splunk Log forwarding from GOV.UK - Part 2

smford:
This PR removes the govuk_splunk module from puppet, and stops the
calculators using this module.

Pair: @smford & @fredericfran-gds


fredericfran-gds:
This is ready to run after:
1. #1235 (with fix #1235) in staging to remove the Splunk agent and dependencies.
2. since the Splunk module was not written with removal in mind, had to also applied temporary PR: #11240

Next time, the best way around might be to delete the Splunk code in Puppet (becomes unmanaged) and then recreate the the `calculators_frontend` machines in Staging which are the only machines where Splunk is installed.

benjamineskola (approved):

PR #11237, opened 2021-08-11
Set the correct instance id in RDS checks for content-data-api

leenagupte:
# What's changed and why?

This value used to be set to title, but title was [updated to be unique](ec0525e38ea5ed509cf0e945a37ee56ad4276f4f) to allow more instances to be monitored.

We need to allow title to remain unique but still point the check at the correct instance.

See errors:
https://alert.staging.govuk.digital/cgi-bin/icinga/extinfo.cgi?type=2&host=ip-10-12-5-36.eu-west-1.compute.internal&service=content-data-api-postgresql-primary_ip-10-12-4-10+-+AWS+RDS+CPU+Utilization
https://alert.blue.production.govuk.digital/cgi-bin/icinga/extinfo.cgi?type=2&host=ip-10-13-5-163.eu-west-1.compute.internal&service=content-data-api-postgresql-primary_ip-10-13-4-79+-+AWS+RDS+Memory+Utilization
https://alert.blue.production.govuk.digital/cgi-bin/icinga/extinfo.cgi?type=2&host=ip-10-13-5-163.eu-west-1.compute.internal&service=content-data-api-postgresql-primary_ip-10-13-4-79+-+AWS+RDS+Storage+Utilization

### DIFF
--- a/modules/govuk/manifests/apps/content_data_api/db.pp
+++ b/modules/govuk/manifests/apps/content_data_api/db.pp
@@ -16,6 +16,7 @@ class govuk::apps::content_data_api::db (
   }
 
   @@monitoring::checks::rds_config { "content-data-api-postgresql-primary_${::hostname}":
+    instance_id      => 'content-data-api-postgresql-primary',
     memory_warning   => 2,
     memory_critical  => 1,
     storage_warning  => 50,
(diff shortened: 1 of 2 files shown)


richardTowers:
The thing I'm not sure about here is this bit:

https://github.com/alphagov/govuk-puppet/blob/main/modules/monitoring/manifests/checks/rds.pp#L53-L55

Which calls this with a list of `$servers`.  I think this might need updating to pass `instance_id`, but I don't know how to do that...

richardTowers:
Superseded by #11238

PR #11241, opened 2021-08-13
Fix hieradata for RDS checks

richardTowers:
I'd added content-data-api to common.yaml, but I hadn't realised it was overridden to the same value in staging.yaml and production.yaml.

Because I'm overriding this specifically in `modules/monitoring/manifests/checks/rds.pp` this causes puppet to error on staging and production (where the value is missing).

To fix this, I've removed the unnecessary overrides in staging / prod, and added a comment to warn people that the override exists.

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -1433,7 +1433,11 @@ monitoring::checks::cache::servers:
   - 'backend-redis-002'
 
 monitoring::checks::rds::servers:
+   # NOTE: content-data-api-postgresql-primary is overridden in
+   # modules/monitoring/manifests/checks/rds.pp. If it is removed from this
+   # list, the override will also need to be removed.
    - 'content-data-api-postgresql-primary'
+
    - 'postgresql-primary'
    - 'postgresql-standby'
    - 'transition-postgresql-primary'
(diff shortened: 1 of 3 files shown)


hieradata_aws/common.yaml
```diff
@@ -1433,7 +1433,11 @@ monitoring::checks::cache::servers:
   - 'backend-redis-002'
 
 monitoring::checks::rds::servers:
+   # NOTE: content-data-api-postgresql-primary is overridden in
+   # modules/monitoring/manifests/checks/rds.pp. If it is removed from this
+   # list, the override will also need to be removed.
    - 'content-data-api-postgresql-primary'
+
```
fredericfran-gds, line 1440: I wonder what will be the effect of this empty line?

sihugh (approved):

fredericfran-gds (commented):
Thanks for fixing, added one comment

hieradata_aws/common.yaml
fredericfran-gds: accoridng to yq, this is allowed

richardTowers: Yeah, I'm pretty sure it has no effect on the semantics. I just wanted it there to make it clear which line the comment referred to.

PR #11245, opened 2021-08-17
Update ckan and harvest processes collectd regex

kentsanggds:
## What

Update the collectd regex to correctly pick up the processes running on the server.

Checked the metrics on graphite integration and data points are appearing.

## Reference 

https://trello.com/c/ziy3xJoN/2654-fix-alert-on-ckan-harvest-workers-not-found


owenatgov (approved):

fredericfran-gds (approved):
looks good, thanks for testing and fixing

PR #11248, opened 2021-08-18
Install Terraform version 0.12

bilbof:
We need TF version 12 to upgrade from 0.11 to 0.13. HashiCorp don't recommend upgrading more than one major (minor) version at a time.


fredericfran-gds (approved):
approve, might need to give a bit more context on which project you want to update.

PR #11252, opened 2021-08-19
Upgrade all Jenkins plugins without compatibility issues

benjamineskola:
These are all the plugins with available upgrades, excluding the ones marked as incompatible (which will need manual intervention).

https://trello.com/c/pCKvKF2J/2583-upgrade-jenkins-plugins-to-versions-without-security-vulnerabilities-5


ChrisBAshton (approved):
LGTM 👍 has this been rolled out to all Jenkins environments?

benjamineskola:
@ChrisBAshton No, it's currently on Staging deploy and Integration CI .

PR #11257, opened 2021-08-26
Fix problem when provisioning new `monitoring` machine instance

brucebolt:
Puppet fails when a new instance of the `monitoring` machine class is created.  This causes multiple problems, including the inability to SSH into the new machine.

The errors were caused by Icinga check titles either not being unique (where they should've been) or including the hostname where this caused a check to not be found in the Puppet catalog.

This fixes the errors by ensuring the titles for Icinga check resources are unique where required, and not unique for checks running solely on the `monitoring` class where the test including the other machine instance(s) hostnames would not be found in the Puppet catalog.

[Trello card](https://trello.com/c/R5qp0I7o)

### DIFF
--- a/lib/tasks/icinga_checks.rake
+++ b/lib/tasks/icinga_checks.rake
@@ -1,13 +1,13 @@
 desc "Test icinga::checks are unique per machine"
 task :icinga_checks do
   $stderr.puts '---> Checking icinga::check titles are sufficiently unique'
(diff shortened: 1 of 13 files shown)


benjamineskola (approved):

ChrisBAshton (commented):
Nice job, @brucebolt 👏

PR #11258, opened 2021-08-31
Run app healthchecks as part of CD pipeline

benthorner:
https://trello.com/c/USH1B9K1/243-investigate-drying-up-deployment-healthchecks-out-of-smokey

One of the consequences of the CD RFC [1].

Currently these are written as bespoke scenarios in Smokey [2],
which is more work and confuses the purpose of the smoke tests
and leads to duplicate Icinga alerts. This makes the healthchecks
part of the CD pipeline so we can remove the bespoke scenarios.

Some apps don't support a web-based healthcheck, such as Cache
Clearing Service. In such cases we can use an empty array of URLs
to represent this. Other apps use a custom (Contacts Admin) or
even multiple domains, which we need to set explicitly.

Note that, unlike the existing scenario for Whitehall [3], this
tests both the Admin and Frontend instances due to the distinct
functionality. Conversely, it's not worth testing draft instances
of frontend apps as they behave the same as non-draft instances.

[1]: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-128-continuous-deployment.md#healthy-deployments
[2]: https://github.com/alphagov/smokey/blob/cae82f827db5153f5545c342bd0f6ff19289e189/features/authenticating_proxy.feature
[3]: https://github.com/alphagov/smokey/blob/cae82f827db5153f5545c342bd0f6ff19289e189/features/whitehall.feature#L37-L41

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
@@ -30,13 +30,29 @@
           fi
 
           # Workaround for our inconsistent repo vs. deployment naming
+          # and multiple / different healthcheck URLs
           case "$TARGET_APPLICATION" in
           <% @applications.keys.each do |app| %>
+            <% default_healthcheck_urls = [
+              "https://#{app}.#{@app_domain}/healthcheck/ready"
+            ] %>
+
             <%= app %>)
               REPO="<%= @applications[app].fetch("repository", app) %>"
+
+              HEALTHCHECK_URLS="<%= @applications[app].fetch(
+                "healthcheck_urls", default_healthcheck_urls
+              ).join(" ")%>"
               ;;
           <% end %>
           esac
+
+          <% if @smokey_pre_check %>
+          # Check healthcheck endpoints return 200 OK
+          for url in $HEALTHCHECK_URLS; do
+            curl --fail $url;
+          done
+          <% end %>
 
           # Check release to deploy is genuine and we're not going backwards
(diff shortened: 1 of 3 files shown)


benthorner:
Adding "DNM" as commit temporarily removes Slack notification (for testing).

modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb
```diff
@@ -30,14 +30,30 @@
...
+                "healthcheck_urls", default_healthcheck_urls
+              ).join(" ")%>"
               ;;
           <% end %>
           esac
 
+          <% if @smokey_pre_check %>
+          # Check healthcheck endpoints return 200 OK
+          for url in $HEALTHCHECK_URLS; do
+            curl $url;
```
barrucadu, line 53: I think you'll want `--fail` or `--fail-with-body` here, to make `curl` return a nonzero exit code

benthorner: Good spot - I've added `--fail` and tested it works as expected.

benthorner:
Removing "DNM" as the commit is now ready to go.

Testing on Integration with a variety of apps:

- All apps that have a healthcheck URL override ✅.
- A few apps that don't have an override ✅ .
- An artificially failing healthcheck ([log here](https://deploy.integration.publishing.service.gov.uk/job/Deploy_App_Downstream/3555/console)) ✅ .

barrucadu (approved):
Looks good to me

PR #11259, opened 2021-09-02
Set account manage / security URI env vars for static in integration

barrucadu:
Note: the security page isn't implemented yet, but this will be the URI.

These env vars are used by govuk_personalisation to determine the URLs
which ultimately appear in the account template returned by static.

When Digital Identity settle on a permanent URI scheme, we may be able
to simplify further (eg, if their stuff is all at
`https://manage.${environment}.account.gov.uk/${path}`).

---

[Trello card](https://trello.com/c/ZvDGaba4/958-use-di-auth-in-integration-give-feedback-on-their-documentation)


aldavidson (approved):
Looks good to me :+1:

PR #11261, opened 2021-09-03
Rename account-api session secret env var

barrucadu:
It's not a key, it's the secret used to derive the key.


richardTowers (approved):
Looks good to me - don't forget to release account-api before you deploy this :)

PR #11263, opened 2021-09-06
Enable CD for email alert service

MuriloDalRi:
https://trello.com/c/AGNZWDFJ/2693-enable-continuous-deployment-for-email-alert-service-3


benthorner:
@MuriloDalRi you'll need to change this as https://github.com/alphagov/govuk-puppet/pull/11258 is now merged.

By the way, this app still seems to be missing [a required healthcheck](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-128/compatibility.md) for connectivity to Redis, which will be a little be tricky to implement as it's not a web app - you'll need to do it as part of the startup code.

PR #11266, opened 2021-09-08
Delay slow-running jobs to start at 7am instead of 3am

ChrisBAshton:
Since upgrading our core Jenkins and our Jenkins plugins - though
not necessarily related - we're seeing issues with some jobs
getting 'stuck', meaning there are fewer executors available and
there can be a build up of jobs. Certain jobs, if not run
sufficiently quickly, trigger PagerDuty alerts, causing our
out-of-hours developers to get called:

`CRITICAL: 'Travel Advice email alert check' on 'ip-10-13....'`

The short term fix is to cancel any hanging jobs, to free up the
executors in running the time-sensitive jobs. Whilst we investigate
the root cause of why jobs are hanging, we don't want any more
unnecessary call-outs to our on-call developers, so will change
the start time of these particular jobs to be 7am instead of 3am.

These jobs take around 2 hours to complete, so any callout would
now happen at around 9am instead of 5am, but the chances of a
call-out are diminished because the actual urgent jobs would have
executed by then.
The jobs we've delayed here don't seem particularly time-sensitive,
so we can afford to delay their start time.

### DIFF
--- a/spec/fixtures/hieradata/common.yaml
+++ b/spec/fixtures/hieradata/common.yaml
@@ -71,7 +71,7 @@ govuk_jenkins::config::github_web_uri: wibble
 govuk_jenkins::jobs::deploy_app::applications: *deployable_applications
 govuk_jenkins::jobs::deploy_app_downstream::applications: *deployable_applications
 govuk_jenkins::jobs::run_rake_task::applications: *deployable_applications
-govuk_jenkins::jobs::content_data_api_re_run::re_run_rake_etl_main_process_cron_schedule: '0 3 * * *'
+govuk_jenkins::jobs::content_data_api_re_run::re_run_rake_etl_main_process_cron_schedule: '0 7 * * *'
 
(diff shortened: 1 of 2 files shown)


benjamineskola (approved):

spec/fixtures/hieradata/common.yaml
```diff
@@ -71,7 +71,7 @@ govuk_jenkins::config::github_web_uri: wibble
 govuk_jenkins::jobs::deploy_app::applications: *deployable_applications
 govuk_jenkins::jobs::deploy_app_downstream::applications: *deployable_applications
 govuk_jenkins::jobs::run_rake_task::applications: *deployable_applications
-govuk_jenkins::jobs::content_data_api_re_run::re_run_rake_etl_main_process_cron_schedule: '0 3 * * *'
+govuk_jenkins::jobs::content_data_api_re_run::re_run_rake_etl_main_process_cron_schedule: '0 7 * * *'
```
ChrisBAshton, line 74: In hindsight, this didn't work, because it was a change to the specfile rather than the actual config 🤦

PR #11269, opened 2021-09-09
Downgrade priority of content-data-api healthcheck failures

bilbof:
When the health check fails we'll no longer get paged.
The checks now use the govuk_regular_service service template,
so we'll see the alert only if we check Icinga.

Content Data API not having the most up to date data is not a critical problem and can be retrospectively fixed, so does not need the immediate attention of the on-call/2nd-line developers


ChrisBAshton (approved):
The change is good 👍 might be worth expanding on why we've come to this decision though. Something like: "Content Data API not having the most up to date data is not a critical problem and can be retrospectively fixed, so does not need the immediate attention of the on-call/2nd-line developers."

PR #11270, opened 2021-09-09
Update jenkins config to match Jenkins

chao-xian:
The difference is causing Jenkins to reload. Similar to what occurred in https://github.com/alphagov/govuk-puppet/pull/11073

### DIFF
--- a/modules/govuk_jenkins/templates/config/config.xml.erb
+++ b/modules/govuk_jenkins/templates/config/config.xml.erb
@@ -1,13 +1,12 @@
 <?xml version='1.1' encoding='UTF-8'?>
 <hudson>
   <disabledAdministrativeMonitors>
-    <string>hudson.diagnosis.ReverseProxySetupMonitor</string>
+    <string>jenkins.security.RekeySecretAdminMonitor</string>
     <string>hudson.diagnosis.TooManyJobsButNoView</string>
     <string>OldData</string>
-    <string>jenkins.security.RekeySecretAdminMonitor</string>
+    <string>hudson.diagnosis.ReverseProxySetupMonitor</string>
   </disabledAdministrativeMonitors>
   <version><%= @version %></version>
-  <installStateName>RESTART</installStateName>
   <numExecutors><%= @executors %></numExecutors>
   <mode>NORMAL</mode>
   <useSecurity>true</useSecurity>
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/templates/config/config.xml.erb
```diff
@@ -102,9 +101,7 @@
   <workspaceDir>${JENKINS_HOME}/workspace/${ITEM_FULLNAME}</workspaceDir>
   <buildsDir>${ITEM_ROOTDIR}/builds</buildsDir>
   <systemMessage>&lt;h2 style=&quot;background-color: <%= @banner_colour_background -%>; color: <%= @banner_colour_text -%>; padding: 0.5em;&quot;&gt;<%= @banner_string -%>&lt;/h2&gt;</systemMessage>
-  <markupFormatter class="hudson.markup.RawHtmlMarkupFormatter" plugin="antisamy-markup-formatter@<%= @markup_formatter_version -%>">
-    <disableSyntaxHighlighting>false</disableSyntaxHighlighting>
-  </markupFormatter>
+  <markupFormatter class="hudson.markup.RawHtmlMarkupFormatter" plugin="antisamy-markup-formatter@2.1">
```
bilbof, line 104: Perhaps we should change `@markup_formatter_version` here?

modules/govuk_jenkins/templates/config/config.xml.erb
```diff
@@ -167,15 +164,21 @@
           <default>
             <comparator class="hudson.util.CaseInsensitiveComparator"/>
           </default>
+          <%- if @aws_migration -%>
```
bilbof, line 167: I'm not sure if we need this and the lines below it

ChrisBAshton (commented):
Nice job, @chao-xian 👏 Is it worth updating the [test](https://github.com/alphagov/govuk-puppet/pull/11073/files#diff-e163a5e6383f0c7a220a9b687601a96a44c29d33710b16a76025dda16658bfb8R38-R44) so that we catch this sort of thing from happening again?

It may need to be a pre-commit hook, as [suggested in the previous PR](https://github.com/alphagov/govuk-puppet/pull/11073#issuecomment-828535515)... 🤔

modules/govuk_jenkins/templates/config/config.xml.erb
chao-xian: No idea where these came from... 😕

chao-xian:
@ChrisBAshton thanks but I'm not clear what to test for. I'd like to get deploys fixed first off.

bilbof (approved):
:shipit:

PR #11273, opened 2021-09-10
Ensure the app memory Icinga checks are absent

ChrisBAshton:
These were removed in #11034, but they weren't set to
`ensure => 'absent'`, therefore the files themselves weren't
removed; Puppet just no longer cares about the files. This means
we're still seeing Icinga alerts concerning 'high memory for
whitehall-admin-procfile-worker', for example.

We first need to ensure absence, and once this has run in all
environments, we should be safe to remove this code.


benjamineskola (approved):

ChrisBAshton:
I think this was a spike and seem to remember it didn't fix the issue. Closing.

PR #11275, opened 2021-09-13
Add Alex Whitehead-Smith

alex9smith:
Add myself to the integration environment.

Note that I've created everything under my old surname (Smith). This is so my SSH username matches the username on my GDS laptop, which will hopefully prevent me having to type in my username every time I want to use it. The email address given here is an alias to the one on my Google account.

Hopefully this doesn't cause problems with the details in the [user monitor](alphagov/govuk-user-reviewer#568). If it does, I'll update that as well so it's all the same.

### DIFF
--- a/modules/users/manifests/alexsmith.pp
+++ b/modules/users/manifests/alexsmith.pp
@@ -1,0 +1,8 @@
+# Create the alexsmith user
+class users::alexsmith {
+  govuk_user { 'alexsmith':
+    fullname => 'Alex Smith',
+    email    => 'alex.smith@digital.cabinet-office.gov.uk',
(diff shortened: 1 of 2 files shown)


modules/users/manifests/alexsmith.pp
```diff
@@ -0,0 +1,8 @@
+# Create the alexsmith user
+class users::alexsmith {
+  govuk_user { 'alexsmith':
+    fullname => 'Alex Smith',
+    email    => 'alex.smith@digital.cabinet-office.gov.uk',
```
barrucadu, line 5: I think the email address in the user reviewer is only used in the AWS checks.  So if your AWS account is `alex.smith@...`, we'll have to update the reviewer.  If your AWS account is `alex.whitehead-smith@...`, I think it'd be better to use the same email address here for consistency.

alex9smith: Cool, that makes sense. I only just put in a request to create a `gds-users` AWS account (the DMp managed their own accounts), so I'll put in a new one that uses `alex.smith@...` and it should all be ok 🤞

barrucadu (approved):

PR #11279, opened 2021-09-17
Disable Signon data sync & create Cyber users for ITHC

barrucadu:


hieradata_aws/class/staging/db_admin.yaml
```diff
@@ -288,7 +288,7 @@ govuk_env_sync::tasks:
     url: "govuk-staging-database-backups"
     path: "mysql"
   "pull_signon_production_daily":
-    ensure: "present"
+    ensure: "disabled"
```
barrucadu, line 291: https://github.com/alphagov/govuk-puppet/blob/738b59aa50da0c1620b718e0500c4215170b283a/modules/govuk_env_sync/manifests/task.pp#L38-L40

huwd (approved):

PR #11286, opened 2021-09-28
Increase nginx timeout duration on hmrc manuals api 

hannako:
Publication of HMRC manuals via HMRC Manuals API is currently degraded. See zendesk [ticket](https://govuk.zendesk.com/agent/tickets/4712404)

See the following error:
```
ec2-integration-blue-backend-ip-10-1-6-26:/var/log/nginx$ tail -n 50 hmrc-manuals-api-error.log
upstream timed out (110: Connection timed out) while reading response header from upstream
```

### DIFF
--- a/modules/govuk/manifests/apps/hmrc_manuals_api.pp
+++ b/modules/govuk/manifests/apps/hmrc_manuals_api.pp
@@ -63,6 +63,7 @@ class govuk::apps::hmrc_manuals_api(
     has_readiness_health_check => true,
     log_format_is_json         => true,
     override_search_location   => $override_search_location,
+    read_timeout               => 300,
   }
 
   unless $ensure == 'absent' {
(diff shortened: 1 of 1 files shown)


modules/govuk/manifests/apps/hmrc_manuals_api.pp
```diff
@@ -63,6 +63,8 @@
     has_readiness_health_check => true,
     log_format_is_json         => true,
     override_search_location   => $override_search_location,
+    read_timeout               => 300,
+    send_timeout               => 300,
```
bilbof, line 67: 5 minutes seems like quite a large timeout given we cap the file size at 20mb. Would 30 seconds suffice?

bilbof (approved):
Configuration looks good so feel free to merge, just a question about the timeout.

modules/govuk/manifests/apps/hmrc_manuals_api.pp
brucebolt: I think we added a `send_timeout` more for trial and error purposes, because we were still seeing 60s timeouts after increasing the `read_timeout` (it turned out to be the load balance).  That probably wasn't needed, so I'll try this without changing the `send_timeout` and see we see an improvement.

brucebolt: I've confirmed we don't need to increase the `send_timeout`, so have reverted this PR back to only increasing the `read_timeout` and unicorn timeout.

bilbof (approved):

PR #11310, opened 2021-11-01
Update the on-call alerting times for the change to UTC

barrucadu:
This reverts commit da857b426d53751f93f80af872e433d31af5aed3.


brucebolt (approved):

brucebolt:
I wonder if this could be simplified with a boolean for whether DST is relevant at the moment or not?

PR #11317, opened 2021-11-11
Fix Licensing config following licensify#508

richardTowers:
In https://github.com/alphagov/licensify/pull/508 we changed the config
value we read the secret from application.secret to play.crypto.secret
in preparation for the upgrade from Play 2.5 to Play 2.6.

This looks like it has caused the code to fall back to the version of
the config in the licensing git repo, instead of the one in puppet,
which means that pre-deployment session cookies aren't decrypting
properly.

Using play.crypto.secret will mean the code gets the "correct" original
value for the cookie (from puppet). New sessions since the broken
release will be broken, but old sessions should work again.


dgm34 (approved):
👍

PR #11318, opened 2021-11-12
Add dana.cotoran to email-alert-api allowlist in staging

barrucadu:
This is so we can test the new single-page subscriptions feature.


barrucadu:
@danacotoran This will need deploying to staging before testing email subscriptions.

sihugh (approved):
You know about this though, right? https://docs.publishing.service.gov.uk/apps/email-alert-api/receiving-emails-from-email-alert-api-in-integration-and-staging.html

danacotoran:
Thanks @sihugh! I didn't know about that group actually... but I think Michael must have since he's a member. He's on holiday now, can't ask him! 
I actually think we could test what we need to using `email-alert-api-staging@digital.cabinet-office.gov.uk` so I will give that a go before I merge this in. 🤔

barrucadu:
Yeah, it was handy for me to have a bunch of email addresses in this list, because I wanted to be able to test several different scenarios.  But we've ben able to handle the other testing with the email-alert-api-staging group.

PR #11322, opened 2021-11-22
Update tag search to reenable Continuous Deployment

Tetrino:
A GitHub API change that flipped the Latest Tag list from Date to
Alphabetical order broke our continuous deployment script. This commit
alters the script such that we always retrieve the latest release tag
using graphql.

This commit also makes the same change to retrieving the commit tag for
the `main` branch, for consistency and future proofing.


ChrisBAshton (approved):
LGTM 👍 well done for getting to the bottom of the GraphQL/Curl fun!

Before merging, it would be great to test a continuous deployment from integration to staging, by building this branch of Puppet to both environments. I'd suggest trying both a standard release as well as a weirdly alphabetically named release.

PR #11324, opened 2021-11-23
Bugfix: Clear Static templates in clear-template-cache job

bilbof:
The templates were added to and renamed over the past 8 months.
This led to a confusing to investigate problem where old templates were served by Static due to caching.

This clears all cached templates when running the clear-template-cache job.

This is a short-term fix. The longer term fix is to stop storing the static templates in four layers of caches.


richardTowers (approved):
I did not know Rails let you cache files like this. Wow. Just writing them to `public`.

Looks like it's a whole thing though: https://github.com/rails/actionpack-page_caching

Definitely a good idea to clear all the files from that cache rather than cherry picking a few.

PR #11325, opened 2021-11-24
Exclude electoral page from traffic replay

sihugh:
It uses an external API that doesn't need to be hit twice for each request.

Is tested by https://github.com/alphagov/smokey/pull/874


hannako (approved):
👍

PR #11326, opened 2021-11-25
Deploy Static when adding/removing the emergency banner

bilbof:
Rather than clear four layers of caches when deploying or removing the emergency banner, we will instead reduce the
TTL of GOV.UK pages to 5 minutes, and wait for the caches to expire.

Static's cache does not expire automatically. Rather than implement a custom sweeper to clear Static's cache (as spiked in https://github.com/alphagov/static/pull/2659) we can simply deploy Static which will purge its cache and regenerate the
templates.

This partially implements GOV.UK RFC 144: https://github.com/alphagov/govuk-rfcs/pull/144


fredericfran-gds (approved):
Good job. 

You need to remove the reference to the clear_template_cache from the hiera too, see: https://github.com/alphagov/govuk-puppet/search?q=clear_template_cache.

bilbof:
Thanks - fixed

PR #11330, opened 2021-11-29
Add Docker powered PostgreSQL & MySQL servers to CI Agents

kevindew:
Trello: https://trello.com/c/oRrpsveM/64-investigating-running-current-version-of-postgresql-and-mysql-on-govuk-ci

This adds puppet defined types to run and monitor PostgreSQL and MySQL
servers via Docker, without a peristent storage layer and with no access
control 😬.

It then uses these defined types to install PostgreSQL 13 and MySQL 8 on all
the CI Agent boxes, as part of the configuration for other database
tooling. Those versions have been chosen as they're the DB versions
we're aspiring to upgrade GOV.UK to use.

The expectation is that apps will make use of these databases by setting
environment variables in their puppet configuration. These are some
examples I've tested:

```
govuk.setEnvar("TEST_DATABASE_URL", "postgresql://postgres@127.0.0.1:54313/content-publisher-test")
govuk.setEnvar("TEST_DATABASE_URL", "mysql2://root:root@127.0.0.1:33068/whitehall_test")
```

Note: MySQL seems to struggle with localhost hostname but works with an
IP addreess (I think it tries to use a local socket when it sees the
hostname)

I've tried this out on a couple of builds: https://github.com/alphagov/content-publisher/commit/29adf45706b340ed3f00412cb27f15a8bea9f545 and https://github.com/alphagov/whitehall/commit/49729b8cdcec1091d6be9ec7729a8927c1dafbc4

---

These changes have been made so that applications can run their tests against
database versions that are newer than what Ubuntu Trusty supports. I've
put this together with a degree of optimism that the configuration for
these versions will work for other versions, so that the same defined
type can be used multiple times.

A couple of caveats:

- I run the database servers on the port that we expose rather than
  mapping their original port to the exposed one (-p 33068:3306) as the
  images are configured to expose their default port and this clashes
  with pre-existing MySQL/PostgreSQL installations.
- For older versions of MySQL client to connect to newer ones I seem to
  need to set `--default-authentication-plugin=mysql_native_password`
  otherwise there's an error of `Authentication plugin
  'caching_sha2_password' cannot be loaded`.

### DIFF
--- a/modules/govuk_ci/manifests/agent/mysql.pp
+++ b/modules/govuk_ci/manifests/agent/mysql.pp
@@ -3,6 +3,14 @@
 # Installs and configures mysql-server
 #
 class govuk_ci::agent::mysql {
+  # Docker instances of MySQL server versions that are installed in parallel
+  # with the Ubuntu Trusty version
+  include ::govuk_docker
+  ::govuk_containers::ci_mysql { 'ci-mysql-8':
+    tag  => '8',
+    port => 33068,
+  }
+
   contain ::govuk_mysql::server
 
   # We want to run MySQL on a ramdisk, so this creates the directory so that it
--- a/modules/govuk_ci/manifests/agent/postgresql.pp
+++ b/modules/govuk_ci/manifests/agent/postgresql.pp
@@ -14,6 +14,14 @@ class govuk_ci::agent::postgresql (
   $mapit_role_password = 'mapit',
   $email_alert_api_role_password = 'email-alert-api',
 ) {
+  # Docker instances of PostgreSQL server versions that are installed in
+  # parallel with the Ubuntu Trusty version
+  include ::govuk_docker
+  ::govuk_containers::ci_postgresql { 'ci-postgresql-13':
+    version => '13',
+    port    => 54313,
+  }
+
   contain ::govuk_postgresql::mirror
   include ::govuk_postgresql::server::standalone
   include ::postgresql::server::contrib
--- a/modules/govuk_containers/manifests/ci_mysql.pp
+++ b/modules/govuk_containers/manifests/ci_mysql.pp
@@ -1,0 +1,32 @@
+# == Resource: govuk_containers::ci_mysql
+#
+# Install and run a dockerised MySQL server, intended for CI purposes
+#
+# === Parameters
+#
+# [*version*]
+#   The version of the MySQL Docker image to use
+#
+# [*port*]
+#   The port that the MySQL server will be exposed on
+#
+define govuk_containers::ci_mysql(
+  $ensure = 'present',
+  $version = '8',
+  $port = 33068
+) {
+  ::docker::run { $title:
+    ensure  => $ensure,
+    ports   => ["127.0.0.1:${port}:3306"],
+    image   => "mysql:${version}",
+    env     => ['"MYSQL_ROOT_PASSWORD=root"'],
+    command => '--default-authentication-plugin=mysql_native_password',
+  }
+
+  @@icinga::check { "check_${title}_running_${::hostname}":
+    ensure              => $ensure,
+    check_command       => "check_nrpe!check_tcp!127.0.0.1 ${port}",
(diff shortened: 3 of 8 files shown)


modules/govuk_ci/manifests/agent/postgresql.pp
```diff
@@ -14,6 +14,14 @@
   $mapit_role_password = 'mapit',
   $email_alert_api_role_password = 'email-alert-api',
 ) {
+  # Docker instances of PostgreSQL server versions that are installed in
+  # parallel with the Ubtunu Trusty version
```
brucebolt, line 18:
```suggestion
  # parallel with the Ubuntu Trusty version
```

modules/govuk_ci/manifests/agent/mysql.pp
```diff
@@ -3,6 +3,14 @@
 # Installs and configures mysql-server
 #
 class govuk_ci::agent::mysql {
+  # Docker instances of MySQL server versions that are installed in parallel
+  # with the Ubtunu Trusty version
```
brucebolt, line 7:
```suggestion
  # with the Ubuntu Trusty version
```

brucebolt (approved):

modules/govuk_containers/manifests/ci_mysql.pp
```diff
@@ -0,0 +1,34 @@
...
+#
+define govuk_containers::ci_mysql(
+  $ensure = 'present',
+  $version = '8',
+  $port = 33068
+) {
+  ::docker::run { $title:
+    ensure  => $ensure,
+    net     => 'host',
+    expose  => [$port],
```
sengi, line 21: Do you need this to be reachable from the host network or only from within other Docker containers?

kevindew: Thanks for taking a look @sengi. I do need it from the host network. 

I started writing a very different comment but I did some research while trying to backup my understanding, which meant my understanding totally changed. So this comment will now be very different.

So, this currently works and it didn't work when I did the more traditionally ports mapping argument (`-p 33068:3306`) it didn't and I think I now understand why. I understand why this works because we're exposing 33068 and the net argument is set to the host network, this would mean that the Docker container exposes 33068 on the host network.

When I tried to the original port mapping I hit problems with a port conflict on `3306` which conflicted with the host's MySQL installation also using port 3306.

I imagine I can actually make this more idiomatic by not using the host network (which should hopefully resolve the port conflict) and then doing a more traditional docker port mapping. I'll try that out.

modules/govuk_ci/manifests/agent/postgresql.pp
kevindew: Oh dear! thank you @brucebolt

modules/govuk_containers/manifests/ci_mysql.pp
kevindew: Yup, removing the `net     => 'host',` meant I could do everything in a more idiomatic way. It's all updated now.

modules/govuk_containers/manifests/ci_mysql.pp
```diff
@@ -0,0 +1,33 @@
...
+    command => '--default-authentication-plugin=mysql_native_password',
+  }
+
+  @@icinga::check { "check_${title}_running_${::hostname}":
+    ensure              => $ensure,
+    check_command       => "check_nrpe!check_proc_running!${title}",
+    service_description => "Docker MySQL ${version} is not running",
+    host_name           => $::fqdn,
+    notes_url           => monitoring_docs_url(check-process-running),
+  }
```
kevindew, line 32: Turns out this doesn't work - I'm going to change this to a TCP port check

modules/govuk_containers/manifests/ci_mysql.pp
sengi: LGTM 👍

sengi (approved):

brucebolt (approved):

PR #11336, opened 2021-12-07
Prepare to remove mysql replica

richardTowers:
Remove some confusing hieradata, monitoring of mysql-replica, and a confusing README.

This is preparation for https://github.com/alphagov/govuk-aws/pull/1491 in which I propose we remove the mysql replicas.

More detail in the commit messages.

### DIFF
--- a/hieradata_aws/class/frontend.yaml
+++ b/hieradata_aws/class/frontend.yaml
@@ -1,8 +1,3 @@
 ---
 govuk::apps::collections::memcache_servers: 'frontend-memcached:11211'
-
-govuk::apps::contacts::db_hostname: 'mysql-replica'
-govuk::apps::contacts::db_username: 'contacts_fe'
-govuk::apps::contacts::db_password: "%{hiera('govuk::apps::contacts::db::mysql_contacts_frontend')}"
-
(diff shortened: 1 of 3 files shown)


hieradata_aws/class/frontend.yaml
```diff
@@ -1,8 +1,3 @@
 ---
 govuk::apps::collections::memcache_servers: 'frontend-memcached:11211'
-
-govuk::apps::contacts::db_hostname: 'mysql-replica'
-govuk::apps::contacts::db_username: 'contacts_fe'
-govuk::apps::contacts::db_password: "%{hiera('govuk::apps::contacts::db::mysql_contacts_frontend')}"
```
richardTowers, line 6: It looks like these were originally added in 9e6193ec3c, at which point there was a `contacts` app on the frontend machines. 

Then in e6f575ab0 the `contacts` application was removed from the frontend machines, but these settings were left behind.

And now here they still are, doing nothing but confuse us.

sengi (approved):

hieradata_aws/class/frontend.yaml
sengi: Cool, thanks - was gonna ask about that :D

PR #11339, opened 2021-12-10
Port SpeedCurve LUX version check to Jenkins

kevindew:
This ports the concourse pipeline from
https://github.com/alphagov/govuk-speedcurve-lux-js-monitor into a
Jenkins job. This has been done due to the imminent retirement of the
GDS Concourse pipeline.

The check itself is a little unusual. It checks that the version of
lux.js which SpeedCurve serve matches the one this script has. The
theory is that if this job starts alerting someone will go update the
lux.js file in govuk_publishing_components and then update the version
number in here. There isn't however any guarantees that the version
number in this file matches what GOV.UK is serving.

I've set this check to only run in production, this seems the most
appropriate place for something that is environment ambivalent.


alex-ju (approved):
Looks good! Thank you for porting this @kevindew! Will review the docs too

PR #11342, opened 2021-12-10
Configure account_api_db_admin machine class

ChrisBAshton:
This creates a dedicated database admin machine for the Account
API, following the content_data_api_db_admin convention. This is
in order to facilitate the implementation of [RFC-143][], which
concludes that every GOV.UK relational database should have its
own RDS instance.

Trello: https://trello.com/c/kMRRFmAh/43-configure-puppet-for-new-db-admin-instances


hieradata_aws/class/integration/account_api_db_admin.yaml
```diff
@@ -0,0 +1,23 @@
...
+    ensure: "present"
+    hour: "5"
+    minute: "0"
+    action: "push"
+    dbms: "postgresql"
+    storagebackend: "s3"
+    database: "account-api_production"
+    temppath: "/tmp/account_api_production"
+    url: "govuk-integration-database-backups"
+    path: "account-api-postgresql"
```
kevindew, line 23: Let's not push in the first iteration as otherwise we'll have both the main db-admin and this trying to write the same file.

hieradata_aws/class/production/account_api_db_admin.yaml
```diff
@@ -0,0 +1,12 @@
...
+    ensure: "present"
+    hour: "23"
+    minute: "0"
+    action: "push"
+    dbms: "postgresql"
+    storagebackend: "s3"
+    database: "account-api_production"
+    temppath: "/tmp/account_api_production"
+    url: "govuk-production-database-backups"
+    path: "account-api-postgresql"
```
kevindew, line 12: Same point about push

hieradata_aws/class/staging/account_api_db_admin.yaml
```diff
@@ -0,0 +1,23 @@
...
+    ensure: "present"
+    hour: "5"
+    minute: "0"
+    action: "push"
+    dbms: "postgresql"
+    storagebackend: "s3"
+    database: "account-api_production"
+    temppath: "/tmp/account_api_production"
+    url: "govuk-staging-database-backups"
+    path: "account-api-postgresql"
```
kevindew, line 23: More on push

hieradata_aws/common.yaml
```diff
@@ -782,7 +782,7 @@ govuk::apps::transition::redis_port: "%{hiera('sidekiq_port')}"
 govuk::apps::travel_advice_publisher::redis_host: "%{hiera('sidekiq_host')}"
 govuk::apps::travel_advice_publisher::redis_port: "%{hiera('sidekiq_port')}"
 
-govuk::apps::account_api::db_hostname: "postgresql-primary"
+govuk::apps::account_api::db_hostname: "account-api-postgresql-primary"
```
kevindew, line 785: I don't think we should change this straight away - as we'd want the postgresql-primary to still be the db in use until we're happy to switchover.

modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -338,7 +338,9 @@ function restore_elasticsearch {
 
 function dump_postgresql {
   # Check which postgres instance the database needs to restore into
-  if [ "${database}" == 'content_data_api_production' ]; then
+  if [ "${database}" == 'account-api_production' ]; then
+    db_hostname='account-api-postgresql-primary'
+  elif [ "${database}" == 'content_data_api_production' ]; then
```
kevindew, line 343: Hmm one of the things that comes to mind here is that when we flip this we'll find the existing db instance isn't sync'd. I wonder if there's a way we can get this to run both in old world and new world as this gradually fans out.

kevindew (commented):
This looks good Chris, just a couple of comments on the push that should be simple to resolve and a thornier one on the env sync script for the transitory phase.

modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -338,7 +338,9 @@ function restore_elasticsearch {
 
 function dump_postgresql {
   # Check which postgres instance the database needs to restore into
-  if [ "${database}" == 'content_data_api_production' ]; then
+  if [ "${database}" == 'account-api_production' ]; then
```
ChrisBAshton, line 341: Note to self - need to update the `restore_postgresql` function too.

modules/govuk_env_sync/files/govuk_env_sync.sh
ChrisBAshton: Resolved by #11343 ⭐ thanks!

kevindew:
The hieradata config changes look good to me 👍

hieradata_aws/class/production/account_api_db_admin.yaml
```diff
@@ -0,0 +1,13 @@
+govuk_env_sync::tasks:
```
kevindew, line 1: Should we comment this line out too? I'm not too sure how YAML interprets this when everything is commented - I'd guess it doesn't set to an object but instead sets it to null?

kevindew (approved):
Looks great - thanks; just one minor thing

hieradata_aws/class/production/account_api_db_admin.yaml
ChrisBAshton: Fair point. I'd assume it would just be treated as an empty array, but it could be interpreted as a null or an empty string. Probably safer to keep the whole thing commented out.

PR #11343, opened 2021-12-10
Configure govuk_env_sync to accept a database hostname

kevindew:
Trello: https://trello.com/c/kMRRFmAh/43-configure-puppet-for-new-db-admin-postgresql-instances

This replaces the conditionals govuk_env_sync with a database hostname
argument (-H or --database_hostname).

This has been done to remove the conditionals needed to map database to
a particular hostname - which would be to grow in size with the current
work to create new RDS instances.

A motivation for this work is to allow us to configure new database RDS
instances to pull from existing databases so they can be populated
automatically.


modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -713,6 +695,11 @@ else
   : "${url?"No storage url specified (pass -u option)"}"
   : "${path?"No storage path specified (pass -p option)"}"
 
+  if [[ "$dbms" == "postgresql" ]] && [ -z "${database_hostname+x}" ]; then
```
sengi, line 698: Simplifies to
```suggestion
  if [[ "$dbms" == "postgresql" && -z "$database_hostname" ]]; then
```

sengi (approved):
Cool — this is way better, thanks!

modules/govuk_env_sync/files/govuk_env_sync.sh
kevindew: Great - thank you. That was the end result of me Googling and a lot of trial and error 😆

kevindew: Hmm I tried that out in https://github.com/alphagov/govuk-puppet/pull/11343/commits/d48360132cb203f2c488f2fcd727a39f1ed43379 and it brought back the error I faced earlier of: 

```
/usr/local/bin/govuk_env_sync.sh: line 698: database_hostname: unbound variable
```

I'm going to revert that change now, but still welcome suggestions.

sengi: Oh right, sorry I didn't test my suggestion with `set -u`. Try `"${database_hostname:-}"`

kevindew: That works 👍  Thanks 🏅

ChrisBAshton (approved):
Nicely done 👏 Resolves https://trello.com/c/LGV0e6O4/45-refactor-how-govukenvsync-hosts-are-handled

PR #11345, opened 2021-12-13
DRY up the app-specific DB admin box configuration

ChrisBAshton:
We're about to create many app-specific DB admin boxes, all of
which will have the same configuration on the node itself, so it
makes sense to encapsulate this in one place.

I did consider leaving the existing db_admin node untouched, as
we will eventually remove the Postgres code from it when every
Postgres-powered app is using its own RDS instance. However, it
uses the same code we've just encapsulated (albeit in a slightly
different order), so there's little harm in making this optimisation
now.

### DIFF
--- a/modules/govuk/manifests/node/s_content_data_api_db_admin.pp
+++ b/modules/govuk/manifests/node/s_content_data_api_db_admin.pp
@@ -31,50 +31,14 @@ class govuk::node::s_content_data_api_db_admin(
   include govuk_env_sync
   include ::govuk::node::s_base
 
-  # This allows easy administration of the PostgreSQL backend:
-  # https://www.postgresql.org/docs/9.3/static/libpq-pgpass.html
-  file { '/root/.pgpass':
-    ensure  => present,
-    mode    => '0600',
-    content => "${postgres_host}:5432:*:${postgres_user}:${postgres_password}",
-  }
-
-  # Unfortunately, the prior art for configuring db-admin style
-  # machines seems to involve a redundant PostgreSQL service, just to
-  # satisfy the Puppet module used to configure PostgreSQL running on
-  # the RDS instance. Some of the below configuration relates to this.
-
-  # Connect to the RDS instance when performing Puppet operations
-  $default_connect_settings = {
-    'PGUSER'     => $postgres_user,
-    'PGPASSWORD' => $postgres_password,
-    'PGHOST'     => $postgres_host,
-    'PGPORT'     => $postgres_port,
-  }
-
-  apt::source { 'postgresql':
-    ensure       => present,
-    location     => "http://${apt_mirror_hostname}/postgresql",
-    release      => "${::lsbdistcodename}-pgdg",
-    architecture => $::architecture,
-    key          => 'B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8',
+  # include the common config/tooling required for our app-specific DB admin class
+  class { '::govuk::nodes::postgresql_db_admin':
+    postgres_host       => $postgres_host,
+    postgres_user       => $postgres_user,
+    postgres_password   => $postgres_password,
+    postgres_port       => $postgres_port,
+    apt_mirror_hostname => $apt_mirror_hostname,
   } ->
-
-  # We don't actually want to run a local PostgreSQL server, just
-  # configure the RDS one
-  class { '::postgresql::server':
-    default_connect_settings => $default_connect_settings,
-    service_manage           => false,
-  } ->
-
-  service { 'postgresql':
-    ensure  => stopped,
-  }
-
-  include ::govuk_postgresql::server::not_slave
-
-  # Ensure the client class is installed
-  class { '::govuk_postgresql::client': } ->
 
   # include all PostgreSQL classes that create databases and users
   class { '::govuk::apps::content_data_api::db': }
--- a/modules/govuk/manifests/node/s_db_admin.pp
+++ b/modules/govuk/manifests/node/s_db_admin.pp
@@ -29,6 +29,8 @@ class govuk::node::s_db_admin(
   include ::govuk::node::s_base
   include govuk_env_sync
 
+  ### MongoDB ###
+
   apt::source { 'mongodb41':
     ensure       => 'absent',
     location     => "http://${apt_mirror_hostname}/mongodb4.1",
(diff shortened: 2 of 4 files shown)


barrucadu:
There's also https://github.com/alphagov/govuk-puppet/blob/main/modules/govuk/manifests/node/s_db_admin.pp

ChrisBAshton:
Thanks @barrucadu. I'm deliberately leaving that file out, as it is not an application-specific DB admin machine - it's the generic shared DB admin that is (currently) responsible for managing all the MySQL RDS instances, MongoDB/DocumentDB/Elasticsearch clusters, and all PostgreSQL RDS instances apart from Content Data API and Transition.

Once all Postgres apps have switched to their own app-specific DB admin node, there should be a follow-up piece of work to cut out all the Postgres logic from the 'shared' DB admin node. Does that make sense?

barrucadu (dismissed):
Looks good to me

modules/govuk_postgresql/manifests/db_admin_common_config.pp
```diff
@@ -0,0 +1,79 @@
...
+#
+# [*postgres_password*]
+#   The password corresponding to the above `postgres_user`.
+#   Default: undef
+#
+# [*postgres_port*]
+#   The port with which to connect to the `postgres_host`.
+#   Default: '5432'
+#
+class govuk_postgresql::db_admin_common_config(
```
kevindew, line 25: I don't think this is the right place for this class, since govuk_postgresql is for installing postgresql (as I recall).

I'd suggest perhaps creating a directory `modules/govuk/manifests/nodes` (similar to `modules/govuk/manifests/apps`) for specific node convenience. 

I'd have thought we'd want to be able to import this file into db_admin too - since it has postgres. If there are reason that stop that it shouldn't be in this file (in my opinion).

(great idea by the way - quick drive by review)

ChrisBAshton: Thanks, I've renamed the class, and, when I realised the Postgres config in the main db_admin node is the same as the per-app configs, I sided towards making the same change there too.

I tried your suggestion from Slack to use inheritance, e.g.

```
# == Class: govuk_node::s_content_data_api_db_admin
#
# This machine class is used to administer the Content Data API
# PostgreSQL RDS instances.
#
class govuk::node::s_content_data_api_db_admin inherits govuk::node::s_postgresql_db_admin_base {
  # include all PostgreSQL classes that create databases and users
  class { '::govuk::apps::content_data_api::db': }
}
```

...but got a `Must pass apt_mirror_hostname to Class[Govuk::Nodes::S_postgresql_db_admin_base]` error. It [doesn't seem possible to inherit parameterized classes](https://serverfault.com/a/443778). (I did also try [inheriting each parameter explicitly](https://serverfault.com/a/900345)).

modules/govuk/manifests/nodes/s_postgresql_db_admin_base.pp
```diff
@@ -0,0 +1,79 @@
+# == Class: govuk::nodes::s_postgresql_db_admin_base
+#
+# This machine class defines the common configuration used to
+# administer the app-specific DB-admin machines for PostgreSQL
```
kevindew, line 4:
```suggestion
# This class installs the tooling for a node to be a PostgreSQL
# db_admin machine
```

Common configuration doesn't seem quite way to phrase it, since there's more than configuration

modules/govuk/manifests/node/s_db_admin.pp
```diff
@@ -26,8 +26,28 @@
   $postgres_password    = undef,
   $postgres_port        = '5432',
 ) {
-  include ::govuk::node::s_base
-  include govuk_env_sync
```
kevindew, line 30: Let's keep these here, it seems confusing to expect the `s_postgresql_db_admin_base` to provide these generic things - hopefully this is without consequence and puppet accepts that two places can define this.

modules/govuk/manifests/nodes/s_postgresql_db_admin_base.pp
```diff
@@ -0,0 +1,79 @@
...
+#
+# [*postgres_password*]
+#   The password corresponding to the above `postgres_user`.
+#   Default: undef
+#
+# [*postgres_port*]
+#   The port with which to connect to the `postgres_host`.
+#   Default: '5432'
+#
+class govuk::nodes::s_postgresql_db_admin_base(
```
kevindew, line 25:
```suggestion
class govuk::nodes::postgresql_db_admin(
```

Since this isn't an actual node I think we can drop the weird `s_` (not too sure what that is for) and I think removing `_base` makes since since we don't expect this to be an inherited class any more.

kevindew (commented):
Thanks Chris, sorry to send you down a bad pathway with inheritance.

modules/govuk/manifests/nodes/s_postgresql_db_admin_base.pp
```diff
@@ -0,0 +1,79 @@
...
+#   Default: undef
+#
+# [*postgres_password*]
+#   The password corresponding to the above `postgres_user`.
+#   Default: undef
+#
+# [*postgres_port*]
+#   The port with which to connect to the `postgres_host`.
+#   Default: '5432'
+#
```
kevindew, line 24: Oh should we have apt_mirror_hostname documented too?

ChrisBAshton: Sure 👍 I've taken the description from https://github.com/alphagov/govuk-puppet/blob/e8c541a58ec0cec12b781688eb77b28e50220ecc/modules/govuk_jenkins/manifests/package.pp#L8-L9.

modules/govuk/manifests/node/s_content_data_api_db_admin.pp
```diff
@@ -31,51 +31,15 @@
   include govuk_env_sync
   include ::govuk::node::s_base
```
kevindew, line 32: Did we hit problems having these defined in modules/govuk/manifests/nodes/postgresql_db_admin.pp ?

kevindew (approved):
Thanks, looks good 👍

modules/govuk/manifests/node/s_content_data_api_db_admin.pp
ChrisBAshton: We didn't, but I kept these here based on your [comment](https://github.com/alphagov/govuk-puppet/pull/11345#discussion_r767847067). I see the comment was only on the generic `s_db_admin.pp` but it seems to apply to the other DB admin machines too. I was in two minds about whether or not to keep them in the first place. There's something quite nice about keeping the non-Postgres related stuff out of `postgresql_db_admin.pp`, even if it means a bit more repetition.

kevindew: Yup, that sounds fair. Thanks

PR #11351, opened 2021-12-14
Add Icinga alerting to Ask Export

kevindew:
This add Icinga notifying into the process of the job completing or
failing, without this we won't actually receive any alerts if the
service fails or stops running - the check will just remain as a forever
pending state.

I've changed the passive check from being conditional to only having the
freshness check being conditional - this is so this can still alert in
integration and having the alerting tested there.


modules/govuk_jenkins/manifests/jobs/ask_export.pp
```diff
@@ -69,14 +69,17 @@
...
+    false => '',
+  }
+
+  @@icinga::passive_check { "ask_export_${::hostname}":
+      ensure              => present,
+      service_description => $service_description,
+      host_name           => $::fqdn,
+      freshness_threshold => $freshness_threshold,
+      action_url          => "https://${deploy_jenkins_domain}/job/ask-export/",
+      # notes_url           => monitoring_docs_url(ask-export);
```
JonathanHallam, line 83: If i can get https://github.com/alphagov/govuk-developer-docs/pull/3348 approved we can uncomment this :)

modules/govuk_jenkins/manifests/jobs/ask_export.pp
```diff
@@ -69,14 +69,17 @@
...
-        freshness_threshold => 86400, # 24 Hours
-        action_url          => "https://${deploy_jenkins_domain}/job/ask-export/",
-        # notes_url           => monitoring_docs_url(ask-export);
-    }
+  $freshness_threshold = $run_daily ? {
+    true  => 86400, # 24 hours
+    false => '',
+  }
+
+  @@icinga::passive_check { "ask_export_${::hostname}":
```
JonathanHallam, line 77: Is there a functional difference between not running this block at all and setting the freshness threshold to an empty string?

modules/govuk_jenkins/manifests/jobs/ask_export.pp
kevindew: 👀

JonathanHallam (approved):
Looks good, small question added but just for my own knowledge :)

modules/govuk_jenkins/manifests/jobs/ask_export.pp
JonathanHallam: :D Thanks!

modules/govuk_jenkins/manifests/jobs/ask_export.pp
kevindew: Yeah - this sets up the check in integration and it'll show green if the job succeeds and red if the job fails, but has no check for freshness.

JonathanHallam: nice one, thanks :)

modules/govuk_jenkins/manifests/jobs/ask_export.pp
JonathanHallam: I've merged that now so if you could uncomment this we're all good! :)

kevindew: Great just pushed up a commit to add that

PR #11353, opened 2021-12-14
Configure collections_publisher_db_admin machine class

ChrisBAshton:
This creates a dedicated database admin machine for Collections
Publisher, following the Account API convention devised in #11342.
This is in order to facilitate the implementation of [RFC-143][],
which concludes that every GOV.UK relational database should have
its own RDS instance.


modules/govuk/manifests/node/s_collections_publisher_db_admin.pp
```diff
@@ -0,0 +1,45 @@
...
+  $apt_mirror_gpg_key_fingerprint,
+) {
+  include ::govuk::node::s_base
+  include govuk_env_sync
+
+  file { '/root/.my.cnf':
+    ensure  => 'present',
+    owner   => 'root',
+    group   => 'root',
+    content => template('govuk/node/s_db_admin/.my.cnf.erb'),
```
ChrisBAshton, line 40: No need for a MySQL equivalent to the [govuk::nodes::postgresql_db_admin class](https://github.com/alphagov/govuk-puppet/blob/f8a80afb0fef55bd2cd61090ae2e7ae8c07e0217/modules/govuk/manifests/nodes/postgresql_db_admin.pp), as the only common config we could extract are the 5-6 lines above, which saves us nothing as we'd have to [pass the constructor arguments](https://github.com/alphagov/govuk-puppet/blob/10871f283d3e8c1c848d89dbba7610b52f5669e6/modules/govuk/manifests/node/s_account_api_db_admin.pp#L35-L41) to the included class.

NB, we're reusing the same `s_db_admin/.my.cnf.erb` file for all, as it contains generic config that will apply to each DB admin machine. The naming is potentially problematic here, as this is no longer being used only by the `s_db_admin` node, but that feels like a refactor we could make later.

hieradata_aws/common.yaml
```diff
@@ -468,6 +468,8 @@ govuk::apps::ckan::s3_aws_secret_access_key: "%{hiera('s3_aws_secret_access_key'
 
 govuk::apps::collections::unicorn_worker_processes: 8
 
+# TODO: switch to "collections-publisher-mysql" and uncomment the 'push'
+# `govuk_env_sync::tasks` tasks when we're ready to switch to the dedicated RDS instance
 govuk::apps::collections_publisher::db_hostname: "mysql-primary"
 govuk::apps::collections_publisher::jwt_auth_secret: "%{hiera('jwt_auth_secret')}"
 govuk::apps::collections_publisher::redis_host: "%{hiera('sidekiq_host')}"
```
ChrisBAshton, line 475: Unlike the PostgreSQL equivalent, we _don't_ need to set a `s_collections_publisher_db_admin::apt_mirror_hostname` [here](https://github.com/alphagov/govuk-puppet/blob/10871f283d3e8c1c848d89dbba7610b52f5669e6/hieradata_aws/common.yaml#L862), as it isn't used anywhere in the `govuk::node::s_collections_publisher_db_admin` class.

modules/govuk/manifests/node/s_collections_publisher_db_admin.pp
kevindew: Yup that sounds reasonable, are you going to change that with a follow up PR? I thought that we could add it as a story to [TODO (Later)](https://trello.com/b/eFUUbwPM/database-migration-dec-2021-jan-2022) but it seemed a bit small. I imagine it could just be: `govuk-puppet/modules/govuk/templates/mysql_my.cnf.erb`

kevindew (approved):
Looks good - thanks Chris

modules/govuk/manifests/node/s_collections_publisher_db_admin.pp
ChrisBAshton: I hesitated because I wasn't sure _where_ the shared template should live, but I like your suggestion. I've raised #11356, and will rebase this PR when that's merged.

kevindew (approved):
Nice one

PR #11357, opened 2021-12-14
Configure support_api_db_admin machine class

ollietreend:
This creates a dedicated database admin machine for the Support API, following the content_data_api_db_admin convention.

This is in order to facilitate the implementation of [RFC-143][], which concludes that every GOV.UK relational database should have its own RDS instance.


ChrisBAshton (commented):
Thanks @ollietreend - this looks great, just one comment! Apologies for not explaining where the `database_name` value comes from.

ollietreend:
Closing as superseded by #11358

PR #11360, opened 2021-12-15
Configure contacts_admin_db_admin machine class

ollietreend:
This creates a dedicated database admin machine for Contacts Admin, following the Account API convention devised in #11342.

This is in order to facilitate the implementation of [RFC-143][], which concludes that every GOV.UK relational database should have its own RDS instance.

## `contacts` vs `contacts-admin`

For historial reasons, the name of this app is inconsistent. It's usually called `contacts-admin` nowadays, but in a past life it was known as `contacts`. There's still some old Puppet config lingering which still calls it that.

I've used the name `contacts-admin` on everything that's new, but have stopped short of renaming existing things.


modules/govuk/manifests/apps/contacts/db.pp
```diff
@@ -1,6 +1,6 @@
 # == Class: govuk::apps:contacts:db
 #
-# MySQL to store content for the Contacts app
+# MySQL to store content for the Contacts Admin app
```
ChrisBAshton, line 3: 🎉

ChrisBAshton (approved):
This looks good to me 👍 Thanks!

PR #11361, opened 2021-12-15
Configure ckan_db_admin machine class

ChrisBAshton:
This creates a dedicated database admin machine for CKAN,
following the content_data_api_db_admin convention. This is
in order to facilitate the implementation of [RFC-143][], which
concludes that every GOV.UK relational database should have its
own RDS instance.

Note that CKAN has two databases - ckan_production, and
ckan_pycsw_production - so I've had to define twice the number of
push/pull govuk_env_sync tasks.

Trello: https://trello.com/c/kMRRFmAh/43-configure-puppet-for-new-db-admin-postgresql-instances


ollietreend (approved):
Looks good to me 👍🏻

richardTowers (approved):

hieradata_aws/class/integration/ckan_db_admin.yaml
```diff
@@ -0,0 +1,49 @@
...
+    hour: "0"
+    minute: "0"
+    action: "pull"
+    dbms: "postgresql"
+    storagebackend: "s3"
+    database: "ckan_production"
+    database_hostname: "ckan-postgresql"
+    temppath: "/tmp/ckan_production"
+    url: "govuk-production-database-backups"
+    path: "ckan-postgresql"
```
kevindew, line 13:
```suggestion
    path: "postgresql-postgresql"
```

There won't be anything in s3://govuk-production-database-backups/ckan-postgresql until production backups start pushing there, so we need to pull from the existing backup source: s3://govuk-production-database-backups/postgresql-primary

I have a feeling this might be a problem we have in all the new db_admins to fix. 

This same comment applies to lots of places.

hieradata_aws/class/integration/ckan_db_admin.yaml
```diff
@@ -0,0 +1,49 @@
...
+  #   url: "govuk-staging-database-backups"
+  #   path: "ckan-postgresql"
+  "pull_ckan_pycsw_production_daily":
+    ensure: "present"
+    hour: "0"
+    minute: "0"
+    action: "pull"
+    dbms: "postgresql"
+    storagebackend: "s3"
+    database: "ckan_pycsw_production"
```
kevindew, line 33: Unfortunately it turns out that we [currently don't push this database](https://github.com/alphagov/govuk-puppet/blob/2609a6ce355fee22457002cd33c3bc7cf4a3c80c/hieradata_aws/class/production/db_admin.yaml#L2-L13) :( so this isn't going to work out the box.

I think the simplest thing is to modify the existing db_admin boxes to start pushing and pulling this strange database and then this can go in. I'll put something together now.

kevindew (commented):
Looks good, nice one Chris. Sorry about the S3 path issue 😬

hieradata_aws/class/integration/ckan_db_admin.yaml
ChrisBAshton: Thanks @kevindew - this should work now that #11362 is merged, right? 🤞

kevindew: Yes - well at least once that has been deployed and had a chance to dump some databases. I'm not sure how graceful/catastrophic things are when there is no database to pull.

hieradata_aws/class/integration/ckan_db_admin.yaml
ChrisBAshton: Should this be `postgresql-backend` @kevindew? I've just logged into AWS and looked in the govuk-production-database-backups S3 and can't find a `postgresql-postgresql` path.

kevindew: Oh gosh sorry! I'm in a meeting but I think it's postgresql-primary not postgresql-backend but worth a check.

ChrisBAshton: I edited the comment to attach a screenshot from AWS - we'll continue on the basis that it's `postgresql-backend` 👍

ChrisBAshton: This has been done now. I've only updated the `pull` tasks, as the `push` ones are 1) commented out anyway, and 2) probably _will_ want to pull from the new path, which can be tackled later.

hieradata_aws/class/integration/ckan_db_admin.yaml
ChrisBAshton: I've pushed up a temporary [comment-out commit](https://github.com/alphagov/govuk-puppet/pull/11361/commits/4fb2028f3dce41a30dcfdfdf47832dd98dcecfeb) to defend against this. We can revert that commit tomorrow.

barrucadu (approved):

PR #11364, opened 2021-12-16
Don't sync account-api data from production to staging / integration

barrucadu:
The account-api database contains real user data (eg email addresses),
and we haven't implemented a sanitisation step (like we have for
email-alert-api).  Furthermore, this data is of limited use in other
environments as the subject identifiers won't match the non-production
Digital Identity service.

So we don't sync this data to staging and integration.

---

[Trello card](https://trello.com/c/kMRRFmAh/43-configure-puppet-for-new-db-admin-postgresql-instances)

### DIFF
--- a/hieradata_aws/class/integration/account_api_db_admin.yaml
+++ b/hieradata_aws/class/integration/account_api_db_admin.yaml
@@ -1,16 +1,4 @@
-govuk_env_sync::tasks:
-  "pull_account_api_production_daily":
-    ensure: "present"
-    hour: "0"
-    minute: "0"
-    action: "pull"
-    dbms: "postgresql"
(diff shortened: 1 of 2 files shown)


hieradata_aws/class/integration/account_api_db_admin.yaml
```diff
@@ -1,16 +1,4 @@
 govuk_env_sync::tasks:
```
kevindew, line 1: I think this line will need commenting out too (or the commented out task being set to `ensure: "disabled"`) otherwise this YAML value will be set to null rather than an empty object - not 100% sure if that actually causes problems though.

richardTowers (approved):

PR #11365, opened 2021-12-16
Add transformation SQL files to the new db-admin machines

barrucadu:
These were missed from the main db-admin hieradata.

---

[Trello card](https://trello.com/c/kMRRFmAh/43-configure-puppet-for-new-db-admin-postgresql-instances)


kevindew (approved):
Nice catch 👍

PR #11359, opened 2021-12-15
Configure machine classes for MySQL DB Admin

ollietreend:
This PR configures puppet for the new DB Admin MySQL instances. There is one commit for each app.

[Trello ticket](https://trello.com/c/HjK4AbUS/49-configure-puppet-for-new-db-admin-mysql-instances)

**Note:** for now, the app **Contacts Admin** has been excluded from this PR because it needs renaming in the existing Puppet config.

### 🤖 Created with a bash script

The config files were created by a bash script to avoid human error. It's a bit hacky and brittle, but it gets the job done.

The bash script is available here: https://gist.github.com/ollietreend/6416b063cd8d33a0732bcf745aa77566

#### Run it yourself

The script makes the following assumptions about your environment:

- The `govuk-puppet` repo is accessible at `~/govuk/govuk-puppet`
- The repo is already on a branch and ready for commits to be added (one per app)
- You have [GNU Sed](https://formulae.brew.sh/formula/gnu-sed) installed and available as `gsed`

Execute it by running:

```
$ ./create-mysql-db-admin.sh "whitehall" "Whitehall"
$ ./create-mysql-db-admin.sh "signon" "Signon"
$ ./create-mysql-db-admin.sh "search_admin" "Search Admin"
$ ./create-mysql-db-admin.sh "release" "Release"
```

PR #11372, opened 2021-12-17
Stop auto-creating MySQL db/users on the new DB admin machines

ChrisBAshton:
In [#11353](https://github.com/alphagov/govuk-puppet/pull/11353) and [#11359](https://github.com/alphagov/govuk-puppet/pull/11359), new DB admin machines were created for apps that use MySQL. This was done as part of [RFC-143](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-143-split-database-instances.md), which agrees that every app should have its own RDS instance.

The [new RDS instances run MySQL version 8](https://github.com/alphagov/govuk-aws-data/blob/3445857dc7afd6644f4e5ccae5c5c6168e9b7281/data/app-govuk-rds/integration/common.tfvars#L47), where the previous shared RDS instance (`blue-mysql-primary`) is on version 5.6.

Traditionally, [we've used Puppet to create the MySQL databases and users](https://github.com/alphagov/govuk-puppet/blob/33dd118f4a37340b402839d790686098f1aac23a/modules/govuk/manifests/node/s_db_admin.pp) via the DB admin machine. However, this approach throws a SQL syntax error when applied to MySQL 8 environments:

```
Error: /Stage[main]/Govuk::Apps::Collections_publisher::Db/Mysql::Db[collections_publisher_production]/Mysql_user[collections_pub@%]/ensure: change from absent to present failed: Execution of '/usr/bin/mysql --defaults-extra-file=/root/.my.cnf --database=mysql -e CREATE USER 'collections_pub'@'%' IDENTIFIED BY PASSWORD '(omitted)'' returned 1: ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'PASSWORD '(omitted)'' at line 1
```

This is because [version 5.6 allowed the IDENTIFIED BY PASSWORD syntax](https://dev.mysql.com/doc/refman/5.6/en/create-user.html), whereas [8.0 is just IDENTIFIED BY](https://dev.mysql.com/doc/refman/8.0/en/create-user.html).

This [isn't overridable in the puppetlabs-mysql module](https://github.com/puppetlabs/puppetlabs-mysql/blob/a48069e89a4c06abccbb3595d2c782c7cd6e3254/lib/puppet/provider/mysql_user/mysql.rb#L66-L78), and whilst the issue has been [fixed](https://github.com/puppetlabs/puppetlabs-mysql/pull/1092) in later versions of the module, we use an old version of Puppet which does not allow us to upgrade the module.

A [spike](https://github.com/alphagov/govuk-puppet/pull/11373) investigated how we could use Puppet to create the database and user(s) in a MySQL 8 environment, but it had a number of unresolved complexities, so it was decided we would manually create the databases instead, given how infrequently we spin up new RDS instances.

### DIFF
--- a/docs/create-mysql-db-and-users.md
+++ b/docs/create-mysql-db-and-users.md
@@ -1,0 +1,74 @@
+# Create MySQL database and user
+
+## How to create a MySQL database/user on RDS instance
+
+You'll need to know three things:
+
+- Database name (`{Snake-cased app name}_production`, e.g. `collections_publisher_production`)
+- Database user (find this in the [config/database.yml](https://github.com/alphagov/collections-publisher/blob/e77c397cef35865aa5198f41e463f2bb7f8e688c/config/database.yml#L4) of your app. Beware that some apps - Whitehall - have multiple users to create)
+- Database user's password (you'll need to extract the value from govuk-secrets, e.g. [govuk::node::s_collections_publisher_db_admin::mysql_db_password](https://github.com/alphagov/govuk-secrets/blob/5981ef269e3e9be50a03c56e81ac530c8973d7b7/puppet_aws/hieradata/integration_credentials.yaml#L89). Make sure you look at the same environment, i.e. look in integration_credentials.yaml when making the database/user on Integration)
+
+Connect to the DB admin machine, e.g.
+
+```
+gds govuk connect ssh -e integration collections_publisher_db_admin
+```
+
+Export the three bits of data:
+
+```
+export TMP_MYSQL_DB_NAME='collections_publisher_production' ;
+export TMP_MYSQL_DB_USER='collections_pub' ;
+export TMP_MYSQL_DB_PASSWORD='THE_PASSWORD' ;
+```
+
+Create the user:
+
+```
+sudo mysql --defaults-extra-file=/root/.my.cnf -e \
+"CREATE USER IF NOT EXISTS '$TMP_MYSQL_DB_USER'@'%' IDENTIFIED WITH 'mysql_native_password' BY '${TMP_MYSQL_DB_PASSWORD}';"
+```
+
+Create the database:
+
+```
+sudo mysql --defaults-extra-file=/root/.my.cnf -e \
+"CREATE DATABASE IF NOT EXISTS ${TMP_MYSQL_DB_NAME};"
+```
+
+Grant the user access to the database:
+
+```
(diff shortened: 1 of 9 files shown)


docs/create-mysql-db-and-users.md
```diff
@@ -0,0 +1,74 @@
...
+export TMP_MYSQL_DB_NAME='collections_publisher_production' ;
+export TMP_MYSQL_DB_USER='collections_pub' ;
+export TMP_MYSQL_DB_PASSWORD='THE_PASSWORD' ;
+```
+
+Create the user:
+
+```
+sudo mysql --defaults-extra-file=/root/.my.cnf -e \
+"CREATE USER IF NOT EXISTS '$TMP_MYSQL_DB_USER'@'%' IDENTIFIED WITH 'mysql_native_password' BY '${TMP_MYSQL_DB_PASSWORD}';"
```
ChrisBAshton, line 29: Note that I originally tried to use [the same command we usually use](https://github.com/alphagov/govuk-puppet/pull/11373/files#diff-b79e25034d8c9b4d13c48d6675cf5c2b1ca6a19175bbd9e691665677242828beR11), i.e. `AS` instead of `BY`:

```diff
- IDENTIFIED WITH 'mysql_native_password' AS '${TMP_MYSQL_DB_PASSWORD}';"
+ IDENTIFIED WITH 'mysql_native_password' BY '${TMP_MYSQL_DB_PASSWORD}';"
```

But this led to an error: `The password hash doesn't have the expected format.`.
I attempted to fix this by changing `TMP_MYSQL_DB_PASSWORD` for `mysql_password(TMP_MYSQL_DB_PASSWORD)` (as per [how we've done things previously](https://github.com/alphagov/govuk-puppet/pull/11373/files#diff-b79e25034d8c9b4d13c48d6675cf5c2b1ca6a19175bbd9e691665677242828beL13)) but found that I'd get `mysql_password: command not found` (even if running the process as `puppet`).

I was concerned that perhaps the password is stored unhashed in the database, but running `sudo mysql --defaults-extra-file=/root/.my.cnf -e "SELECT * FROM mysql.user WHERE user='collections_pub';"`, I can't see the original password in the output, so I don't _think_ we're losing any security in taking this approach.

richardTowers (approved):
Looks good to me. Nice job documenting this.

PR #11375, opened 2021-12-20
WIP: Remove 'whitehall_fe' MySQL user

ChrisBAshton:
Searching for 'whitehall_fe' across the entire GOV.UK codebase
results in just the references removed in this commit. It seems
that the plan was to have a whitehall_fe user, reading from the
MySQL replica database, but this never happened (and the replicas
have now been removed [CITE]).

In da5dd0b35de257f775e1e19d50c2f9cf76d4a970, the whitehall_fe
settings were moved (four years ago), with the commit message even
then hinting that this might not be being used anywhere.
It looks like the user was originally added in 2013, in c6bad68c0005a06e4bd0419b26494a509920a3b3.
There is no corresponding PR and no explanation other than
"Puppet definitions for new whitehall mysql boxes".

I believe the 'whitehall_fe' user is not currently used (and am
not sure if it was ever used). We can simplify the setup of our
DB admin / RDS instances by cutting out this unnecessary user.

A complication is in the hieradata_aws/common.yaml, where we now
have `whitehall::admin_db_name` and `whitehall::db_name` (and
passwords). It seems the `db_name` variants were intended for use
by this 'frontend' user, and so can be dropped, leaving us with
just the 'admin' variants. But by convention, we should then
rename 'admin_db_name' to just 'db_name', in line with other apps.

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -591,18 +591,14 @@ govuk::apps::whitehall::admin_key_space_limit: '262144'
 govuk::apps::whitehall::admin_db_name: whitehall_production
 govuk::apps::whitehall::admin_db_password: "%{hiera('govuk::apps::whitehall::db::mysql_whitehall_admin')}"
 govuk::apps::whitehall::admin_db_username: whitehall
-# TODO this should be using a replica, but we can change that when we have deployed one
 # TODO: switch to "whitehall-mysql" and uncomment the 'push'
 # `govuk_env_sync::tasks` tasks when we're ready to switch to the dedicated RDS instance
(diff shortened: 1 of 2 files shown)


ChrisBAshton:
I just tried SSH'ing into a `whitehall_frontend` machine, then running `app-dbconsole whitehall` and then executing `SELECT USER(),CURRENT_USER();`, which confirms I was logged in as the `whitehall_fe` user. It may be that we don't need two users for Whitehall, but I guess it makes sense that the frontend app uses a user which has read-only permissions, which wouldn't be possible if we just had them use the same shared admin user.

PR #11381, opened 2021-12-21
Point Link Checker API to dedicated RDS instance on Integration

ollietreend:
We're not ready to apply this to all environment yet, so can't make the changes in the [common.yaml][].

### Testing Link Checker API

I've tested this by using the "Check Links" button in Whitehall on integration. My assumption is that Whitehall (integration) uses Link Checker API (integration), and therefore is an effective way to test that the app still works as expected.

I've also used the `dbconsole` to verify that the `version()` is now Postgres 13.


ChrisBAshton (approved):
Nice one ⭐

PR #11389, opened 2021-12-21
Point Whitehall to dedicated RDS instance on Integration

ollietreend:
We're not ready to apply this to all environment yet, so can't make the changes in the [common.yaml]( https://github.com/alphagov/govuk-puppet/blob/fdf2678eab4f498f1daa0bc3e765996c6002d665/hieradata_aws/common.yaml#L589-L597).

### Whitehall 'frontend' vs 'backend'

Note that there are two machine classes which run Whitehall: `whitehall_backend` and `whitehall_frontend`.

They each use different database config keys in the hieradata:

- `whitehall_backend` uses `govuk::apps::whitehall::admin_db_hostname`
- `whitehall_frontend` uses `govuk::apps::whitehall::db_hostname`

---

### DIFF
--- a/hieradata_aws/integration.yaml
+++ b/hieradata_aws/integration.yaml
@@ -91,6 +91,10 @@ govuk::apps::whitehall::basic_auth_credentials: "%{hiera('http_username')}:%{hie
 govuk::apps::whitehall::highlight_words_to_avoid: true
 govuk::apps::whitehall::govuk_notify_template_id: "759acac6-da53-4a19-b591-b7538c7c39de"
 govuk::apps::whitehall::aws_s3_bucket_name: 'govuk-integration-whitehall-csvs'
+# DB hostname for whitehall backend
+govuk::apps::whitehall::admin_db_hostname: 'whitehall-mysql'
+# DB hostname for whitehall frontend
+govuk::apps::whitehall::db_hostname: 'whitehall-mysql'
 
 licensify::apps::configfile::mongo_database_reference_name: 'licensify-refdata'
 licensify::apps::configfile::mongo_database_audit_name: 'licensify-audit'


kevindew:
Rather than creating the two yaml files hieradata_aws/class/integration/whitehall_backend.yaml  and hieradata_aws/class/integration/whitehall_frontend.yaml, it would be a bit simpler to add `govuk::apps::whitehall::admin_db_hostname: 'whitehall-mysql'` to the entries in: https://github.com/alphagov/govuk-puppet/blob/2696946db44d57c6118d9924602c4e29d6e6cf51/hieradata_aws/integration.yaml#L88-L91

hieradata_aws/integration.yaml
```diff
@@ -91,6 +91,10 @@ govuk::apps::whitehall::basic_auth_credentials: "%{hiera('http_username')}:%{hie
 govuk::apps::whitehall::highlight_words_to_avoid: true
 govuk::apps::whitehall::govuk_notify_template_id: "759acac6-da53-4a19-b591-b7538c7c39de"
 govuk::apps::whitehall::aws_s3_bucket_name: 'govuk-integration-whitehall-csvs'
+# DB hostname for whitehall backend
+govuk::apps::whitehall::admin_db_hostname: 'whitehall-mysql'
+# DB hostname for whitehall frontend
+govuk::apps::whitehall::db_hostname: 'whitehall-mysql'
```
ChrisBAshton, line 97: As this is MySQL, we'll also need to provide the passwords for each DB, like in https://github.com/alphagov/govuk-puppet/pull/11407/files#diff-32004bbd56816fd524912aacd8994cc4ebcc03172b2fec591f92134b60b9b152R47. 👀 

Apologies if you're already on the case!

beccapearce: Hmm I've added in the passwords and it's made whitehall, quite unhappy!

I don't suppose you know if I've named something wrong that would be causing those problems?

ChrisBAshton (approved):
LGTM (provided it works on Integration) - nice one ⭐

hieradata_aws/integration.yaml
ChrisBAshton: Ah, I approved too soon then 😁 no, the changes here look good to me. What errors were you seeing?

PR #11391, opened 2021-12-22
Point Account API to dedicated RDS instance on Integration

ChrisBAshton:

Note that we do not have to worry about the data sync - see
https://github.com/alphagov/govuk-puppet/pull/11364.

[common.yaml]: https://github.com/alphagov/govuk-puppet/blob/fdf2678eab4f498f1daa0bc3e765996c6002d665/hieradata_aws/common.yaml#L817

### DIFF
--- a/hieradata_aws/class/integration/account.yaml
+++ b/hieradata_aws/class/integration/account.yaml
@@ -1,0 +1,1 @@
+govuk::apps::account_api::db_hostname: "account-api-postgres"


richardTowers (approved):
Nice - have we already populated the new RDS instance in integration then? Manually?

ChrisBAshton:
Right now it's just an empty database @richardTowers , no tables or anything. I think we'd need to run `db:create` in a rails console on the `account` box after merging this.

richardTowers:
I wonder if we should dump the data from big postgres in integration and restore it into this? Not sure if there's any valuable data for this app in integration, but if there is it would be a shame to go back to an empty DB.

barrucadu:
It's not valuable, we can lose the data in integration.

PR #11402, opened 2021-12-31
Point Release to dedicated RDS instance on Integration

ChrisBAshton:

Note that we had to override the password that is used, otherwise
it would use the `govuk::apps::release::db::mysql_release` value
set [here](https://github.com/alphagov/govuk-puppet/blob/2a74c8b85a43a5193a1b196c6f36387cfb57c6da/hieradata_aws/common.yaml#L675),
otherwise we'd get the following error:

```
ERROR 1045 (28000): Access denied for user 'release'@'10.1.6.26' (using password: YES
```


beccapearce (approved):
LGTM

PR #11403, opened 2021-12-31
Point Collections Publisher to dedicated RDS instance on Integration

beccapearce:

[common.yaml]: 
https://github.com/alphagov/govuk-puppet/blob/fdf2678eab4f498f1daa0bc3e765996c6002d665/hieradata_aws/common.yaml#L473-475

Trello: 
https://trello.com/c/OnPpEYlk/59-run-integration-applications-on-postgres-13-and-mysql-8


ChrisBAshton (approved):
PR description has a stray line `Note that we do not have to worry about the data sync - see` - otherwise all good!

PR #11406, opened 2021-12-31
Point Signon to dedicated RDS instance on Integration

ChrisBAshton:

Note that we had to override the password that is used, otherwise
it would use the `govuk::apps::signon::db::mysql_signonotron` value
set [here](https://github.com/alphagov/govuk-puppet/blob/2a74c8b85a43a5193a1b196c6f36387cfb57c6da/hieradata_aws/common.yaml#L757),
otherwise we'd get the following error:

```
ERROR 1045 (28000): Access denied for user 'signon'@'10.1.6.26' (using password: YES
```

Note also that we've applied this to hieradata_aws/integration.yaml,
rather than hieradata_aws/class/integration/backend.yaml (where
[signon is hosted](https://docs.publishing.service.gov.uk/apps/signon.html)),
because the signon DB credentials are configured globally:
https://github.com/alphagov/govuk-puppet/blob/2a74c8b85a43a5193a1b196c6f36387cfb57c6da/hieradata_aws/common.yaml#L755-L757

This goes against the convention established for other apps
([example](https://github.com/alphagov/govuk-puppet/pull/11407/files)),
but is necessary for Signon.

### DIFF
--- a/docs/create-mysql-db-and-users.md
+++ b/docs/create-mysql-db-and-users.md
@@ -61,7 +61,7 @@ sudo mysql --defaults-extra-file=/root/.my.cnf -e "SHOW DATABASES;"
 sudo mysql --defaults-extra-file=/root/.my.cnf -e "SELECT host, user FROM mysql.user;"
 
 # See which tables are in the database, e.g.
-sudo mysql --defaults-extra-file=/root/.my.cnf -e "USE collections_publisher_sproduction; SHOW TABLES;"
+sudo mysql --defaults-extra-file=/root/.my.cnf -e "USE collections_publisher_production; SHOW TABLES;"
 # Note that this will only be populated after the first data sync has run.
 ```
 
--- a/hieradata_aws/class/integration/signon_db_admin.yaml
+++ b/hieradata_aws/class/integration/signon_db_admin.yaml
@@ -9,7 +9,7 @@ govuk_env_sync::tasks:
     database: "signon_production"
     database_hostname: "signon-mysql"
     temppath: "/tmp/signon_production"
-    url: "govuk-production-database-backups"
+    url: "govuk-integration-database-backups"
     path: "mysql"
   # "push_signon_production_daily":
   #   ensure: "present"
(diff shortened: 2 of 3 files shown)


ChrisBAshton:
Do not merge yet. Building to Integration causes a `We're sorry, but something went wrong.` message at https://signon.integration.publishing.service.gov.uk/users/sign_in when attempting to sign in.

Steps to reproduce:

```
# build `enable-new-rds-on-signon` branch to integration: https://deploy.integration.publishing.service.gov.uk/job/Deploy_Puppet/

gds govuk connect ssh -e integration signon_db_admin

# see that no tables exist
sudo mysql --defaults-extra-file=/root/.my.cnf -e "use signon_production; SHOW TABLES;"

# attempt a manual env sync
sudo envdir /etc/govuk_env_sync/env.d /usr/local/bin/govuk_env_sync.sh -a "pull" -D "mysql" -S "s3" -T "/tmp/signon_production" -d "signon_production" -u "govuk-integration-database-backups" -p "mysql" -H "signon-mysql"

# process hangs - need to run CTRL+C
# but the command above now shows newly populated tables
```

I suspect it's something to do with the env sync not fully working.

hieradata_aws/class/integration/backend.yaml
```diff
@@ -44,4 +44,6 @@ govuk::apps::local_links_manager::db_hostname: "local-links-manager-postgres"
 govuk::apps::release::db_hostname: "release-mysql"
 govuk::apps::release::db_password: "%{hiera('govuk::node::s_release_db_admin::mysql_db_password')}"
 govuk::apps::service_manual_publisher::db_hostname: "service-manual-publisher-postgres"
+govuk::apps::signon::db_hostname: "signon-mysql"
+govuk::apps::signon::db_password: "%{hiera('govuk::node::s_signon_db_admin::mysql_db_password')}"
```
kevindew, line 48: May be that the horse has bolted on this, but any reason why this is in hieradata_aws/class/integration/backend.yaml rather than hieradata_aws/integration.yaml?

ChrisBAshton: This only applies the `signon::db_{hostname/password}` values to the `backend` nodes, rather than to _all_ nodes.

kevindew: Don't we want it to all nodes though as per common.yaml?

ChrisBAshton: Ah, good spot. I've updated it now 👍

kevindew (approved):

PR #11439, opened 2022-01-07
Switch Support API to use the new RDS instance in Production 

hannako:
Trello: https://trello.com/c/sX5WXAJZ/65-plan-production-upgrade-for-support-api-postgresql-database

This is opened as a draft PR so that it can be merged in at the point of switching production over to this new instance.

This completes the switch across the three environments by setting the same value in all environments and allowing the environmental specific configuration to be removed.


kentsanggds:
Super-ceded by https://github.com/alphagov/govuk-puppet/pull/11447

PR #11441, opened 2022-01-07
Replace piped sql restore data transfers with an exchange file

rtrinque:
With big database the pipe transfer doesn't play well with docker
This solution solves the problem.


modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -358,7 +358,28 @@ function dump_postgresql {
...
+    *)
+      >&2 echo "${DUMPFILE_VERSION} is not a supported dump file version"
+      exit 1
+      ;;
+  esac
+  sudo docker run --rm --net=host  \
+	  -v "${tempdir}:/tmp/" \
+	  -v "/root/.pgpass:/tmp/.pgpass" -e PGPASSFILE=/tmp/.pgpass \
+	  "postgres:$PG_RESTORE_VERSION" \
+	  pg_restore -j 2  -f "/tmp/restore.sql" "/tmp/${filename}" | sed -r "${sed_cmds}"
```
kevindew, line 381: I wonder if we want restore.sql to be project specific so it doesn't risk any clashes or using the wrong file on a db_admin talking to multiple databases?

modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -358,7 +358,28 @@ function dump_postgresql {
...
+      exit 1
+      ;;
+  esac
+  sudo docker run --rm --net=host  \
+	  -v "${tempdir}:/tmp/" \
+	  -v "/root/.pgpass:/tmp/.pgpass" -e PGPASSFILE=/tmp/.pgpass \
+	  "postgres:$PG_RESTORE_VERSION" \
+	  pg_restore -j 2  -f "/tmp/restore.sql" "/tmp/${filename}" | sed -r "${sed_cmds}"	
+
   if [ "${transformation_sql_file:-}" ]; then
```
kevindew, line 383: This transformation file sql will need to be appended to the restore file?

modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -387,9 +408,9 @@ function filtered_postgresql_restore {
     single_transaction=''
   fi
 
-  output_restore_sql \
-    | sudo psql -U aws_db_admin -h "${database_hostname}" "${single_transaction}" \
-      --no-password -d "${database}" 2>&1
+  output_restore_sql
+  sudo psql -U aws_db_admin -h "${database_hostname}" "${single_transaction}" \
+    --no-password -d "${database}" -f "${tempdir}/restore.sql" 2>&1
```
kevindew, line 413: I wonder if life would be easier if we change pg_dump to not be format custom and just output SQL, then we'd not need to use pg_restore at all?

It'd be a breaking change but we could just skip a day of db pulling to populate the new databases.

richardTowers: I guess we'd have an issue with file sizes, since `pg_dump --format=plain` doesn't do any compression.

We could compress them with gzip before uploading them to S3, but we'd have to be careful to avoid the same issue we're having here with piping large amounts of data from docker.

Maybe we'd be better off building a debian package for postgres13 and sacking off docker after all...

kevindew: Would agree that having the package installed is best, not sure how much of a time sink that would be.

We take the plain and gzip approach already for MySQL (and I think other dump formats) we could always unzip the file first and put it in a temp file. That's essentially what this change is doing except it's using pg_restore as an agent to decompress the SQL.

kevindew:
Assuming this is superseded by: https://github.com/alphagov/govuk-puppet/pull/11454

PR #11447, opened 2022-01-11
Switch Support API to use the new RDS instance in production

kentsanggds:
This completes the switch across the three environments by setting the same value in all environments and allowing the environmental specific configuration to be removed.

Trello: https://trello.com/c/sX5WXAJZ/65-plan-production-upgrade-for-support-api-postgresql-database

PR #11440, opened 2022-01-07
Support API post-DB switch configuration changes

hannako:
This sets up all env sync config for the new Publishing API RDS instance and removes the previous config from the main DB Admin instance.

Trello: https://trello.com/c/sX5WXAJZ/65-plan-production-upgrade-for-support-api-postgresql-database

Related PR's to be merged first:
- [Use new RDS instance in staging](https://github.com/alphagov/govuk-puppet/pull/11446)
- [Use new RDS instance in production](https://github.com/alphagov/govuk-puppet/pull/11447)

PR #11449, opened 2022-01-11
Support API db post-upgrade configuration cleanup

edwardkerry:
This sets up all env sync config for the new Publishing API RDS instance and removes the previous config from the main DB Admin instance.

Trello: https://trello.com/c/sX5WXAJZ/65-plan-production-upgrade-for-support-api-postgresql-database

Related PR's to be merged first:
- [Use new RDS instance in staging](https://github.com/alphagov/govuk-puppet/pull/11446)
- [Use new RDS instance in production](https://github.com/alphagov/govuk-puppet/pull/11447)

Replaces https://github.com/alphagov/govuk-puppet/pull/11440


hieradata_aws/class/integration/db_admin.yaml
```diff
@@ -1,16 +1,4 @@
 govuk_env_sync::tasks:
-  "pull_support-contacts_integration_daily":
-    ensure: "present"
```
kevindew, line 3: We'll need to mark this (and all other task removals) as `ensure: "absent"` so they are deleted from the machine, and then we can remove them.

kevindew (commented):
Looks good, just some ensure: absents to add

hieradata_aws/class/integration/db_admin.yaml
edwardkerry: thanks Kevin, I've re-added these tasks and marked them as absent

kevindew (approved):

PR #11454, opened 2022-01-12
Replace piped sql restore data transfers with a fifo exchange file

rtrinque:
With big database the pipe transfer doesn't play well with docker
This solution solves the problem.

### DIFF
--- a/modules/govuk_env_sync/files/govuk_env_sync.sh
+++ b/modules/govuk_env_sync/files/govuk_env_sync.sh
@@ -375,7 +375,8 @@ function output_restore_sql {
   esac
   sudo docker run --rm --net=host  \
 	  -v "${tempdir}:/tmp/" \
-	  -v "/root/.pgpass:/tmp/.pgpass" -e PGPASSFILE=/tmp/.pgpass \
+	  -v "/root/.pgpass:/tmp/.pgpass" \
+	  -e PGPASSFILE=/tmp/.pgpass \
 	  "postgres:$PG_RESTORE_VERSION" \
 	  pg_restore -j 2  -f "/tmp/sed_pipe" "/tmp/${filename}" &
   sed -r "${sed_cmds}" < "${tempdir}/sed_pipe" > "${tempdir}/output_pipe"
--- a/modules/govuk_env_sync/files/govuk_env_sync.sh
+++ b/modules/govuk_env_sync/files/govuk_env_sync.sh
@@ -358,8 +358,7 @@ function dump_postgresql {
 }
 
 function output_restore_sql {
-
-  PG_RESTORE_VERSION='13.4'
+  PG_RESTORE_VERSION='' # this gets set in the case statement below
   DUMPFILE_VERSION=$(file "${dumpfile}" | awk '{print $NF}')
 
   case $DUMPFILE_VERSION in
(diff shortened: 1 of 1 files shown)


modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -358,14 +358,36 @@ function dump_postgresql {
 }
 
 function output_restore_sql {
-  pg_restore -j 2 "${dumpfile}" | sed -r "${sed_cmds}"
+
+  PG_RESTORE_VERSION='13.4'
```
ChrisBAshton, line 362: Rather than set the version here and have it immediately overwritten, I'd just make an empty variable:

```suggestion
  PG_RESTORE_VERSION='' # this gets set in the case statement below
```

modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -358,14 +358,36 @@ function dump_postgresql {
...
+      PG_RESTORE_VERSION='13.4'
+      ;;
+    *)
+      >&2 echo "${DUMPFILE_VERSION} is not a supported dump file version"
+      exit 1
+      ;;
+  esac
+  sudo docker run --rm --net=host  \
+	  -v "${tempdir}:/tmp/" \
+	  -v "/root/.pgpass:/tmp/.pgpass" -e PGPASSFILE=/tmp/.pgpass \
```
ChrisBAshton, line 379: For consistency & neatness, newline for each arg:

```suggestion
	  -v "/root/.pgpass:/tmp/.pgpass" \
	  -e PGPASSFILE=/tmp/.pgpass \
```

modules/govuk_env_sync/files/govuk_env_sync.sh
```diff
@@ -358,14 +358,36 @@ function dump_postgresql {
...
+	  "postgres:$PG_RESTORE_VERSION" \
+	  pg_restore -j 2  -f "/tmp/sed_pipe" "/tmp/${filename}" &
```
ChrisBAshton, line 381:
```suggestion
	  pg_restore -j 2 -f "/tmp/sed_pipe" "/tmp/${filename}" &
```

Out of interest, why did you choose to make this a background process with the `&`?
Same question for line 412?

ChrisBAshton (commented):
Well done for getting to the bottom of this ⭐

modules/govuk_env_sync/files/govuk_env_sync.sh
rtrinque: When sending stuff to a fifo file you need to have a process on the other side that will read from it, that's why the original process is put into the background so we can start the process that will read from the fifo file. Otherwise the initial process would never stop

modules/govuk_env_sync/files/govuk_env_sync.sh
rtrinque: When sending stuff to a fifo file you need to have a process on the other side that will read from it, that's why the original process is put into the background so we can start the process that will read from the fifo file. Otherwise the initial process would never stop

ChrisBAshton: Makes sense, thanks.

ChrisBAshton (approved):
Not sure if other devs want to weigh in too, but LGTM 👀

PR #11457, opened 2022-01-13
Add go version 1.17.6 for Router

hannako:
Related PR to update router to use go version 1.17.6: https://github.com/alphagov/router/pull/241


sengi (approved):
You mentioned on Slack that you've already checked that nothing's using 1.17.5, so LGTM 👍

PR #11458, opened 2022-01-13
Configure support-api env sync jobs

edwardkerry:
This PR removes Support-API environment sync jobs that have been marked absent on the db_admin machine.

It enables these jobs on the new suport_api_db_admin machine, where they are currently disabled.


kentsanggds (approved):
LGTM

PR #11459, opened 2022-01-13
Fix cron schedule for the LTR jenkins job

theseanything:
This fixes the invalid cron pattern which caused the build trigger to not be added to the Search API Learn to rank (LTR) Jenkins job.

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/search_api_learn_to_rank.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/search_api_learn_to_rank.yaml.erb
@@ -18,7 +18,7 @@
     triggers:
       - timed: |
           TZ=Europe/London
-          0 22 * * SUN
+          0 22 * * 7
     properties:
       - build-discarder:
           days-to-keep: 30


modules/govuk_jenkins/templates/jobs/search_api_learn_to_rank.yaml.erb
```diff
@@ -18,7 +18,7 @@
     triggers:
       - timed: |
           TZ=Europe/London
-          0 22 * * SUN
+          H 22 * * 7
```
bilbof, line 21: Interesting, `0 22 * * SUN` seems valid cron to me (https://crontab.guru/#0_22_*_*_SUN), I guess Jenkins requires the `7` instead of `SUN`.

I needed to look up what the `H` was:

> To allow periodically scheduled tasks to produce even load on the system, the symbol H (for “hash”) should be used wherever possible.
> The H symbol can be thought of as a random value over a range, but it actually is a hash of the job name, not a random function, so that the value remains stable for any given project.

bilbof (approved):

modules/govuk_jenkins/templates/jobs/search_api_learn_to_rank.yaml.erb
theseanything: Ah sorry the H was left in by mistake - I don't think there's any need to vary the start times, as its a long running which we run out of hours - so likely to have capacity anyways. I've replaced it with a `0` just keep things simple.

PR #11461, opened 2022-01-14
Remove Postgres config from db admin machine

1pretz1:
More info in the commits ->

Trello: https://trello.com/c/ewzHAr6A/56-remove-postgresql-primary-and-postgresql-standby-infrastructure-and-configuration

### DIFF
--- a/modules/govuk/manifests/node/s_db_admin.pp
+++ b/modules/govuk/manifests/node/s_db_admin.pp
@@ -83,7 +83,4 @@ class govuk::node::s_db_admin(
     postgres_port       => $postgres_port,
     apt_mirror_hostname => $apt_mirror_hostname,
   }
-
-  # include all PostgreSQL classes that create databases and users
-  -> class { '::govuk::apps::ckan::db': }
 }
--- a/modules/govuk/manifests/node/s_db_admin.pp
+++ b/modules/govuk/manifests/node/s_db_admin.pp
@@ -76,6 +76,7 @@ class govuk::node::s_db_admin(
 
   # include the common config/tooling required for our DB admin class
   class { '::govuk::nodes::postgresql_db_admin':
+    ensure              => absent,
     postgres_host       => $postgres_host,
     postgres_user       => $postgres_user,
     postgres_password   => $postgres_password,
--- a/modules/govuk/manifests/nodes/postgresql_db_admin.pp
+++ b/modules/govuk/manifests/nodes/postgresql_db_admin.pp
@@ -74,17 +74,17 @@ class govuk::nodes::postgresql_db_admin(
     service_manage           => false,
   }
 
-  if $ensure == present {
-    service { 'postgresql':
-      ensure  => stopped,
-      require => Class['postgresql::server'],
-    }
-
-    include ::govuk_postgresql::server::not_slave
-  }
-
-  # Ensure the client class is installed
-  class { '::govuk_postgresql::client':
-    ensure => $ensure,
-  }
+  # if $ensure == present {
+  #   service { 'postgresql':
+  #     ensure  => stopped,
+  #     require => Class['postgresql::server'],
+  #   }
+  #
+  #   include ::govuk_postgresql::server::not_slave
+  # }
+  #
+  # # Ensure the client class is installed
+  # class { '::govuk_postgresql::client':
+  #   ensure => $ensure,
+  # }
 }
--- a/modules/govuk/manifests/nodes/postgresql_db_admin.pp
+++ b/modules/govuk/manifests/nodes/postgresql_db_admin.pp
@@ -72,11 +72,12 @@ class govuk::nodes::postgresql_db_admin(
     package_ensure           => $ensure,
     default_connect_settings => $default_connect_settings,
     service_manage           => false,
-  } ->
+  }
 
   if $ensure == present {
     service { 'postgresql':
-      ensure => stopped,
+      ensure  => stopped,
+      require => Class['postgresql::server'],
     }
 
     include ::govuk_postgresql::server::not_slave
--- a/modules/govuk/manifests/nodes/postgresql_db_admin.pp
+++ b/modules/govuk/manifests/nodes/postgresql_db_admin.pp
@@ -5,6 +5,9 @@
 #
 # === Parameters
 #
+# [*ensure*]
+#   Whether to configure Postgres.
+#
 # [*postgres_host*]
 #   Hostname of the RDS database to use.
 #   Default: undef
@@ -26,16 +29,18 @@
 #   packages
 #
 class govuk::nodes::postgresql_db_admin(
+  $ensure               = present,
   $postgres_host        = undef,
   $postgres_user        = undef,
   $postgres_password    = undef,
   $postgres_port        = '5432',
   $apt_mirror_hostname,
 ) {
+
   # This allows easy administration of the PostgreSQL backend:
   # https://www.postgresql.org/docs/9.3/static/libpq-pgpass.html
   file { '/root/.pgpass':
-    ensure  => present,
+    ensure  => $ensure,
     mode    => '0600',
     content => "${postgres_host}:5432:*:${postgres_user}:${postgres_password}",
   }
@@ -54,7 +59,7 @@ class govuk::nodes::postgresql_db_admin(
   }
 
   apt::source { 'postgresql':
-    ensure       => present,
+    ensure       => $ensure,
     location     => "http://${apt_mirror_hostname}/postgresql",
     release      => "${::lsbdistcodename}-pgdg",
     architecture => $::architecture,
@@ -64,16 +69,21 @@ class govuk::nodes::postgresql_db_admin(
   # We don't actually want to run a local PostgreSQL server, just
   # configure the RDS one
   class { '::postgresql::server':
+    package_ensure           => $ensure,
     default_connect_settings => $default_connect_settings,
     service_manage           => false,
   } ->
 
-  service { 'postgresql':
-    ensure  => stopped,
+  if $ensure == present {
+    service { 'postgresql':
+      ensure => stopped,
+    }
+
+    include ::govuk_postgresql::server::not_slave
   }
 
-  include ::govuk_postgresql::server::not_slave
-
   # Ensure the client class is installed
-  class { '::govuk_postgresql::client': }
+  class { '::govuk_postgresql::client':
+    ensure => $ensure,
+  }
 }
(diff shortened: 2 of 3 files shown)


kevindew (commented):
Thanks for this Pete, I think there’s a few postgresql bits that’ll be left on the host

modules/govuk/manifests/node/s_db_admin.pp
```diff
@@ -76,13 +76,11 @@
 
   # include the common config/tooling required for our DB admin class
   class { '::govuk::nodes::postgresql_db_admin':
+    enabled             => false,
```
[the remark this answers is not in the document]
1pretz1: This was my thoughts originally but noticed other files using 'enabled'. Done!

modules/govuk/manifests/nodes/postgresql_db_admin.pp
```diff
@@ -53,27 +63,39 @@
...
+      architecture => $::architecture,
+      key          => 'B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8',
+    } ->
 
-  service { 'postgresql':
-    ensure  => stopped,
-  }
+    # We don't actually want to run a local PostgreSQL server, just
+    # configure the RDS one
+    class { '::postgresql::server':
```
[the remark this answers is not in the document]
1pretz1: Ah ha, so it can. Have added the param `package_ensure`.

modules/govuk/manifests/nodes/postgresql_db_admin.pp
```diff
@@ -53,27 +63,39 @@
...
+    service { 'postgresql':
+      ensure  => stopped,
+    }
 
-  # Ensure the client class is installed
-  class { '::govuk_postgresql::client': }
+    include ::govuk_postgresql::server::not_slave
+
+    # Ensure the client class is installed
+    class { '::govuk_postgresql::client': }
```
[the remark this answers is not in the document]
1pretz1: Ah okay, the ensure variable is now being passed to both of these packages which accept the param `package_ensure`.

1pretz1:
Thank you for the review @kevindew, your comments have been very useful! Sorry, a lot of this has been guess work as we didn't know Puppet well enough. 

Think I've made the changes you've requested and some more which may be needed, could you take another look please? Thanks!

modules/govuk/manifests/nodes/postgresql_db_admin.pp
```diff
@@ -64,16 +69,21 @@
...
 
   service { 'postgresql':
-    ensure  => stopped,
+    ensure => stopped,
   }
 
-  include ::govuk_postgresql::server::not_slave
+  class { '::govuk_postgresql::server::not_slave':
+    ensure => $ensure,
+  }
```
kevindew, line 83: I think we might have to put this bit within a conditional of if ensure is present. I'm expecting that `service 'postgresql' ...` may error if that service doesn't exist and it looks like the not_slave only runs queries on a postgres service so an `ensure => absent` on it is a bit misleading

modules/govuk_postgresql/manifests/client.pp
```diff
@@ -1,5 +1,23 @@
...
+
+  # installs postgresql-client-common, which includes psql
+  class { 'postgresql::client':
+    package_ensure => $ensure,
+  }
+
+  # installs libpq-dev package needed for pg gem
+  class { 'postgresql::lib::devel':
+    package_ensure => $ensure,
+  }
```
kevindew, line 22: This looks great 👍

modules/govuk_postgresql/manifests/server/not_slave.pp
```diff
@@ -3,10 +3,20 @@
...
+  $ensure = present,
+) {
+
+  if $ensure == present {
+    Postgresql::Server::Role           <| tag == $name |>
+    Govuk_postgresql::Rds_sql          <| tag == $name |>
+    Postgresql::Server::Db             <| tag == $name |>
+    Postgresql::Server::Database_grant <| tag == $name |>
+    Govuk_postgresql::Extension        <| tag == $name |>
+  }
```
kevindew, line 21: I don't think we should have an ensure here since it seems to just be queries this does (I think) rather than install software.

kevindew (commented):
Great stuff Pete. This looks almost there now to my eyes.

Have you had a chance to try it on integration? There's a part of me that wonders if all the assumptions I've made in guiding you hold and it'd be worth testing out. (I imagine we can undo any deployments by re-deploying the main branch)

modules/govuk/manifests/nodes/postgresql_db_admin.pp
1pretz1: Ah okay, good point. Have added a conditional here, instead of in the `not_slave` class.

1pretz1:
Awesome - thanks for the informative review Kev!

I've not tried it out yet, as I wanted to wait until I was confident in the changes. However, I think we're there now and so will do tomorrow 🧪

modules/govuk/manifests/nodes/postgresql_db_admin.pp
```diff
@@ -64,16 +69,21 @@
...
   } ->
 
   service { 'postgresql':
-    ensure  => stopped,
+    ensure => stopped,
   }
 
-  include ::govuk_postgresql::server::not_slave
+  if $ensure == present {
+    include ::govuk_postgresql::server::not_slave
```
kevindew, line 82: I think you'll want the `service { 'postgresql'` block in this conditional too. I don't know how that will behave if the service is removed.

kevindew:
> Awesome - thanks for the informative review Kev!
> 
> I've not tried it out yet, as I wanted to wait until I was confident in the changes. However, I think we're there now and so will do tomorrow 🧪

Great, aside from my last comment this looks great. Good luck with the integration run, hopefully I've not given you any bad advice.

modules/govuk/manifests/nodes/postgresql_db_admin.pp
1pretz1: Good shout!

1pretz1:
I tried this out on integration and it mostly worked.

For the app db admin machines, everything works fine and all the Postgres resources and config remain. For the generic db admin machine most of the resources were removed, but not all. The client software was removed and most of the dirs / files were deleted (`/root/.pgpass`, `/etc/postgresql/`, `/var/lib/postgresql/`). However the `/etc/init.d/postgresql` still persisted and thus upstart still recognised postgresql as a service.

Digging into the puppet output, everything seems to be going fine until `Error: Cannot create /var/lib/postgresql/9.6/main; parent directory /var/lib/postgresql/9.6 does not exist`. It seems that something is still trying to add something to the Postgres 9.6 directory. The next line is: `Error: /Stage[main]/Postgresql::Server::Initdb/File[/var/lib/postgresql/9.6/main]/ensure: change from absent to directory failed: Cannot create /var/lib/postgresql/9.6/main; parent directory /var/lib/postgresql/9.6 does not exist` which explains why the init script remained.

```
Notice: /Stage[main]/Postgresql::Client/Package[postgresql-client]/ensure: removed

Notice: /Stage[main]/Postgresql::Lib::Devel/Package[postgresql-devel]/ensure: removed

Info: Computing checksum on file /root/.pgpass

Info: /Stage[main]/Govuk::Nodes::Postgresql_db_admin/File[/root/.pgpass]: Filebucketed /root/.pgpass to puppet with sum 45024caa072ee16f75ff2ed78231f7f8

Notice: /Stage[main]/Govuk::Nodes::Postgresql_db_admin/File[/root/.pgpass]/ensure: removed

Info: Computing checksum on file /etc/apt/sources.list.d/postgresql.list

Info: /Stage[main]/Govuk::Nodes::Postgresql_db_admin/Apt::Source[postgresql]/Apt::Setting[list-postgresql]/File[/etc/apt/sources.list.d/postgresql.list]: Filebucketed /etc/apt/sources.list.d/postgresql.list to puppet with sum 46696071f38bc9b07946afc1626f6ffc

Notice: /Stage[main]/Govuk::Nodes::Postgresql_db_admin/Apt::Source[postgresql]/Apt::Setting[list-postgresql]/File[/etc/apt/sources.list.d/postgresql.list]/ensure: removed

Info: /Stage[main]/Govuk::Nodes::Postgresql_db_admin/Apt::Source[postgresql]/Apt::Setting[list-postgresql]/File[/etc/apt/sources.list.d/postgresql.list]: Scheduling refresh of Class[Apt::Update]
Info: Class[Apt::Update]: Scheduling refresh of Exec[apt_update]

Notice: /Stage[main]/Apt::Update/Exec[apt_update]: Triggered 'refresh' from 1 events

Notice: /Stage[main]/Postgresql::Server::Install/Package[postgresql-server]/ensure: ensure changed '9.6.9-2.pgdg14.04+1' to 'purged' [server removed]

Error: Cannot create /var/lib/postgresql/9.6/main; parent directory /var/lib/postgresql/9.6 does not exist

Error: /Stage[main]/Postgresql::Server::Initdb/File[/var/lib/postgresql/9.6/main]/ensure: change from absent to directory failed: Cannot create /var/lib/postgresql/9.6/main; parent directory /var/lib/postgresql/9.6 does not exist

Notice: /Stage[main]/Postgresql::Server::Config/Postgresql::Server::Config_entry[data_directory]/Exec[postgresql_stop_data_directory]: Dependency File[/var/lib/postgresql/9.6/main] has failures: true

Warning: /Stage[main]/Postgresql::Server::Config/Postgresql::Server::Config_entry[data_directory]/Exec[postgresql_stop_data_directory]: Skipping because of failed dependencies

Notice: /Stage[main]/Postgresql::Server::Config/Postgresql::Server::Pg_hba_rule[deny access to postgresql user]/Concat::Fragment[pg_hba_rule_deny access to postgresql user]/Concat_fragment[pg_hba_rule_deny access to postgresql user]: Dependency File[/var/lib/postgresql/9.6/main] has failures: true
```

kevindew:
Thanks Pete, looks like you're super close 🎉 

I'm wondering if the problem is this line: https://github.com/alphagov/govuk-puppet/blob/aa6e626ff8ab2c5e50d8b72e2c1dce2cc32497f8/modules/govuk/manifests/node/s_db_admin.pp#L88

(which probably should have been deleted in an earlier clean up but must have been missed)

That class has https://github.com/alphagov/govuk-puppet/blob/064405329eb2cb78290e5ef000a27bd9c821770b/modules/govuk/manifests/apps/ckan/db.pp#L31 which I imagine has a bunch of postgresql requires hidden away.

kevindew:
Superseded by https://github.com/alphagov/govuk-puppet/pull/11506

PR #11463, opened 2022-01-14
Remove references to the previous RDS instances

davidgisbey:
Our new RDS instances hostnames are split between these two conventions:

- app-name-postgres
- app-name-postgresql-primary

As the first is the most common one, I've updated the docs to use that as the default.

postgresql-primary and postgresql-standby are referenced in a couple of files. As we're removing the instances they can
should be removed or updated.


brucebolt (approved):

hannako (approved):
LGTM

davidgisbey:
Thanks!

PR #11464, opened 2022-01-14
Enable sync pull postgres

kentsanggds:
## What 

Enable govuk_env_sync script to be run on postgres machines.

## Reference 

https://trello.com/c/9tdzkMIG/31-ensure-the-govukenvsync-process-works-with-postgresql-13-and-mysql-8

### DIFF
--- a/hieradata_aws/class/integration/transition_db_admin.yaml
+++ b/hieradata_aws/class/integration/transition_db_admin.yaml
@@ -1,6 +1,18 @@
 govuk_env_sync::tasks:
+  "pull_transition_integration_daily":
+    ensure: "present"
+    hour: "6"
+    minute: "0"
+    action: "pull"
+    dbms: "postgresql"
+    storagebackend: "s3"
+    database: "transition_production"
+    database_hostname: "transition-postgresql-primary"
+    temppath: "/tmp/transition_production"
+    url: "govuk-production-database-backups"
+    path: "transition-postgresql"
   "push_transition_integration_daily":
-    ensure: "absent"
+    ensure: "present"
     hour: "3"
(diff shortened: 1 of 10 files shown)


hieradata_aws/class/integration/transition_db_admin.yaml
```diff
@@ -1,26 +1,25 @@
...
     temppath: "/tmp/transition_production"
-    url: "govuk-integration-database-backups"
+    url: "govuk-production-database-backups"
     path: "transition-postgresql"
-  "pull_transition_integration_daily":
-    # Temporarily disabled due to issues restoring Postgres 13 databases (https://trello.com/c/9tdzkMIG)
-    ensure: "disabled"
-    hour: "6"
+  "push_transition_integration_daily":
+    ensure: "absent"
```
brucebolt, line 15: Should this task be enabled here?

```suggestion
    ensure: "present"
```

kentsanggds: Is this a question around whether we should push the integration db to s3? I think this is done as a way to back it up so that it can be restored? Other apps are also doing this, otherwise we should probably disable it in Integration for all apps

brucebolt: Yes, I think we push dumps from integration to S3 daily, so the `push_transition_integration_daily` task should probably be enabled rather than being removed.

kentsanggds: Sorry, get what you mean now, I've rebased with the change

brucebolt (approved):

PR #11465, opened 2022-01-17
Update grafana dashboards

hannako:
- Remove some unused dashboards, see [commit](https://github.com/alphagov/govuk-puppet/pull/11465/commits/24aa68cb95acc6d58c403e94ed43d52c71b4d4a1).
- Add link to the cloudwatch metrics to application dashboards instead of machine dashboards.

https://trello.com/c/ewzHAr6A/56-remove-postgresql-primary-and-postgresql-standby-infrastructure-and-configuration


kevindew:
I've deployed this to integration to try it out and I think there's a few issues to resolve (and various problems with our Grafana integration set-up :( )

- Remove AWS link from machine - works
- Cloudwatch metric links - doesn't seem to work for me. I can't see the postgres links and I think non postgres boards are raising an error as they're to show a panel with a height of 0 (there's also Elasticsearch query issues unreleated to these changes only present in integration)
- Email Alert API link - this works to a staging or prod destination, it doesn't look like the cloudwatch metric source is working in integration (do we want to shave that yak)
- Removed dashboards - this works

1pretz1:
Hey @kevindew 👋 I've fixed the issues you've outlined and added config to Puppet to ensure we have the correct datasources and values if they're wiped from the machine (think this is what happened with CloudWatch in Integration). 

The datasource values listed in the [datasources manifest](https://github.com/alphagov/govuk-puppet/blob/main/modules/grafana/manifests/datasources.pp) can still fall out sync with the actual datasources in Grafana though, as changes in the manifest will not be reflected unless the datasource is deleted from the Grafana UI first. Similarly, any datasources updated in the UI will not be overwritten by Puppet. 

We probably want Puppet to overwrite existing datasources, but I think that would be a separate piece of work.

Let me know if you have Q's

kevindew (approved):
Looks great, thanks 👍

PR #11467, opened 2022-01-17
Reduce content-store ttl to 20 mins

bilbof:
This will reduce the TTL on content items to 20 minutes.

Content Store provides a max-age directive which is honoured by frontend applications.


kentsanggds (approved):
LGTM

PR #11469, opened 2022-01-17
Reduce Content Store TTL to 10 minutes

bilbof:
We saw no impact to our cache hit ratio at the CDN when
reducing the TTL from 30 to 20 minutes.

This partially implements RFC 144 - the goal
is to get down to 5 minutes as the default for
most pages.


kentsanggds (approved):
LGTM

PR #11481, opened 2022-01-19
Whitehall staging: configure env_sync for new RDS instance

ollietreend:
This commit configures env_sync tasks to run on the new Whitehall DB Admin instance in staging, and removes them from the main DB Admin machine.

I've re-used the [existing configuration][1] as it was on the main DB Admin machine – which is why the scheduled task times have changed.

[1]: https://github.com/alphagov/govuk-puppet/blob/e7825bc45fbf455f50575de29a4b165741700bb8/hieradata_aws/class/staging/db_admin.yaml#L121-L145


kevindew (approved):
Looks good - thanks 👍

PR #11482, opened 2022-01-19
Whitehall: use new RDS instance in all environments

ollietreend:
This PR switches **all environments** of Whitehall over to use their new RDS instances – **including production**.

Only merge this when the production database is ready to be switched over.


kevindew (approved):
Looks good, could you rebase this before the end of the day?

PR #11492, opened 2022-01-20
Update Search Admin to use the new RDS instance on staging

davidgisbey:
This PR updates the app to use the new RDS instance. Uncomments the env sync push job and marks the previous env sync jobs as absent on the previous db admin.

They can be removed after puppet has run.


hieradata_aws/class/staging/release_db_admin.yaml
```diff
@@ -11,15 +11,15 @@ govuk_env_sync::tasks:
...
-  #   minute: "0"
-  #   action: "push"
-  #   dbms: "mysql"
-  #   storagebackend: "s3"
-  #   database: "release_production"
-  #   database_hostname: "release-mysql"
-  #   temppath: "/tmp/release_production"
-  #   url: "govuk-staging-database-backups"
-  #   path: "release-mysql"
+  "push_release_production_daily":
```
brucebolt, line 14: I think you meant to edit the `search_admin_db_admin.yaml` file here 😄.

brucebolt (changes requested):

hieradata_aws/class/staging/release_db_admin.yaml
davidgisbey: yea sorry good spot!

PR #11499, opened 2022-01-21
Configure Whitehall env_sync tasks

ollietreend:
This PR configures env_sync tasks for Whitehall across all environments. These sync tasks are slightly unusual compared to those for most other apps, because there's a delicately timed dance required to support the desired data scrubbing behaviour.

Unlike other apps, Whitehall's env_sync process flows through all three environments: production => staging => integration

This provides the following:
- Staging will contain an unredacted copy of production data
- Integration will contain a redacted copy of staging data (which is 1 day old production data)

The data scrubbing happens on the staging instance prior to dumping to S3, so potentially sensitive data never makes it into the integration environment.


hieradata_aws/class/integration/whitehall_db_admin.yaml
```diff
@@ -1,25 +1,25 @@
 govuk_env_sync::tasks:
-  "pull_whitehall_production_daily":
+  "pull_whitehall_staging_daily":
```
kevindew, line 2: You'll need to make "pull_whitehall_production_daily" ensure absent as part of this.

kevindew (commented):
Looks good, just the one ensure absent to add.

kevindew (approved):
Nice one 👍

PR #11514, opened 2022-01-25
Rename env_sync tasks for consistency

ollietreend:
This commit renames the `env_sync` tasks defined on `*_db_admin` machines to improve their consistency.

Because of the way puppet works, these tasks need to be renamed in two deployment steps:

1. Create new tasks with the desired name, and mark the old tasks as "absent".
2. Remove the configuration for the "absent" tasks.

The changes in this pull request address **step 1** of the renaming process.

The changes include:
- Naming the tasks according to the environment they're pushing to or pulling from.
- Configuring the push/pull tasks to use different `temppath` temporary directories.

These conventions were lost in the transition to multiple DB Admin machines, which was scripted and didn't vary these values as was subsequently realised to be necessary.

Trello: https://trello.com/c/Ur7mYFEc


brucebolt (approved):
⭐️

PR #11516, opened 2022-01-25
Remove redundant env_sync tasks from DB Admin

ollietreend:
This commit tidies up the DB Admin (integration) configuration by removing the env_sync tasks which have been marked as absent.

They are no longer needed as they've been replaced by env_sync tasks on individual DB Admin instances belonging to each app.

Trello: https://trello.com/c/Ur7mYFEc


brucebolt (approved):
Looks like these were for _very_ old applications that are now switched off and archived.  Good spot.

PR #11521, opened 2022-01-27
Add Icinga alert that checks S3 static mirror freshness

MuriloDalRi:
Change the govuk_sync_mirror script to write a file with timestamps every time the script runs. It also adds a new alert that checks the last time this file was modified.

[Trello](https://trello.com/c/GaLJbJOp/2797-improve-monitoring-for-process-which-creates-the-static-mirror-fallback-of-govuk-5)


modules/govuk_crawler/templates/govuk_sync_mirror.erb
```diff
@@ -41,10 +41,25 @@ function get_error_page() {
   curl -sf https://${STATIC_HOSTNAME}/templates/503.html.erb -o ${MIRROR_ROOT}/error/503.html
 }
 
+function write_timestamps() {
+  website_mirror_dir=$(find "$MIRROR_ROOT" -name "www.*" | head -n 1)
+  hompage_last_modified=$(stat -c '%Y' ${website_mirror_dir}/index.html)
```
nicholsj, line 46: is `hompage` a typo or deliberate?

MuriloDalRi: Definitely a typo!

deborahchua (approved):
Looks good to me, it's also worth updating the dev docs about this alert.

modules/monitoring/files/usr/lib/nagios/plugins/check_s3_mirror_freshness
```diff
@@ -0,0 +1,17 @@
...
+
+difference=$(($datenow - $last_modified_date))
+
+if [[ $(($difference)) > 86400 ]]; then
+  echo "The static mirror has not been updated in ${difference} seconds"
+  exit 1
+fi
+
+echo 'The static mirror was updated in the last 24 hours'
+exit 0
```
ChrisBAshton, line 17: Nice job re-writing this in bash 👍

modules/govuk_crawler/templates/govuk_sync_mirror.erb
```diff
@@ -41,10 +41,25 @@ function get_error_page() {
   curl -sf https://${STATIC_HOSTNAME}/templates/503.html.erb -o ${MIRROR_ROOT}/error/503.html
 }
 
+function write_timestamps() {
+  website_mirror_dir=$(find "$MIRROR_ROOT" -name "www.*" | head -n 1)
+  homepage_last_modified=$(stat -c '%Y' ${website_mirror_dir}/index.html)
+  upload_time=$(date +%s)
+
+  cat <<EOF > timestamps.txt
+  homepage_last_modified=$homepage_last_modified
```
ChrisBAshton, line 50: It doesn't look like we're using this value anywhere in the `check_s3_mirror_freshness` script.

Should we just delete lines 46 and 50?

Out freshness checker makes sure that the `timestamps.txt` file is regularly updated, which proves that the mirror script is working, and by definition, that the homepage remains up to date. We don't need to track the homepage age anymore.

ChrisBAshton (commented):
Looks good - just one comment about whether we still need to query "homepage last modified".

modules/govuk_crawler/templates/govuk_sync_mirror.erb
MuriloDalRi: I agree, will remove those lines 👍

MuriloDalRi:
> Looks good to me, it's also worth updating the dev docs about this alert.

I was thinking of adding some info to [this page](https://github.com/alphagov/govuk-developer-docs/blob/main/source/manual/alerts/mirror-sync.html.md) but I might add a new page just for this alert to avoid confusion

ChrisBAshton (approved):

PR #11523, opened 2022-01-27
Add dashboard for Publishing Service

bilbof:
This adds a dashboard for tracking metrics relevant to the publishing service team. In particular this shows '[fact check emails alert](https://github.com/alphagov/govuk-puppet/blob/734b1feea335dc9ba1c9dbeff28830272216982a/modules/govuk/manifests/apps/publisher/unprocessed_emails_count_check.pp#L21)', which allows us to delete the icinga alert for this problem. It requires the service team to monitor this dashboard/problem, in keeping with the devops/service model the organisation is moving towards.

I've set the traffic light colour scheme based on gut feel rather than any SLOs set by the team, e.g. a 5% error rate will yield yellow, over 10% is red - these should be easy to change once the team has some objectives in place.

I removed the fact check emails alert that 2nd line is paged with in another commit.

Temporary view of dashboard available at: https://grafana.blue.production.govuk.digital/dashboard/db/temporary-publishing-service?orgId=1&from=now-1h&to=now

Dashboard screenshot:

publishing-service-slo

### DIFF
--- a/modules/grafana/files/dashboards/service-publishing.json
+++ b/modules/grafana/files/dashboards/service-publishing.json
@@ -1,0 +1,905 @@
+{
+  "annotations": {
+    "list": []
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "hideControls": false,
+  "id": 69,
+  "links": [],
+  "refresh": false,
+  "rows": [
+    {
+      "collapse": false,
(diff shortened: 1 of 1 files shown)


sengi (approved):
Cool, dashboard looks good (haven't read the YAML in any details as it's auto-generated). Thanks for eliminating another crufty alert!

PR #11525, opened 2022-01-27
Add automated rake task

bilbof:
This duplicates the run-rake-task Jenkins job which has been run 225,038 times in one environment.

We'd like to see what jobs are run manually vs automated so we have a better idea of the necessity for a manual job runner and the features we'll need to implement to replace the manual rake tasks.

The second commit changes all the automated jobs to trigger the automated-rake-task job so we don't need to update our docs / manual workflow to use a different job.

Example of the job in integration (deployed branch): https://deploy.integration.publishing.service.gov.uk/job/search_api_index_checks/145154/downstreambuildview/


richardTowers (approved):
Nice, this will make it much easier to audit what's happening manually.

PR #11534, opened 2022-01-28
Only do daily backups of MongoDB in integration

kevindew:
Trello: https://trello.com/c/rFIPmgBl/107-turn-off-hourly-mongodb-backups

We only run this script in integration and it only seems to have the
purpose of populating the files that are used for a govuk-docker
replication task [1] (which itself is a bit weird, it's unclear why it
doesn't use env sync backups like everything else... but yaks).

Because of this it seems rather superfluous to be generating backups
every 30 mins.

This changes the script to only run once a day, it is set to run at 8am
to be out the way of the database syncs.

I simplified the test a bit as it seemed unnecessary for the test to be
coupled with the time this task runs.

[1]: https://github.com/alphagov/govuk-docker/blob/6004cdc44ff60fc59e9b7aa156737621ff780a6d/bin/replicate-mongodb.sh#L52-L54

### DIFF
--- a/modules/mongodb/manifests/aws_backup.pp
+++ b/modules/mongodb/manifests/aws_backup.pp
@@ -1,8 +1,6 @@
 # == Class: mongodb::aws_backup
 #
-# This class handles everything related to regular backups
-# in AWS. While it's similar to the current MongoDB to S3 backups,
-# there are enough differences to create a new, simpler class.
+# This class handles daily backups from MongoDB to AWS S3.
 #
 # === Parameters:
 #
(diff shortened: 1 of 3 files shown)


davidgisbey (approved):
This one i can review! LGTM

One question. Can the original class/backup mechanism now be removed?

kevindew:
> This one i can review! LGTM
> 
> One question. Can the original class/backup mechanism now be removed?

No, unfortunately not. It seems that both types of MongoDB backup is used (govuk_env sync for syncing between environments and this for developer data in integration). Hopefully I understand your question correctly, let me know if not I could have missed something.

davidgisbey:
You do. Thanks!

PR #11544, opened 2022-02-11
Generate Jenkins' public ssh key

richardTowers:
Previously, the govuk_jenkins ssh_key class required you to pass in the
private key and the public key separately.

This meant configuring the public keys in hieradata_aws, and the private
keys in govuk-secrets, and making sure you got the right public keys for
the right private keys.

Predictably, we did not get this right:

* https://github.com/alphagov/govuk-puppet/pull/11543
* https://github.com/alphagov/govuk-aws-data/pull/980

There's no need to ask for both the private key and the public key.
Since the private key _contains_ the public key we can just generate it.

The only reason I can think of for specifying them separately would be
if we wanted to allow the mistake of having mismatched keys.

Using Puppet's Exec resource, we can run `ssh-keygen` to create the
public key from the private key. The `creates` property ensures this
runs if the key is not already there.

Using the `subscribe` property means that the exec resource should be
notified if the private key resource changes, and run the command again.


richardTowers:
Tested this by deploying the branch to integration, running puppet (which was a noop), deleting the id_rsa.pub file, and then running puppet again:

```
$ sudo rm /var/lib/jenkins/.ssh/id_rsa.pub

$ govuk_puppet --test

...

Notice: /Stage[main]/Govuk_jenkins::Job_builder_update/Exec[jenkins_jobs_update]/returns: executed successfully

...

$ sudo cat /var/lib/jenkins/.ssh/id_rsa.pub
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDQBl40cv64wBa1zEG3dIOwsTTcJsMybZW0nPmCLBqS9/xzv4WoW5VzvID6yrSlg5XfX1Qxq8FmFGIDaAhb1fna2Z05EAC1Jh8EnCSFK8Q6NaUGxlyYoHRD06kZI8ZdAj3Ct8Hsqa0YaWKa/vSIWKIRtboVKm6SMbNxcLwQ04AG2zP2wtnGpyDKBPZol/L3jxVExx1B2lIww0drSKNFKQzM9kijZyAmhu8ocClNl19Rv86q44v0PcDIv5hkW5bEbsavTghnLNXad2dmiSP5Se68NscumyboetuG+o0lOFbFjuHk8NaXklOWiFZxJaJXiOVLihXHVhpDcuXEzwNoOKhYEzA06vHBVXbngBuEsgns/Hgpz4we2H4y4k9w9eJ4rKNhTvrfAzcYzEsnmhbNtQMZaLbqKnWBt2+X6lKTYUBpnUWXwLMaAb5dqEqD+LGiDxcfJ4b6UctSR7+CF29gRChwv0HUO1NdiVzZ2AMrqsYp9QtCWnfNipveGZl9Rqox3JSt4u/+7+I9xw0d8bFp8xCPxan78eMu42i3jNm4qcbbXGvPU6WFP0htjZZ8S0Fq7Dss4AbADrLxwepW8n7E+PozZRjH2P7TgmZ+wQXS6aUNHdgDeYsv5070NYK33wuE2f9GNVuN35/5ImB9PuyxDNSdHIPXTABMOZk7fVQUqXLCRw==
```

bilbof (approved):
Fantastic 👍

PR #11553, opened 2022-02-17
Backfill alerts for smokey loop features

benthorner:
https://trello.com/c/D2Yt6TrG/251-audit-tests-in-the-smokey-repo

Depends on: https://github.com/alphagov/smokey/pull/886 - where we start
running search_api.feature in Integration.

Please review commit by commit.


ChrisBAshton (approved):

benthorner:
@ChrisBAshton thanks for the review!

I've deployed this on a branch to Integration to verify the two new alerts work ✅. The "publisher" one looks like it might be flakey, but that's just a reason to push forward so that the problem is more visible.

I don't have access to check Staging or Production:

- For Staging I would verify that a new [alert](https://alert.staging.govuk.digital/) exists for "search_api" Smokey loop and is "OK".
- For Production I would just check some existing alerts are still there and are as "OK" as they were before.

Would you like to merge and deploy this all the way?

PR #11554, opened 2022-02-17
Correct signon database sync path

kevindew:
This was reading from the old location for database dumps and thus restoring signon from a progressively older database (25th January). This corrects the path to the new one defined in https://github.com/alphagov/govuk-puppet/blob/b9cecd0885b573e29a5042cef3258ce260909023/hieradata_aws/class/integration/signon_db_admin.yaml#L13


ollietreend (approved):
Great find ⭐ 🚀

PR #11556, opened 2022-02-21
Remove spurious extra "ssh-rsa" from public key.

sengi:
This was causing reduplication of the "ssh-rsa" part in the key file, since the Puppet class prepends it also.

The other environments are ok, just prod was wrong.


brucebolt (approved):

bilbof (approved):
Verified that integration doesn't have `ssh-rsa` prepended at https://github.com/alphagov/govuk-puppet/blob/main/hieradata_aws/integration.yaml#L182

PR #11558, opened 2022-02-23
Add Sam Simpson's user to integration

samsimpson1:


modules/users/manifests/samsimpson.pp
```diff
@@ -0,0 +1,8 @@
+# Creates the samsimpson user
+class users::samsimpson {
+  govuk_user { 'samsimpson':
+    fullname => 'Sam Simpson',
+    email    => 'sam.simpson@digital.cabinet-office.gov.uk',
+    ssh_key  => 'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIHGtiWG65GGmpzqG48BWeB6TXTx8hdR26uu2MyG+JCMa sam.simpson@digital.cabinet.office.gov.uk',
```
sengi, line 6: We're on a pretty ancient Ubuntu (14.04 ESM) but I *think* ed25519 should just about work 😅

sengi (approved):

modules/users/manifests/samsimpson.pp
sengi: Oh I see there are already a few ed25519 keys - I think it's a pretty safe bet that our openssh is new enough.

PR #11559, opened 2022-02-24
Ship jumpbox logs to cloudwatch

richardTowers:
When a user uses ssh to connect to a linux server, the server stores a log in /var/log/auth.log

We should ship these logs to Splunk, so we have a proper audit trail of who's used ssh to connect to which machines.

The cyber security team recommend shipping logs to AWS CloudWatch Logs first, and then using their [centralised-security-logging-service](https://github.com/alphagov/centralised-security-logging-service) to ship them on to Splunk.

Initially just doing this for the jumpboxes is valuable enough. We may decide to do it for the other servers later on.

Depends on https://github.com/alphagov/govuk-aws/pull/1549

https://trello.com/c/3PEIdbKp/2608-ship-ssh-logs-from-jumpboxes-to-splunk

### DIFF
--- a/modules/govuk/files/node/s_jumpbox/amazon-cloudwatch-agent.json
+++ b/modules/govuk/files/node/s_jumpbox/amazon-cloudwatch-agent.json
@@ -1,0 +1,18 @@
+{
+  "agent": {
+    "run_as_user": "root"
+  },
+  "logs": {
+    "logs_collected": {
+      "files": {
+        "collect_list": [
+          {
+            "file_path": "/var/log/auth.log",
+            "log_group_name": "auth-log",
+            "log_stream_name": "{instance_id}"
+          }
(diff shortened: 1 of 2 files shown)


modules/govuk/files/node/s_jumpbox/amazon-cloudwatch-agent.json
```diff
@@ -0,0 +1,18 @@
+{
+  "agent": {
+    "run_as_user": "root"
```
sihugh, line 3: We may not need this, as the [default user should already be root](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html)

sihugh (approved):
This will ship logs every 60s

PR #11563, opened 2022-02-28
Update icinga graphite links to correct URL.

KludgeKML:
## What
Point icinga graphite links to the AWS origin (production.govuk.digital)

## Why
Icinga graphite links point to the publishing hostname, so currently broken.


maxgds (changes requested):
Looks like a test might want an update?

PR #11565, opened 2022-03-01
Add go version 1.17.7 for Router

hannako:
Related PR to update router to use go version 1.17.7: https://github.com/alphagov/router/pull/254

### DIFF
--- a/modules/golang/manifests/init.pp
+++ b/modules/golang/manifests/init.pp
@@ -7,16 +7,14 @@ class golang {
   include govuk_ppa
 
   class { 'goenv':
-    global_version => '1.15.8',
+    global_version => '1.17.7',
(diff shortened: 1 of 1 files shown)


modules/golang/manifests/init.pp
```diff
@@ -17,6 +17,7 @@
     '1.12.1',  # Used by alphagov/govuk-csp-forwarder
     '1.15.8',  # Used by alphagov/router
```
sihugh, line 18: might be able to remove some of these old versions?

sihugh: Also worth updating the `global_version` on line 10?

sihugh (approved):

PR #11567, opened 2022-03-02
Update config for imminence so integration will connect to the shared-documentdb cluster

KludgeKML:
As part of moving our apps away from Mongo 2.6, it would be useful to test moving the imminence app in integration away from the EC2 Mongo 2.6 cluster and onto the DocumentDB shared-documentdb-integration cluster. 

This change adds a new imminence-specific file to the heiradata in the integration class that should (for integration) override the existing mongodb cluster information.

When deployed, imminence will lose the existing information in integration until an env sync is done, at which point the sync should be going into the shared documentdb cluster.

Concerns:
- https://github.com/alphagov/govuk-puppet/compare/move-imminence-to-shared-documentdb?expand=1#diff-5b5aa948d59a4b8ab656f91eee5d216d21aade137e36c531106d581d32b23847R130-R131. I've left password and username unset in the hope that if they're not set (as in staging and production), the app won't try to send them (username and password appear to not be used for the mongo EC2 cluster). If it does send them (as.. blank? null?), there's a change we might see a problem in staging/production if this version of puppet is deployed there.


KludgeKML:
Obsoleted by 11573.

PR #11568, opened 2022-03-03
Install ChromeDriver alongside Google Chrome on test/CI machines

ollietreend:
This adds a module `govuk_chromedriver` which can be included to install Google Chrome alongside a compatible version of ChromeDriver, ready for use by Selenium browser tests.

This is being introduced so that there is always a usable version of ChromeDriver available on CI machines which need to run tests. In the past, CI machines have had a system-wide version of Google Chrome available for use, but no accompanying ChromeDriver. Instead each application has had to use the `webdrivers` gem to install its own version of ChromeDriver locally.

This change represents a deliberate shift in responsibility. Where previously applications were responsible for sourcing their own ChromeDriver (and hence depended on the `webdrivers` gem), we've now decided that it be the responsibility of the underlying operating environment to provide it.

In real terms, this means that the GOV.UK development environment – [govuk-docker] – and the CI environment need to have ChromeDriver installed.

[govuk-docker]: https://github.com/alphagov/govuk-docker/pull/577

---

Trello: https://trello.com/c/qC0SfSAm

### DIFF
--- a/modules/govuk/manifests/node/s_jenkins.pp
+++ b/modules/govuk/manifests/node/s_jenkins.pp
@@ -22,7 +22,7 @@ class govuk::node::s_jenkins (
   $jenkins_api_token = undef,
   $environment_variables = {},
 ) inherits govuk::node::s_base {
-  include google_chrome
+  include govuk_chromedriver
   include nginx
   include govuk_postgresql::client
   include govuk_rbenv::all
--- a/modules/govuk_chromedriver/manifests/init.pp
+++ b/modules/govuk_chromedriver/manifests/init.pp
@@ -1,0 +1,18 @@
+# == Class: govuk_chromedriver
+#
+# Installs Google Chrome and its associated ChromeDriver
+# Used by Selenium tests which need to drive a browser
+#
+class govuk_chromedriver {
+  require google_chrome
+
+  # Copy the install-chromedriver.sh script on to the machine
+  file { '/usr/sbin/install-chromedriver':
+    ensure => file,
+    source => 'puppet:///modules/govuk_chromedriver/install-chromedriver.sh',
+    mode   => '0755',
+  } # and then execute it (-> denotes a dependency)
+  -> exec { 'install-chromedriver':
+    command => '/usr/sbin/install-chromedriver',
+  }
+}
--- a/modules/govuk_chromedriver/spec/classes/govuk_chromedriver_spec.rb
+++ b/modules/govuk_chromedriver/spec/classes/govuk_chromedriver_spec.rb
@@ -1,0 +1,14 @@
+require_relative '../../../../spec_helper'
+
+describe 'govuk_chromedriver', :type => :class do
+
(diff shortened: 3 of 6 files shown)


modules/govuk_chromedriver/manifests/init.pp
```diff
@@ -0,0 +1,18 @@
...
+  # Copy the install-chromedriver.sh script on to the machine
+  file { '/usr/sbin/install-chromedriver':
+    ensure => file,
+    source => 'puppet:///modules/govuk_chromedriver/install-chromedriver.sh',
+    mode   => '0755',
+  } # and then execute it (-> denotes a dependency)
+  -> exec { 'install-chromedriver':
+    command => '/usr/sbin/install-chromedriver',
+  }
+}
```
kevindew, line 18: I feel like we should add a rudimentary test to go with this, something like this might work?

```
require_relative '../../../../spec_helper'

describe 'govuk_chromedriver', :type => :class do

  it { is_expected.to compile }

  it { is_expected.to compile.with_all_deps }

  it { is_expected.to contain_class('govuk_chromedrive') }

  it { is_expected.to contain_class(/usr/sbin/install-chromedriver').with_ensure('file') }

  it { is_expected.to contain_exec('install-chromedriver') }
end

```

kevindew (commented):
Looks super - just a couple of minor things. I gave it a spin on integration and didn't manage to break it.

modules/govuk_chromedriver/manifests/init.pp
ollietreend: Thanks for the nudge – I'd forgotten about this 🤦🏻

Your test worked pretty much out-the-box, so thank you for this.

modules/govuk/manifests/node/s_jenkins.pp
```diff
@@ -22,7 +22,7 @@
   $jenkins_api_token = undef,
   $environment_variables = {},
 ) inherits govuk::node::s_base {
-  include google_chrome
+  include ::govuk_chromedriver
```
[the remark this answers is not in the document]
ollietreend: That makes sense! Thanks, I've taken the `::`s out.

modules/govuk_chromedriver/spec/classes/govuk_chromedriver_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+  it { is_expected.to compile }
+
+  it { is_expected.to compile.with_all_deps }
+
+  it { is_expected.to contain_class('govuk_chromedriver') }
+
+  it { is_expected.to contain_file('/usr/sbin/install-chromedriver').with_ensure('file') }
+
+  it { is_expected.to contain_exec('install-chromedriver') }
+end
```
kevindew, line 14: Minor looks like you've missed a trailing new line here.

kevindew (approved):
Super 🤞 that this does the trick everywhere.

PR #11571, opened 2022-03-04
Add Feedback accessible format request email template and reply to ids

edwardkerry:
Adding AccesssibleFormatRequest Notify variables for Feedback for Staging and Production.

This PR also updates Integration's reply_to id.

This follows https://github.com/alphagov/govuk-puppet/pull/11300 which first introduced the env vars and added them to the manifest.

[tello](https://trello.com/c/7JeL3Ob8/1113-add-notify-email-template-and-reply-to-ids-to-puppet-for-staging-and-production)


KludgeKML (approved):
Looks good to me.

PR #11573, opened 2022-03-07
Update config for imminence so integration will connect to a new imminence-documentdb cluster

KludgeKML:
As part of moving our apps away from Mongo 2.6, it would be useful to test moving the imminence app in integration away from the EC2 Mongo 2.6 cluster and onto the DocumentDB imminence-documentdb-integration cluster that will be created by this PR:

https://github.com/alphagov/govuk-aws/pull/1555

This change adds a new imminence-specific file to the heiradata in the integration class that should (for integration) override the existing mongodb cluster information.

When deployed, imminence will lose the existing information in integration until an env sync is done, at which point the sync should be going into the imminence documentdb cluster.

Concerns:

https://github.com/alphagov/govuk-puppet/pull/11573/files#diff-5b5aa948d59a4b8ab656f91eee5d216d21aade137e36c531106d581d32b23847R62-R63. I've left password and username unset in the hope that if they're not set (as in staging and production), the app won't try to send them (username and password appear to not be used for the mongo EC2 cluster). If it does send them (as.. blank? null?), there's a change we might see a problem in staging/production if this version of puppet is deployed there.

### DIFF
--- a/modules/govuk/manifests/apps/imminence.pp
+++ b/modules/govuk/manifests/apps/imminence.pp
@@ -26,6 +26,14 @@
 # [*mongodb_name*]
 #   The name of the MongoDB database to use
 #
+# [*mongodb_username*]
+#   The username to use when logging to the MongoDB database,
+#   only needed if the app uses documentdb rather than mongodb
+#
+# [*mongodb_password*]
+#   The password to use when logging to the MongoDB database
+#   only needed if the app uses documentdb rather than mongodb
+#
 # [*oauth_id*]
 #   The OAuth ID used by GDS-SSO to identify the app to GOV.UK Signon
 #
@@ -52,6 +60,8 @@ class govuk::apps::imminence(
   $sentry_dsn = undef,
   $mongodb_nodes = undef,
   $mongodb_name = 'imminence_production',
+  $mongodb_username = '',
+  $mongodb_password = '',
   $redis_host = undef,
   $redis_port = undef,
   $oauth_id = undef,
@@ -118,6 +128,8 @@ class govuk::apps::imminence(
   govuk::app::envvar::mongodb_uri { $app_name:
     hosts    => $mongodb_nodes,
     database => $mongodb_name,
+    username => $mongodb_username,
+    password => $mongodb_password,    
   }
 
   govuk::procfile::worker { $app_name:
--- a/hieradata_aws/class/integration/backend.yaml
+++ b/hieradata_aws/class/integration/backend.yaml
@@ -1,5 +1,11 @@
 ---
 
+govuk::apps::imminence::mongodb_name: 'imminence_production'
+govuk::apps::imminence::mongodb_nodes:
+  - 'imminence-documentdb'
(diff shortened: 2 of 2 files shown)


hieradata_aws/common.yaml
```diff
@@ -610,6 +610,7 @@ govuk::apps::govuk_crawler_worker::root_urls:
   - "https://assets.%{hiera('app_domain')}"
   - "https://www.%{hiera('app_domain')}"
 
+govuk::apps::imminence::mongodb_name: 'imminence_production'
```
fredericfran-gds, line 613: why not keep it as the default db name in `[modules/govuk/manifests/apps/imminence.pp](https://github.com/alphagov/govuk-puppet/pull/11573/files#diff-5b5aa948d59a4b8ab656f91eee5d216d21aade137e36c531106d581d32b23847)`

modules/govuk/manifests/apps/imminence.pp
```diff
@@ -50,8 +57,10 @@
   $port,
   $enable_procfile_worker = true,
   $sentry_dsn = undef,
-  $mongodb_nodes = undef,
-  $mongodb_name = 'imminence_production',
+  $mongodb_nodes,
+  $mongodb_name,
+  $mongodb_username,
+  $mongodb_password,
```
fredericfran-gds, line 63: same as for username, need an empty default

fredericfran-gds (changes requested):
Thanks for doing this. See my comments inline.

modules/govuk/manifests/apps/imminence.pp
```diff
@@ -50,8 +57,10 @@
   $port,
   $enable_procfile_worker = true,
   $sentry_dsn = undef,
-  $mongodb_nodes = undef,
-  $mongodb_name = 'imminence_production',
+  $mongodb_nodes,
+  $mongodb_name,
+  $mongodb_username,
```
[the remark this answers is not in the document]
KludgeKML: I had blank defaults for this originally, but what I'm slightly worried about is that if I set a default username and password, potentially the client in staging/production will try to use that to connect to the old cluster, because this file (unlike the others) doesn't seem to be specific to integration. It might be fine, or it might be that the empty string triggers an attempt to login which may or may not work (compare it with, say, a badly written basic auth handler that treats an empty string as different from nil, and then tries to login with a basic auth hash made from an empty username and password). I'm sure the likely thing is that it'll be fine, I just don't understand enough about how this or the clients work and it makes me a bit nervous.

hieradata_aws/common.yaml
fredericfran-gds: I get it, I would have kept it in the class definition, to be honest. They are the same across all environments. Up to you.

KludgeKML: If it's the same everywhere, class definition makes more sense. I'll remove it from here and from the /integration/imminence.yaml file.

KludgeKML:
Closing as documentdb is not suitable for imminence (it requires 2d index support, which DocumentDB does not have)

PR #11592, opened 2022-03-21
Remove old cookie exceptions from varnish config

smford:
These two cookie exceptions are no longer required because the services
that used them have been decomissioned.

Removing /find-coronavirus-support alphagov/smart-answers#5585
Removing /brexit-eu-funding alphagov/frontend#2263


sengi (approved):

smford:
abandoning, new pr will be raised.

PR #11603, opened 2022-03-28
Standardise Smokey environments for Staging / Prod

benthorner:
https://trello.com/c/LnE7dZdj/257-survey-actions-for-smokey

These are the same [^1] [^2]. Now that we operate entirely on AWS
we can dispense with the "_aws" suffix.

[^1]: https://github.com/alphagov/smokey/blob/0c09a9e40a7fa8984ffc9264aab2b892a6cb0735/config/cucumber.yml#L4-L7
[^2]: https://github.com/alphagov/smokey/blob/ba3829983c91478c0702442d9c03f85c49904bd4/features/support/env.rb#L19

### DIFF
--- a/hieradata_aws/production.yaml
+++ b/hieradata_aws/production.yaml
@@ -323,7 +323,7 @@ govuk_jenkins::jobs::signon_cron_rake_tasks::rake_users_send_suspension_reminder
 
 govuk_jenkins::jobs::smart_answers_broken_links_report::cron_schedule: '30 1 1,15 * *' # every 1st and 15th of the month at 1:30 AM
 
-govuk_jenkins::jobs::smokey::environment: production_aws
+govuk_jenkins::jobs::smokey::environment: production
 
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
⭐

benthorner:
Thanks for the review @ChrisBAshton !

Would you mind merging / deploying this one? I don't have enough access to. I'd recommend running Puppet on the Staging monitoring machine and then checking the "Smokey" job and Smokey Loop still run OK.

ChrisBAshton:
Looks like the Smokey loop is still working:

```
$ ls -lsa /tmp/smokey.json
596 -rw-r--r-- 1 smokey smokey 606719 Mar 30 09:00 /tmp/smokey.json

$ sudo service smokey-loop status
smokey-loop start/running, process 10349

$ ls -lsa /tmp/smokey.json
596 -rw-r--r-- 1 smokey smokey 606692 Mar 30 09:03 /tmp/smokey.json
```

PR #11607, opened 2022-03-28
Remove unused packages: Selenium, QT and Xvfb

ollietreend:
This PR removes Selenium Standalone Server, QT and Xvfb from machines that run automated browser tests.

These packages were previously needed by [capybara-webkit][], which is no longer used. Capybara tests now use ChromeDriver, which is installed by the [govuk_chromedriver][] module.

[capybara-webkit]: https://github.com/thoughtbot/capybara-webkit
[govuk_chromedriver]: https://github.com/alphagov/govuk-puppet/blob/ff1632486beab74677bb5e9ad27e8a5036cf3061/modules/govuk_chromedriver/manifests/init.pp

### Step one: ensure absent

This is step one of a two-step process. It ensures the packages have been removed from affected machines.

The next step, to follow in another PR, is to remove the Puppet configuration for these resources entirely.

---

Trello: https://trello.com/c/8IaaOY1h/298-puppet-clean-up-remove-selenium-and-qt-from-machines


kevindew (approved):
This looks good to me, have you been able to run it on integration?

PR #11609, opened 2022-03-29
Change S3 path for backup to new path

rtrinque:
The path where we store backup have been changed, each database
has its own dir now.


sihugh (approved):
👍

AgaDufrat (approved):
Thanks!

PR #11610, opened 2022-03-29
Remove post-deploy Icinga alert for Smokey

benthorner:
https://trello.com/c/vIt6bDXI/258-stop-full-post-deploy-runs-of-smokey

Smokey is run twice after most deployments:

1. A full run triggered by the Deploy_App job [^1].

2. A partial run triggered by the Deploy_App_Downstream
   job (for CD apps).

This commit removes a single Icinga alert that would be updated
every time the job runs. The introduction of (2) has made this
meaningless: it's unclear if the status of the alert relates to
a full or partial run of Smokey.

We also run Smokey in a continuous "loop", with an Icinga alert
for each feature. These more fine-grained Smokey Loop alerts also
make the general one we're removing here redundant i.e. we should
just remove it instead of fixing it.

As I was making this change I noticed a bunch of variables that
aren't used anymore so I've removed them.

[^1]: https://github.com/alphagov/govuk-puppet/blob/a5d4360ff18dc64e4edef3992875d978f8492874/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb#L73

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/smokey.pp
+++ b/modules/govuk_jenkins/manifests/jobs/smokey.pp
@@ -10,26 +10,12 @@
 class govuk_jenkins::jobs::smokey (
   $environment = 'production',
 ) {
-  $environment_variables = $govuk_jenkins::environment_variables
-  $smokey_task = "test:${environment}"
-  $deploy_jenkins_domain = hiera('deploy_jenkins_domain')
-
   include govuk::apps::smokey
-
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):

benthorner:
Thanks for reviewing @ChrisBAshton !

Would you mind deploying this, since I don't have access?

I've already tested it on Integration so I don't you'll need to do any more testing.

PR #11611, opened 2022-03-30
Stop duplicate full post-deploy Smokey runs

benthorner:
https://trello.com/c/vIt6bDXI/258-stop-full-post-deploy-runs-of-smokey

**Deadline for discussion: 2022-04-13**

Currently we run Smokey in 3 different ways:

1. A "full" run of Smokey after each deployment, triggered by the
Deploy_App job (see commit). This runs all of the scenarios.

2. A partial run of Smokey as part of a Continuous Deployment. This
is triggered by the Deploy_App_Downstream job [^1].

3. Continuous "full" runs of Smokey as part of the Smokey Loop [^2].
Each feature has its own alert in Icinga.

This commit removes the first way we run Smokey:

- It's not necessary for continuously deployed apps, where the 
full, duplicate run blocks the CD pipeline.

- For manually deployed apps, it's still possible to run Smokey
manually after a deployment has completed.

Making this change will help to simplify and clarify the purpose of
Smokey as we work towards guidance on what tests go in it and a
general high-level testing strategy for GOV.UK. It will not affect
the third use case (alerting) of Smokey[^3].

Note about manual deployments
=============================

While having to run Smokey manually for manually deployed apps is
slightly less convenient, it also has advantages:

- It ensures the deployer is looking at a run of Smokey that is
definitely associated with their deployment. At the moment it's too
easy to not check if your deployment triggered the last run.

- It allows the deployer to do a partial run of Smokey - assuming
they've added support for this [^4].*

- It creates a stronger mental association between doing a deploy
and running Smokey - as opposed to passively checking it's passed,
which is easier to forget as it requires less thought.

*We could change Deploy_App to do this but we would have to do it
for every app. It makes more sense to invest effort in moving apps
to Continuous Deployment instead.

[^1]: https://github.com/alphagov/govuk-puppet/blob/77f85be19ceed239f6d652cc07207266b165fb9f/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb#L14
[^2]: https://docs.publishing.service.gov.uk/manual/alerts/smokey-loop-tests.html
[^3]: https://github.com/alphagov/govuk-puppet/pull/11610
[^4]: https://github.com/alphagov/smokey/blob/main/docs/writing-tests.md

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
@@ -69,9 +69,6 @@
                     current-parameters: true
         <% end %>
     publishers:
-        - trigger:
-            project: Smokey
-            threshold: UNSTABLE
         - text-finder:
             regexp: "DOCKER TAG FAILED"
             also-check-console-output: true


kevindew:
This sounds reasonable to me. I see the scenario you've referenced as something of a false friend - it looks like a safety net but I don't think people pay attention to it and a failure doesn't have any meaningful consequences.

ChrisBAshton (approved):
Makes sense to me 👍 It has the added bonus of further incentivising people to move apps to CD - one less part of the workflow to do manually!

We may need to spread awareness of the change though, as some developers will be surprised that their merge no longer triggers a Smokey run. In a worst case scenario, they may even look at the most recent build and assume that that was the run triggered by their merge.

1pretz1 (approved):
Change looks good 👍 Although it will be a bit more of a pain to deploy things manually, hopefully it will encourage the migration to CD and force people to actually use and check Smokey (if they don't already). 

+1 to Chris's comment concerning communication. We'd need to ensure this is known amongst all the developers and seen as mandatory step of manually deploying a change, as it'd be easy to deploy something and not run Smokey, as there is nothing forcing you to do it.

benthorner:
Thanks for the reviews so far everyone!

In response to @1pretz1 and @ChrisBAshton I'm suggesting we make this change: https://github.com/alphagov/release/pull/952. Aside from promoting this in Slack, is there anything else we could do to raise awareness of the change?

(I won't be around for the next GOV.UK Tech Fortnightly, but that would be a good forum to announce it verbally.)

PR #11613, opened 2022-03-31
Part 2: Remove unused config: Selenium, QT and Xvfb

ollietreend:
This PR removes the configuration that was used for Selenium Standalone Server, QT and Xvfb on machines that run automated browser tests.

These packages were previously needed by [capybara-webkit][], which is no longer used. Capybara tests now use ChromeDriver, which is installed by the [govuk_chromedriver][] module.

[capybara-webkit]: https://github.com/thoughtbot/capybara-webkit
[govuk_chromedriver]: https://github.com/alphagov/govuk-puppet/blob/ff1632486beab74677bb5e9ad27e8a5036cf3061/modules/govuk_chromedriver/manifests/init.pp

### Step two: clean up config

This is step two of a two-step process. The files and packages were removed from machines by #11607, which was tagged as [release_24137](https://github.com/alphagov/govuk-puppet/releases/tag/release_24137).

This second step cleans up the now redundant configuration file from this Puppet repository.

---

Trello: https://trello.com/c/8IaaOY1h/298-puppet-clean-up-remove-selenium-and-qt-from-machines


kevindew (approved):
Nice one. Feel free to merge and I'll work this through with the current deploy.

PR #11628, opened 2022-04-06
Update assets path for contacts-admin

kevindew:
This corrects the previously incorrect asset path (contacts defaults to "/assets") which has meant that the assets haven't been set with the correct cache headers. This updates this path to match the new one introduced in: https://github.com/alphagov/contacts-admin/pull/959


sengi (approved):
Good find!

PR #11630, opened 2022-04-08
Auto-delete jobs removed from `govuk_ci::master::pipeline_jobs`

ChrisBAshton:
When we remove jobs from the hieradata ([1]), the job doesn't
actually get deleted in Jenkins ([2]). The process to remove the
job is manual ([3]). Consequently, there are a number of jobs in
Jenkins which should be deleted:

> Screenshot of Jenkins jobs, ordered by "Last Success"
> Just 18 out of 51 repositories have had a (successful) build in the past 3 months.
> This is a sign that many of these are no longer active, and should be deleted.

Rather than rely on developers manually removing things from two
places, we can use Puppet to manage the resources. Taking
inspiration from a StackOverflow suggestion ([4]), updated to use
the older Puppet 3 syntax, we can have Puppet manage the
`/var/lib/jenkins/jobs` directory on the `ci_master` machine,
instructing it to purge any directories it finds within, _apart_
from those we've explicitly told it to keep.

[1]: https://github.com/alphagov/govuk-puppet/commit/876b7e188bcb3564557dad301bf4ccbd14a856a6
[2]: https://ci.integration.publishing.service.gov.uk/job/govuk_app_config/
[3]: https://docs.publishing.service.gov.uk/manual/retiring-an-application.html#12-remove-jobs-in-ci
[4]: https://stackoverflow.com/questions/44721855/delete-unmanaged-files-of-a-directory-via-puppet

### DIFF
--- a/modules/govuk_ci/manifests/master.pp
+++ b/modules/govuk_ci/manifests/master.pp
@@ -46,6 +46,15 @@ class govuk_ci::master (
 
   # Add pipeline jobs from applications hash in Hieradata
   create_resources(govuk_ci::job, $pipeline_jobs)
+  # Manually delete all jobs which aren't in the hash
+  file { '/usr/local/bin/remove_old_jenkins_jobs':
+    ensure  => present,
+    mode    => '0755',
+    content => template('govuk_ci/usr/local/bin/remove_old_jenkins_jobs.sh.erb'),
+  }
+  exec { 'single run of remove_old_jenkins_jobs after adding pipeline jobs':
+    command => '/usr/local/bin/remove_old_jenkins_jobs',
+  }
 
   # Only add this configuration for CI master Jenkins instances
   file { '/var/lib/jenkins/github-plugin-configuration.xml':
--- a/modules/govuk_ci/manifests/master.pp
+++ b/modules/govuk_ci/manifests/master.pp
@@ -4,6 +4,12 @@
 #
 # === Parameters:
 #
+# [*pipeline_jobs*]
+#   Hash of applications to create jobs for
+#
+# [*job_builder_jobs*]
+#   Array of jobs created with the Job Builder plugin
+#
 # [*github_client_id*]
 #   The Github client ID is used as the user to authenticate against Github.
 #
@@ -18,6 +24,7 @@ class govuk_ci::master (
   $github_client_secret_encrypted,
   $jenkins_api_token,
   $pipeline_jobs = {},
+  $job_builder_jobs = [],
   $environment_variables = {},
   $ci_agents = {},
   $credentials_id,
--- a/modules/govuk_ci/templates/usr/local/bin/remove_old_jenkins_jobs.sh.erb
+++ b/modules/govuk_ci/templates/usr/local/bin/remove_old_jenkins_jobs.sh.erb
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+dirs_to_keep='<%= @pipeline_jobs.keys.join("\n") %>'
+
+ls -A /var/lib/jenkins/jobs | grep -vF "$dirs_to_keep" | xargs rm -rf
--- a/modules/govuk_ci/templates/usr/local/bin/remove_old_jenkins_jobs.sh.erb
+++ b/modules/govuk_ci/templates/usr/local/bin/remove_old_jenkins_jobs.sh.erb
@@ -2,4 +2,4 @@
 
 dirs_to_keep='<%= (@pipeline_jobs.keys + @job_builder_jobs.map { |job| job.split("::").last }).join("\n") %>'
 
-ls -A /var/lib/jenkins/jobs | grep -vF "$dirs_to_keep" | xargs rm -rf
+ls -A /var/lib/jenkins/jobs | grep -ivxF "$dirs_to_keep" | xargs rm -rf
--- a/modules/govuk_ci/templates/usr/local/bin/remove_old_jenkins_jobs.sh.erb
+++ b/modules/govuk_ci/templates/usr/local/bin/remove_old_jenkins_jobs.sh.erb
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-dirs_to_keep='<%= @pipeline_jobs.keys.join("\n") %>'
+dirs_to_keep='<%= (@pipeline_jobs.keys + @job_builder_jobs.map { |job| job.split("::").last }).join("\n") %>'
 
 ls -A /var/lib/jenkins/jobs | grep -vF "$dirs_to_keep" | xargs rm -rf
--- a/hieradata_aws/class/integration/ci_master.yaml
+++ b/hieradata_aws/class/integration/ci_master.yaml
@@ -139,10 +139,11 @@ govuk_jenkins::config::views:
 govuk_jenkins::config::create_agent_role: true
 govuk_jenkins::config::executors: '0'
 
-govuk_jenkins::job_builder::jobs:
+govuk_jenkins::job_builder::jobs: &job_builder_jobs
   - govuk_jenkins::jobs::deploy_app_downstream
   - govuk_jenkins::jobs::deploy_puppet_downstream
   - govuk_jenkins::jobs::build_fpm_package
+govuk_ci::master::job_builder_jobs: *job_builder_jobs
 
 govuk_jenkins::jobs::deploy_app_downstream::jenkins_downstream_api_password: "%{hiera('govuk_jenkins::jobs::deploy_app_downstream::ci_jenkins_downstream_api_password')}"
 


modules/govuk_ci/manifests/master.pp
```diff
@@ -46,6 +46,14 @@
 
   # Add pipeline jobs from applications hash in Hieradata
   create_resources(govuk_ci::job, $pipeline_jobs)
+  # Manually delete all jobs which aren't in the hash
+  # lint:ignore:single_quote_string_with_variables
+  $full_rm_command = sprintf('%s%s%s', 'key_str=$(IFS=\|; printf \'%s\' "${', keys($pipeline_jobs), '[*]}"); ls -a | grep -vE "$key_str|\.|\.\." | xargs rm -rf')
```
sengi, line 51: This is very hard to read and even harder to test. How about making it a separate shell script? You could pass the list of directories to keep via a Puppet template.

ChrisBAshton: Thanks @sengi - I've moved this to its own script now. Do you see any problems with the approach before I give it a whirl on Integration?

modules/govuk_ci/templates/usr/local/bin/remove_old_jenkins_jobs.sh.erb
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+key_str=<%= @pipeline_jobs.keys.join("|") %>
+ls -a /var/lib/jenkins/jobs | grep -vE "$key_str|\.|\.\." | xargs rm -rf
```
sengi, line 4: I'd recommend using the `-F` fixed string option to grep (takes a newline-separated list of fixed strings to match against), to avoid potential issues with quoting. Also `ls -A` excludes the current and parent dir for you so you can simplify the grep a bit. So something like

```suggestion
dirs_to_keep='<%= @pipeline_jobs.keys.join("\n") %>'

ls -A /var/lib/jenkins/jobs | grep -vF "$dirs_to_keep" | xargs rm -rf
```

sengi: (I'm assuming you want to delete directories that start with `.`. If not then just omit the `-A` altogether of course.)

modules/govuk_ci/manifests/master.pp
```diff
@@ -46,6 +46,15 @@
...
   # Add pipeline jobs from applications hash in Hieradata
   create_resources(govuk_ci::job, $pipeline_jobs)
+  # Manually delete all jobs which aren't in the hash
+  file { '/usr/local/bin/remove_old_jenkins_jobs':
+    ensure  => present,
+    mode    => '0755',
+    content => template('govuk_ci/usr/local/bin/remove_old_jenkins_jobs.sh.erb'),
+  }
+  exec { 'single run of remove_old_jenkins_jobs after adding pipeline jobs from applications hash':
+    cwd => '/usr/local/bin/remove_old_jenkins_jobs',
```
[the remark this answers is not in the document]
ChrisBAshton: 🤦 good spot!

modules/govuk_ci/templates/usr/local/bin/remove_old_jenkins_jobs.sh.erb
ChrisBAshton: Nice 👍

sengi (approved):
LGTM - be sure to test it carefully though! 😅

ChrisBAshton:
I've now tested the theory:

```
$ gds govuk connect -e integration ssh ci_master

$ cd /var/lib/jenkins/jobs
$ rm -rf govuk_app_config
# https://ci.integration.publishing.service.gov.uk/job/govuk_app_config/ still there

# restart Jenkins

# https://ci.integration.publishing.service.gov.uk/job/govuk_app_config/ now 404's 🎉
```

So now deploying the script to integration, without the `rm -rf`, just to check that it's going to remove the right directories.

ChrisBAshton:
```
$ /usr/local/bin/remove_old_jenkins_jobs 
backdrop-write
build_fpm_package
bulk-merger
content-audit-tool
Deploy_App_Downstream
Deploy_Puppet_Downstream
email-alert-monitoring
gapy
gds-scala-common
govspeak
govuk_admin_template
govuk-app-deployment
govuk-aws
govuk-cdn-logs-monitor
govuk-content-schema-test-helpers
govuk-delivery
govuk-dependencies
govuk-deploy-lag-badger
govuk-developer-docs
govuk-diff-pages
govuk-dns
govuk-dns-config
govuk-docker
govuk-lint
govuk_mirrorer
govuk_navigation_helpers
govuk_publishing_components
govuk-saas-config
govuk-secrets
govuk_security_audit
govuk-tagging-monitor
govuk-user-reviewer
performance-datastore
rubocop-govuk
rummager
seal
spotlight
spotlight-performance
trigger-deploy
```

...ahhhh - I hadn't factored in the jobs that _aren't_ [created via multibranch pipeline](https://github.com/alphagov/govuk-puppet/blob/575728073c94588e12d6e91ce82f9ab3abeefa24/hieradata_aws/common.yaml#L938-L1033). These are [Deploy_App_Downstream](https://ci.integration.publishing.service.gov.uk/job/Deploy_App_Downstream/), [Deploy_Puppet_Downstream](https://ci.integration.publishing.service.gov.uk/job/Deploy_Puppet_Downstream/) and [Build fpm package](https://ci.integration.publishing.service.gov.uk/job/build_fpm_package/), and luckily they're all defined [here](https://github.com/alphagov/govuk-puppet/blob/4245835e9b69d803e020a2c7fb177f3e4379f880/hieradata_aws/class/integration/ci_master.yaml#L142-L145), so hopefully I can pass those to the script too.

Everything else on the list looks ripe for deletion. There are others that look suitable for deletion which aren't on the output above, but they appear in `govuk_ci::master::pipeline_jobs`, so that's just a config issue.

ChrisBAshton:
Branch tidied up now. Tested on integration, pretty confident it'll do what we hope 👍 Are you able to take another look @sengi?
Also happy to defer deploying this until after the Easter weekend if that would be wise!

modules/govuk_ci/templates/usr/local/bin/remove_old_jenkins_jobs.sh.erb
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+dirs_to_keep='<%= (@pipeline_jobs.keys + @job_builder_jobs.map { |job| job.split("::").last }).join("\n") %>'
```
sengi, line 3: It would be much better to put this logic in the manifest rather than in the template. The template wants to be receiving a simple list to fill in here. Putting the logic in the Puppet file will allow you to write it more clearly (e.g. by giving meaningful names to intermediate values if it helps understanding) and avoid one-liners like this.

ChrisBAshton: I completely agree with you, except that I'm not sure we can accomplish this sort of thing on our version of Puppet. We don't even have a working `keys()` method in Puppet 3 😢 

Part of the complexity lies in the two different data structures:

- `pipeline_jobs` == `{ job_name: 'foo' }`
- `job_builder_jobs` == `["foo::bar::job_name"]`

I attempted to keep things DRY by [reusing the `govuk_jenkins::job_builder::jobs` hieradata](https://github.com/alphagov/govuk-puppet/blob/acafcb6eb24b855dc5fd245ea2353b5337d55836/hieradata_aws/class/integration/ci_master.yaml#L146). But on reflection, a small bit of duplication to tidy up the template and make it readable seems a net win.

ChrisBAshton: Hmmm even the simple hash concatenation appears not to be supported, even if I [enable 'future' parser mode](https://github.com/alphagov/govuk-puppet/pull/11630/commits/6348a9671848418b12277e57b6e449e925eb4ca4). Any thoughts, @sengi?

sengi: Oh gosh, sorry had no idea that would be a rabbithole. I wouldn't worry about it then, let's just do it in the template 😅

ChrisBAshton: Cool, I've removed those commits now. For future reference, they were https://github.com/alphagov/govuk-puppet/commit/d2093fb4477f2e001cadde4850312b13a6b6c057 and https://github.com/alphagov/govuk-puppet/commit/6348a9671848418b12277e57b6e449e925eb4ca4.

Could you re-review @sengi, and I'll look at getting this deployed today?

sengi (approved):
LGTM - thanks!

PR #11633, opened 2022-04-13
Redis access for Feedback in integration only

KludgeKML:
## WHAT

Allow Feedback to use the shared REDIS backend so that Rack::Attack has a backing cache.

## WHY

Adding rate-limiting to Feedback app via Rack::Attack - for testing we just need quick access to a redis cluster, so use the shared one. We won't do this in production because we want to avoid shared backing instances, especially one that might have to defend against a DoS.

https://trello.com/c/9qqjfqVo/1186-add-app-level-rate-limiting-to-feedback

Feedback PR: https://github.com/alphagov/feedback/pull/1390

### DIFF
--- a/modules/govuk/manifests/apps/feedback.pp
+++ b/modules/govuk/manifests/apps/feedback.pp
@@ -36,6 +36,14 @@
 # [*govuk_notify_accessible_format_request_reply_to_id*]
 #   Accessible format request email address reply_to ID for GOV.UK Notify
 #
+# [*redis_host*]
+#   Redis host for Rack::Attack.
+#   Default: undef
+#
+# [*redis_port*]
+#   Redis port for Rack::Attack.
+#   Default: undef
+#
 class govuk::apps::feedback(
   $port,
   $sentry_dsn = undef,
@@ -51,6 +59,8 @@ class govuk::apps::feedback(
   $govuk_notify_survey_signup_reply_to_id,
   $govuk_notify_accessible_format_request_template_id = undef,
   $govuk_notify_accessible_format_request_reply_to_id = undef,
+  $redis_host = undef,
+  $redis_port = undef,
 ) {
   $app_name = 'feedback'
 
@@ -139,5 +149,10 @@ class govuk::apps::feedback(
   govuk::app::envvar { "${title}-GOVUK_NOTIFY_ACCESSIBLE_FORMAT_REQUEST_REPLY_TO_ID":
     varname => 'GOVUK_NOTIFY_ACCESSIBLE_FORMAT_REQUEST_REPLY_TO_ID',
     value   => $govuk_notify_accessible_format_request_reply_to_id,
+  }
+
+  govuk::app::envvar::redis { $app_name:
+    host => $redis_host,
+    port => $redis_port,
   }
 }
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/feedback.pp
```diff
@@ -140,4 +150,9 @@
     varname => 'GOVUK_NOTIFY_ACCESSIBLE_FORMAT_REQUEST_REPLY_TO_ID',
     value   => $govuk_notify_accessible_format_request_reply_to_id,
   }
+
+  govuk::app::envvar::redis { $app_name:
+    host => $redis_host,
+    port => $redis_port,
```
richardTowers, line 156: I think these [will default to 127.0.0.1 / 6379](https://github.com/alphagov/govuk-puppet/blob/main/modules/govuk/manifests/app/envvar/redis.pp#L24-L25) in non-integration environments. Is that going to cause trouble?

You could wrap this with an `if $redis_host != undef {` just to be sure?

richardTowers (commented):
This looks about right to me - one cautious comment.

I assume that network wise, the backend server can already connect to the shared redis on integration (i.e. the security groups are already set up).

modules/govuk/manifests/apps/feedback.pp
KludgeKML: I... don't know - hopefully by the time it gets to non-integration environments we'll have a better change here (because we don't want rate-limiting cache stuff pumped into the redis instance everything else shares), but yeah, no harm in being cautious and putting in that wrapper.

KludgeKML:
I assume so - I think feedback is in the same security group as the other backend services, so it can actually talk to the redis host, it just doesn't know about it in-app. I'll try SSH-ing from one to the other to double-check.

modules/govuk/manifests/apps/feedback.pp
richardTowers: >  by the time it gets to non-integration environments

Always better to be careful with these things - if it gets merged to main, you should probably assume someone will deploy it everywhere. Thanks for the wrapper :)

richardTowers (approved):

KludgeKML:
Oh wait, I see your point - dang, feedback isn't a backend service, it's a frontend app, so it's probably in a sec group that doesn't have access to that redis instance.

 Hmm. So I need to see if there is an existing frontend redis instance (maybe there isn't). Maybe I have to fall back to doing this the painful but long-term correct way.

richardTowers:
> dang, feedback isn't a backend service, it's a frontend app, so it's probably in a sec group that doesn't have access to that redis instance.

I wouldn't necessarily be averse to opening up a security group so whichever machine feedback is running on can talk to the redis. It's not particularly elegant, but if it saves you time I don't mind being pragmatic.

KludgeKML:
Ah, wait - it seems like some existing frontend apps (frontend, email-alert-frontend) are set up to use redis, and they do point to the backend-redis instance (at least in integration, I don't have the permissions to check staging/prod). So I think this should be okay for the current purpose. I'll try deploying this branch and see what we get! :)

KludgeKML:
Hmm. Unless I'm missing some procedure, this doesn't appear to be all that's necessary.

- This branch of puppet deployed to integration
- wait 30 minutes
- deploy feedback branch to integration
- `% gds govuk connect app-console --environment integration feedback`
- `> ENV.key?("REDIS_URL")` is false, expecting it to be true. (there are no other redis env vars either)

KludgeKML:
Going to merge this and then remove the changes in a later merge - too painful to try to synchronise this and the feedback app.

PR #11634, opened 2022-04-13
Remove thread count alerts

sihugh:
These are rarely actionable, so let's try removing them.

The default warning level of 100 threads is a bit arbitrary and it's not clear what value to increase things to if this is consistently breached.

We currently have an alert for a router instance because the thread count is 102.

These are the current levels for apps where specified:

- Asset manager: 165
- Cache clearing service: 250
- Content store: 155
- Email alert api: 100

Defaults are:

- apps: 100
- workers: 50


richardTowers (approved):
Strongly agree with removing the alerts.

Will you do a follow up PR to tidy up the puppet?  Arguably not worth it I suppose - happy to leave that up to you.

PR #11639, opened 2022-04-14
 Reduce default local log retention to 7 days

spikeheap:
GOV.UK uses Logit for centralised log management and retention, and local logs will disappear with the migration to Kubernetes, so these logs don't form part of our log retention strategy. 

This commit brings the default local  log retention down from 31 days to 7, which will resolve the recurring out-of-disk errors  for Content Store. This also makes the default consistent with a couple of existing apps which override the default (search and cache), and makes it more obvious that these logs aren't necessary by pushing people to Logit for log queries.

This doesn't affect the local backup retention for other things (e.g.local Mongo).

#### Potential improvement

By removing some redundant logging config (e.g. in `hieradata_aws/class/integration/cache.yaml`), we end up with empty files. If it's safe to remove them, I'll do that as part of this PR, but I wasn't confident Puppet didn't rely on the files presence for anything.


ChrisBAshton (approved):
LGTM ✅ I can't see any historic justification for the previous values. The 31 day threshold was introduced in 21b3bfc20b70296c67a89b54e253b3f6a8bb1cdf, and made parameterisable in 50266cc44d53fdb52c408378e6d48989c62c9499: "We want to be able to increase the number of days that logrotate keeps logs for on some machines."

The machine referenced is the `backend_lb`, which wanted 45 days of logs: 84abbebd7f965bc4d115464c265bc4b4e6d18f40. Doesn't look like that's a thing anymore though.

sihugh (approved):
This is probably weeknote-worthy, but looks good to me 👍

spikeheap:
Excellent, thanks everyone 🙇. @kevindew confirmed the empty YAML files aren't needed, so I've added a commit to remove them.

PR #11645, opened 2022-04-20
Remove redundant env vars for Smokey

benthorner:
https://trello.com/c/DL5omRGP/230-smokey-needs-java-and-a-glitchy-proxy-server-to-run-tests

These became redundant in [1] and [2].

Note that "http_username/password" are still used elsewhere [3],
but "smokey_rate_limit_token" should be removed from secrets -
it's passed to Smokey and to Nginx [5].

[1]: https://github.com/alphagov/smokey/pull/918
[2]: https://github.com/alphagov/smokey/pull/913
[3]: https://github.com/alphagov/govuk-puppet/blob/91164cb02f9504c8d7245de0444ff61a7ae3f957/hieradata_aws/common.yaml#L849
[4]: https://github.com/alphagov/govuk-puppet/blob/91164cb02f9504c8d7245de0444ff61a7ae3f957/modules/router/manifests/nginx.pp#L31


KludgeKML:
Some of these things are in govuk-secrets, so perhaps add an issue there when this is merged to clean them up?

KludgeKML (approved):
Looks good to me.

benthorner:
Thanks for the review and the nudge @KludgeKML !

I was thinking someone else would need to do the govuk-secrets change but then realised it's just a straight deletion, so I can do it myself: https://github.com/alphagov/govuk-secrets/pull/1288

PR #11647, opened 2022-04-20
Prune rotated files older than rotate value.

spikeheap:
This commit introduces a clean-up to delete any left-over rotated files that are older than
the current `rotate` value.

Under normal circumstances, logrotate will remove logs as they rotate off the end. We set the maximum
number of rotations (in our case days) with the 'rotate' value.

However, an edge-case exists where older files will not be removed if we've reduced the value of rotate
between runs. This is because logrotate only looks at `rotate + 1` to check for removal, it doesn't
attempt to manage all files in that directory.

This clean-up mitigates that edge-case, ensuring that all files older than the maximum number of days are
removed. There are a couple of caveats:

1. The clean-up uses modification time, so if we touch any older files they won't be removed.
2. We assume that logrotate will continue to be daily. Any other setting will break the clean-up. This is not
   a configurable parameter in our Puppet setup, so this risk feels acceptable

### Handy bits for testing
For local testing, I used `touch`:

```bash
mkdir -p /var/log/test-app/

touch -d "1 days ago" /var/log/test-app/app.log.1
touch -d "2 days ago" /var/log/test-app/app.log.2
touch -d "3 days ago" /var/log/test-app/app.log.3
touch -d "4 days ago" /var/log/test-app/app.log.4
touch -d "5 days ago" /var/log/test-app/app.log.5

find /var/log/test-app/*.log* -maxdepth 1 -mtime +2 -name "*.log.*" -delete &>/dev/null

# Should show only up to .2
ls /var/log/test-app/
```

### Testing on integration

There are a load of old unrotated files on integration. I tested this change by manually patching `govuk-publisher`'s log rotate config on `ec2-integration-blue-backend-ip-10-1-5-190`, and force-running logrotate. 

> A 😡 lesson: logrotate's `--debug` argument does provide debug output... but it _also_ doesn't actually run the command, somewhat surprising if you're wondering why files aren't being removed but it says it's running the script.

If you know a nicer way to do this, let me know in the comments below. Don't forget to like and subscribe 😛

### DIFF
--- a/modules/logrotate/templates/logrotate.conf.erb
+++ b/modules/logrotate/templates/logrotate.conf.erb
@@ -2,7 +2,10 @@
   <%- if @user && @group -%>
   su <%= @user -%> <%= @group %>
   <%- end -%>
+
+  # To use anything other than daily, the clean up script in lastaction MUST be updated
   daily
+
   missingok
   rotate <%= @days_to_keep %>
   compress
@@ -40,5 +43,11 @@
     # Remove empty .1 log files. There seems to be an issue where logrotate
     # will refuse to rotate the logs if an empty .1 log file exists.
     find <%= @matches %>* -wholename '/var/log/*' -name '*.1' -size 0 -delete
+  endscript
+  lastaction
+    # Clean up files older than maximum rotation. No-op unless "rotate" value is reduced,
+    # because logrotate only deletes rotate+1, not all older files.
+    # Thanks to https://serverfault.com/questions/636844/making-logrotate-remove-old-logs-after-reducing-rotate-value
+    find <%= @matches %>.*  -maxdepth 1 -mtime +<%= @days_to_keep %> -delete -printf "deleting old file: %p\n"
   endscript
 }


modules/logrotate/templates/logrotate.conf.erb
```diff
@@ -41,4 +44,10 @@
     # will refuse to rotate the logs if an empty .1 log file exists.
     find <%= @matches %>* -wholename '/var/log/*' -name '*.1' -size 0 -delete
   endscript
+  lastaction
+    # Clean up files older than maximum rotation. No-op unless "rotate" value is reduced,
+    # because logrotate only deletes rotate+1, not all older files.
+    # Thanks to https://serverfault.com/questions/636844/making-logrotate-remove-old-logs-after-reducing-rotate-value
+    find <%= @matches %>.*  -maxdepth 1 -mtime +<%= @days_to_keep %> -delete &>/dev/null
```
kevindew, line 51: Rather than redirecting stdout to null I think it'd be slightly nicer if we had stdout that said what this was doing. From a quick test:

```
find <%= @matches %>.* -maxdepth 1 -mtime +<%= @days_to_keep %> -delete -printf "deleting old file: %p\n"
```

should add to the output e.g:

```
...
old log /var/log/content-store/app.out.log.0.gz does not exist
copying /var/log/content-store/app.out.log to /var/log/content-store/app.out.log.1
truncating /var/log/content-store/app.out.log
compressing log with: /bin/gzip
removing old log /var/log/content-store/app.out.log.8.gz
running last action script
deleting old file: /var/log/content-store/app.err.log.9.gz
deleting old file: /var/log/content-store/app.out.log.9.gz
```

spikeheap: Great point @kevindew, I've updated things to run exactly that. 🙇

PR #11648, opened 2022-04-20
Add manually added jobs to the 'remove_old_jenkins_jobs' allowlist

ChrisBAshton:
The 'remove_old_jenkins_jobs' script is removing the govuk-dns
and govuk-dns-config jobs because they don't appear in either the
`pipeline_jobs` hash nor the `job_builder_jobs` array. govuk-dns
and govuk-dns-config are special cases in that they are the only
jobs declared with the `govuk_ci::job` helper.


sengi (approved):
Thanks!

PR #11550, opened 2022-02-15
Add GTM env vars to government frontend

hannako:
We want to test GA4 tracking in integration only.

Environment variables provided here: https://trello.com/c/YgdDINY2/61-set-up-gtm-container-with-custom-environments
Trello: https://trello.com/c/kBaHV7MG/784-ga4-add-gtm-environment-variables-to-integration

PR #11653, opened 2022-04-22
Add GTM env vars to frontend and collections

andysellick:
## What
Add some env vars to `frontend` and `collections` for GA4 testing purposes.

## Why
We want to test some of the new GA4 stuff on integration only in these applications and the code will depend on these variables to work. This involves some repetition, which ideally would be avoided, but this is a temporary measure for testing purposes.

PR based on previous work: https://github.com/alphagov/govuk-puppet/pull/11550

### DIFF
--- a/modules/govuk/manifests/apps/collections.pp
+++ b/modules/govuk/manifests/apps/collections.pp
@@ -31,6 +31,15 @@
 # [*memcache_servers*]
 #   URL of a shared memcache cluster.
 #
+# [*google_tag_manager_id*]
+#   The ID for the Google Tag Manager account
+#
+# [*google_tag_manager_preview*]
+#   Allows a tag to be previewed in the Google Tag Manager interface
+#
+# [*google_tag_manager_auth*]
+#   The identifier of an environment for Google Tag Manager
+#
 class govuk::apps::collections(
   $vhost = 'collections',
   $port,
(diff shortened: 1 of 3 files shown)


modules/govuk/manifests/apps/collections.pp
```diff
@@ -31,6 +31,15 @@
...
 #
+# [*google_tag_manager_id*]
+#   The ID for the Google Tag Manager account
+#
+# [*google_tag_manager_preview*]
+#   Allows a tag to be previewed in the Google Tag Manager interface
+#
+# [*google_tag_manager_auth*]
+#   The identifier of an environment for Google Tag Manager
+#
```
gclssvglx, line 42: I wonder if `google_tag_manager_env_id` would be more descriptive for this?

gclssvglx (commented):
This is looking really good @andysellick 👍  I'm just being a really picky on the name of one of the env vars.

modules/govuk/manifests/apps/collections.pp
andysellick: Yeah it probably is, but `google_tag_manager_auth` is already used by content-data-admin and content-publisher (two things I can't change) so I think I'd rather stick with consistency for the moment.

gclssvglx: Fair point - at least it's consistently a bad name 😉

gclssvglx (approved):
Great stuff 👍

modules/govuk/manifests/apps/collections.pp
andysellick: Thanks. These should be temporary anyway while we test this stuff, it'll be removed before go live.

PR #11663, opened 2022-05-03
Make CDN cache clear Jenkins task equivalent to cache-clear service

kevindew:
Trello: https://trello.com/c/TX9wLihh/101-cache-clearing-app-only-clears-path-not-query-strings

This updates the "Clear CDN Cache" Jenkins task so that it can be equivalent of functionality to the cache-clearing-service's: [`cache:clear`task](https://github.com/alphagov/cache-clearing-service/blob/64a4bebe97497102f286bfab6e613e33f61f0183/lib/tasks/cache.rake#L18-L22 which is done to reduce dependencies on that tool ahead of ideas to retire it.

There's more details in the commits, but essentially what this does is:

- moves the hardcoded urls into a Jenkins parameter, so any collection of URLs can be cleared
- automatically clears from internal varnish too, where applicable
- will accept a full URL (as per cache-clearing service) to clear items such as assets.

Once this is merged, I'll look at removing references in dev docs to cache-clearing service to reduce the coupling of that to day-to-day operations


ChrisBAshton (approved):
LGTM 👍

PR #11666, opened 2022-05-03
Skip SSH key strength check for ECDSA keys

benjamineskola:
This appears only to be relevant for RSA and DSA keys and was already being ignored for newer Ed25519 keys, but ECDSA keys were still being checked.


AgaDufrat (approved):
👍  Makes sense to change this test. 

On a separate note, I'm wondering if we should re-visit this guidance:
https://github.com/alphagov/govuk-developer-docs/blob/main/source/manual/get-started.html.md?plain=1#L155-L159

benjamineskola:
@AgaDufrat I think you're right about updating that documentation. The second line in particular seems misleading: it won't necessarily have `==` in (that's just random chance that base16 worked out that way) and it doesn't have to end with an email address, although it does by default. I'll take a look at that at some point.

PR #11671, opened 2022-05-09
Make logrotate aware of disk space

rtrinque:
Delete logs files above a threshold in size in chronological order
while disk space usage is above a given threshold.

### DIFF
--- a/modules/logrotate/templates/logrotate.conf.erb
+++ b/modules/logrotate/templates/logrotate.conf.erb
@@ -43,6 +43,16 @@
     # Remove empty .1 log files. There seems to be an issue where logrotate
     # will refuse to rotate the logs if an empty .1 log file exists.
     find <%= @matches %>* -wholename '/var/log/*' -name '*.1' -size 0 -delete
+    # Remove files bigger than $SIZE in chronological order while use of  
+    # disk space on the logging partition is greater than $MAX_USE
+    SIZE='300M'
+    MAX_USE=85
+    for file in $(find <%= @matches %>* -type f -size +$SIZE -exec ls -tr {} +); do
+      USE=$(df /var/log --output=pcent | sed -n '2p' | tr -d %)
+        if [ $USE -gt $MAX_USE ]; then
+          rm $file
+        fi
+    done;
   endscript
   lastaction
     # Clean up files older than maximum rotation. No-op unless "rotate" value is reduced,


modules/logrotate/templates/logrotate.conf.erb
```diff
@@ -43,6 +43,16 @@
     # Remove empty .1 log files. There seems to be an issue where logrotate
     # will refuse to rotate the logs if an empty .1 log file exists.
     find <%= @matches %>* -wholename '/var/log/*' -name '*.1' -size 0 -delete
+    # Remove files bigger than $SIZE in chronological order while use of  
+    # disk space on the logging partition is greater than $MAX_USE
+    SIZE='300M'
+    MAX_USE=85
```
ChrisBAshton, line 49: Nice one putting this together! 👏 Can you add a bit more detail to the commit message? I think it's missing:

1) why we're making this change (ci_master machines running out of space)
2) where this gets applied (all machines? Only certain classes?)
3) why we chose a 300M threshold (e.g. "only n% of the logs are above 300M, but account for y% of the disk space")

modules/logrotate/templates/logrotate.conf.erb
```diff
@@ -43,6 +43,16 @@
...
+    for file in $(find <%= @matches %>* -type f -size +$SIZE -exec ls -tr {} +); do
+      USE=$(df /var/log --output=pcent | sed -n '2p' | tr -d %)
+        if [ $USE -gt $MAX_USE ]; then
+          rm $file
```
ChrisBAshton, line 53: If I understand this correctly, we loop over every log file (that is more than 300M), and check if disk space is at >85%. If it is, remove the file. Once we reach <=85%, we won't be removing any more, but we'll still be looping over the rest of the files.

I think it would be logically easier to follow by inverting it like this:

```
while (not enough disk space)
  remove the oldest file that is >300mb
```

But this is quite nitpicky 😅 also not sure how slow the `find` query is, in which case my suggestion would be less efficient. So happy to leave as-is 👍

ChrisBAshton (commented):
LGTM 👍 but would like a bit more commit message detail before merge.

modules/logrotate/templates/logrotate.conf.erb
rtrinque: Commit message amended.

modules/logrotate/templates/logrotate.conf.erb
rtrinque: Yes that's the idea, and I agree the logic of the loop is a bit backward.
The reason for that is mainly to keep it simple as find gives us a nice list to go over in a simple way, using the disk space left to control the loop would make more sense but would introduce more control work as we would need some bit of logic to break the loop if we run out of files to delete.
As it is this we just get a list of files we think we can do something about, work and them, and then we've done what we could to alleviate the space issue.

ChrisBAshton (approved):
Thanks 👍

Issue #11658, opened 2022-04-27
Remove temporary changes from PR 11633


KludgeKML:
This PR is needed temporarily to allow testing in integration without env vars being clobbered all the time, but should be removed when testing is complete to allow a proper fix to the problem:

https://github.com/alphagov/govuk-puppet/pull/11633

If this issue exists after 12th May 2022, contact me to deal with it!

PR #11672, opened 2022-05-09
Move from backend-redis to rate-limit-redis in integration.

KludgeKML:
Move feedback from using the shared backend redis cluster to a new cluster (only in integration for the moment).

Fixes temporary work in https://github.com/alphagov/govuk-puppet/pull/11633 
...and resolves https://github.com/alphagov/govuk-puppet/issues/11658

https://trello.com/c/9qqjfqVo/1186-add-app-level-rate-limiting-to-feedback

### DIFF
--- a/hieradata_aws/class/integration/feedback.yaml
+++ b/hieradata_aws/class/integration/feedback.yaml
@@ -1,4 +1,0 @@
----
-
-govuk::apps::feedback::redis_host: 'backend-redis'
-govuk::apps::feedback::redis_port: '6379'
--- a/hieradata_aws/class/integration/frontend.yaml
+++ b/hieradata_aws/class/integration/frontend.yaml
@@ -1,0 +1,4 @@
+---
+
+govuk::apps::feedback::redis_host: 'rate-limit-redis'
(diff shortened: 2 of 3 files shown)


hieradata_aws/class/integration/feedback.yaml
```diff
@@ -1,4 +1,4 @@
 ---
 
-govuk::apps::feedback::redis_host: 'backend-redis'
```
1pretz1, line 3: Are these variables / file necessary, as they're already defined in [hieradata_aws/integration.yaml](https://github.com/alphagov/govuk-puppet/pull/11672/files#diff-769a9254659e825834006a559d95c14cdc1292e0a48758685f42b4d99dafd296)?

KludgeKML: Good question! Probably not. This was where I originally put the variables, but I think maybe neither of those places is right - it should probably be in hieradata_aws/class/integration/frontend.yaml. Let me check with someone who might know what the ideal is here.

KludgeKML: Fixed - by moving it into the right file.

PR #11673, opened 2022-05-09
Set an environment variable for the Sidekiq logfile path.

sengi:
This will be needed as of alphagov/publisher#1603. See that PR for context.

https://trello.com/c/Jef1t4xU


kevindew (approved):
This looks good to me 👍 

I've had something of an internal debate whether to push for it to be sidekiq.json.log for wider consistency, but I think your choice might well be the right one since the application decides whether it is JSON or not.

PR #11677, opened 2022-05-12
Update LUX check to use latest version number

injms:
Update the LUX check to use the latest version number, which is updated in https://github.com/alphagov/govuk_publishing_components/pull/2773.

Speedcurve's minified JavaScript also had the variable name changed, so this has been updated as well.


Nooshu (approved):
Matches the other PR, so approved.

PR #11679, opened 2022-05-16
Add new Jenkins deploy role in Staging

deborahchua:
We are adding a new level of access, which allows engineers to deploy
code but not administer related systems. This uses the new Github
team https://github.com/orgs/alphagov/teams/gov-uk-production-deploy

`hudson.model.Item.Build` - This permission grants the ability to start a new build.
`hudson.model.Item.Cancel` - This permission grants the ability to cancel a scheduled, or abort a running, build.
`hudson.model.Item.Create` - Create a new job.
`hudson.model.Item.Discover` - This permission grants discover access to jobs. Lower than read permissions, it allows you to redirect anonymous users to the login page when they try to access a job url. Without it they would get a 404 error and wouldn't be able to discover project names. This permission is implied by Job/Read.
`hudson.scm.SCM.Tag` - This permission allows users to create a new tag in the source code repository for a given build.

Trello card: https://trello.com/c/docwZ4Gm/2892-implement-production-deploy-access-5

### DIFF
--- a/hieradata_aws/class/staging/jenkins.yaml
+++ b/hieradata_aws/class/staging/jenkins.yaml
@@ -26,6 +26,16 @@ jenkins_admin_permission_list: &jenkins_admin_permission_list
   - 'hudson.model.View.Read'
   - 'hudson.scm.SCM.Tag'
 
+jenkins_deploy_permission_list: &jenkins_deploy_permission_list
+  - 'hudson.model.Hudson.Read'
+  - 'hudson.model.Item.Build'
+  - 'hudson.model.Item.Cancel'
+  - 'hudson.model.Item.Create'
+  - 'hudson.model.Item.Discover'
+  - 'hudson.model.Item.Read'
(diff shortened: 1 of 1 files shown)


hieradata_aws/class/staging/jenkins.yaml
```diff
@@ -44,6 +54,9 @@ govuk_jenkins::config::user_permissions:
     permissions:
       - 'hudson.model.Item.Build'
       - 'hudson.model.Item.Read'
+  -
+    user: 'alphagov*GOV.UK Production Deploy'
+    permissions: *jenkins_admin_permission_list
```
catalinailie, line 59: I think here we need `jenkins_deploy_permission_list`? But I might be wrong

deborahchua: Thanks yes you're right!

catalinailie (approved):

PR #11684, opened 2022-05-19
Add `locations_api` machine class to rake task options

ChrisBAshton:
We now need to be able to run rake tasks on the `locations_api`
machines, which host the `locations-api` app.

This was the original intention of #11683, which edited the wrong file - although that file needed updating too, so no harm done.

https://trello.com/c/iy6P2rvX/2875-warm-up-locations-api-cache-5

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/run_rake_task.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/run_rake_task.yaml.erb
@@ -37,8 +37,9 @@
             - draft_frontend
             - email_alert_api
             - frontend
-            - router_backend
+            - locations_api
             - publishing_api
+            - router_backend
             - search
             - whitehall_backend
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/templates/jobs/run_rake_task.yaml.erb
```diff
@@ -37,8 +37,8 @@
             - draft_frontend
             - email_alert_api
             - frontend
-            - router_backend
```
catalinailie, line 40: `locations_api` should be added here

ChrisBAshton: 🤦  sorry. Thanks. Clearly need another coffee!

catalinailie (approved):

PR #11687, opened 2022-05-20
Add Jenkins "view only" permissions in Staging

deborahchua:
As covered in our access levels spreadsheet (https://docs.google.com/spreadsheets/d/1oqy7tKpB8mHBhHQ9jAZu0NR0GKKZXOqtQGBKHYVnpmk/edit?pli=1#gid=0),
those with Integration access should be able to view builds in Jenkins
Staging and Production. This uses the GOV.UK Github team which people
are added to when they first join the programme.

### DIFF
--- a/hieradata_aws/class/staging/jenkins.yaml
+++ b/hieradata_aws/class/staging/jenkins.yaml
@@ -36,6 +36,11 @@ jenkins_deploy_permission_list: &jenkins_deploy_permission_list
   - 'hudson.model.View.Read'
   - 'hudson.scm.SCM.Tag'
 
+jenkins_integration_permission_list: &jenkins_integration_permission_list
+  - 'hudson.model.Hudson.Read'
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
Nice one! This will also need applying in Production too, right? Any reason why that can't be done in this PR?

Code changes look good but will need a manual test on Staging first 👀

deborahchua:
> Nice one! This will also need applying in Production too, right? Any reason why that can't be done in this PR?
> 
> Code changes look good but will need a manual test on Staging first 👀

I was going to merge and test this in Staging first - I considered deploying my branch to staging to test but don't want to block anyone from deploying as it takes half an hour between environments. I also need to find someone who only has integration access at the moment to test this with.

ChrisBAshton:
Makes sense - thanks!

PR #11689, opened 2022-05-23
Add Jenkins "view only" permissions in Production

deborahchua:
As covered in our access levels spreadsheet (https://docs.google.com/spreadsheets/d/1oqy7tKpB8mHBhHQ9jAZu0NR0GKKZXOqtQGBKHYVnpmk/edit?pli=1#gid=0),
those with Integration access should be able to view builds in Jenkins
Staging and Production. This uses the GOV.UK Github team which people
are added to when they first join the programme.

This has already been tested and added for Staging in https://github.com/alphagov/govuk-puppet/pull/11687/commits/8c9bbcae6c5d15d00ee4c07d446d10b1a022fdaa

Trello card: https://trello.com/c/docwZ4Gm/2892-implement-production-deploy-access-5

### DIFF
--- a/hieradata_aws/class/production/jenkins.yaml
+++ b/hieradata_aws/class/production/jenkins.yaml
@@ -29,6 +29,11 @@ jenkins_deploy_permission_list: &jenkins_deploy_permission_list
   - 'hudson.model.View.Read'
   - 'hudson.scm.SCM.Tag'
 
+jenkins_integration_permission_list: &jenkins_integration_permission_list
+  - 'hudson.model.Hudson.Read'
+  - 'hudson.model.Item.Read'
+  - 'hudson.model.View.Read'
+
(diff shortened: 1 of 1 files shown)


catalinailie (approved):

hieradata_aws/class/production/jenkins.yaml
```diff
@@ -29,6 +29,11 @@ jenkins_deploy_permission_list: &jenkins_deploy_permission_list
   - 'hudson.model.View.Read'
   - 'hudson.scm.SCM.Tag'
 
+jenkins_integration_permission_list: &jenkins_integration_permission_list
```
catalinailie, line 32: Is it ok `integration` here or should be `production`?

catalinailie: 👍 yes, makes sense. Looks good to me ✅

PR #11690, opened 2022-05-23
Remove redundant fallback for app healthchecks

benthorner:

This was added in [^1] without explanation.

I've checked and all apps have has_readiness_health_check set to
true wherever has_liveness_health_checkset is also set to true,
so this conditional branch is never used in practice.

I think it's clearer to have the alert work with just one type of
healthcheck and require someone adding an app to add both. This
means we can simplify the documentation for the alert.

[^1]: https://github.com/alphagov/govuk-puppet/commit/50c30a600617df809238715cb6c398b60cdea2aa

### DIFF
--- a/modules/govuk/manifests/app/config.pp
+++ b/modules/govuk/manifests/app/config.pp
@@ -434,17 +434,8 @@ define govuk::app::config (
         contact_groups      => $additional_check_contact_groups,
       }
     }
-  } elsif $has_liveness_health_check {
-    @@icinga::check { "check_app_${title}_up_on_${::hostname}":
-      ensure              => $ensure,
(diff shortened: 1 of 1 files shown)


KludgeKML:
In the comment you mention has_readiness_health_check twice - just to confirm you actually mean that everything with has_readiness_health_check also has has_liveness_health_check set, right?

benthorner:
Oops, thanks for spotting.

I've updated the commit and PR description.

benthorner:
Tested on Integration: leaving in place for 2 hours - enough time for Puppet state to update (1 hour) + reapply on monitoring (1 hour) - shows no change in the 135 total healthcheck alerts.

PR #11696, opened 2022-05-25
Upgrade kernel to 4.4.0-224

rtrinque:
This should fix this kernel bug https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1715812


fredericfran-gds (approved):
looks good

PR #11708, opened 2022-06-07
Add golang 1.18.3 for router.

KludgeKML:
To support https://github.com/alphagov/router/pull/274

### DIFF
--- a/modules/golang/manifests/init.pp
+++ b/modules/golang/manifests/init.pp
@@ -14,9 +14,8 @@ class golang {
   goenv::version { [
     '1.7.1',   # Used by alphagov/govuk_crawler_worker
     '1.12.1',  # Used by alphagov/govuk-csp-forwarder
-    '1.17.6',  # Used by alphagov/router (remove once router is running latest)
-    '1.17.7',  # Used by alphagov/router (remove once router is running latest)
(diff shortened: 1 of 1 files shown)


modules/golang/manifests/init.pp
```diff
@@ -14,9 +14,9 @@
   goenv::version { [
     '1.7.1',   # Used by alphagov/govuk_crawler_worker
     '1.12.1',  # Used by alphagov/govuk-csp-forwarder
-    '1.17.6',  # Used by alphagov/router (remove once router is running latest)
-    '1.17.7',  # Used by alphagov/router (remove once router is running latest)
-    '1.17.8',  # Used by alphagov/router
+    '1.17.8',  # Used by alphagov/router (remove once router is running latest)
+    '1.18.3',  # Used by alphagov/router
+    '1.',
```
AgaDufrat, line 19: Do we need this `'1.',` ? 🤔 

```suggestion

```

KludgeKML: You're right, I don't think we do. That's odd, I was under the impression that that was there in the previous version, but clearly it shows up as me having added it. I'll remove and rebase.

AgaDufrat (approved):

PR #11711, opened 2022-06-09
Revert "Retire manuals-frontend (step 2)"

1pretz1:
Reverts alphagov/govuk-puppet#11588


1pretz1:
This has been deployed and run in integration which has fixed the issue.

PR #11712, opened 2022-06-10
Increase smartanswers unicorn works 4 => 8

edwardkerry:
There is a chance that a new smart-answer may attract an increase in traffic. 
This PR is prepared in case we need to quickly act and increase capacity by doubling unicorn workers.

[trello](https://trello.com/c/kFcg5KBm/91-prep-a-pr-to-double-the-number-of-unicorns-just-in-case-unprecedented-traffic)


leenagupte (approved):

edwardkerry:
this pre-emptive PR was not needed, so closing it

PR #11713, opened 2022-06-15
Remove deleted features from smokey loop checks

benthorner:

Relates to [^1].

[^1]: https://github.com/alphagov/smokey/pull/972


ollietreend (approved):
Looks good to me 👍🏻

PR #11715, opened 2022-06-15
Increase proxy buffer size in nginx for draft-cache machines

rtrinque:
We've experienced the following error with nginx on draft-cache machines:

“upstream sent too big header while reading response header from upstream”

The origin of the problem is unknown but simply increasing the buffer size in nginx makes it go away.


hieradata_aws/class/draft_cache.yaml
```diff
@@ -35,6 +35,9 @@ router::nginx::app_specific_static_asset_routes:
 
 router::nginx::check_requests_warning: '@0'
 router::nginx::check_requests_critical: '@0'
+router::nginx:::proxy_busy_buffers_size: '512k'
+router::nginx:::proxy_buffers: '4 512k'
+router::nginx:::proxy_buffer_size: '256k'
```
ollietreend, line 40: Are there any (obvious) drawbacks to setting these buffers so much higher? e.g. are there any performance hits due to more memory being allocated on all requests, even if they don't need it, etc.?

If so, I wonder if we'd get away with setting the buffer to be, say, double the 8k default rather than 64x bigger.

rtrinque: Good point, testing shows that doubling seems to solve the problem. Will amend the buffers size values accordingly

ollietreend (approved):
Looks good to me 🚀

PR #11718, opened 2022-06-16
Fix typo in healtyhosts_ignore

brucebolt:
There was a typo in https://github.com/alphagov/govuk-puppet/pull/11716.


1pretz1 (approved):
Missed this as well!

PR #11725, opened 2022-06-21
Filecheck requires the govuk_awsclloudwatch modules

MahmudH:
Adding a `require` will ensure the modules are installed before the file check. Fixing the ordering issue.


rtrinque (approved):
Looks good

PR #11727, opened 2022-06-22
Re-sync Smokey Loop alerts with new file structure

benthorner:

This updates the alerts to match the new feature files [^1][^2].
Please see the commit messages for more details.

Note: by doing a full re-sync this PR also backfills some missing 
alerts that I was proposing to fix in https://github.com/alphagov/govuk-puppet/pull/11726.

## Testing

I've tested this on Integration in tandem with the associated branch 
of Smokey. All new alerts worked as expected (see below)*.

The only two alerts not covered are for DGU and mirrors, which are 
unchanged from before so I'm not expecting any issues.

*I accidentally missed the "origin" feature in the version of the code 
I used for the test, but I don't think that's worth a re-test.

Screenshot 2022-06-22 at 14 44 04

[^1]: https://github.com/alphagov/smokey/pull/976
[^2]: https://github.com/alphagov/smokey/tree/organising-tests/features

### DIFF
--- a/hieradata_aws/integration.yaml
+++ b/hieradata_aws/integration.yaml
@@ -283,6 +283,8 @@ monitoring::uptime_collector::environment: 'integration'
 monitoring::contacts::slack_username: 'Integration'
 
 monitoring::checks::smokey::features:
+  check_service_manual_frontend:
+    feature: apps/service_manual_frontend
   check_ab_testing:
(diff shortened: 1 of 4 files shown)


KludgeKML (approved):
LGTM

KludgeKML (approved):
Looks okay.

PR #11729, opened 2022-06-23
Push mongo databases to integration backup directory

kevindew:
It seems that in the integration environment we've not configured these
databases to be pushed to the backups folder and thus they can't be used
by developers to replicate data locally.

I've configured these to have a window between the production pull and
the integration dump similar to that of production.


fredericfran-gds (approved):
Nice work, I guess these were missed after the migration to DocumentDB.

PR #11733, opened 2022-06-27
Set APPLICATION_FORM_AWS_REGION for Licensify

richardTowers:
This is mandatory following the upgrade to Play 2.7. Currently this is
the same in all environments, so it never needs to be passed as a parameter.


sihugh (approved):

sihugh:
Edit suggestion:
>Currently this is the same in all ~regions~ `environments`

?

PR #11740, opened 2022-07-05
Ensure CSV previews are rendered by Whitehall on draft origin

davidgisbey:
## Description

Currently, the custom routing set out for CSV preview rendering that is in place for assets origin is not in place for draft assets. This means that CSV previews hit the assets manager application rather than being passed to the CSV preview controller in Whitehall.

## Trello card

https://trello.com/b/0n3p84mY/govuk-publishing-service-publisher-experience-team-doing


davidgisbey:
I have 👍 it's working as expected on Int.

PR #11743, opened 2022-07-05
Change log group to auth-log

MahmudH:
I believe logs are not being sent as the log group must be auth-log as seen on:
https://github.com/alphagov/govuk-aws/blob/main/terraform/projects/infra-cyber-cloudwatch-to-splunk/main.tf#L31

trello: https://trello.com/c/t8t3sT51

### DIFF
--- a/modules/govuk/files/node/s_base/amazon-cloudwatch-agent.json
+++ b/modules/govuk/files/node/s_base/amazon-cloudwatch-agent.json
@@ -12,7 +12,7 @@
             "log_stream_name": "{instance_id}"
           },
           {
-            "file_path": "/var/log/secure",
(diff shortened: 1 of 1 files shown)


modules/govuk/files/node/s_base/amazon-cloudwatch-agent.json
```diff
@@ -13,7 +13,7 @@
           },
           {
             "file_path": "/var/log/secure",
-            "log_group_name": "secure",
+            "log_group_name": "auth-log",
```
ChrisBAshton, line 16: While we're here, can we fix the file_path on line 15 too? (`secure` => `secure.log`).

PR #11750, opened 2022-07-12
Use new Docker Enterprise user for builds

deborahchua:
We have a new user `govukdockerci` which has been added to our
Enterprise hub and is an owner of the `govuk` organisation.
This should allow us to push without any limits. We will continue
to push to the existing govuk organisation that is owned by the
`govukci` user. These credentials have been stored in Jenkins CI
and govuk-secrets.

Related commit: https://github.com/alphagov/govuk-jenkinslib/commit/0617724150162937cc86517a6db1c04b79714543

Trello card: https://trello.com/c/I2IRWNSU/2897-look-into-docker-pull-rate-limits-3

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_app.yaml.erb
@@ -85,6 +85,11 @@
                 credential-id: govukci-docker-hub
                 username: DOCKER_HUB_USERNAME
                 password: DOCKER_HUB_PASSWORD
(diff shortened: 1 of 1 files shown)


MuriloDalRi (approved):

kevindew (approved):
Looks good. 

I did wonder if we hit problems with both this and above using `DOCKER_HUB_(USERNAME|PASSWORD)` but I don't think it's a problem unless they were used together and that seems very unlikely.

PR #11760, opened 2022-07-25
Add general cleanup class to CI to prevent FS resources exhaustion

rtrinque:
This add a daily run script that get rid of workspaces in jenkins
that are older than jenkins_worspace_max_age.
Over time the accumulation of files in the workspaces can lead to
inodes or space exhaustion on the filesystem. This script should
keep this under control.

Trello: https://trello.com/c/JwTIJsk1/2920-stop-low-available-disk-inodes-alert-3

### DIFF
--- a/hieradata_aws/class/integration/ci_agent.yaml
+++ b/hieradata_aws/class/integration/ci_agent.yaml
@@ -7,6 +7,7 @@ mongodb::server::replicaset_members:
 
 clamav::use_service: false
 
+govuk_ci::agent::cleanup::jenkins_workspace_max_age: '30'
 govuk_ci::agent::postgresql::mapit_role_password: 'mapit'
 govuk_ci::agent::postgresql::email_alert_api_role_password: 'email-alert-api'
 govuk_ci::agent::gdal_version: "2.4.4"
--- a/modules/govuk_ci/manifests/cleanup.pp
+++ b/modules/govuk_ci/manifests/cleanup.pp
@@ -1,0 +1,18 @@
+# == Class: govuk_ci::cleanup
+#
+#  General cleanup of filesystem.
+#
+# === Parameters:
+#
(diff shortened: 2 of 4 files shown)


modules/govuk_ci/manifests/cleanup.pp
```diff
@@ -0,0 +1,18 @@
...
+# [*jenkins_workspace_max_age*]
+#   Maximum age -in days- of Jenkins workspaces before they are deleted by the cleanup job.
+#
+class govuk_ci::cleanup (
+  $jenkins_workspace_max_age = 90,
+){
+  cron::crondotdee { 'clean_up_workspace':
+    command => 'find /var/lib/jenkins/workspace/ -maxdepth 1 -mindepth 1 -type d -mtime +$jenkins_workspace_max_age -exec rm -rf {} \;',
+    hour    => 7,
+    minute  => 25,
```
ChrisBAshton, line 16: Any particular reason for this time of day?

ChrisBAshton (approved):
LGTM 👍 ran the command on all environments (without the `rm -rf`) and played around with the ages, looks like it should clear up a lot of files.

modules/govuk_ci/manifests/cleanup.pp
rtrinque: There already a cleaning job running a little later, so just offset it to a little earlier

hieradata_aws/class/integration/ci_agent.yaml
```diff
@@ -7,6 +7,7 @@ mongodb::server::replicaset_members:
 
 clamav::use_service: false
 
+govuk_ci::agent::cleanup::jenkins_workspace_max_age: '30'
```
ChrisBAshton, line 10: 👍 thanks

PR #11766, opened 2022-07-26
Point authenticating-proxy to documentdb in integration

KludgeKML:
As part of moving our apps away from Mongo 2.6, it would be useful to test moving the authenticating proxy app in integration away from the EC2 Mongo 2.6 cluster and onto the DocumentDB authenticating-proxy-documentdb-integration cluster that will be created by this PR:

https://github.com/alphagov/govuk-aws/pull/1556

This change adds a new draft-cache yaml file to the heiradata in the integration class that should (for integration) override the existing mongodb cluster information.

When deployed, authenticating proxy will lose the existing information in integration. There's no env sync for AP from staging down to integration, so it won't automatically be populated. However, it's possible this might not make any difference (since the data in AP is transient anyway), and it should be possible to export from the existing mongo and import manually.

https://trello.com/c/lSpntlfk/81-mongo-26-has-reached-end-of-life

### DIFF
--- a/hieradata_aws/class/integration/draft_cache.yaml
+++ b/hieradata_aws/class/integration/draft_cache.yaml
@@ -1,0 +1,7 @@
+---
+
+govuk::apps::authenticating_proxy::mongodb_name: 'authenticating_proxy_production'
+govuk::apps::authenticating_proxy::mongodb_nodes:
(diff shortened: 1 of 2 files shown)


KludgeKML:
Closing as PG solution preferred.

PR #11775, opened 2022-08-03
Authenticate Fastly cache purge requests

robinjam:
[Trello card](https://trello.com/c/O5Wy2X9w/2946-authenticate-fastly-purge-requests-with-an-api-token-3)

We are updating our Fastly configuration to authenticate cache purge requests using API tokens instead of an IP address allowlist[^1].

This PR updates the Jenkins job for clearing the CDN cache to use API tokens already added to the `govuk-secrets` repo[^2] to authenticate the purge requests. Because we are now sending secret API tokens with the purge requests, a check has been added to the script to ensure that requests are only sent over HTTPS.

As a result of this change, we no longer need to `ssh` to an allowlisted node before issuing a `PURGE` request to Fastly, so this extra step has been removed.

This branch has been deployed to integration for testing purposes: https://deploy.integration.publishing.service.gov.uk/job/clear-cdn-cache/

[^1]: https://docs.fastly.com/en/guides/authenticating-api-purge-requests
[^2]: https://github.com/alphagov/govuk-secrets/pull/1325

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
@@ -30,6 +30,12 @@
               exit 1
             fi
 
+            if [[ "${cdn_url:0:8}" != "https://" ]]; then
+              # Ensure the Fastly API token is not sent over a plaintext connection
+              echo "expected $cdn_url to be an HTTPS URL"
+              exit 1
+            fi
+
             if [ -n "$local_varnish_url" ]; then
               echo $local_varnish_url
               for node in `govuk_node_list -c cache`; do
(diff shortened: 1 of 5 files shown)


modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
```diff
@@ -30,6 +30,12 @@
               exit 1
             fi
 
+            if [[ "${cdn_url:0:8}" != "https://" ]]; then
+              # Ensure the Fastly API token is not sent over a plaintext connection
+              echo "expected $cdn_url to be an HTTPS URL"
+              exit 1
+            fi
```
ChrisBAshton, line 37: Nice 👍

ChrisBAshton (approved):
LGTM ⭐

PR #11780, opened 2022-08-16
Add postgis ci container

KludgeKML:
Adds a container running postgres 14 with postgis 3.2 (by default) to the CI agents. This is necessary to run the tests for the Mongo->Postgresql conversion for imminence. If we decide to go ahead with that, it'll be necessary to merge this into the main branch.

Related PR:
https://github.com/alphagov/imminence/pull/766


KludgeKML:
This might not be necessary? I guess things are working in the CI now (both Github and Jenkins), so maybe we don't need to care. Going to close.

PR #11781, opened 2022-08-18
Add unreleased_features_flag for collections

chao-xian:
On integration only, we want to be able to show new features such as the Cost of living support page, and hide on staging and production.

[Trello](https://trello.com/c/0rJHy0zc/1236-page-is-on-production-and-hidden-viewable-only-on-integration-m)


AgaDufrat:
I think we also need to add a parameter in `[modules/govuk/manifests/apps/collections.pp](https://github.com/alphagov/govuk-puppet/blob/main/modules/govuk/manifests/apps/collections.pp#L35)
```
  $unreleased_features_enabled = false,
```

chao-xian:
Ah thanks @AgaDufrat ! Didn't spot that from the searching and PRs I looked through.

AgaDufrat (approved):
It think that's all we need 👍 

I forgot to mention – it would also be nice if we could add a comment [here](https://github.com/alphagov/govuk-puppet/blob/main/modules/govuk/manifests/apps/collections.pp#L33) with something like:
```

as it seems to be the pattern.
# [*unreleased_features_enabled*]
#   Whether unreleased features should be enabled
```

PR #11784, opened 2022-08-19
Disable unattended reboots

leenagupte:
We want to reboot the docker_management management machine. The
docker_management machine manages locks for unattended reboots of other
machines. If this machine is down, then multiple machines in high
availability groups may  choose to reboot themselves at the same time.


leenagupte:
The docker_management machine has been rebooted, so closing this PR.

PR #11794, opened 2022-08-30
Remove puppet_auth_token and hardcode jenkins_downstream_api_user

MuriloDalRi:
puppet_auth_token was added in https://github.com/alphagov/govuk-puppet/pull/11792 but isn't actually needed.

https://trello.com/c/JcY1L7so/2942-automatically-run-smokeydeploy-when-a-smokey-change-is-merged-3

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/deploy_smokey_downstream.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/deploy_smokey_downstream.yaml.erb
@@ -12,7 +12,7 @@
           # Deploy to downstream environment
           CRUMB=$(curl https://<%= @jenkins_downstream_api_user %>:<%= @jenkins_downstream_api_password %>@<%= @deploy_url %>/crumbIssuer/api/json | jq --raw-output '. | .crumb')
 
-          curl -f -H "Jenkins-Crumb:$CRUMB" -XPOST https://<%= @jenkins_downstream_api_user %>:<%= @jenkins_downstream_api_password %>@<%= @deploy_url %>/job/Smokey_Deploy/build -d token=<%= @puppet_auth_token %>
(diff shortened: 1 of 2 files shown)


modules/govuk_jenkins/templates/jobs/deploy_smokey_downstream.yaml.erb
```diff
@@ -12,7 +12,7 @@
           # Deploy to downstream environment
           CRUMB=$(curl https://<%= @jenkins_downstream_api_user %>:<%= @jenkins_downstream_api_password %>@<%= @deploy_url %>/crumbIssuer/api/json | jq --raw-output '. | .crumb')
 
-          curl -f -H "Jenkins-Crumb:$CRUMB" -XPOST https://<%= @jenkins_downstream_api_user %>:<%= @jenkins_downstream_api_password %>@<%= @deploy_url %>/job/Smokey_Deploy/build -d token=<%= @puppet_auth_token %>
```
ChrisBAshton, line 15: This is a straight revert of #11792, apart from [we're keeping the removal of the JSON variable](https://github.com/alphagov/govuk-puppet/pull/11792/files#diff-825e79bea45602dd599d2ba3ab655b4aec50aae15f555f46735ac6754ca46299L12)... right?

ChrisBAshton (approved):
Would be good to document why this was added & why it isn't needed.
PR title should be updated too 👍

modules/govuk_jenkins/templates/jobs/deploy_smokey_downstream.yaml.erb
MuriloDalRi: Yes, we don't actually need either of those. I've added a new commit now

PR #11801, opened 2022-08-31
Remove ASSET_HOST for Static

theseanything:
By removing ASSET_HOST assets URLs and compiled asset URLs (e.g. in
css) will become relative. This works as assets are served from the same
domain in production (however with the path prefix /assets/static/*).

Having ASSET_HOST set and using absolute URLs prevents us creating re-usable
container images, as the assets URLs are environment dependent.

We set ASSET_HOST for Static so that local development and heroku
deployments can send requests to Static hosted in Production, instead of
running a local version. However, this problem can be solved by proxing
requests via the frontend app.


sengi:
Turns out this was a load-bearing piece of tech debt 😭🤣🙈

PR #11805, opened 2022-09-02
Remove unreleased_features_flag for collections

chao-xian:
## What

Remove the env var set in Puppet which triggered whether the Cost of living page was shown on each environment.

## Why

This is un-used code and we should tidy up.

## How

Revert PRs:

* https://github.com/alphagov/govuk-puppet/pull/11781
* https://github.com/alphagov/govuk-puppet/pull/11786

[Trello](https://trello.com/c/b4INV89F/1298-remove-puppet-env-var-for-feature-flag)


chao-xian:
@leenagupte has pointed out we can leave this in.

PR #11806, opened 2022-09-02
Licensify - Remove deprecated configuration property

brandonmoss-99:
As part of the Licensify Play 2.5 to 2.6 migration, we added the new play.http.secret.key parameter, and kept the existing one so that we could seamlessly change version

Now that we're running on newer Play versions, the older play.crypto.secret is no longer needed, and is bringing up deprecation warnings when deploying a new release of Licensify

### DIFF
--- a/modules/licensify/templates/gds-licensify-admin-config.conf.erb
+++ b/modules/licensify/templates/gds-licensify-admin-config.conf.erb
@@ -8,7 +8,6 @@ include "application"
 # ~~~~~
 # The secret key is used to secure cryptographics functions.
 # If you deploy your application to several instances be sure to use the same key!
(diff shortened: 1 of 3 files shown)


sihugh:
Confirm that this key was swapped out in https://github.com/alphagov/licensify/commit/a3b2a8af649eb8c33430f12964cb833c310c4963

sihugh:
Have pushed following [the docs](https://docs.publishing.service.gov.uk/manual/merge-pr.html#a-change-from-an-external-contributor) to kick CI off

sihugh:
Thanks for the tidy up @brandonmoss-99!

racoon sweeping

sihugh (approved):

PR #11811, opened 2022-09-12
Stop spamming #govuk-developers

robinjam:
A lot of `Deploy_App_Downstream` jobs are failing at the moment due to issues with Smokey - each failure causes a new notification to be sent to `#govuk-developers`, which is noisy and unhelpful. Let's stop that, at least until we've fixed the issue with Smokey.

### DIFF
--- a/modules/govuk_jenkins/manifests/jobs/deploy_app_downstream.pp
+++ b/modules/govuk_jenkins/manifests/jobs/deploy_app_downstream.pp
@@ -30,7 +30,7 @@ class govuk_jenkins::jobs::deploy_app_downstream (
   $github_api_token = undef,
   $smokey_pre_check = true,
   $release_app_bearer_token = undef,
-  $slack_channel = 'govuk-deploy',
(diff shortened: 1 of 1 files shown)


catalinailie (approved):

modules/govuk_jenkins/manifests/jobs/deploy_app_downstream.pp
```diff
@@ -30,7 +30,7 @@
   $github_api_token = undef,
   $smokey_pre_check = true,
   $release_app_bearer_token = undef,
-  $slack_channel = 'govuk-deploy,govuk-developers',
+  $slack_channel = 'govuk-deploy',
```
ChrisBAshton, line 33: I'd suggest this instead tbh:

```suggestion
  $slack_channel = 'govuk-deploy,govuk-2ndline-tech',
```

catalinailie (approved):

ChrisBAshton (approved):

PR #11812, opened 2022-09-12
Add staging eks as a target for goreplay

desenhar:
https://trello.com/c/bIw2FkZ9/992-look-into-goreplay-and-add-eks-as-a-target

Following similar logic here,
https://github.com/alphagov/govuk-puppet/commit/52929e9c1d495ab2c20c82a8621a21536b08e142
setting eks staging router as a target from aws staging router.


sengi (approved):
Cool

PR #11816, opened 2022-09-20
Remove FACT_CHECK_ADDRESS_FORMAT env var

kevindew:
This env var has no longer had an effect since the corresponding code was removed in August 2020 [1]. We can now remove this env var.

The motivation for making this change was to make it clearer that the alphagov.co.uk is not used as part of fact-checking as this was a recent point of confusion.

[1]: https://github.com/alphagov/publisher/commit/c72119346523e3a89067ad3fac1d74abf5dd0df1


ollietreend (approved):
Great stuff ✨🧹

PR #11818, opened 2022-09-21
Add Referrer-Policy header

MahmudH:
The Referrer-Policy HTTP header controls how much referrer information (sent with the Referer header) should be included with requests.

This commit sets the header to "strict-origin-when-cross-origin" to aid in security.

The header will not allow any information to be sent when a scheme downgrade happens (the user is navigating from HTTPS to HTTP).

Trello: https://trello.com/c/2Q31FJne/2927-investigate-adding-referrer-policy-header


ChrisBAshton (approved):
LGTM 👍 worth deploying this branch of Puppet to Integration and checking that it sets the header correctly (if you haven't already).

PR #11819, opened 2022-09-21
Show env in Slack message when deploy_app_downstream fails

ChrisBAshton:
Currently not clear which environment the job failed in. We care more about failed deploys to Production!


brucebolt (approved):
Looks good to me, this will be a great improvement 🎖️

PR #11828, opened 2022-09-29
Ensure the Fastly API token is only sent to GOV.UK URLs

MahmudH:
This PR is to allow Fastly API tokens to be sent to GOV.UK URLs only in the Clear CDN Cache job in Jenkins. 

Trello: https://trello.com/c/Fs4Kz4Ky/2978-update-jenkins-clear-cdn-cache-job-to-only-accept-govuk-urls-2

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
@@ -35,6 +35,13 @@
               echo "expected $cdn_url to be an HTTPS URL"
               exit 1
             fi
+            
+            cdn_domain=`echo $cdn_url | awk -F[/:] '{print $4}'`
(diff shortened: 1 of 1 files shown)


modules/govuk_jenkins/templates/jobs/clear_cdn_cache.yaml.erb
```diff
@@ -35,6 +35,13 @@
               echo "expected $cdn_url to be an HTTPS URL"
               exit 1
             fi
+            
+            cdn_domain=`echo $cdn_url | awk -F[/:] '{print $4}'`
+            if [[ "${cdn_domain: -6}" != "gov.uk" ]]; then
```
MuriloDalRi, line 40: Is the -6 needed here? It will only keep the last six chars of that string

MuriloDalRi (approved):

PR #11833, opened 2022-10-03
Show env and tag in Slack message when deploying Puppet

ChrisBAshton:
Takes inspiration from #11819. See at a glance which environment and which release is being deployed.


ChrisBAshton:
Didn't work. Not high value enough to investigate.

PR #11836, opened 2022-10-05
Change verification strategy for ci-master nodes

rtrinque:
Change the verification strategy for ci-master nodes from "known hosts file verification strategy" to
"non verifying verification strategy".
This will prevent errors when jenkins tries to connect to one of its nodes which ssh key has changed because it was reprovisioned. Note that in general it's a bad idea to disable
'StrictHostKeyChecking' but in this case we were already trying to bypass this check by deleting the known_hosts file. By disabling the check we don't have to care about managing the known_hosts file.

Trello: https://trello.com/c/RUWdTgH6/2975-change-how-ssh-host-keys-are-managed-on-jenkins-5


ChrisBAshton (approved):
Makes sense to me 👍 sounds like it doesn't increase the risk either. Nice one!

PR #11837, opened 2022-10-05
Pass through existing X-Forwarded-Host header

theseanything:
For assets and www pass through the exisiting X-Forwarded-Host header if present on the client request. This ensures X-Forwarded-Host represents the original host (from Fastly). Previously we re-set the header to the Nginx host, which break redirects in Rails as it uses the header to determine the original host for the client request.


sengi (approved):
LGTM but it's been a while since I've worked much with nginx config.

I'm assuming you've tested thoroughly 😅

PR #11839, opened 2022-10-06
Remove ASSET_HOST for Static

theseanything:
This is the second attempt at removing ASSET_HOSTS for Static. See commit ba498b06303d2be55bdda94093fd6e06d9486d48 for the original context. This change was orginially rolled back as it caused leaking of internal domains in the redirection URL produced for the favicon.ico on the assets domain. The internal domain was then flagged by Google's Unsafe Browsing feature and prevented users from loading any page using the favico.ico on the assets domain.

This leak was caused by Rails falling back to using the X-Forwarded-Host header as the redirect host. This was being incorrectly set as our internal origin domain by Nginx.

We've now implemented [changes to Fastly](https://github.com/alphagov/govuk-cdn-config/pull/414) and [Ngnix](https://github.com/alphagov/govuk-puppet/pull/11837) to properly forward the correct X-Forwarded-Host header and should prevent the leaked domain from occuring again. 

This reverts commit bdd867fd71f753fc56d356cba9303e713208cd53, reversing changes made to dff49b92ab41910133ef8a899d5e728d81daa6d5.

### DIFF
--- a/hieradata_aws/class/draft_frontend.yaml
+++ b/hieradata_aws/class/draft_frontend.yaml
@@ -7,6 +7,5 @@ govuk::apps::government_frontend::vhost: 'draft-government-frontend'
 govuk::apps::service_manual_frontend::vhost: 'draft-service-manual-frontend'
 govuk::apps::smartanswers::vhost: 'draft-smartanswers'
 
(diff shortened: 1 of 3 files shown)


nsabri1 (approved):
OK

PR #11842, opened 2022-10-09
Fixed insecure coding pattern

akondasif:
Fixed one suspicious comment and one default admin


ChrisBAshton:
Hello 👋  The existing code is intentional, so we will close this PR, but thank you for your contribution.

PR #11849, opened 2022-10-13
Use the governmentdigitalservice docker hub account instead of govuk

MuriloDalRi:
We're consolidating all of GDS' Docker account usage so that it goes through governmentdigitalservice. This org has multiple owners and a process for requesting additional seats, etc, unlike the govuk org which has a 3 seat limit.

https://trello.com/c/6JRYK8hU/2982-push-to-pull-from-governmentdigitalservice-docker-org-3

### DIFF
--- a/modules/govuk_containers/manifests/gemstash.pp
+++ b/modules/govuk_containers/manifests/gemstash.pp
@@ -11,7 +11,7 @@
 #   The docker image version use.
 #
 class govuk_containers::gemstash(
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
This repo already exists in the new Docker org, so should be safe to merge:

https://hub.docker.com/repository/docker/governmentdigitalservice/gemstash-alpine

PR #11857, opened 2022-10-19
Revert "Add gtag env var details"

andysellick:
Reverts alphagov/govuk-puppet#11838 as we no longer need to be testing GTAG on integration.

Also reverts https://github.com/alphagov/govuk-puppet/pull/11826 - these were related changes that should have been in one PR originally.


AshGDS (approved):
All good :+1:

PR #11863, opened 2022-10-20
Remove mapit stage2

KludgeKML:
Remove mapit from govuk-puppet, including mapit node.

https://trello.com/c/LMRiT7yz/1581-retire-mapit-and-mapit-scripts


1pretz1 (approved):
Nice! 🍏

PR #11875, opened 2022-10-24
Update CKAN auth config for creating users

kentsanggds:


maxgds (approved):
Approved, but see question on Slack - wondering if we need to do comms so user admins know they can't manage their org.

PR #11876, opened 2022-10-27
Remove 'network_config_deploy' job and config

ChrisBAshton:
I can't find this job in any of the Jenkins environments. Other than the [hieradata config](https://github.com/alphagov/govuk-puppet/blob/4101a4d6e473629c266d55a51b245918bee56e6a/hieradata_aws/staging.yaml#L287-L290) hanging around in `staging.yaml` (referencing Carrenza exclusively), it's not on the
[list of jobs that get created and imported into Jenkins](https://github.com/alphagov/govuk-puppet/blob/4101a4d6e473629c266d55a51b245918bee56e6a/hieradata_aws/class/staging/jenkins.yaml#L69).

Looks like it [used to be pulled into jobs in jenkins.yaml](https://github.com/alphagov/govuk-puppet/commit/a9957b8b9671aa7ca7b18e42453604c70140236a) but was (mistakenly?) never added to the `staging.yaml` jobs list.

In any case, looks like this config can be safely removed, as the job no longer exists in any instance of Jenkins.

Trello: https://trello.com/c/DMA2e3Jp/3023-investigate-vcloud-core-and-vcloud-edgegateway-gems


sengi (approved):
Looks very dead to me 🦜💀

Thanks for the cleanup!

PR #11878, opened 2022-10-27
Reduce storage of closed PR branches

kevindew:
We currently store up to 50 builds for a project after a PR has been merged and for up-to 30 days.

This seems unnecessary as we rarely look at these after they are merged.

The motivation for this change is the frequency that we run out of disk space on Jenkins.

### DIFF
--- a/modules/govuk_ci/templates/application_build_job.xml.erb
+++ b/modules/govuk_ci/templates/application_build_job.xml.erb
@@ -17,8 +17,8 @@
   </icon>
   <orphanedItemStrategy class="com.cloudbees.hudson.plugins.folder.computed.DefaultOrphanedItemStrategy" plugin="cloudbees-folder@6.5.1">
     <pruneDeadBranches>true</pruneDeadBranches>
-    <daysToKeep>30</daysToKeep>
-    <numToKeep>50</numToKeep>
+    <daysToKeep>5</daysToKeep>
+    <numToKeep>10</numToKeep>
   </orphanedItemStrategy>
   <triggers/>
   <disabled>false</disabled>


ChrisBAshton (dismissed):
Just to double-check - this is just about removing the workspaces for old branches, and doesn't delete the _logs_ for said builds?

kevindew:
> Just to double-check - this is just about removing the workspaces for old branches, and doesn't delete the _logs_ for said builds?

This would remove the logs for those builds too, they'd 404 if you tried to view them. There doesn't seem to be much of a concept in Jenkins of deleting the workspace without the logs (seems mostly handled by plugins like [workspace cleanup](https://plugins.jenkins.io/ws-cleanup/))

ChrisBAshton:
Hmm, I'd be a bit hesitant about this change then 😬 I find it super useful being able to debug build issues, which we quite often don't get around to until some time after-the-fact.

Perhaps a compromise is to reduce the `numToKeep` value, but keep the `daysToKeep` as-is? If I understand correctly, that would keep fewer builds, but not arbitrarily delete them just if they pass a certain age?

kevindew:
> Hmm, I'd be a bit hesitant about this change then 😬 I find it super useful being able to debug build issues, which we quite often don't get around to until some time after-the-fact.
> 
> Perhaps a compromise is to reduce the `numToKeep` value, but keep the `daysToKeep` as-is? If I understand correctly, that would keep fewer builds, but not arbitrarily delete them just if they pass a certain age?

Do you do that with closed branches? I couldn't think why people would be looking at branches for PRs that have been merged after the time.

ChrisBAshton (approved):
Fair enough - 'closed branches' not too much of a concern 👍

kevindew:
> Fair enough - 'closed branches' not too much of a concern 👍

Thanks, there's always a chance I didn't fully understand the Jenkins config on this, so do prod me if this affects any broader. But I understand both config values to only apply once the branch is removed from GitHub.

PR #11881, opened 2022-10-31
Absent account team

huwd:
GOV.UK Account team is now a part of Identity, Follow up GOV.UK work is no longer being prioritised so I'm removing our access here.


AgaDufrat (approved):

sihugh:
Looks like this has happened in parallel. Thanks both.

PR #11886, opened 2022-11-02
Use updated bundler command

1pretz1:
bundle install flags `--path` and `--deployment` are deprecated so we should stop using them. We also think that this command could be causing an issue when installing Ruby 3.1.2 gems for Email Alert Monitor.

Deprecate warnings:

```
[DEPRECATED] The `--deployment` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set deployment 'true'`, and stop using this flag
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set path '/var/lib/jenkins/bundles/medical-safety-email-alert-check'`, and stop using this flag
```

Trello:
https://trello.com/c/wbq6cawk/1611-upgrade-ruby-for-email-alert-monitoring

### DIFF
--- a/modules/govuk_jenkins/templates/jobs/medical_safety_email_alert_check.yaml.erb
+++ b/modules/govuk_jenkins/templates/jobs/medical_safety_email_alert_check.yaml.erb
@@ -21,7 +21,9 @@
               ) || echo "Cloning failed. Re-using directory from previous run."
 
               cd email-alert-monitoring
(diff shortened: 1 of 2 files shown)


KludgeKML (approved):
LGTM

PR #11891, opened 2022-11-03
Revert "Add pentester usernames to hieradata"

MahmudH:
Reverts alphagov/govuk-puppet#11865


ChrisBAshton:
I've merged these changes into #11890 directly. Cleaner to just do it in one PR.

PR #11902, opened 2022-11-16
switch publishing api to read from its own version of schemas

BeckaL:
This is part of work to merge content schemas into publishing api. As publishing api is the only application for which content schemas is a runtime dependency, this is the only change required to switch an application over outside of CI configuration.

Do not merge yet: depends on [the PR in publishing api](https://github.com/alphagov/publishing-api/pull/2184) that actually merges content schemas in.

[Trello](https://trello.com/c/ZUK2duYI/361-add-govuk-content-schemas-to-publishing-api-repo)


JonathanHallam (approved):
Awesome, cheers :)

PR #11904, opened 2022-11-17
Set correct repo name for licensify-admin and licensify-feed.

sengi:
This should fix the failures in the Deploy_App_Downstream job for Licensify / GOV.UK Licensing.


dj-maisy (approved):
🚀

PR #11905, opened 2022-11-17
Set repo for licensify-admin and licensify-feed in Deploy_App_Downstream.

sengi:
This should fix Deploy_App_Downstream failing because it infers the wrong value for `$REPO`.


dj-maisy (approved):
👌

PR #11907, opened 2022-11-17
Override repo for licensify admin & feed apps

dj-maisy:
## What?
We're going to try adding the repo overrides for `licensify-admin` and `licensify-feed` to the YAML file specifically for `ci_master`, to see if we can get a fix for the Jenkins Job that is (still) failing.

## Why?
Well, because this block here...

https://github.com/alphagov/govuk-puppet/blob/ee86ba7434c7020b5bf5eeee5f80eb5ac96a6d03/modules/govuk_jenkins/templates/jobs/deploy_app_downstream.yaml.erb#L34-L48

... seems to be populated by a list of applications, and is still assuming the wrong repo name.


sengi (approved):

sengi:
Ahaha, good find there 😅

PR #11908, opened 2022-11-18
Remove MySQL XtraBackup code

ChrisBAshton:
XtraBackup is an open source utility, designed to perform a hot backup of MySQL data while the system is running:
https://www.percona.com/software/mysql-database/percona-xtrabackup

This harks back to an era where GOV.UK managed its own MySQL clusters in VMs spread across EC2 instances. With that infra, we had to handle our own replication and backup logic. GOV.UK moved to using a managed RDS service sometime after August 2017: https://docs.publishing.service.gov.uk/repos/govuk-aws/architecture/decisions/0018-use-rds-instead-of-provisioned-ec2-databases.html

Self-managed MySQL no longer exists in GOV.UK's infrastructure except for on the Jenkins CI agents. However, as these are only used for running tests, there is no need to worry about hot backups.


kevindew (commented):
This looks good, I'm just wondering is it installed anywhere where we need to run `ensure: absent`? hopefully not as I imagine that's a pain.

kevindew (approved):
Great stuff

PR #11914, opened 2022-11-21
Set python37 version to 1 to deploy on CI agents

kentsanggds:
## What

Update version number to 1 on CI agents for govuk-python-3.7 as it was was packaged as version 1

https://github.com/alphagov/packager/blob/main/fpm/recipes/govuk-python-3.7.15/recipe.rb#L4


KludgeKML (approved):
LGTM

PR #11915, opened 2022-11-22
Remove govuk content schemas

BeckaL:
We have now merged content schemas into publishing api, so we can remove:

* reference to govuk-content-schemas in deployable apps and CI. This should stop jenkins from running (we have removed write access to the repo)
* reference to govuk-content-schema in the publishing api machines
* remove the manifest file for content schemas

[Trello](https://trello.com/c/ZUK2duYI/361-add-govuk-content-schemas-to-publishing-api-repo)


brucebolt (approved):
All looks good to me.  I can't see anything where we might need to mark it as absent first, before removal.

PR #11920, opened 2022-11-29
Revert "Set repo for licensify-admin and licensify-feed in Deploy_App_Downstream."

sihugh:
Unfortunately, this turns on continuous deployment for these apps which we're not ready for yet.

We also spotted that the [licensify-feed auto-deploy fails](https://deploy.integration.publishing.service.gov.uk/job/Deploy_App_Downstream/9608/console):

> Could not resolve host: licensify-feed.integration.<internal-domain>

[Slack thread from when someone noticed](https://gds.slack.com/archives/C02KHSF1CMC/p1669715863378059)

Reverts alphagov/govuk-puppet#11905

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -1051,10 +1051,6 @@ govuk_jenkins::jobs::deploy_app_downstream::applications:
   hmrc-manuals-api: {}
   imminence: {}
   info-frontend: {}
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):

sengi:
oh no :( thanks for spotting!

PR #11926, opened 2022-12-01
Forward draft-assets traffic to internal LBs instead of public LBs.

sengi:
The public load balancers for whitehall-frontend were removed in https://github.com/alphagov/govuk-aws/pull/1572.

This should fix asset preview pages in draft (Content Preview), which have been broken since 5 Apr when the above was merged. (https://github.com/alphagov/govuk-aws/commit/f122845).

Examples of broken draft assets preview pages:
- https://draft-assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1120394/HMRC_ePCS_spending_over_500_for_October_2022.csv/preview
- https://draft-assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1106432/HMRC_Senior_Officials_Travel_April_to_June_2022.csv/preview

Rollout: will test on staging to confirm the fix.

### DIFF
--- a/modules/router/manifests/draft_assets.pp
+++ b/modules/router/manifests/draft_assets.pp
@@ -18,14 +18,8 @@ class router::draft_assets(
 ) {
   $whitehall_uploaded_assets_routes = hiera('router::assets_origin::whitehall_uploaded_assets_routes', [])
   $asset_manager_uploaded_assets_routes = hiera('router::assets_origin::asset_manager_uploaded_assets_routes', [])
-
   $upstream_ssl = true
-
-  if $::aws_environment == 'integration'{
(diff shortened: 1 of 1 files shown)


MuriloDalRi (approved):

sengi:
Turns out that Whitehall no longer serves those asset preview pages anyway, so unfortunately this doesn't solve the actual user issue.

It does still fix a bit of config that had been wrong since the AWS migration though 😅

sengi:
cc @brucebolt in case there's somewhere else we could forward these `://draft-assets.publishing.service.gov.uk/government/uploads/.../preview` requests?

Context is https://govuk.zendesk.com/agent/tickets/5141774

PR #11929, opened 2022-12-08
Use new Content Security Policy report-uri endpoints

kevindew:
Trello: https://trello.com/c/lxxx5XLZ/178-govuk-has-a-half-implemented-content-security-policy-csp

This is a draft until https://github.com/alphagov/govuk-aws/pull/1664 has been applied in all environments.

This configures our environments to use the new CSP endpoints that have been created from: https://github.com/alphagov/govuk-aws/pull/1664

As each URI is defined in each environment I've moved the report only setting to be paired with, just to avoid a situation of report only without a URI.


rtrinque (approved):
Looks good

PR #11928, opened 2022-12-07
Update RDS sync to include Imminence PG

KludgeKML:
Relies on https://github.com/alphagov/imminence/pull/766 and related infrastructure PRs.

Once Imminence is moved to RDS, this sets up the RDS sync for the new imminence DB.

PR #11930, opened 2022-12-12
Use amazonmq instead of rabbit mq on integration [Do not merge yet]

aldavidson:
Supply `RABBITMQ_URL` env var to publishing-api, search-api, email-alert-service, cache-clearing-service, and content-data-api on integration only. 

This URL points to the new AmazonMQ broker in integration. When this is merged, publishing-api will start publishing messages to AmazonMQ instead of the existing RabbitMQ. 

The consuming apps will ignore the `RABBITMQ_URL` env var and stay consuming from the existing RabbitMQ until the [corresponding new version of govuk_message_queue_consumer](https://github.com/alphagov/govuk_message_queue_consumer/pull/80) is released.

### DIFF
--- a/hieradata_aws/common.yaml
+++ b/hieradata_aws/common.yaml
@@ -446,6 +446,7 @@ govuk::apps::bouncer::unicorn_worker_processes: "8"
 
 govuk::apps::cache_clearing_service::enabled: true
 govuk::apps::cache_clearing_service::rabbitmq::enabled: true
+govuk::apps::cache_clearing_service::rabbitmq_url: ""
 govuk::apps::cache_clearing_service::rabbitmq_hosts: [rabbitmq]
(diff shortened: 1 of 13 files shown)


sengi (approved):
Rest of the changes (9eb9297 onwards) LGTM 👍

aldavidson:
> > I'm a bit confused by the imminence/postgres changes ([f70d9c4](https://github.com/alphagov/govuk-puppet/commit/f70d9c44645d530cd9c801338e2a5f6c5ee3671c), [bd452a7](https://github.com/alphagov/govuk-puppet/commit/bd452a71c93b172f1f4ff4d0d0ffd4862cb44889)) - are those intentionally included here?
> 
> Ah, good spot - Keith was working on integration for #11928 the other day, and we rebased those changes in to avoid conflicting. I'll rebase them back out again.

Done, thanks :+1:

PR #11931, opened 2022-12-12
Remove TLS V1 and TLS V1.1 from the list of ssl_protocols in nginx

MuriloDalRi:
Those protocols are now considered unsafe and should not be used.

https://trello.com/c/Eqxi6ssR/3054-remove-tlsv1-and-tlsv11-from-the-list-of-sslprotocols-in-nginx


sengi (approved):
Change LGTM; I'll leave fixing the test failure to you.

PR #11932, opened 2022-12-14
Stop monitoring rabbitmq for consumerless queues on integration

aldavidson:
In integration, all the apps that were publishing to, or consuming from, self-hosted RabbitMQ, are [now connecting to AmazonMQ](https://github.com/alphagov/govuk-puppet/pull/11930) instead. 

Icinga in integration has since been showing lots of alerts, as there are monitoring processes still configured to check RabbitMQ queues for activity, and alert if there are no consumers.

This PR introduces a `$monitor_rabbitmq_consumers` variable in hiera data for each application, which is then set to false in integration only. This should stop those alerts.


aldavidson:
closing un-merged - this PR has been superceded

PR #11937, opened 2022-12-19
add json logging variable

JonathanHallam:
This sets a default variable to be used by the latest sidekiq upgrade. 

https://github.com/alphagov/govuk_sidekiq/pull/63

### DIFF
--- a/modules/govuk/manifests/deploy/config.pp
+++ b/modules/govuk/manifests/deploy/config.pp
@@ -17,7 +17,7 @@
 #   Default: 'production'
 #
 # [*govuk_environment_name*]
-#   Name of the environment used to populate the environment label for 
+#   Name of the environment used to populate the environment label for
 #   publishing apps.
 #
 # [*website_root*]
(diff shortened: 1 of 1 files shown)


brucebolt (dismissed):
Will this be set as an environment variable (or used elsewhere in Puppet) in a later commit?

kevindew:
I'm a bit confused about what we're doing here as this variable doesn't seem to do anything? 

I'm assuming we're trying to set an env var? I'm not sure we actually need this as something optional in puppet. I imagine we probably want to set the environment variable up in: https://github.com/alphagov/govuk-puppet/blob/5fc81d2e5eace5d36358aa3f5b6d6c84a982ea9c/modules/govuk/manifests/deploy/config.pp#L121-L134 with the other `/etc/govuk/env.d` global vars - and just always have it set as a `1` value

modules/govuk/manifests/app.pp
```diff
@@ -257,6 +264,7 @@
   $port = 0,
   $command = undef,
   $create_pidfile = 'NOTSET',
+  $govuk_sidekiq_json_logging = false,
```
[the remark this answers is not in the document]
JonathanHallam: Nope, I think your comment is pretty on the money and I've misunderstood how puppet's configuring things 😬 I've amended the commit, this is going off the comment you made on the sidekiq pr 
https://github.com/alphagov/govuk_sidekiq/pull/63#issuecomment-1331947639

modules/govuk/manifests/deploy/config.pp
```diff
@@ -17,7 +17,7 @@
 #   Default: 'production'
 #
 # [*govuk_environment_name*]
-#   Name of the environment used to populate the environment label for 
+#   Name of the environment used to populate the environment label for
```
JonathanHallam, line 20: I assume my editor did this automatically, not mad though

modules/govuk/manifests/deploy/config.pp
```diff
@@ -129,14 +134,15 @@
...
-    'SIDEKIQ_LOGFILE': value         => $sidekiq_logfile;
+    'ERRBIT_ENVIRONMENT_NAME': value    => $errbit_environment_name;
+    'SENTRY_CURRENT_ENV': value         => $errbit_environment_name;
+    'GOVUK_ENVIRONMENT_NAME': value     => $govuk_environment_name;
+    'GOVUK_ASSET_ROOT': value           => $asset_root;
+    'GOVUK_WEBSITE_ROOT': value         => $website_root;
+    'GOVUK_CSP_REPORT_ONLY': value      => $csp_report_only_value;
+    'GOVUK_CSP_REPORT_URI': value       => $csp_report_uri;
+    'SIDEKIQ_LOGFILE': value            => $sidekiq_logfile;
+    'GOVUK_SIDEKIQ_JSON_LOGGING': value => $govuk_sidekiq_json_logging
```
kevindew, line 145:
```suggestion
    'GOVUK_SIDEKIQ_JSON_LOGGING': value => '1'
```

We've not actually set up a `$govuk_sidekiq_json_logging` variable. I don't think we really need to have it as a variable though as we just want it set everywhere.

modules/govuk/manifests/deploy/config.pp
```diff
@@ -43,6 +43,11 @@
 # [*sidekiq_logfile*]
 #   Path to the logfile for Sidekiq (async work queue library).
 #   Default: 'log/sidekiq.log'
+#
+# [*govuk_sidekiq_json_logging*]
+#   Whether the app outputs logs in JSON format. Only applications running
+#   govuk_sidekiq > 6 will need this
+#   Default: false
```
kevindew, line 50:
```suggestion
```

As per the previous comment that we don't actually have a variable for this and I don't think we need one.

brucebolt (approved):
Looks good to me, as we want this environment variable to always be set 👍.

PR #11939, opened 2022-12-19
Update regex for sidekiq-monitoring collectd metrics

samsimpson1:
Update regex to match the sidekiq-monitoring processes in CollectD.

CollectD not being able to match the processes was causing Unknown errors on every environment (1 for each backend machine), with the line: UNKNOWN: INTERNAL ERROR: RuntimeError: no valid datapoints. You could replicate this by running the graphite check manually on monitoring, and you could see there were no datapoints in the graphite graph. Post this change you see data in the graph, the unknown error clears, and running the graphite check on monitoring gives you a metric.


kashifatcha (approved):

KludgeKML (approved):

brucebolt (commented):
Looks good to me.  Would you be able to add some extra information to the commit message explaining why we're making this change, in case anyone stumbles across it again in the future.

PR #11941, opened 2022-12-19
Stop checking status of the now-empty RabbitMQ load balancer in integration

aldavidson:
Part of [Trial AmazonMQ deployment process on integration](https://trello.com/c/oqsXryjz/389-update-monitoring-alerting-from-rabbitmq-amazonmq) epic. 

We've [removed all existing RabbitMQ EC2 instances on integration](https://github.com/alphagov/govuk-aws/pull/1676), but now the monitoring server is triggering alerts that the corresponding load balancer does not have any healthy instances attached.

This PR removes `blue-rabbitmq-internal` from the integration load balancer checks


KludgeKML (approved):
LGTM!

PR #11943, opened 2022-12-20
Introduce monitoring_host variable

BeckaL:
This will allow us to switch from self-hosted rabbitMQ (for which we install the monitoring plugin on the same EC2 instance as the rabbitMQ process, and therefore points to localhost) and AmazonMQ (for which we will monitor from publishing api, so the host needs to point to the AmazonMQ instance).


aldavidson:
Closing this PR unmerged - we've come up with a better model for monitoring AmazonMQ via Cloudwatch. This PR is now superceded by #11958

PR #11953, opened 2023-01-03
Add unless_vg/createonly to physical volume creation to fix ci-agent 5

KludgeKML:
Agent 5 has the pvs for the volume groups the wrong way round, and is failing to swap them, which causes a puppet run failure. This addition tells puppet only to create the volume groups if they don't exist, don't try to fiddle with existing VGs.


KludgeKML:
Just going to rebase/rewrite the commit so it is a bit nicer.

ChrisBAshton (approved):
Great work getting to the bottom of this!

PR #11956, opened 2023-01-09
Raise threshold for the gov-frontend CPU usage alert.

sengi:
This alert is unactionable, but deleting it is more faff than just increasing the threshold.


modules/govuk/manifests/apps/government_frontend.pp
```diff
@@ -44,8 +44,8 @@
   $account_api_bearer_token = undef,
   $publishing_api_bearer_token = undef,
   $unicorn_worker_processes = undef,
-  $cpu_warning = 150,
-  $cpu_critical = 200,
+  $cpu_warning = 300,
+  $cpu_critical = 300,
```
sengi, line 48: The units are % of a core, so the new threshold is 3 cores.

edwardkerry (approved):

PR #11962, opened 2023-01-12
Add AmazonMQ Grafana dashboard

aldavidson:
[Trello card](https://trello.com/c/Dj9II4RC/418-repoint-the-grafana-dashboards-at-amazonmq)

Replicate as much as possible from the existing RabbitMQ dashboard into a new AWS/AmazonMQ dashboard, pulling metrics from CloudWatch.

This new dashboard shows message counts (ready / unack'ed / total) and consumer counts for each queue.
It was not possible to bring across the exchange publish rates from the existing RabbitMQ dashboard, as CloudWatch does not expose those metrics.

It's been [deployed on integration](https://grafana.blue.integration.govuk.digital/dashboard/file/aws-amazonmq.json?orgId=1) and works (although there hasn't been much activity to graph, we published some docs around 12 noon on Jan 12th):
image

### DIFF
--- a/modules/grafana/files/dashboards_aws/aws-amazonmq.json
+++ b/modules/grafana/files/dashboards_aws/aws-amazonmq.json
@@ -126,7 +126,7 @@
                 "showDisabledItems": false
               },
               "period": "",
-              "refId": "A",
+              "refId": "B",
               "region": "$region",
               "statistics": [
                 "Average"
@@ -159,7 +159,7 @@
                 "showDisabledItems": false
               },
               "period": "",
-              "refId": "A",
+              "refId": "C",
               "region": "$region",
               "statistics": [
                 "Average"
@@ -192,7 +192,7 @@
                 "showDisabledItems": false
               },
               "period": "",
-              "refId": "A",
+              "refId": "D",
               "region": "$region",
               "statistics": [
                 "Average"
--- a/modules/grafana/files/dashboards_aws/aws-amazonmq.json
+++ b/modules/grafana/files/dashboards_aws/aws-amazonmq.json
@@ -1,0 +1,382 @@
+{
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "4.5.2"
+    },
+    {
+      "type": "panel",
+      "id": "graph",
+      "name": "Graph",
+      "version": ""
+    },
+    {
+      "type": "datasource",
+      "id": "cloudwatch",
+      "name": "CloudWatch",
+      "version": "1.0.0"
+    }
+  ],
+  "annotations": {
+    "list": []
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "hideControls": false,
+  "id": null,
+  "links": [],
+  "refresh": "10s",
+  "rows": [
+    {
+      "editable": true,
+      "collapse": false,
+      "height": "300px",
+      "panels": [
+        {
+          "aliasColors": {},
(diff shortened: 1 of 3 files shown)


modules/grafana/manifests/datasources.pp
```diff
@@ -29,7 +29,7 @@
     \"name\":\"CloudWatch\",
     \"type\":\"cloudwatch\",
     \"access\":\"proxy\",
-    \"jsonData\":{\"authType\":\"arn\", \"defaultRegion\":\"eu-west-1\", \"assumeRoleArn\":\"\", \"timeField\":\"@timestamp\"}
+    \"jsonData\":{\"authType\":\"arn\", \"defaultRegion\":\"eu-west-1\", \"assumeRoleArn\":\"\", \"timeField\":\"@timestamp\", \"customMetricsNamespaces\": \"AWS/AmazonMQ\"}
```
BeckaL, line 32: Probs just a q for my understanding: Is this custom namespace doing anything, given we add it in on the templating anyway?

modules/grafana/files/dashboards_aws/aws-amazonmq.json
```diff
@@ -0,0 +1,382 @@
...
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "hideControls": false,
+  "id": null,
+  "links": [],
+  "refresh": "10s",
+  "rows": [
+    {
+      "editable": true,
```
BeckaL, line 34: Was this line required, or was this the one I added in when I forgot you had to be signed in to edit 🙃 ?

modules/grafana/files/dashboards_aws/aws-amazonmq.json
```diff
@@ -0,0 +1,382 @@
...
+                "filter": ""
+              },
+              "metricName": "MessageReadyCount",
+              "mode": 0,
+              "namespace": "AWS/AmazonMQ",
+              "options": {
+                "showDisabledItems": false
+              },
+              "period": "",
+              "refId": "A",
```
BeckaL, line 129: I'm guessing these ref ids should be unique (even though we're not using them for anything atm)?

modules/grafana/files/dashboards_aws/aws-amazonmq.json
aldavidson: This line has always been there, AFAICT - all the other dashboards have `"editable": true` lines, so I suggest we stick with the convention

modules/grafana/files/dashboards_aws/aws-amazonmq.json
aldavidson: AFAICT there's no problem with the refIDs not being unique (for instance, the content_data_api application dashboard is all "A"'s), but I've made them unique anyway

modules/grafana/manifests/datasources.pp
aldavidson: According to the docs this is what makes the metrics available in the online editor UI. 
There's some weird caching going on that means it's difficult to test this with/without the line - I can't get rid of it in the 'Edit datasource' UI now, even after deploying a revision without it.

So I suggest we err on the side of caution and keep it in?

sengi:
No longer shows up on integration Grafana, but screenshots look good so 👍

sengi (approved):

modules/grafana/manifests/datasources.pp
aldavidson: I've now managed to test this without the custom namespace in 'Datasources' - and it still seems to work, so I've removed it from the code.

PR #11968, opened 2023-01-24
Ship procfile worker logs to Logit to replace lack of Sidekiq file logs

kevindew:
Trello: https://trello.com/c/TEXZQpu9/439-sidekiq-6-log-files-arent-reaching-logit

This configures govuk::procfile::worker to utilise filebeat to ship log
files. The motivation for adding this is that GOV.UK apps that use
Sidekiq >= 6 are not having their logs shipped to logit.

The most common usage of govuk::procfile::worker is for Sidekiq and
prior to version 6 GOV.UK wrote Sidekiq logs to a file
(log/sidekiq.json). With Sidekiq 6 there is no longer the ability to
write to a logfile [1]. While there have previously been stdout and
stderr log files created as part of a govuk::procfile::worker process
these have not been configured to use filebeat.

While Sidekiq is the most common usage of govuk::procfile::worker it is
not the only one - it is used for running other background tasks like
RabbitMQ consumers - and thus some compromises were made so it wasn't
Sidekiq specific. This mostly impacts the tagging of logs which will now
use the process name from the procfile (generally "worker") rather than
a tag of "sidekiq".

The expectation is that GOV.UK Sidekiq logging will be in JSON so this
is the default in the class. For other uses of this class there's a good
chance that this will need to be set to false (see next commit).

In experimenting with this I investigated whether to ship stderr logs as
well as stdout and became concerned that this might create a yak shave.
From my brief testing there seemed to be a lot of junk in stderr
(search-api logged a Ruby warning every 4 seconds 😭) and there seemed
to be some duplications (presumably a consequence of our stdout/stderr
munging in GovukLogging [2]). As these logs weren't previously shipped
adding this seemed to be more adding a new feature that fixing something
lost so I stepped away from it.

[1]: https://www.mikeperham.com/2019/09/03/welcome-to-sidekiq-6.0/
[2]: https://github.com/alphagov/govuk_app_config/blob/13083f8f8563c4703c14fe8175ca35646ec64442/lib/govuk_app_config/govuk_logging.rb#L7-L26

### DIFF
--- a/modules/govuk/manifests/app.pp
+++ b/modules/govuk/manifests/app.pp
@@ -438,6 +438,7 @@ define govuk::app (
     fields => {'application' => $title},
   }
 
+  # This configuration will become redundant once all Sidekiq versions are >= 6
   @filebeat::prospector { "${title}_sidekiq_json_log":
     paths  => ["/var/apps/${title}/log/sidekiq*.log"],
     fields => {'application' => $title},
--- a/modules/govuk/manifests/apps/ckan.pp
+++ b/modules/govuk/manifests/apps/ckan.pp
@@ -146,28 +146,31 @@ class govuk::apps::ckan (
     }
 
     govuk::procfile::worker { 'harvester_fetch_consumer':
-      ensure         => $toggled_harvester_fetch,
-      setenv_as      => 'ckan',
-      enable_service => $enable_harvester_fetch,
-      process_type   => 'harvester_fetch_consumer',
-      respawn_count  => 15,
-      process_regex  => $fetch_process_regex,
+      ensure          => $toggled_harvester_fetch,
+      setenv_as       => 'ckan',
+      enable_service  => $enable_harvester_fetch,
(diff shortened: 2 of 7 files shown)


BeckaL (approved):
Makes sense to me. Thanks for comprehensive details of compromises made / testing, very helpful. Approving but a question - do we have an idea of the volume of logs that already get sent to stdout? Partly wondering whether we might see issues if there's a high volume of logging.

kevindew:
> Makes sense to me. Thanks for comprehensive details of compromises made / testing, very helpful. Approving but a question - do we have an idea of the volume of logs that already get sent to stdout? Partly wondering whether we might see issues if there's a high volume of logging.

Good question, in theory it should be in a very similar ballpark to that which was previously used for sidekiq logging so should hopefully make a negligible difference to overall logit usage. There is a slight risk existing things that logged to stdout could be noisy and increase things a bit, but I didn't see evidence of that in my testing (stderr on the other hand 😬)

PR #11971, opened 2023-01-27
Enable Crumb Issuer proxy compatibility

deborahchua:
This should hopefully fix the 403 response error we have been seeing when trying to load pages in Jenkins.
note from https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/security/csrf/DefaultCrumbIssuer/help-excludeClientIPFromCrumb.html using this option makes the nonce value easier to forge.

Screenshot 2023-01-27 at 10 08 02


catalinailie (approved):

ChrisBAshton (approved):
Worth saying you've tested it on Integration and it seems to have worked for people 🎉

PR #11972, opened 2023-01-27
Run a Redis 6 container instance in CI 

kevindew:
Trello: https://trello.com/c/JXAvbpOX/1103-upgrade-apps-to-sidekiq-v6

On our Jenkins CI machines we run Redis 2.8.4 😭, this finally caught up
to us with tests for GOV.UK apps that use Sidekiq 6 and makes use of
newer Redis features.

This applies the same approach we've used to get around ancient versions
of PostgreSQL, MongoDB and MySQL and sets up a containerised version of
Redis so our apps can use an equivalent to version as the production
runtime.

This sets-up a Redis 6 container that runs on the CI machines (we are
using Redis 6 via ElastiCache in production). Rails applications can use
this by setting up an env var e.g:

```
  govuk.setEnvar("REDIS_URL", "redis://127.0.0.1:63796")
```

I set the port number to be 63796 to reflect 6379 as the regular Redis
port and 6 to signify the version.

An example of this running successfully is: https://ci.integration.publishing.service.gov.uk/job/travel-advice-publisher/job/sidekiq-6-kevin/6/

### DIFF
--- a/modules/govuk_ci/manifests/agent.pp
+++ b/modules/govuk_ci/manifests/agent.pp
@@ -20,7 +20,6 @@ class govuk_ci::agent(
 
   include ::clamav
   include ::clamav::cron_freshclam
-  include ::govuk_ci::agent::redis
   include ::govuk_ci::agent::docker
   include ::golang
   include ::govuk_ci::agent::gcloud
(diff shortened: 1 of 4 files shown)


davidgisbey (approved):
Looks good to me. Thanks so much for sorting this 👍

PR #11975, opened 2023-02-01
Update RDS sync to include Authenticating Proxy PG

KludgeKML:
Relies on https://github.com/alphagov/authenticating-proxy/pull/350 and related infrastructure PRs.

Once Authenticating proxy is moved to RDS, this sets up the RDS sync for the new authenticating proxy DB.


hieradata_aws/class/integration/authenticating_proxy_db_admin.yaml
```diff
@@ -0,0 +1,25 @@
+govuk_env_sync::tasks:
+  "pull_authenticating_proxy_production_daily":
+    ensure: "present"
+    hour: "0"
+    minute: "0"
+    action: "pull"
+    dbms: "postgresql"
+    storagebackend: "s3"
+    database: "_production"
```
AgaDufrat, line 9: Don't we need a full name here? 🤔 
```suggestion
    database: "authenticating_proxy_production"
```

AgaDufrat (approved):
Left a small comment/question otherwise seems fine

KludgeKML:
You're absolutely right, we did need a full name there. Good catch! Fixed now.

PR #11976, opened 2023-02-02
Repoint publishing-api at AmazonMQ in staging

aldavidson:
[Trello card](https://trello.com/c/Qv3u0M0c/442-package-2-point-publishing-api-at-amazonmq), part of [run the AmazonMQ migration process in staging](https://trello.com/c/8a87xtft/428-run-amazonmq-migration-process-on-staging)

Depends on the govuk-secrets PR to [add AmazonMQ credentials in staging](https://github.com/alphagov/govuk-secrets/pull/1441) (now merged)

To be merged as the first part of a suite of several PRs, aiming to perform the switchover as smoothly and quickly as possible

PR #11977, opened 2023-02-02
Add rabbitmq_url pointing to AmazonMQ for consuming apps in staging

aldavidson:
[Trello card](https://trello.com/c/tMid1Zae/448-package-3-point-consuming-apps-at-amazonmq-try-all-in-one-release) , part of [run the AmazonMQ migration process on staging](https://trello.com/c/8a87xtft/428-run-amazonmq-migration-process-on-staging)

To be merged & deployed _after_ [repointing publishing-api](https://github.com/alphagov/govuk-puppet/pull/11976)


BeckaL (approved):
Looks good to me, checked this is all the services we need to do and that each url looks correct

PR #11989, opened 2023-02-09
Switch apps which consume publishing RabbitMQ, to consume the new AmazonMQ in production 

aldavidson:
[Trello card](https://trello.com/c/ZMgbl4GI/462-point-consuming-apps-at-amazonmq-in-production-all-in-one-release), part of [running the AmazonMQ migration process in production](https://trello.com/c/to4fHvoa/458-run-amazonmq-migration-process-on-production)

For reference, here's the [corresponding PR for staging](https://github.com/alphagov/govuk-puppet/pull/11977/files)

Do not merge yet, needs to be deployed at the right stage of the above migration process


BeckaL (approved):
Changes look good to me. I think it might be worth updating the commit message (and PR message?) to explain that this will actually switch the applications over to consuming from the new amazonmq queues, as it might be non-obvious to people without context.

aldavidson:
> Changes look good to me. I think it might be worth updating the commit message (and PR message?) to explain that this will actually switch the applications over to consuming from the new amazonmq queues, as it might be non-obvious to people without context.

Good call, thanks - done

PR #11991, opened 2023-02-10
Stop daily backups of MongoDB in integration

deborahchua:
This script is only used to populate data that govuk-docker used for data replication - this has recently been updated so that it now uses data from the govuk-env-sync process.

Once we can confirm this job no longer runs we can proceed to remove the associated code.

See https://github.com/alphagov/govuk-docker/pull/640/commits/c998b4d9811c59b642f4d5389781128eebad3227 for more information.

Trello card: https://trello.com/c/mOYRWDOc/3084-modify-govuk-docker-so-that-it-uses-the-govuk-env-sync-backups-3

### DIFF
--- a/modules/mongodb/manifests/aws_backup.pp
+++ b/modules/mongodb/manifests/aws_backup.pp
@@ -11,7 +11,7 @@
 #   Name of the bucket to backup to.
 #
 class mongodb::aws_backup (
-  $ensure     = 'present',
+  $ensure     = 'absent',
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
LGTM 👍 I think you'd also be safe to remove https://github.com/alphagov/govuk-puppet/blob/main/modules/mongodb/templates/mongodump-to-s3.erb at this point, but no harm in waiting until the next PR if you like.

PR #11996, opened 2023-02-13
Bump up ruby version to 3.1.3

rtrinque:
Bump up ruby version to 3.1.3


modules/govuk_rbenv/manifests/all.pp
```diff
@@ -39,7 +39,7 @@
     '3.0.3',
     '3.0.4',
     '3.0.5',
-    '3.1.2',
```
catalinailie, line 42: We might want to keep '3.1.2' here

rtrinque: ooops

catalinailie (approved):

PR #11999, opened 2023-02-16
Switch on 2SV features for Signon in integration

brucebolt:
We have recently introduced some new 2-step verification (2SV) features in Signon that are currently behind feature flags.

This sets the environment variables to switch them on in integration, so they can be demonstrated outside of production.


BeckaL (approved):
Looks good to me

PR #12005, opened 2023-02-22
Add session key for sidekiq-monitoring app

brucebolt:
These will be used to allow us to upgrade the app to use Sidekiq 6, by creatinga session cookie.

The values have been added in https://github.com/alphagov/govuk-secrets/pull/1450.

[Trello card](https://trello.com/c/aVx282Rt)

### DIFF
--- a/modules/govuk/manifests/apps/sidekiq_monitoring.pp
+++ b/modules/govuk/manifests/apps/sidekiq_monitoring.pp
@@ -144,6 +144,10 @@
 #   Redis port for Search API Sidekiq.
 #   Default: undef
 #
+# [*session_key*]
+#   Session key secret for this application.
+#   Default: undef
+#
 # [*signon_redis_host*]
 #   Redis host for Signon Sidekiq.
 #   Default: undef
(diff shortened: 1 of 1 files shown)


BeckaL (approved):
Looks good to me, was thinking we'd have to also read the secrets in as env variables such as [here](https://github.com/alphagov/govuk-puppet/blob/696c12ddfe54f0f714b1a282f67a526ac12bcab5/hieradata_aws/integration.yaml#L17) but actually this is because we're interpolating the value, so I assume that we don't actually need to do anything more than this?

brucebolt:
> was thinking we'd have to also read the secrets in as env variables such as [here](https://github.com/alphagov/govuk-puppet/blob/696c12ddfe54f0f714b1a282f67a526ac12bcab5/hieradata_aws/integration.yaml#L17) but actually this is because we're interpolating the value, so I assume that we don't actually need to do anything more than this?

Yes, nothing needed in puppet as the values from secrets will get added in here on deployment.

PR #12013, opened 2023-02-28
Update Golang to 1.20.1

Tetrino:
Major version update to match new version in Router.


AgaDufrat (approved):
👍  Shall we remove 1.17.8 and 1.18.3 in this commit?

Tetrino:
@AgaDufrat I'd love to, but I need to audit the versions as doing so breaks something and it's a real pain to find out exactly what. That's a background task for the future.

PR #12017, opened 2023-03-02
Add chriswhitehouse to integration.yml

chriswhitehouse-tw:


ollietreend (commented):
Thanks for opening this @chriswhitehouse-tw.

You'll also need to add your SSH public key at the path:
```
modules/users/manifests/chriswhitehouse.pp
```

To see an example of how to structure that file, take a look at this recent PR from Korny: #12011

ollietreend (approved):

PR #12010, opened 2023-02-24
Allow integration to pull data from staging not prod

MahmudH:
Integration is restricted in pulling data from our production environment, therefore, this config will allow staging to push data into staging to allow integration to access it. 

Trello: https://trello.com/c/42WqbBy7/3082-lock-down-access-to-production-database-dumps-investigation

PR #12018, opened 2023-03-03
Revert "Allow integration to pull data from staging not prod"

ChrisBAshton:
Reverts alphagov/govuk-puppet#12010

We're seeing issues with the env sync roughly corresponding with the merge date of this PR, so will try reverting, as it's still not obvious what the cause is.

Trello: https://trello.com/c/42WqbBy7/3082-lock-down-access-to-production-database-dumps-investigation


MahmudH (approved):
👍🏽

PR #12028, opened 2023-03-14
Set Rory Hurley to absent

andysellick:
As per step 2 of https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html

Trello card: https://trello.com/c/1c06t71t/1856-leaver-rory-hurley-tech-role

### DIFF
--- a/modules/users/manifests/roryhurley.pp
+++ b/modules/users/manifests/roryhurley.pp
@@ -1,6 +1,7 @@
 # Creates the roryhurley user
 class users::roryhurley {
   govuk_user { 'roryhurley':
+    ensure   => absent,
     fullname => 'Rory Hurley',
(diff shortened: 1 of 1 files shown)


MuriloDalRi (approved):

modules/users/manifests/roryhurley.pp
```diff
@@ -1,6 +1,7 @@
 # Creates the roryhurley user
 class users::roryhurley {
   govuk_user { 'roryhurley':
+    ensure => absent,
```
MuriloDalRi, line 4: Puppet is complaining about the indentation here

andysellick: Good shout, thanks, I see it now. Have corrected.

PR #12033, opened 2023-03-17
Disable workers for migrated EKS applications

theseanything:
This disables the Sidekiq workers for application that are now running on EKS. This prevents workers in EC2 from processing jobs, as they still have access to the same Redis instance with the queues.


sengi (approved):
I briefly thought we might wanna do this environment by environment, but in practice I think just switching them all off at once like this makes a lot of sense — saves time, can still roll back easily and I can't think of an actual practical downside. LGTM 👍

PR #12036, opened 2023-03-20
Remove Rory Hurley's and Sid Elangovan's manifestos

davidgisbey:
## Description

As per the documents https://docs.publishing.service.gov.uk/manual/removing-a-user-from-puppet.html. This removes their manifestos and removes Rory from the integration users.

Sid is not present in this list of users.

## Trello card

https://trello.com/c/0WFQYdlL/1845-leaver-sid-elangovan-tech-role
https://trello.com/c/1c06t71t/1856-leaver-rory-hurley-tech-role

### DIFF
--- a/hieradata_aws/integration.yaml
+++ b/hieradata_aws/integration.yaml
@@ -355,7 +355,6 @@ users::usernames:
   - rebeccapearce
   - richardtowers
   - rochtrinque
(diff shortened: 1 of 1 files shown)


kentsanggds (approved):
LGTM

PR #12051, opened 2023-04-14
Disable Sentry for Smokey

theseanything:
As we still deploy to EC2, Smokey still runs and reports errors to Sentry. However, these are just noise as we've migrated to EKS.


ChrisBAshton (approved):
Nice 👍

PR #12056, opened 2023-04-18
Remove Smokey from the Deploy_App_Downstream job

deborahchua:
We no longer need to run Smokey in EC2 as it's now running on Kubernetes. This should stop it from running as part of the deploy app downstream job.

Trello card: https://trello.com/c/ahTC4Wjj/3134-disable-smokey-in-jenkins-2


MahmudH (approved):
LGTM 👍🏽

PR #12063, opened 2023-04-25
Add Farah Mehmood to integration

farahTW:
Pull request to Add farah mehmood to integration as per step 7 mentioned in the below
https://docs.publishing.service.gov.uk/manual/get-started.html#generate-a-ssh-key

### DIFF
--- a/modules/users/manifests/farahmehmood.pp
+++ b/modules/users/manifests/farahmehmood.pp
@@ -2,7 +2,7 @@
 class users::farahmehmood {
   govuk_user { 'farahmehmood':
     fullname => 'Farah Mehmood',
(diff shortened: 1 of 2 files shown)


modules/users/manifests/farahmehmood.pp
```diff
@@ -0,0 +1,8 @@
+# Creates the farahmehmood user
+class users::farahmehmood {
+  govuk_user { 'farahmehmood':
+    fullname => 'Farah Mehmood',
+    email    => 'farahm.mehmood@digital.cabinet-office.gov.uk',
```
davidgisbey, line 5: I think this should be 
```
'farah.mehmood@digital.cabinet-office.gov.uk',
```

davidgisbey (approved):

PR #12064, opened 2023-04-25
Remove govuk-mirror-repos Jenkins job

theseanything:
This has been replaced with functionality in govuk-infrastructure. CodeCommit repositories are now manged with Terraform and the git repos kepts in sync by the mirror-repo GitHub Action.


sengi (approved):
Awesome!

PR #12095, opened 2023-06-01
Remove the govuk-dns Jenkins job and CI job

theseanything:
This removes the remaining resources for supporting govuk-dns as this has been retried in favour for govuk-dns-tf.


sengi (approved):
Thanks!

PR #12105, opened 2023-07-04
Remove cache_clearing_service AmazonMQ monitoring

AgaDufrat:
The app was retired and queues removed so we don't need to monitor it anymore.
It also marks cache_clearing_service as disabled. I'm not sure if it's needed as the
app was migrated to Kubernetes, so it might only serve as documentation.

[GIFT week summer 2023 card](https://trello.com/c/ENPwB2vT/41-retire-cache-clearing-service)
[Tech debt card](https://trello.com/c/TX9wLihh/101-cache-clearing-app-exists-and-we-dont-think-its-necessary-replacing-cache-clearing-app-only-clears-path-not-query-strings)


PeterHattyar (approved):
Looks good, nice one!

PR #12110, opened 2023-07-20
Revert "Add user danmoore to integration"

kevindew:
Reverts alphagov/govuk-puppet#12108

Dan has now left, potentially temporarily but we're still doing the off-boarding process.

This wasn't ever run in any environments (puppet seems to have failed to auto-deploy) so we can skip the ensure => absent process


kevindew:
Argh forgot we need to remove first

hannako (approved):

PR #12120, opened 2023-08-16
Remove conorlowney user

dnkrj:
# What
Add `ensure => absent` to Conor Lowney's entry

# Why
He has left GDS


MartinJJones (approved):

MartinJJones (commented):
Looks like we need to align the arrow `=>` with the other arrows in the list, for example - https://github.com/alphagov/govuk-puppet/pull/12078/files

MartinJJones (approved):

PR #12124, opened 2023-08-30
Match pull/push time for Publishing API and Content Store

brucebolt:
As part of the daily data sync between our environments, we currently pull the Publishing API database before Content Store.

This means the data within can be out-of-sync when pushed to the next environment.  One issue that arises is the `payload_version` (a value sent from Publishing API to Content Store, to make sure we don't overwrite newer content) is mismatched.

Therefore updating the data sync so we pull from Content Store and Publishing API at exactly the same time (00:00), then push to the next environment at exactly the same time (01:30).

This will resolve any issues resulting from data mismatch.

[Trello card](https://trello.com/c/HRwJTf6S)

### DIFF
--- a/hieradata_aws/class/integration/mongo.yaml
+++ b/hieradata_aws/class/integration/mongo.yaml
@@ -1,8 +1,8 @@
 govuk_env_sync::tasks:
   "pull_content_store_daily":
     ensure: "present"
-    hour: "4"
-    minute: "16"
+    hour: "1"
+    minute: "30"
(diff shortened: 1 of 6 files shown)


sengi (approved):
No problem changing this but just checking you're aware: this doesn't solve the underlying race condition and consistency issue, right? :)

sengi:
"A man with a watch knows what time it is. A man with two watches is never sure."

PR #12148, opened 2023-10-27
Remove mirror monitoring

theseanything:
These have been replaced with new monitoring in Prometheus.


sengi (approved):
Puppet's broken in prod at the moment btw — feel free to merge and roll out but it just won't take effect until we fix the Puppetmaster.

Issue #12155, opened 2023-11-10
Public visibility of https://github.com/alphagov/govuk-puppet/blob/main/hieradata_aws/vagrant_credentials.yaml


annaobarz:
I stumbled across this publicly accessible page, and was wondering whether it should be publicly accessible?

sengi:
I think the Vagrant environment hasn't been used in a long time and this is some disused config related to it, but we really appreciate your vigilance — thanks for raising this!

I'll double-check that those private keys etc. really aren't used anywhere. (And of course we'll nuke the file in the meantime.)

sengi:
Also huge thanks for putting me onto the fact that — somehow — we don't appear to have secret scanning enabled on this repo 😅 (it's supposed to be enabled+enforced org-wide, plus I thought public repos were meant to be covered by default — clearly not always, apparently!)

annaobarz:
@sengi happy to help!

PR #12158, opened 2023-11-10
Delete old Vagrant credentials.

sengi:
The Vagrant environment is long-dead (and only ever would have been run locally) but it's not a good look to have any kind of private key or password checked into a public repo (even if it's just for testing), so let's clean it up.

Addresses #12155, with credit to @annaobarz for bringing this to our attention.

### DIFF
--- a/hieradata_aws/vagrant.yaml
+++ b/hieradata_aws/vagrant.yaml
@@ -7,19 +7,10 @@ backup::offsite::monitoring::offsite_fqdn: backup.gov.uk
 backup::offsite::monitoring::offsite_hostname: backup0.provider0
 backup::offsite::monitoring::monitor_cdn_logs_disk: true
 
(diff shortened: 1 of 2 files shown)


sengi:
Considered expunging it from the history, but I don't think it's worth it given the specifics of this case.

sengi:
Ah _of course_ it breaks the tests. Sigh.

sengi:
Let's try that again.

sengi:
Just to clarify, these keys etc. aren't used anywhere as far as I can tell, so there's nothing left to revoke etc. This is about removing a bunch of defunct (but still scary-looking) key material from HEAD so that a) it's clear that it's no longer in use and b) we have a record that we checked it's no longer in use.

sihugh (approved):
All I see is `*******`

PR #12159, opened 2023-11-17
Add user pete goddard to integration

peteglondon:
Added the pp file to the users directory and petergoddard classname to the integration yaml in hieradata_aws


Rosa-Fox (approved):
Thanks Pete!

